Issue #9, opened 2020-03-12
Design proxy route config format


anurse:
Config based proxies are common and we'll need to support at least basic proxy scenarios from config. Here are some initial considerations:
- Define routes based on host and/or path
- A restart should not be needed to pick up config changes
- You should be able to augment a route with code/middleware. Kestrel does something similar using named endpoints.
- List multiple back-ends per route for load balancing
- Configure a load balancing strategy

Tratcher:
Feedback from https://github.com/microsoft/reverse-proxy/pull/47#discussion_r418632319
> Without looking through the implementation, it's not clear why the Action takes a string. It looks like it's for the Backend id. Would it be better to put an id property on Backend similar to what we do for ProxyRoute?

This is a reflection of the asymmetry we have in the config structures. Backends is a set of named objects (with the names on the outside), but routes are an array of objects where the name is a field. There was some concern that routes needed to be specifically ordered because of how route matching works, but backends weren't considered ordered and named objects made the config cleaner. We haven't yet decided if consistency is needed here.
```
"Backends": {
      "backend1": {
        "LoadBalancing": {
          "Mode": "Random"
        },
        "Endpoints": {
          "backend1/endpoint1": {
            "Address": "https://localhost:10000/"
          },
          "backend1/endpoint2": {
            "Address": "http://localhost:10010/"
          }
        }
      },
      "backend2": {
        "Endpoints": {
          "backend2/endpoint1": {
            "Address": "https://localhost:10001/"
          }
        }
      }
    },
    "Routes": [
      {
        "RouteId": "backend1/route1",
        "BackendId": "backend1",
        "Match": {
          "Methods": [ "GET", "POST" ],
          "Host": "localhost",
          "Path": "/{**catchall}"
        }
      }
    ]
```

Tratcher:
Discussion notes from the above feedback: We can keep the config structure as is and update the Backend and Endpoint classes to have a field for their IDs so it's more consistent with Routes when you're working with them in code. We just have to do a little fixup after the default config binding to set the id fields.

That should be the last item to address in this config epic.

PR #42, opened 2020-04-03
Restructure backend/endpoint config, config reload

Tratcher:
#9 Restructure backend/endpoint config as discussed in https://github.com/microsoft/reverse-proxy/blob/master/docs/config.md#backend-configuration, nesting endpoints under their backends, and moving from list to object/dictionary formats. After Consulting with David I also took the opportunity to simplify the internal object model to match.

Also moved the config reload logic into the product assemblies.


samples/ReverseProxy.Sample/Startup.cs
```diff
@@ -30,12 +29,7 @@ public Startup(IConfiguration configuration)
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddControllers();
-            services.AddReverseProxy();
-
-            // The following 2 lines are all that we need to react to config changes on the fly.
-            // You can then change appsettings.json on disk and we will apply the new configs without a restart.
-            services.Configure<ProxyConfigRoot>(_configuration.GetSection("ReverseProxy"));
-            services.AddHostedService<ProxyConfigApplier>();
+            services.AddReverseProxy().LoadFromConfig(_configuration.GetSection("ReverseProxy"), reloadOnChange: true);
```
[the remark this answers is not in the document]
Tratcher: > This brings up a question for services like Antares, we probably don't want to fetch the entire route table from the db, but instead have a pull and cache for each hostname model. Is that basically a replacememt router?

Hmm, that doesn't fit well into the current model. I'll file a separate issue to evaluate that.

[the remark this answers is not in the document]
Tratcher: @samsp-msft well done, you broke it. I've filed https://github.com/microsoft/reverse-proxy/issues/46 for that scenario. I think supporting it would require a much different architecture.

src/ReverseProxy.Core/Configuration/ProxyConfigLoaderOptions.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Core.Configuration
+{
+    internal class ProxyConfigLoaderOptions
+    {
+        public bool ReloadOnChange { get; set; }
```
halter73, line 8: Why not add this to ProxyConfigOptions?

src/ReverseProxy.Core/Configuration/ProxyConfigLoader.cs
```diff
@@ -10,73 +10,77 @@
...
 
         public Task StartAsync(CancellationToken cancellationToken)
         {
-            // Nothing to start
-            return Task.CompletedTask;
+            if (_options.Value.ReloadOnChange)
+            {
+                _subscription = _proxyConfig.OnChange((newConfig, name) => _ = ApplyAsync(newConfig));
+            }
+            return ApplyAsync(_proxyConfig.CurrentValue);
```
halter73, line 74: I know this isn't new, but could there be overlapping calls to ApplyAsync? Should we use a semaphore or something?

Tratcher: In theory there could be multiple calls, but everything ApplyAsync calls is already heavily guarded. The config objects all deep clone, the route and backeds repos lock on set and get, and the various collections used by ApplyConfigurationsAsync also lock. It's not an efficient process, but hopefully your config isn't changing that frequently.

src/ReverseProxy.Core/Configuration/ProxyConfigLoaderOptions.cs
Tratcher: Hmm, we could. It seems a little out of place when the rest of ProxyConfigOptions are bound from config. I'll try.

PR #47, opened 2020-04-10
Augment code from config

Tratcher:
This provides a series of APIs that let you run a configuration action when all or a specific route/backend are loaded. The actions rerun if the config is reloaded. The main scenario covered so far is applying global or targeted policies and reacting to additional config metadata. This system is not tightly coupled to our config file datasource, it could be used with other custom sources as well.

What's not covered here is injecting middleware on a per route or backend basis. That needs to be designed differently because you don't want to rebuild a full middleware pipeline each time config changes. One option is to put the relevant middleware into the main pipeline and have them enable/disable per request based on endpoint matadata. We can also use a UseWhen wrapper to enable/disable middleware that are not endpoint aware.


samples/ReverseProxy.Sample/Startup.cs
```diff
@@ -30,7 +33,39 @@ public Startup(IConfiguration configuration)
...
             services.AddControllers();
-            services.AddReverseProxy().LoadFromConfig(_configuration.GetSection("ReverseProxy"), reloadOnChange: true);
+            services.AddMemoryCache();
+            services.AddReverseProxy()
+                .LoadFromConfig(_configuration.GetSection("ReverseProxy"), reloadOnChange: true)
+                .ConfigureBackendDefaults((id, backend) =>
+                {
+                    backend.HealthCheckOptions ??= new HealthCheckOptions();
+                    // How to use custom metadata to configure backends
+                    if (backend.Metadata?.TryGetValue("CustomHealth", out var customHealth) ?? false
```
davidfowl, line 43: Is this the API? Looks hard to use.

Tratcher: Metadata? It's an `IDictionary<string, string>` which conveniently binds from config. We have a lot of experience adding sugar to IDictionary, but the underlying structure does what we need.

davidfowl: What do nested configuration sections look like?

samples/ReverseProxy.Sample/Startup.cs
```diff
@@ -30,7 +33,39 @@ public Startup(IConfiguration configuration)
...
+                        && string.Equals(customHealth, "true", StringComparison.OrdinalIgnoreCase))
+                    {
+                        backend.HealthCheckOptions.Enabled = true;
+                    }
+                })
+                .ConfigureBackend("backend1", backend =>
```
Kahbazi, line 49: What if "backend1" does not exists in config? Should this api creates the backend?

samples/ReverseProxy.Sample/Startup.cs
Tratcher: Hmm, looks config can't bind nested sections into a `<string, string>` structure, which seems odd because that's config's data model.
```
System.InvalidOperationException: 'Cannot create instance of type 'System.String' because it is missing a public parameterless constructor.'
```

Speaking of sugar, you can convert the Metadata object to a config object to take advantage of all of our config sugar.
```
var config = new ConfigurationBuilder().AddInMemoryCollection(backend.Metadata).Build();
backend.HealthCheckOptions.Enabled = config.GetValue<bool>("CustomHealth");
```

davidfowl: Why not expose IConfiguration here directly?

samples/ReverseProxy.Sample/Startup.cs
Tratcher: This is designed strictly as an augmentation mechanic, not a creation one. While we may add API to create these objects, I don't think we should mix it with these configuration APIs.

Create-if-it-doesn't-exist behavior could have surprising side effects. Consider if I had "backend1" defined in my data source and then used this config to change one setting. Then I remove "backend1" from my data source but it's still here in the code. I'd end up with a backend object missing most of its configuration.

Tratcher: The other problem is that these callbacks only run in response to some external event (e.g. config re/load), you can't invoke them when you want or change these callbacks after startup.

A code API for adding/removing backends and routes would need to control when those happened.

samples/ReverseProxy.Sample/Startup.cs
davidfowl: As in backend.Metadata is an IConfiguration

Tratcher: Configuration can't bind to an IConfiguration property.

Also, we'll have data sources that aren't config based (e.g. database).

davidfowl: IConfiguration is a an abstraction over a hierarchy of name value pairs, it doesn't need to be config backed. The current API is hard to use and doesn't intrinsically support nesting. This seems extremely limiting.

Tratcher: Sure, but this is the limit of what automatic config binding supports. To change this we're going to need custom config bindings.

You also have this problem:
https://github.com/microsoft/reverse-proxy/blob/a7b44f66651571846bf3bcb1b0bdbdee5c5828e8/src/ReverseProxy.Core/Abstractions/BackendDiscovery/Contract/Backend.cs#L63
This entire object model is designed around being able to be deeply cloned to ensure atomic updates.

Tratcher: https://github.com/microsoft/reverse-proxy/issues/121

halter73: > Configuration can't bind to an IConfiguration property.

Can we fix this? It would be nice to expose an IConfigurationSection like we do in Kestrel.

Tratcher: We'll investigate as part of #121.

src/ReverseProxy.Core/Configuration/DependencyInjection/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -48,5 +52,233 @@ public static IReverseProxyBuilder LoadFromConfig(this IReverseProxyBuilder buil
...
         }
+
+        /// <summary>
+        /// Provides a configuration action that runs on all routes each time the configuration is generated. This can be called more than once and
+        /// the Actions are run in the given order.
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="configure"></param>
+        /// <returns></returns>
+        public static IReverseProxyBuilder ConfigureRouteDefaults(this IReverseProxyBuilder builder, Action<ProxyRoute> configure)
```
halter73, line 63: Should we be filling out the param and returns doc comments?

src/ReverseProxy.Core/Configuration/DependencyInjection/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -48,5 +52,233 @@ public static IReverseProxyBuilder LoadFromConfig(this IReverseProxyBuilder buil
...
+        }
+
+        /// <summary>
+        /// Provides a configuration action that runs on all backends each time the configuration is generated. This can be called more than once and
+        /// the Actions are run in the given order.
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="configure"></param>
+        /// <returns></returns>
+        public static IReverseProxyBuilder ConfigureBackendDefaults(this IReverseProxyBuilder builder, Action<string, Backend> configure)
```
halter73, line 177: Without looking through the implementation it's not clear why the Action takes a string. It looks like its' for the Backend id. Would it be better to put an id property on Backend similar to what we do for ProxyRoute?

halter73: Also, should these configuration callbacks by async?

davidfowl: This is why callbacks suck.

src/ReverseProxy.Core/Abstractions/Config/IConfigErrorReporter.cs
```diff
@@ -1,6 +1,8 @@
-﻿// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
+
 namespace Microsoft.ReverseProxy.Core.Abstractions
```
davidfowl, line 6: Hopefully we plan to revisit all of these public APIs and make them all internal.

Tratcher: We'll see. Our extensibility story is going to require a lot of stuff to be public.

davidfowl: As long as its thoughtful. It seems everything is public now and has had no API review. I want to be intentional about what is public (like what we did in 3.0) rather than have extension methods on random types exposed from this library.

src/ReverseProxy.Core/Configuration/DependencyInjection/ReverseProxyServiceCollectionExtensions.cs
Tratcher: > Without looking through the implementation, it's not clear why the Action takes a string. It looks like it's for the Backend id. Would it be better to put an id property on Backend similar to what we do for ProxyRoute?

This is a reflection of the asymmetry we have in the config structures. Backends is a set of named objects (with the names on the outside), but routes are an array of objects where the name is a field. There was some concern that routes needed to be specifically ordered because of how route matching works, but backends weren't considered ordered and named objects made the config cleaner. We haven't yet decided if consistency is needed here.
```
"Backends": {
      "backend1": {
        "LoadBalancing": {
          "Mode": "Random"
        },
        "Endpoints": {
          "backend1/endpoint1": {
            "Address": "https://localhost:10000/"
          },
          "backend1/endpoint2": {
            "Address": "http://localhost:10010/"
          }
        }
      },
      "backend2": {
        "Endpoints": {
          "backend2/endpoint1": {
            "Address": "https://localhost:10001/"
          }
        }
      }
    },
    "Routes": [
      {
        "RouteId": "backend1/route1",
        "BackendId": "backend1",
        "Match": {
          "Methods": [ "GET", "POST" ],
          "Host": "localhost",
          "Path": "/{**catchall}"
        }
      }
    ]
```

Tratcher: > Also, should these configuration callbacks be async?

> This is why callbacks suck.

- The callbacks are called from an async code path in DynamicConfigBuilder so it is possible for them to be async.
- It's unclear if this scenario benefits from being async. You're applying policies to available config objects, but maybe you need to pull that policy info from an async source?
- If it's async then it should also be cancelable and that bloats the signature.
- Callbacks do kindof suck.

While writing code like this in Startup has its demo appeal, maybe it's not a good scalable solution? What if I throw all of this out and resolve an IEnumerable of filters from DI instead?

```
public interface IProxyConfigFilter
{
  Task ConfigureBackendAsync(string id, Backend backend, CancellationToken cancel) => Task.CompletedTask;
  Task ConfigureRouteAsync(ProxyRoute route, CancellationToken cancel)  => Task.CompletedTask;
}
```
Yes I'm using DIMs so you only have to implement the one you want.

The biggest difference here is that it wouldn't filter by backend/route id for you, you'd have to check that yourself. Unless we made those optional interface properties?
```
public interface IProxyConfigFilter
{
  string ForBackendId => null;
  string ForRouteId => null;
  Task ConfigureBackendAsync(string id, Backend backend, CancellationToken cancel) => Task.CompletedTask;
  Task ConfigureRouteAsync(ProxyRoute route, CancellationToken cancel)  => Task.CompletedTask;
}
```

Thoughts?

Those Startup ConfigureBackend/Route/Defaults could still be implemented on top of this, but I'd rather replace them all with a singleton service registration:
```
IReverseProxyBuilder.AddConfigurationFilter<TFilter>() where TFilter is IProxyConfigFilter;
```

halter73: I prefer the interface to the callbacks.

We should find a way to add the backend id to the Backend type despite the backing configuration structure, but the interface having named parameters makes not having this not as bad.

I prefer just checking the ids in the ConfigureBackend/RouteAsync methods rather than having the ForBackend/RouteId properties.

I wonder if this should be two interfaces instead. That negates the need for DIMs. It complicates registering a single service instance for both interfaces, but that would still be possible, and I'm not sure how necessary that is anyway.

Tratcher: I ended up removing the DIMs, they have usability issues (https://github.com/dotnet/roslyn/issues/36505).

I want to keep the methods together on one interface since I expect it to be more common to have a single filter that covers both backends and routes than to have multiple filters for one or the other.

Tratcher:
How does that look?

src/ReverseProxy.Core/Configuration/DependencyInjection/ReverseProxyServiceCollectionExtensions.cs
Tratcher: Lets take the discussion of backend id back to #9.

src/ReverseProxy.Core/Configuration/DependencyInjection/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -36,10 +38,7 @@ public static IReverseProxyBuilder AddReverseProxy(this IServiceCollection servi
         /// <summary>
         /// Loads routes and endpoints from config.
         /// </summary>
-        /// <param name="builder"></param>
-        /// <param name="config"></param>
-        /// <param name="reloadOnChange"></param>
-        /// <returns></returns>
+        /// <param name="reloadOnChange">Indicates if the proxy config should be reloaded when the config file is changed. The default is true.</param>
```
halter73, line 41: I think reloadOnChange should be false by default maintain consistency our other reloadOnChange parameters. We can update samples/templates to pass in true. It looks ReverseProxy.Sample already does this.

src/ReverseProxy.Core/Service/Config/DynamicConfigBuilder.cs
```diff
@@ -47,40 +51,80 @@ public async Task<Result<DynamicConfigRoot>> BuildConfigAsync(IConfigErrorReport
...
+                    foreach (var filter in _filters)
+                    {
+                        await filter.ConfigureBackendAsync(id, backend, cancellation);
+                    }
+
+                    configuredBackends[id] = backend;
+                }
+                catch (Exception ex)
+                {
+                    errorReporter.ReportError(ConfigErrors.ConfigBuilderBackendException, id, "An exception was thrown from the configuration callbacks.", ex);
```
halter73, line 72: Does this swallow and log an error if a filter throws at startup? It seems better to let the exception bubble up to the caller if it's during server startup rather than config reload like Kestrel does.

test/ReverseProxy.Core.Tests/Service/Config/DynamicConfigBuilderTests.cs
```diff
@@ -138,61 +155,138 @@ public async Task BuildConfigAsync_ValidRoute_Works()
...
+        }
+
+        [Fact]
+        public async Task BuildConfigAsync_ConfigFilterBackendActionThrows_BackendSkipped()
+        {
+            var errorReporter = new TestConfigErrorReporter();
+            var configBuilder = CreateConfigBuilder(CreateOneBackend(), new TestRoutesRepo(),
+                proxyBuilder =>
+                {
+                    proxyBuilder.AddProxyConfigFilter<BackendAndRouteThrows>();
```
halter73, line 233: Are there tests for multiple filters to verify throwing prevents the execution of subsequent filters?

test/ReverseProxy.Core.Tests/Service/Config/DynamicConfigBuilderTests.cs
```diff
@@ -138,61 +155,138 @@ public async Task BuildConfigAsync_ValidRoute_Works()
         }
 
         [Fact]
-        public async Task BuildConfigAsync_RouteParseError_SkipsRoute()
```
halter73, line 141: Why did this get removed?

test/ReverseProxy.Core.Tests/Service/Config/DynamicConfigBuilderTests.cs
```diff
@@ -138,61 +155,138 @@ public async Task BuildConfigAsync_ValidRoute_Works()
...
+        public async Task BuildConfigAsync_RouteValidationError_SkipsRoute()
         {
-            // Arrange
             var errorReporter = new TestConfigErrorReporter();
-            Mock<IBackendsRepo>()
-                .Setup(r => r.GetBackendsAsync(It.IsAny<CancellationToken>()))
-                .ReturnsAsync(new Dictionary<string, Backend>());
+            var route1 = new ProxyRoute { RouteId = "route1", Match = { Host = "invalid host name" }, Priority = 1, BackendId = "backend1" };
```
halter73, line 161: Might be worth verifying that validation happens after filters are applied.

src/ReverseProxy.Core/Configuration/DependencyInjection/ReverseProxyServiceCollectionExtensions.cs
Tratcher: Consistency aside, I think the expectation for a config based proxy is to reload by default. @samsp-msft?

samsp-msft: Why wouldn't it reload on a change? If not then what will trigger the reload? I could see some kind of timestamp being put in the config for when to apply changes - so it can be coordinated with other systems, but at that point I would suggest its better for the config to be pushed in programatically rather than from a file.

src/ReverseProxy.Core/Service/Config/DynamicConfigBuilder.cs
Tratcher: IConfigErrorReporter is provided by the caller, not as a service. It's purpose is to collect all config errors before acting on them. The current implementation up in ProxyConfigLoader does log and swallow.

src/ReverseProxy.Core/Configuration/DependencyInjection/ReverseProxyServiceCollectionExtensions.cs
Tratcher: The config can contain data other components than the proxy. If you expect your proxy config to remain stable but you still want to reload config for other elements (logging) then you could opt out of proxy config reload for efficiency reasons.

But yes, this was intentionally an opt-out rather than an opt-in. @halter73 would you rather I rename the parameter to that effect? `SuppressReload = false`? I'd rather avoid a double negative.

test/ReverseProxy.Core.Tests/Service/Config/DynamicConfigBuilderTests.cs
Tratcher: It was a leftover from the config route parser we removed. It wasn't testing anything.

test/ReverseProxy.Core.Tests/Service/Config/DynamicConfigBuilderTests.cs
Tratcher: I'm counting the errors thrown (single in this case), so what if I added the same throwing filter twice and made sure I only got one error?

test/ReverseProxy.Core.Tests/Service/Config/DynamicConfigBuilderTests.cs
Tratcher: Added

Tratcher:
Rebased with updates

Issue #49, opened 2020-04-10
How to break up the proxy monolith


Tratcher:
Today routing hands the request off to the ProxyInvoker. This orchestrator is a bit monolithic and has limited pluggability. It is given a list of healthy endpoints from the health service, invokes a load balancer service to resolve an endpoint, and then invokes the proxy logic.

In design meetings we've proposed the concept of a pipeline to allow developers to insert customizations into this process. However, there's a specific service contract between each of these components, not a generic contract between all of them, and it's not well suited to making this a generic pipeline.

Options:
A) Try to refactor this to a more generic object model? (e.g. middleware)
B) Insert targeted extensibility at specific points? (e.g. the authentication handler events model)

Additionally, is it possible to do this customization per route? Or is there a central pipeline that can react to endpoint specific metadata?

Issue #21, opened 2020-03-19
Config-based request and response transformation


davidni:
In Island Gateway we implemented the notion of `Transformations`, which takes inspiration from Traefik's middlewares (though we're not calling them middlewares for obvious reasons). See e.g. [Traefik's StripPrefix](https://docs.traefik.io/middlewares/stripprefix/).

A user can specify e.g. `Transformations=StripPrefix('/api'), AddPrefix('/v2')`, which results in the listed processing steps being applied sequentially at runtime.

The ability to specify simple transformations like these through config is a requirement for us.

Example of a route using transformations:
```
Rule            = Host('example.com') && Path('/myservice/{**catchall}')
Transformations = StripPrefix('/myservice')
```

(this is a feature we already implemented in Island Gateway, filing here as requested to track)

Tratcher:
The verbosity could be reduced:
```
"Transform": [
  { "StripPrefix": "/api/v1/myservice" },
  { "AddPrefix": "/apis" }
  { "AddHeader": { "X-MyHeader": "myVal" } }
  { "StripHeaders": [ "X-SomeUnhelpfulIncomingHeader" ] }
]
```

Tratcher:
This is a space with a lot of prior art, we should do some comparisons.

samsp-msft:
@davidni I suspect I know the reason, but why is this driven by config rather than code? How often are routes created in your world that need this kind of capability? 
The reason I am asking is to try and form a philosophy for what we need to support in config vs writing app code. I think one of the distinguishing features of YARP is that you can create your own project using its libraries as a starting point, and can then add custom logic in code. That's not typically possible in other proxies without re-building the entire proxy from scratch. With YARP you aren't rebuilding the proxy library, just your instantiation of the existing modules, plus your custom glue.

Tratcher:
> How often are routes created in your world that need this kind of capability?

Altering the path is very common, adding or stripping a prefix. Adding and removing headers is also common, usually headers directly related to the proxy activity like x-forwarded-*. I'd expect us to provide built in support for things like x-forwarded-* so manually adding them per route is less of a concern.

Again, this is a case where simple common operations should be possible via config, and advanced ones via code.

Kahbazi:
I will work on this issue if it's ok.

Tratcher:
@Kahbazi it needs some design first. You're welcome to make some proposals on this issue.

Kahbazi:
[...]

Issue #13, opened 2020-03-12
Implement Forwarded Headers


anurse:
* X-Forwarded-Proto
* X-Forwarded-Host
* X-Forwarded-For
* X-Client-Cert
* Forwarded (yes, there's a chicken-and-egg problem but I see no reason we can't break that cycle by implementing it).

Tratcher:
Planning notes: We need a minimal implementation of this for preview1. E.g. automatically append forwarders without any configurability.

davidni:
I can help with the minimal implementation. I already have it coded up to append `X-Forwarded-For`, `X-Forwarded-Host`, `X-Forwarded-Host`, `X-Forwarded-Proto`, as well as [RFC7239](https://tools.ietf.org/html/rfc7239) `Forwarded`.

Because of possible URL transformations, we also add a non-standard header with the full original incoming url.

Tratcher:
I did the minimal work for preview1 in #133. Moving this out of the preview1 milestone for further development.

Tratcher:
Design: I expect this to overlap heavily with the request transformation infrastructure we're designing in https://github.com/microsoft/reverse-proxy/issues/21. We wouldn't necessarily ask you to define these headers in config (though nginx does), but we would use the same infrastructure to actually do the transformations.

Tratcher:
Yes this does align with the transforms, but we don't need another middleware to add transforms per request. We can pre build the transforms when building the config for each route/backend, then we can run them per request with minimal overhead.

damianh:
I think you should also consider supporting `X-Forwarded-PathBase`. It is first class supported in ProxyKit for reasons described in this readme: https://github.com/ProxyKit/HttpOverrides . If this deserves it's own issue, let me know.

Tratcher:
@damianh PathBase makes sense, you can't properly generate public urls without it.

damianh:
@Tratcher Yes am aware of the at-startup approach, however I am suggesting per-request should be supported.

Tratcher:
@damianh sending x-forwarded-PathBase was implemented in YARP as part of #225. Filed https://github.com/dotnet/aspnetcore/issues/23263 for receiving it.

damianh:
Excellent!

PR #50, opened 2020-04-13
Split up the ProxyInvoker into a pipeline model 

Tratcher:
#49 Experimenting with a pipeline model for proxy customization. This creates a single pipeline for all proxy routes. The modules in the pipeline examine the endpoint metadata for route specific configuration.

I was able to split up the ProxyInvoker into three decoupled middleware:
- EndpointInitializerMiddleware: Gets the current healthy endpoints from the route backend and sets AvailableBackendEndpointsFeature.Endpoints.
- LoadBalancingMiddleware: Gets the available endpoints, invokes the ILoadBalancer service to reduce the set to one, and replaces AvailableBackendEndpointsFeature.Endpoints.
- ProxyInvokerMiddleware: Gets the available endpoint (should be singular after load balancing) and proxies to it.

Conceptually this allows us to insert middleware before or after load balancing to vary the behavior. For example we can now add a separate affinity middleware before load balancing. It would pull the affinized endpoint from the available list and replace the list with that single element. This will cause load balancing to naturally short circuit since there's only one option.

The other middleware to consider is for request transformations. It might be easier to transform parts of the the request like the url components using the HttpContext object model than HttpRequestMessage since HttpContext treats the url as decomposed fields. 

Is this a viable approach? Does it fit with our design proposals?


src/ReverseProxy.Core/Middleware/EndpointInitializerMiddleware.cs
```diff
@@ -0,0 +1,74 @@
...
+                throw new InvalidOperationException($"ASP .NET Core Endpoint wasn't set for the current request. This is a coding defect.");
+            }
+
+            var routeConfig = aspNetCoreEndpoint.Metadata.GetMetadata<RouteConfig>();
+            if (routeConfig == null)
+            {
+                throw new InvalidOperationException($"ASP .NET Core Endpoint is missing {typeof(RouteConfig).FullName} metadata. This is a coding defect.");
+            }
+
+            var backend = routeConfig.BackendOrNull;
```
davidfowl, line 45: IS this really the property name 😄

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        private readonly ILogger _logger;
+        private readonly IOperationLogger _operationLogger;
+        private readonly ILoadBalancer _loadBalancer;
+        private readonly RequestDelegate _next;
+
+        public LoadBalancingMiddleware(
+            RequestDelegate next,
+            ILogger<LoadBalancingMiddleware> logger,
+            IOperationLogger operationLogger,
```
davidfowl, line 29: What is an IOperationLogger?

Tratcher: https://github.com/microsoft/reverse-proxy/blob/61055e1dd468ab90083ec4f5fddbce603e11ace5/src/ReverseProxy.Common/Telemetry/TextOperationLogger.cs#L34-L48
It's a log scoping feature that gets you start and stop logs for an arbitrary code block, as well as elapsed time and errors.

halter73: It also has an IOperationContext for adding dimensions to the logs, but that's apparently not implemented yet.

src/ReverseProxy.Core/Middleware/ProxyInvokerMiddleware.cs
```diff
@@ -2,95 +2,76 @@
...
 
-            var endpoint = _operationLogger.Execute(
-                "ReverseProxy.PickEndpoint",
-                () => _loadBalancer.PickEndpoint(dynamicState.HealthyEndpoints, dynamicState.AllEndpoints, in loadBalancingOptions));
-
-            if (endpoint == null)
+            if (endpoints.Count > 1)
             {
-                throw new ReverseProxyException($"No available endpoints.");
+                _logger.LogDebug("More than one endpoint available, choosing the first.");
```
halter73, line 66: If the proxy invoker isn't going to at least round robin, this should at least be logged with higher severity. Having more than one endpoint at this stage is certainly misconfiguration, since the load balancing middleware should always pick a single backend.

src/ReverseProxy.Core/Middleware/AvailableBackendEndpointsFeature.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Microsoft.ReverseProxy.Core.RuntimeModel;
+
+namespace Microsoft.ReverseProxy.Core.Middleware
+{
+    public class AvailableBackendEndpointsFeature
+    {
+        public IReadOnlyList<EndpointInfo> Endpoints { get; set; }
```
halter73, line 11: Should this be an interface? I get that it's really simple, but usually we've tried to stick with interfaces for "features" to get maximum flexibility.

halter73:
> Is this a viable approach? Does it fit with our design proposals?

I think so.

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
```diff
@@ -0,0 +1,69 @@
...
+            _next = next;
+            _logger = logger;
+            _operationLogger = operationLogger;
+            _loadBalancer = loadBalancer;
+        }
+
+        public Task Invoke(HttpContext context)
+        { 
+            var backend = context.Features.Get<BackendInfo>() ?? throw new InvalidOperationException("Backend unspecified.");
+            var endpointsFeature = context.Features.Get<AvailableBackendEndpointsFeature>();
```
samsp-msft, line 46: Are we using features here to store the endpoints, or as a way to supply different features to the pipeline. How would an affinity middleware appear here?

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Microsoft.ReverseProxy.Core.Middleware
+{
+    /// <summary>
+    /// Load balances across the available endpoints.
+    /// </summary>
+    public class LoadBalancingMiddleware
+    {
+        private readonly ILogger _logger;
+        private readonly IOperationLogger _operationLogger;
+        private readonly ILoadBalancer _loadBalancer;
```
samsp-msft, line 23: This may be a stylistic thing for how asp.net stuff is written, is this abstracted out for replacement, or just so the actual loadbalancing code doesn't have to have all the cruft to play as middleware?

src/ReverseProxy.Core/Middleware/AvailableBackendEndpointsFeature.cs
Tratcher: I've added the interface in a later commit.

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
Tratcher: It was abstracted out for replacement in the prior model and I didn't want to change that layering yet. Once we work out the overall pipeline model we can revisit the factoring of individual components.

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
Tratcher: Features are the default storage for strongly typed per-request data. Something early in the pipeline seeds the feature with data and later components can consume it, modify it, etc..

An affinity middleware would work a lot like the load balancing one, taking the whole list of endpoints, finding one that matched it's affinity criteria, and resetting the list so it only contained that endpoint. Establishing affinity is a slightly different trick we'll need to work out. The middleware will need to hook a response starting event, check which endpoint was used, and store the affinity accordingly. We need to formalize how it can tell which endpoint was used.

src/ReverseProxy.Core/Middleware/ProxyInvokerMiddleware.cs
Tratcher: Warning?

Tratcher: > since the load balancing middleware should always pick a single backend.

It's now possible to leave out the load balancing middleware, how can this component tell if that was deliberate?

halter73: I think a warning is fine. Random endpoint selection here might also be a good idea, but I mostly just want to increase the log severity.

halter73: I think it's safe to say that leaving multiple endpoints in the AvailableBackendEndpointsFeature by the time the ProxyInvokerMiddleware executes is a misconfiguration. If you really want to to just pick the first endpoint, you should have middleware that explicitly does that.

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
samsp-msft: Particularly if the connection was retried. Although its probably embedded in the request/response from httpClient.

samples/ReverseProxy.Sample/Startup.cs
```diff
@@ -44,7 +47,18 @@ public void Configure(IApplicationBuilder app)
...
-                endpoints.MapReverseProxy();
+                endpoints.MapReverseProxy(proxyPipeline =>
+                {
+                    proxyPipeline.UseProxyLoadBalancing();
+                    // Customize the request before forwarding
+                    proxyPipeline.Use((context, next) =>
+                    {
+                        var connection = context.Connection;
+                        context.Request.Headers.AppendCommaSeparatedValues("X-Forwarded-For",
+                            new IPEndPoint(connection.RemoteIpAddress, connection.RemotePort).ToString());
```
samsp-msft, line 58: We need to think about the headers in both directions - there probably should be an intermediary collection from the request before it gets given to httpClient, and the same for the response the other way.
Defining the contract for what data can be acted on is going to be an important part of the pipeline. should that get added as comments to the original issue?

Tratcher: Yes, transformations are an independent issue we'll need to work out. I added it here only as an easy example. https://github.com/microsoft/reverse-proxy/issues/13 tracks some well known headers and https://github.com/microsoft/reverse-proxy/issues/21 tracks config based transformations.

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
Tratcher: Yeah, retry is going to be interesting. We'll need to see what component would handle errors, reset state, and retry.

samples/ReverseProxy.Sample/Startup.cs
halter73: I think the question is whether modifying `context.Request.Headers` is the way to modify the request headers sent by the IHttpProxy. Or whether the request headers should come from some separate collection that's originally copied from `context.Request.Headers`. That way you can compare the original request headers to what will be sent to the backend at any stage of the pipeline. It's probably not as important for response headers since that collection starts out empty, but it might be good for consistency.

Tratcher: Copies cause a lot of overhead, but it would may retrying easier. I think the url transformations are more interesting. That said, let's move this discussion to the transformation issues.

src/ReverseProxy.Core/Middleware/EndpointInitializerMiddleware.cs
Kahbazi: Could this project use C# 8 nullable reference types?

samples/ReverseProxy.Sample/Startup.cs
```diff
@@ -44,7 +46,18 @@ public void Configure(IApplicationBuilder app)
...
```
halter73, line 57: Isn't X-Forwarded-For usually just contain the client IP address and not the ephemeral port it used?

src/ReverseProxy.Core/Middleware/EndpointInitializerMiddleware.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            if (dynamicState.HealthyEndpoints.Count == 0)
+            {
+                _logger.LogInformation($"No available healthy endpoints.");
+                context.Response.StatusCode = 503;
+                return Task.CompletedTask;
+            }
+
+            context.Features.Set(backend);
+            context.Features.Set<IAvailableBackendEndpointsFeature>(new AvailableBackendEndpointsFeature() { Endpoints = dynamicState.HealthyEndpoints });
```
halter73, line 59: Should we add a feature for all backend endpoints healthy or not? Or is it rare enough for middleware to be concerned about this, that we're ok with making middleware authors that want this to manually retrieve it from RouteConfig?

src/ReverseProxy.Core/Service/DynamicEndpoints/RuntimeRouteBuilder.cs
```diff
@@ -69,5 +70,13 @@ public RouteConfig Build(ParsedRoute source, BackendInfo backendOrNull, RouteInf
 
             return newRouteConfig;
         }
+
+        // This indirection is needed because on startup the routes are loaded from config and built before the
+        // proxy pipeline gets built.
+        private Task Invoke(HttpContext context)
```
halter73, line 76: Should we always use this incase _pipeline isn't null in Build(), but changes later?

src/ReverseProxy.Core/Middleware/EndpointInitializerMiddleware.cs
Tratcher: That information is already available via the Backend or RouteConfig objects, nor would we expect it to be mutated in the pipeline like IAvailableBackendEndpointsFeature will be.

src/ReverseProxy.Core/Service/DynamicEndpoints/RuntimeRouteBuilder.cs
Tratcher: Once _pipeline is set we don't allow it to be reset to null.

Issue #72, opened 2020-04-20
Implement Round-Robin Load Balancing


anurse:
Load balance by taking the "next" backend in the list for each incoming request. We wrap around to the start of the list when we run out of backends.

Since requests will likely be highly-concurrent, there's some complexity here in how the state is maintained. Also, since backends can change, we'll need to consider how this algorithm behaves in that case. Naively, this would be a simple atomic counter that the load balancer does an atomic increment-and-return, then modulus by the number of backends to figure out which backend to use.

Tratcher:
The challenge in the current object model is finding somewhere to store this per-backend state. Health checks also throw this off since they keep churning the list of available backends.

samsp-msft:
I have started writing an implementation for this.

samsp-msft:
I talked to @anurse about the design, and what we came up with was different from what I was originally thinking. I was originally thinking of having a cache based on a signature for the collection of endpoints storing the counter for a mod operation.
I we think that the round robin should be unique to each cluster of back end endpoints, then it should be stored as part of the Backend.

Implementation wise, would it make sense to have a feature collection on the back end, similar to HttpContext, and proxy code can then stash/fetch into that as needed.
Do we also want to have a dictionary so that config can specify parameters for routes and backends, that are later picked up by custom modules?

Tratcher:
Yes there will be to be some way to track state in the backend, and that will need to be thread safe.

samsp-msft:
Should this just be a counter property on BackEndInfo, or should we have a dictionry? Do we have a thread-safe dictionary?

Tratcher:
ConcurrentDictionary might work, we'll see if we need to go that far. We'll also need to account for the fact that the Backend objects are torn down and rebuilt if there's a config change so any state you add to them is ephemeral. That might be OK to reset RoundRobin on a config change, they're not that common.

There's a type in this repo called AtomicCounter that would be useful for tracking a simple incrementing value. If we're not overly concerned about the endpoints list being unstable then the counter might be all we need.

All of this being said, RoundRobin isn't much better than Random, I'm not sure it's worth adding.

anurse:
> ConcurrentDictionary might work, we'll see if we need to go that far. We'll also need to account for the fact that the Backend objects are torn down and rebuilt if there's a config change so any state you add to them is ephemeral. That might be OK to reset RoundRobin on a config change, they're not that common.

[...]

PR #78, opened 2020-04-21
Basic load balancing

### LINKED ISSUE #73: Implement Random Load-Balancing
This is one of the simplest load-balancers. Select a random backend from the set of healthy backends and route to it. It can be surprisingly effective, so it's a useful algorithm to have. It's also trivial to implement so it gives us a way to exercise our abstractions.

### LINKED ISSUE #74: Implement "Least-Requests" load balancing
Load balance to the backend which has the fewest on-going requests. This requires some state be tracked about the currently load to each backend (though @Tratcher indicated we already have that state).

### LINKED ISSUE #75: "The Power of Two Choices" Load-Balancing Algorithm
This falls somewhere between Random (#73) and Least-Requests (#74). It tends to be easier to implement across a cluster of proxies.

Given that load-balancers don't share state about the load to each backend (this would be very costly), a proxy instance is making it's decision based on the number of outstanding requests **it** has to the backends, and not the number of requests **all** instances have. This means that often multiple proxy instances can see the same nodes as under-utilized and send a storm of requests to it (if the incoming traffic is randomly balanced, as tends to be the case).

The "Power of Two Choices" model (discussed in some sources below) adds a degree of randomness by first selecting two random backends and **then** making a "least-outstanding-requests" decision between those two. This helps to prevent these storms of traffic. This doesn't generally perform better than Least Requests/Connections in a single-node proxy deployment, but it can be better than Random in situations where each node doesn't have perfect knowledge of the load

#### Sources
* [NGINX and the “Power of Two Choices” Load-Balancing Algorithm](https://www.nginx.com/blog/nginx-power-of-two-choices-load-balancing-algorithm/)
* [Test Driving “Power of Two Random Choices” Load Balancing](https://www.haproxy.com/blog/power-of-two-load-balancing/)
* [Michael Mitzenmacher. 2001. The Power of Two Choices in Randomized Load Balancing. IEEE Trans. Parallel Distrib. Syst. 12, 10 (October 2001), 1094–1104. DOI:https://doi.org/10.1109/71.963420](https://www.eecs.harvard.edu/~michaelm/postscripts/tpds2001.pdf)

Tratcher:
I'd added the config model to select a load balancing strategy per backend. I've also implemented most of the strategies.

Fixes #75 - The power of two choices - This is the default, it has the best cost-benefit trade off.
Fixes #74 - Least requests - Has the cost of examining all endpoints.
Fixes #73 - Random - Has degenerate cases where it make assign work to busy endpoints.
- First - This was the only one available before. It doesn't seem like actual load balancing, but it may be useful in the scenario where you have two endpoints and you only want to switch to the second one if the first goes down. First handles that because load balancing only considers healthy endpoints.

#72 - Round-Robin - This is the only one I skipped. It's not clearly better than any of the above strategies, and it's difficult to implement over an unstable list of endpoints. Nor do we have a place to store the per-backend state needed to track which endpoint to select next.

Note this will have merge conflicts with https://github.com/microsoft/reverse-proxy/pull/50.


src/ReverseProxy.Core/Service/RuntimeModel/LoadBalancingMode.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Core.RuntimeModel
+{
+    /// <summary>
+    /// Load balancing strategies for endpoint selection.
+    /// </summary>
+    public enum LoadBalancingMode
```
davidfowl, line 9: Custom?

Tratcher: Custom load balancing would require a few different things.
A) You'd have to replace/wrap the ILoadBalancer service or middleware.
B) Would you only want that for some backends?
C) What if you have more than one type of custom load balancing for different backends?

I did a quick test and you can put `"Mode": "-1"` into config and it successfully binds to the enum. That means you don't need a `custom` value, people can just define their own entries.

src/ReverseProxy.Core/Service/Proxy/LoadBalancer.cs
```diff
@@ -13,25 +14,55 @@ namespace Microsoft.ReverseProxy.Core.Service.Proxy
     /// </summary>
     internal class LoadBalancer : ILoadBalancer
     {
+        private readonly IRandom _random;
```
stephentoub, line 17: Not related to this PR, but since I hadn't seen this IRandom before here... System.Random exposes virtuals that can be overridden.  Do we really need `IRandom`?

Tratcher: This was existing infrastructure I was taking advantage of. On that note, what are your thoughts on ThreadStaticRandom? https://github.com/microsoft/reverse-proxy/blob/9e64458cd08ba2ca328c76213257437fa18c9174/src/ReverseProxy.Utilities/Utils/ThreadStaticRandom.cs#L11

stephentoub: It shouldn't be necessary with .NET Core.  It's existing in order to initialize a per-Thread Random instance with a random seed, which could have been necessary on .NET Framework due to Random's default ctor picking a seed based on Environment.TickCount:
https://referencesource.microsoft.com/#mscorlib/system/random.cs,53
And it looks a lot like:
https://devblogs.microsoft.com/pfxteam/getting-random-numbers-in-a-thread-safe-way/
But in .NET Core, we actually employ a similar randomized-seed scheme:
https://github.com/dotnet/runtime/blob/3275913af8aca0a55f833922d4c6754b0c0dc276/src/libraries/System.Private.CoreLib/src/System/Random.cs#L38
with the seed constructed from a global Random instance:
https://github.com/dotnet/runtime/blob/3275913af8aca0a55f833922d4c6754b0c0dc276/src/libraries/System.Private.CoreLib/src/System/Random.cs#L124-L138
So the ThreadStaticRandom implementation can be condensed to just:
```C#
[ThreadStatic]
private static IRandom t_inst;
public static IRandom Instance => t_inst ??= new RandomWrapper(new Random());
```

src/ReverseProxy.Core/Abstractions/BackendDiscovery/Contract/LoadBalancingOptions.cs
```diff
@@ -1,17 +1,22 @@
...
 // Licensed under the MIT License.
 
+using Microsoft.ReverseProxy.Core.RuntimeModel;
+
 namespace Microsoft.ReverseProxy.Core.Abstractions
 {
     /// <summary>
     /// Load balancing options.
     /// </summary>
     public sealed class LoadBalancingOptions
```
halter73, line 11: Should we have mode-specific options? I could imagine wanting to add a PowerOfNChoices mode where N was some sort of option for example, but I wouldn't want that cluttering up LoadBalancingOptions. Maybe the mode name should be a key in config, and the mode-specific options, if any, could be in the value.

Tratcher: I'd rather not tackle that until we have a concrete need.  I am curious what kind of config would be needed, these algorithms are pretty self-contained. The discussion I saw around PowerOfNChoices was that the benefits dropped off steeply when doing more than 2 choices. If 3 choices was still a good option but more than that wasn't, then I could see just adding a separate enum value for it.

src/ReverseProxy.Utilities/Utils/ThreadStaticRandom.cs
```diff
@@ -10,44 +10,8 @@ namespace Microsoft.ReverseProxy.Utilities
...
-                    var random = new Random(seedForThreadLocalInstance);
-                    currentInstance = new RandomWrapper(random);
-                    _threadLocalRandom = currentInstance;
-                }
-
-                return currentInstance;
-            }
-        }
+        private static Random t_inst;
+        public static Random Instance => t_inst ??= new Random();
```
davidni, line 15: This is bad and leads to bad randomness. If multiple threads that don't have a Random instance yet call this at about the same time (e.g. multiple requests coming in at about the same time), they will all get Random instances with the same seed because default seeding is time-based and not very fine grained. The original code took special care to create more entropy across threads by using a global seed generator (`_globalRandom`)

src/ReverseProxy.Core/Service/Proxy/LoadBalancer.cs
```diff
@@ -13,25 +14,57 @@ namespace Microsoft.ReverseProxy.Core.Service.Proxy
...
+
         public EndpointInfo PickEndpoint(
             IReadOnlyList<EndpointInfo> healthyEndpoints,
             IReadOnlyList<EndpointInfo> allEndpoints,
             in BackendConfig.BackendLoadBalancingOptions loadBalancingOptions)
         {
+            var endpointCount = healthyEndpoints.Count;
+            if (endpointCount == 0)
+            {
+                return null;
```
davidni, line 32: Something to keep in mind is what to do when a large portion of endpoints in a backend are unhealthy. Envoy for example has a `panic mode` where it will prefer to try to route to endpoints it knows to be unhealthy, because it may be better than the alternative of not routing at all. See e.g. [Panic threshold](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/panic_threshold)

src/ReverseProxy.Utilities/Utils/ThreadStaticRandom.cs
stephentoub: This is not the case on .NET Core.

src/ReverseProxy.Core/Service/Proxy/LoadBalancer.cs
```diff
@@ -13,25 +14,57 @@ namespace Microsoft.ReverseProxy.Core.Service.Proxy
...
+                    // Pick two, and then return the least busy. This avoids the effort of searching the whole list, but
+                    // still avoids overloading a single endpoint.
+                    var random1 = _randomFactory.CreateRandomInstance();
+                    var firstEndpoint = healthyEndpoints[random1.Next(endpointCount)];
+                    var secondEndpoint = healthyEndpoints[random1.Next(endpointCount)];
+                    return (firstEndpoint.ConcurrencyCounter.Value <= secondEndpoint.ConcurrencyCounter.Value) ? firstEndpoint : secondEndpoint;
+                case LoadBalancingMode.LeastRequests:
+                    var leastRequestsEndpoint = healthyEndpoints[0];
+                    var leastRequestsCount = leastRequestsEndpoint.ConcurrencyCounter.Value;
+                    for (var i = 1; i < endpointCount; i++)
```
alnikola, line 57: It might be worth storing healthy endpoints in a [Min Heap data structure](https://en.wikipedia.org/wiki/Binary_heap) instead of list when endpointCount is big enough (e.g. > 20). It will allow to find an endpoint with least requests with O(1) complexity in contrast to O(n) for searching in a plain list. Heap can be reconstructed in [BackendProber.ProbeEndpointsAsync](https://github.com/microsoft/reverse-proxy/blob/bc9298322206f35e7d658183464cdf806280143f/src/ReverseProxy.Core/Service/HealthProbe/BackendProber.cs#L119) after each probing iteration.

Tratcher: The tradeoffs being how frequently you have to build and update that structure. Having to update it every time you add and remove a request seems like a lot of overhead.

alnikola: Yes, I understand it. But, do we really want to see an effect of ConcurrencyCounters updates on each request? The current linear search algorithm is not atomic anyways, so it might return endpoint which is not the least loaded one.

Tratcher: Point, it's an approximation of least busy without locking.

Out goal at this point is getting basic implementations we can measure and compare. We'll use that data to decide what to refine later.

davidfowl: @alnikola We should file a bug for this for sure, that's a good idea.

src/ReverseProxy.Utilities/Utils/ThreadStaticRandom.cs
davidni: Nice, thanks for sharing! Sure would be nice if the framework would expose thread safe access to that, then we wouldn't need another ThreadStatic :). The seeding logic already keeps a Random instance for the current thread.

src/ReverseProxy.Core/Service/Proxy/LoadBalancer.cs
Tratcher: Thinking about this more, that kind of structure would be disruptive to the other load balancing algorithms, not to mention unnecessary overhead when you're not using LeastRequests. It's also unclear how it would fit with the winnowing middleware approach.

If the heap were not updated frequently, wouldn't you end up overloading the first node?

davidni: Keep in mind that Power of Two Choices is already O(1) and has no overhead on ConcurrencyCounters updates. It is a good default choice when there is a large number of endpoints.

alnikola: > Thinking about this more, that kind of structure would be disruptive to the other load balancing algorithms, not to mention unnecessary overhead when you're not using LeastRequests.

We can introduce a new abstraction for endpoint collection (e.g. `IEndpointCollection : IReadOnlyList<EndpointInfo>`) exposing GetLeastLoadedEndpoint method with at least two implementations: ListEndpointCollection and MinHeapEndpointCollection. Further, [in the end of probing iteration](https://github.com/microsoft/reverse-proxy/blob/bc9298322206f35e7d658183464cdf806280143f/src/ReverseProxy.Core/Service/HealthProbe/BackendProber.cs#L119) if the current [BackendConfig](https://github.com/microsoft/reverse-proxy/blob/bc9298322206f35e7d658183464cdf806280143f/src/ReverseProxy.Core/Service/HealthProbe/BackendProber.cs#L73).Mode = LeastRequests and endpoint count is big enough, BackendProber will command IEndpointManager to change the underlying collection holding endpoints to new MinHeapEndpointCollection, or to ListEndpointCollection otherwise.
The worst case here is mode changing to LeastRequests from something else just after BackendProber updated the endpoint list, so ListEndpointCollection would have to use a linear search to find the target item. However, it will be the same as it is currently.

> If the heap were not updated frequently, wouldn't you end up overloading the first node?

To avoid overloading, heap update should be trigged by a heuristic based on the real request count [tracked by ProxyInvokerMiddleware](https://github.com/microsoft/reverse-proxy/blob/4520fb1d110f8687647e3b052a40d84d26b2ec7f/src/ReverseProxy.Core/Middleware/ProxyInvokerMiddleware.cs#L86) instead of a simple timer.

Summarily, I will create a separate issue as @davidfowl suggested for this improvement to discuss it separately from this PR.

samsp-msft: Can the same endpoint appear in more than one backend cluster? If so I'm assuming that the load needs to be tracked on a per endpoint basis rather than as part of the backend structure. As the list of backends can change dynamically, managing lists for least loaded is going to be expensive.
This is where the pick the least loaded of 2 is beneficial. I wonder if we should do other algorithms that don't involve sorting all the endpoints - like something that can assess the load in relation to the average, and pick the first that is below average? That way it can drop out on the first good candidate. If that's combined with a round robin seed, so the next search will be based on the last found, then it will spread the load out across the servers.

alnikola: Submitted issue #94

PR #80, opened 2020-04-22
Fix duplex request proxying (especially gRPC)

### LINKED ISSUE #79: Proxying halts or aborts in some configurations (especially gRPC)
### Describe the bug
I will send a PR shortly. The current prpoxying logic does not handle long-running duplex channels effectively:
  * We were not removing ASP .NET Core's limits on incoming request min data rate. That is a DoS protection mechanism that makes sense for normal http requests, but not for duplex or client-streaming channels. Both HTTP1 upgraded connections, as well as the official Microsoft implementation of gRPC server middlewares also remove those limits for incoming requests within some criteria
   * We were not flushing bytes to HttpClient's machinery effectively. HttpClient keeps its own buffers on top of the underlying sockets, causing deadlocks in some situations (client is waiting for bytes that will never come because the server is waiting for bytes that will never come because.... rinse and repeat).

### To Reproduce
* Scenario 1: Attempt proxying a gRPC request that does client-streaming, server-streaming, or duplex. keep the channel open for a while, observe proxy terminates the connection
* Scenario 2; gRPC abort scenarios (e.g. have the gRPC server throw an exception before producing a valid proto result. Observe the client hangs.

davidni:
Fixes #79. At a glance:
   * We were not removing ASP .NET Core's limits on incoming request min data rate. That is a DoS protection mechanism that makes sense for normal http requests, but not for duplex or client-streaming channels. Note that there is precedent to doing what we are doing here. Both HTTP1 upgraded connections, as well as gRPC server middlewares shipped by Microsoft do the same for incoming requests within some criteria
   * We were not flushing bytes to HttpClient's machinery effectively. HttpClient keeps its own buffers on top of the underlying sockets, and we now properly flush bytes when reasonable to do so.


davidni:
This PR is the result of several days of hunting down bugs with proxying gRPC comms. Discalimer: I have not been able to test this in this repo (or even build -- I don't have the preview .NET Core 5 SDK's)  I ported changes I made in our internal implementation "by hand", so expect compile issues and possible runtime issues. @Tratcher would be great if you can help with those as my badnwidth is currently limited to contribute here. Thanks!

anurse:
> Discalimer: I have not been able to test this in this repo (or even build -- I don't have the preview .NET Core 5 SDK's)

Hmm, the build scripts should take care of this automatically. Did you use the instructions in the README to set up for building/testing in VS? I’d like to make sure the repo is accessible as possible for external contributors so feel free to file bugs for issues you encountered trying to build/test the repo.

davidni:
@anurse Thanks! I was worried that this would sully my dev machine, but I see that clearly you put in the effort to make this clean and easy, it worked great. Might be worthwhile making it even clearer in README that the restore script will NOT install anything globally and won't impair previous installations of .NET Core SDK, VS, etc. I pushed a new commit getting a green build + ported over unit tests.

anurse:
Oh yeah, we don’t want to go back to the world of special build shells or VMs to isolate different build environments :). Being well-isolated and safe for any contributor to use without messing with their machine is part of the design goals for the .NET Engineering System ([Arcade](https://github.com/dotnet/arcade)) that we’re using here :).

I’ll make a quick update pass over the README to clarify!

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
```diff
@@ -183,29 +192,41 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
...
             // :::::::::::::::::::::::::::::::::::::::::::::
-            // :: Step 8: Copy response body Downstream ◄-- Proxy ◄-- Upstream
+            // :: Step 7: Copy response body Downstream ◄-- Proxy ◄-- Upstream
             await CopyBodyDownstreamAsync(upstreamResponse.Content, context.Response.Body, proxyTelemetryContext, longCancellation);
 
+            // :::::::::::::::::::::::::::::::::::::::::::::
+            // :: Step 8: Copy response trailer headers and finish response Downstream ◄-- Proxy ◄-- Upstream
+            CopyTrailingHeadersToDownstream(upstreamResponse, context);
+
+            // NOTE: We must call `CompleteAsync` so that Kestrel will flush all bytes to the client.
```
Tratcher, line 215: This doesn't belong in the normal request handling code path, it's very specific to the streaming case and it prevents middleware from adding trailers. This should probably only be called if bodyToUpstreamContent.ConsumptionTask hasn't finished yet.

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
```diff
@@ -415,5 +449,39 @@ private void CopyTrailingHeadersToDownstream(HttpResponseMessage source, HttpCon
                 }
             }
         }
+
+        /// <summary>
+        /// Disable some ASP .NET Core server limits so that we can handle long-running gRPC requests unconstrained.
+        /// Note that the gRPC server implementation on ASP .NET Core does the same for client-streaming and duplex methods.
+        /// Since in Gateway we have no way to determine if the current request requires client-streaming or duplex comm,
```
Tratcher, line 456: @JamesNK

src/ReverseProxy.Core/Util/GRpcProtocolHelper.cs
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.ReverseProxy.Core
+{
+    internal static class GRpcProtocolHelper
```
Tratcher, line 8: This grpc casing is strange.

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
```diff
@@ -133,6 +133,15 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
             Contracts.CheckValue(targetUri, nameof(targetUri));
             Contracts.CheckValue(httpClient, nameof(httpClient));
 
+            // :::::::::::::::::::::::::::::::::::::::::::::
+            // :: Step 0: Disable ASP .NET Core limits for gRPC requests
+            var isIncomingHttp2 = HttpProtocol.IsHttp2(context.Request.Protocol);
+            var isLikelyGrpc = isIncomingHttp2 && GRpcProtocolHelper.IsGRpcContentType(context.Request.ContentType);
```
Tratcher, line 139: @JamesNK this seems overly broad. Is there some way to detect streaming requests? We shouldn't be killing rate and limits for all gRPC requests.

src/ReverseProxy.Core/Util/GRpcProtocolHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        internal const string GrpcContentType = "application/grpc";
+
+        /// <summary>
+        /// Checks whether the provided content type header value represents a gRPC request.
+        /// Takes inspiration from
+        /// <see href="https://github.com/grpc/grpc-dotnet/blob/3ce9b104524a4929f5014c13cd99ba9a1c2431d4/src/Shared/CommonGrpcProtocolHelpers.cs#L26"/>.
+        /// </summary>
+        public static bool IsGRpcContentType(string contentType)
```
Tratcher, line 17: You should be able to use Microsoft.Net.Http.Headers.MediaTypeHeaderValue.IsSubsetOf or similar for this.

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
JamesNK: No. We're able to be granular in gRPC ASP.NET Core because we know what endpoint the request is to (e.g. URL matches a unary call vs a streaming call). You won't have that information in a proxy.

src/ReverseProxy.Core/Util/GRpcProtocolHelper.cs
JamesNK: I didn't use that type because it allocates.

Tratcher: It's all StringSegments, it shouldn't allocate any more than the one class instance, no?

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
Tratcher: That's problematic. We can't just turn this off for all gRPC requests.

What about the response headers? Would they tell us any more? We may be able to delay disabling these until we've seen them.

JamesNK: No difference there either. And `IHttpMinRequestBodyDataRateFeature` is a request feature.

src/ReverseProxy.Core/Util/GRpcProtocolHelper.cs
JamesNK: One. Or use this and have zero 😋

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
davidni: @Tratcher we could make this logic opt-in for specific routes. Then we would get a safe default, and a little bit of friction for anyone who wants to do gRPC or other long running http/2 communications. Something like the following could be configured on each proxy route:

```cs
enum BypassMinRequestRatePolicy
{
   Never, // default
   GRpcContentType, // disables min data rate constraint on HTTP/2 requests with grpc content-type
   AllHttp2, // disables min data rate constraint on all HTTP/2 requests
}
```

src/ReverseProxy.Core/Service/Proxy/AutoFlushingStream.cs
```diff
@@ -0,0 +1,91 @@
...
+namespace Microsoft.ReverseProxy.Core.Service.Proxy
+{
+    /// <summary>
+    /// Delegates to a wrapped stream and calls its
+    /// <see cref="Stream.Flush"/> or <see cref="Stream.FlushAsync(CancellationToken)"/>
+    /// on every write.
+    /// </summary>
+    /// <remarks>
+    /// This is used by <see cref="StreamCopyHttpContent"/> to work around some undesirable behavior
+    /// in the HttpClient machinery (as of .NET Core 3.1.201) where an internal buffer
```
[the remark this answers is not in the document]
davidni: Also see https://github.com/dotnet/runtime/issues/35358

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
Tratcher: It shouldn't even be per route, it should only be disabled based on an exact path match for known streaming endpoints.

src/ReverseProxy.Core/Util/GRpcProtocolHelper.cs
davidfowl: One per request? That's wasteful, zero is better.

Tratcher: Then add an overload to MediaTypeHeaderValue.IsSubsetOf that doesn't need to allocate. We shouldn't be copying this kind of matching logic all over the framework, it's error prone.

src/ReverseProxy.Core/Service/Proxy/AutoFlushingStream.cs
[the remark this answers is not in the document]
halter73: The HttpClient behavior isn't that weird, it just doesn't autoflush. In a meeting where this came up, there didn't seem to be much interest in changing it since we can always manually flush after every write yourself like this Stream does.

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
halter73: I wouldn't mind predicating whether or not to enforce min data rate and max body size on some response header. This would require some coordination with the backend, and should be opt-in/configurable, but it seems like it's probably the easiest/safest option.

davidni: @halter73 that's an interesting proposal, we should consider it further. But isn't it potentially too late to remove min data rate? By the time we get the upstream response headers, we may have already begun reading the incoming request body stream. Not sure how well Kestrel would handle changing the limits after reading has started.

Additionally, how well with this play with multiplexed incoming http2 connections? Does the min data rate in Kestrel apply to the connection as a whole, or to the individual stream of the current request?

Tratcher: I had a long talk with James, special casing the grpc request content-type is likely as good a built-in solution as we're going to get. grpc uses the same content-type for streaming and non-streaming and only the backend knows based on it's API definitions. If we disable the size and rate limits here then they will still be enforced by the backend-server, but that server knows which requests it should enforce it for. We'll have to document this as a requirement for backend gRPC servers.

Note SignalR's new Http2 streaming protocol will need something similar, but at least that one should use a unique content-type https://github.com/dotnet/aspnetcore/issues/21104.

Tratcher, line 139:
```suggestion
            var isStreamingRequest = isIncomingHttp2 && GRpcProtocolHelper.IsGRpcContentType(context.Request.ContentType);
```

Tratcher:
@davidni I think we've answered the hard question here and can proceed. Can you handle the cleanup feedback or would you like us to take over?

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
Tratcher: According to James there isn't a unique response header we can check either so the point is moot.

Tratcher: > Additionally, how well with this play with multiplexed incoming http2 connections? Does the min data rate in Kestrel apply to the connection as a whole, or to the individual stream of the current request?

Rate and size limits are per stream/request, not per connection.

src/ReverseProxy.Core/Util/GRpcProtocolHelper.cs
davidni: @Tratcher if you don't mind, I'll file an issue and leave a comment, but won't address it with this PR.

Tratcher: Understood, this feedback was addressed more at us than you.

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
davidni: Filed https://github.com/microsoft/reverse-proxy/issues/118 for design discussions and I'll resolve this thread in favor of continuing discussions there.

davidni:
@Tratcher done, I think I addressed all comments.

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
```diff
@@ -134,6 +134,18 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
             Contracts.CheckValue(targetUri, nameof(targetUri));
             Contracts.CheckValue(httpClient, nameof(httpClient));
 
+            // :::::::::::::::::::::::::::::::::::::::::::::
+            // :: Step 0: Disable ASP .NET Core limits for streaming requests
+            var isIncomingHttp2 = HttpProtocol.IsHttp2(context.Request.Protocol);
+
+            // NOTE: We heuristically assume gRPC-looking requests may require streaming semantics.
+            // See https://github.com/microsoft/reverse-proxy/issues/118 for design discussion.
+            var isStreamingRequest = isIncomingHttp2 && GrpcProtocolHelper.IsGRpcContentType(context.Request.ContentType);
```
JamesNK, line 143: Change IsGRpcContentType to IsGrpcContentType please.

Repeat for all GRpc.

Tratcher:
Thanks

PR #84, opened 2020-04-23
Remove C#-specific reference

isaacabraham:
Unless the project does not work in F#.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 YARP (which stands for "YARP: A Reverse Proxy") is a project to create a reverse proxy server. We found a bunch of internal teams at Microsoft who were either building a reverse proxy for their service or had been asking about APIs and tech for building one, so we decided to get them all together to work on a common solution, this project.
 
-YARP is a reverse proxy toolkit for building fast proxy servers in C# using the infrastructure from ASP.NET and .NET. The key differentiator for YARP is that it's been designed to be easily customized and tweaked to match the specific needs of each deployment scenario. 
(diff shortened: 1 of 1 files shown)


samsp-msft:
Hmm. This project is not intended to be a perfect multi-language solution - being able to use F# or VB to customize the proxy are not explicit goals, and I don't think we have any demand for it from the customers we have spoken to.
My concern here is in terms of expectation setting - we don't explicitly intend to disable non c# languages - but we are not going to explicitly test them either.

forki:
maybe you have spoken to the wrong customers then? ASP.NET Core is pretty popular in F# land

isaacabraham:
@samsp-msft appreciate your honesty there.

Perhaps try looking at it from another perspective: if someone is looking at adopting F# (and subsequently dotnet) and looks at this project, would you want them to try it out - maybe start using it? Or go away with the impression that this library is only compatible with C#? This is something we see happening more than you might think.

If this is indeed going to be a library that's going to become a part of the asp .net family of packages, it would be unfortunate if people got the wrong idea that ASP .NET = C# only.

dustinmoris:
Please don't make me have to create "Pony - A functional wrapper for YARP".

I've got no time for that atm :P

samsp-msft:
My hesitancy was from an expectation management perspective. We are not trying to exclude F# or VB, and we don't think that there is anything that will prevent usage of those languages - but creating templates, doing testing etc to have a really welcoming experience for the F# user is not in our short term roadmap.

isaacabraham:
@samsp-msft understood :-) I think no one in the F#-specific userbase will expect some custom F# wrapper for this (of course, that would be lovely!). People will create their own if needed; F# can interop with standard C# pretty well when you need to.

nbevans:
Lack of templates/sample code is not a deterrent since F# developers can almost always read C# code fluently and repurpose it.

PR #95, opened 2020-04-24
Fix blog post link in AsyncSemaphore

ReubenBond:
Perhaps it's a misconfigured redirect and the correct resolution is not to change the code, but the current link currently redirects to https://devblogs.microsoft.com/pfxteam/https:/devblogs.microsoft.com/iotdev/https:/devblogs.microsoft.com/iotdev/https:/devblogs.microsoft.com/iotdev/building-async-coordination-primitives-part-5-asyncsemaphore/

### DIFF
--- a/src/ReverseProxy.Utilities/Utils/AsyncSemaphore.cs
+++ b/src/ReverseProxy.Utilities/Utils/AsyncSemaphore.cs
@@ -10,7 +10,7 @@ namespace Microsoft.ReverseProxy.Utilities
     /// Alternative to SemaphoreSlim that respects the current thread scheduler.
     /// </summary>
     /// <remarks>
-    /// Based on <c>https://blogs.msdn.microsoft.com/pfxteam/2012/02/12/building-async-coordination-primitives-part-5-asyncsemaphore/</c>.
+    /// Based on <c>https://devblogs.microsoft.com/pfxteam/building-async-coordination-primitives-part-5-asyncsemaphore/</c>.
     /// </remarks>
     public sealed class AsyncSemaphore
     {


Tratcher:
Thanks @ReubenBond.

@samsp-msft any guess who runs that blog and mangled the redirect so badly?

Tratcher:
@anurse the CI seems to be stuck?

davidfowl:
There's a bug in the blogs, lets not fix this.

davidfowl:
Why do we even have this type?

Tratcher:
> Why do we even have this type?

*shrug* heath checks seems to use it.

davidfowl:
Doesn't SemaphoreSlim work?

Tratcher:
> Doesn't SemaphoreSlim work?

I don't know why they chose this, you'd have to ask @davidni.

davidni:
Determinism in unit tests. AsyncSempahore respects the current TaskScheduler when scheduling continuations, and enables us to write precise and deterministic unit tests that simulate passage of time. SemaphoreSlim escapes the current TaskScheduler and therefore makes writing those kinds of tests impossible or extremely hard.

Having said that.... `AsyncSemaphore` is a pretty naïve implementation and doesn't perform well with many waiters. I think it would be wise to fix the bottlenecks, or get rid of it entirely if we find that we can exercise all relevant unit test scenarios without it.

davidfowl:
I'd get rid of it entirely and use `SemaphoreSlim`, are you using a custom task scheduler in the unit tests?

davidfowl:
Blog posts fixed

davidni:
are you using a custom task scheduler in the unit tests?

@davidfowl yep. I've shared internal links privately with some examples.

PR #96, opened 2020-04-24
Round robin load balancing 

Tratcher:
#72 While we debated more complex backend data storage mechanisms, I wanted to try something simple first. Built in algorithms can avail themselves of built in structures so I added the round robin state tracker to the existing BackendLoadBalancingOptions structure.

### DIFF
--- a/src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
+++ b/src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
@@ -41,7 +41,8 @@ namespace Microsoft.ReverseProxy.Core.Middleware
             var endpoints = endpointsFeature?.Endpoints
                 ?? throw new InvalidOperationException("The AvailableBackendEndpoints collection was not set.");
 
-            var loadBalancingOptions = backend.Config.Value?.LoadBalancingOptions ?? default;
+            var loadBalancingOptions = backend.Config.Value?.LoadBalancingOptions
+                ?? new BackendConfig.BackendLoadBalancingOptions(default);
 
             var endpoint = _operationLogger.Execute(
                 "ReverseProxy.PickEndpoint",
--- a/src/ReverseProxy.Core/Service/RuntimeModel/BackendConfig.cs
+++ b/src/ReverseProxy.Core/Service/RuntimeModel/BackendConfig.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.ReverseProxy.Core.Abstractions;
+using Microsoft.ReverseProxy.Core.Util;
 
 namespace Microsoft.ReverseProxy.Core.RuntimeModel
 {
@@ -79,9 +80,13 @@ namespace Microsoft.ReverseProxy.Core.RuntimeModel
             public BackendLoadBalancingOptions(LoadBalancingMode mode)
             {
                 Mode = mode;
+                // Increment returns the new value and we want the first return value to be 0.
+                RoundRobinState = new AtomicCounter() { Value = -1 };
             }
 
             public LoadBalancingMode Mode { get; }
+
+            public AtomicCounter RoundRobinState { get; }
         }
     }
 }
--- a/src/ReverseProxy.Core/Service/RuntimeModel/BackendConfig.cs
+++ b/src/ReverseProxy.Core/Service/RuntimeModel/BackendConfig.cs
@@ -86,7 +86,7 @@ namespace Microsoft.ReverseProxy.Core.RuntimeModel
 
             public LoadBalancingMode Mode { get; }
 
-            public AtomicCounter RoundRobinState { get; }
+            internal AtomicCounter RoundRobinState { get; }
         }
     }
 }
--- a/test/ReverseProxy.Core.Tests/Service/Proxy/LoadBalancerTests.cs
+++ b/test/ReverseProxy.Core.Tests/Service/Proxy/LoadBalancerTests.cs
@@ -135,6 +135,29 @@ namespace Microsoft.ReverseProxy.Core.Service.Proxy.Tests
             Assert.Same(result, endpoints[1]);
         }
 
+        [Fact]
+        public void PickEndpoint_RoundRobin_Works()
+        {
+            var loadBalancer = Create<LoadBalancer>();
+            var endpoints = new[]
+            {
+                new EndpointInfo("ep1"),
+                new EndpointInfo("ep2"),
+            };
+            endpoints[0].ConcurrencyCounter.Increment();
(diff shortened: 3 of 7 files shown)


src/ReverseProxy.Core/Service/RuntimeModel/BackendConfig.cs
```diff
@@ -79,9 +80,13 @@ public BackendHealthCheckOptions(bool enabled, TimeSpan interval, TimeSpan timeo
             public BackendLoadBalancingOptions(LoadBalancingMode mode)
             {
                 Mode = mode;
+                // Increment returns the new value and we want the first return value to be 0.
```
samsp-msft, line 83: I'd make the first value random - that way all requests don't all start out hitting the same server.

test/ReverseProxy.Core.Tests/Service/Proxy/LoadBalancerTests.cs
```diff
@@ -135,6 +135,29 @@ public void PickEndpoint_LeastRequests_Works()
...
+            };
+            endpoints[0].ConcurrencyCounter.Increment();
+            var options = new BackendConfig.BackendLoadBalancingOptions(LoadBalancingMode.RoundRobin);
+
+            var result0 = loadBalancer.PickEndpoint(endpoints, in options);
+            var result1 = loadBalancer.PickEndpoint(endpoints, in options);
+            var result2 = loadBalancer.PickEndpoint(endpoints, in options);
+            var result3 = loadBalancer.PickEndpoint(endpoints, in options);
+
+            Assert.Same(result0, endpoints[0]);
```
samsp-msft, line 155: This is assuming we start at 0 - check result2 == result0 and result1 == result3

src/ReverseProxy.Core/Service/RuntimeModel/BackendConfig.cs
Tratcher: ? Every request increments the counter for an even distribution. It shouldn't matter where it starts, I only wanted 0 because it conceptually meets expectations (start the the top of the list and cycle through all entries).

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
```diff
@@ -41,7 +41,8 @@ public Task Invoke(HttpContext context)
             var endpoints = endpointsFeature?.Endpoints
                 ?? throw new InvalidOperationException("The AvailableBackendEndpoints collection was not set.");
 
-            var loadBalancingOptions = backend.Config.Value?.LoadBalancingOptions ?? default;
+            var loadBalancingOptions = backend.Config.Value?.LoadBalancingOptions
+                ?? new BackendConfig.BackendLoadBalancingOptions(default);
 
             var endpoint = _operationLogger.Execute(
```
davidfowl, line 47: This allocates a closure per request. I just noticed this... We need to get rid of this operation logger.

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
```diff
@@ -41,7 +41,8 @@ public Task Invoke(HttpContext context)
             var endpoints = endpointsFeature?.Endpoints
                 ?? throw new InvalidOperationException("The AvailableBackendEndpoints collection was not set.");
 
-            var loadBalancingOptions = backend.Config.Value?.LoadBalancingOptions ?? default;
+            var loadBalancingOptions = backend.Config.Value?.LoadBalancingOptions
```
davidfowl, line 44: Why is this being read per request?

Tratcher: Because it can be configured per route/backend.

src/ReverseProxy.Core/Middleware/LoadBalancingMiddleware.cs
Tratcher: https://github.com/microsoft/reverse-proxy/issues/115

src/ReverseProxy.Core/Service/RuntimeModel/BackendConfig.cs
```diff
@@ -79,9 +80,13 @@ public BackendHealthCheckOptions(bool enabled, TimeSpan interval, TimeSpan timeo
...
+                RoundRobinState = new AtomicCounter() { Value = -1 };
             }
 
             public LoadBalancingMode Mode { get; }
+
+            public AtomicCounter RoundRobinState { get; }
```
halter73, line 89: This is temporary until we come up with a better design, right? Until then, can we make it internal?

Tratcher: This seemed to work well, it's unclear if we need a more complex design. I don't see a problem with built in algorithms having built in state.

I don't know about internal or public here, it can be internal for now. We need to do a full API review at some point.

Tratcher, line 89:
```suggestion
            internal AtomicCounter RoundRobinState { get; }
```

LeonHartley: I think some sort of strategy pattern would be nice here. Each mode could have its own processor and be in charge of its own state. 

Doesn't feel great having every load balancing algorithm contained within 1 class. Also feel like it violates the "Open-closed" SOLID principle.

Maybe you could have a Dictionary of LoadBalancer strategies/processors?

Tratcher: We're discussing that in https://github.com/microsoft/reverse-proxy/issues/106.

LeonHartley: Awesome, I didn't see that issue. Thanks mate!

src/ReverseProxy.Core/Service/RuntimeModel/BackendConfig.cs
davidni: In a distributed deployment of YARP, all YARP instances would start out choosing the same target server if initial choice is const. There is some merit to making it random, but i reckon the benefit is minimal to negligible. In a practical deployment, it is unlikely that many instances of YARP would be (re-)started at precisely the same time, and even if they were, there is a lot of entropy that would help minimize hotspots to some extent. Further, anyone doing round-robin probably doesn't care about hotspots anyway.

test/ReverseProxy.Core.Tests/Service/Proxy/LoadBalancerTests.cs
Tratcher: Leaving this unless we decide to change the starting value.

PR #103, opened 2020-04-25
Fix Get-Content encoding error.

littlesmilelove:
When in non-English environment, `Get-Content` will return wrong encoding string, so `Test-Path` will fail.


anurse:
The `eng\common` directory actually gets synced from the [Arcade](https://github.com/dotnet/arcade) tools we use to build the repo. Could you make a PR there instead? Otherwise this change will just get wiped out when we next sync.

littlesmilelove:
> The `eng\common` directory actually gets synced from the [Arcade](https://github.com/dotnet/arcade) tools we use to build the repo. Could you make a PR there instead? Otherwise this change will just get wiped out when we next sync.

Ok.

anurse:
Thanks! Feel free to tag me in your PR there and we’ll try to get it merged to Arcade quickly and synced over here.

Issue #106, opened 2020-04-26
Add more extensible/custom load balancers


lvermeulen:
### What should we add or change to make your life better?
More load balancing strategies.

In short, I would like to:
* Have a generic ILoadBalancingStrategy
* Have several implementations of this ILoadBalancingStrategy
* Make this extensible somehow

### Why is this important to you?
I would like to have more choices when I choose the load balancing strategy.

The scope would initially be:
* A callback load balancing strategy: to allow a function to decide which endpoint gets picked next
* A deficit round robin load balancing strategy: to allow round robin taking into account a quantum (see [Deficit round robin](https://en.wikipedia.org/wiki/Deficit_round_robin)) and thus providing a "fairer" round robin load balancing strategy
* A failover load balancing strategy: to allow an availability predicate to decide which endpoint gets picked next
* A traffic allocation load balancing strategy: to allow inspection into traffic allocation of the endpoints (no matter how that is measured)

I have submitted the algorithms described above (with passing tests). The ILoadBalacingStrategy interface has a single Balance method which takes as parameters the list of available endpoints and the load balancing options.

I am unsure how to handle the configuration for the algorithms. Right now I have added configuration to the BackEndConfig.BackEndLoadBalancingOptions, which felt a little odd. :-)

samsp-msft:
Updated title to be clearer.

Tratcher:
Why these algorithms specifically? Have you had situations elsewhere that these were able to address?

> * A callback load balancing strategy: to allow a function to decide which endpoint gets picked next

If all this does is run a custom user callback then there are several other ways to do that like the new middleware model (https://github.com/microsoft/reverse-proxy/pull/50). It's not clear that it gains anything by being integrated directly into the LoadBalancer.

> * A deficit round robin load balancing strategy: to allow round robin taking into account a quantum (see [Deficit round robin](https://en.wikipedia.org/wiki/Deficit_round_robin)) and thus providing a "fairer" round robin load balancing strategy

The description sounds like a credit based LeastRequests strategy. I don't see any obvious benefits over the existing algorithms?

> * A failover load balancing strategy: to allow an availability predicate to decide which endpoint gets picked next

We expect to address this with health checks, retries, etc., not as part of load balancing.

> * A traffic allocation load balancing strategy: to allow inspection into traffic allocation of the endpoints (no matter how that is measured)

[...]

PR #107, opened 2020-04-26
Add callback, deficit round robin, failover, traffic allocation

lvermeulen:
This is a PR wrt issue [106](https://github.com/microsoft/reverse-proxy/issues/106)

### DIFF
--- a/CallbackLoadBalancingStrategy.cs
+++ b/CallbackLoadBalancingStrategy.cs
@@ -1,0 +1,21 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.ReverseProxy.Core.RuntimeModel;
+
+namespace ReverseProxy.Core.Service.Proxy.LoadBalancingStrategies
+{
+    public class CallbackLoadBalancingStrategy : ILoadBalancingStrategy
+    {
+        private readonly Func<IEnumerable<EndpointInfo>, BackendConfig.BackendLoadBalancingOptions, EndpointInfo> _callback;
+
+        public CallbackLoadBalancingStrategy(Func<IEnumerable<EndpointInfo>, BackendConfig.BackendLoadBalancingOptions, EndpointInfo> callback)
+        {
+            _callback = callback;
+        }
+
+        public EndpointInfo Balance(IReadOnlyList<EndpointInfo> availableEndpoints, BackendConfig.BackendLoadBalancingOptions loadBalancingOptions)
+        {
+            return _callback(availableEndpoints, loadBalancingOptions);
+        }
+    }
+}
--- a/FailOverLoadBalancingStrategy.cs
+++ b/FailOverLoadBalancingStrategy.cs
@@ -1,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.ReverseProxy.Core.RuntimeModel;
+
+namespace ReverseProxy.Core.Service.Proxy.LoadBalancingStrategies
+{
+    public class FailOverLoadBalancingStrategy : ILoadBalancingStrategy
+    {
+        private readonly EndpointInfo _preferredEndpoint;
+        private readonly Predicate<EndpointInfo> _isAvailable;
+        private readonly ILoadBalancingStrategy _fallBackLoadBalancingStrategy;
+
+        public FailOverLoadBalancingStrategy(EndpointInfo preferredEndpoint, Predicate<EndpointInfo> isAvailable, Func<ILoadBalancingStrategy> fallBackLoadBalancingStrategyFactory = null)
+        {
+            _preferredEndpoint = preferredEndpoint;
+            _isAvailable = isAvailable;
+
+            _fallBackLoadBalancingStrategy = fallBackLoadBalancingStrategyFactory();
+        }
+
+        public EndpointInfo Balance(IReadOnlyList<EndpointInfo> availableEndpoints, BackendConfig.BackendLoadBalancingOptions loadBalancingOptions)
+        {
+            var exceptPreferred = availableEndpoints
+                .Except(Enumerable.Repeat(_preferredEndpoint, 1))
+                .ToList();
+
+            return _isAvailable(_preferredEndpoint)
+                ? _preferredEndpoint
+                : _fallBackLoadBalancingStrategy.Balance(exceptPreferred.ToList().AsReadOnly(), loadBalancingOptions);
+        }
+    }
+}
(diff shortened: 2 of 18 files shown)


Tratcher:
Please discuss the design of large features with the team in an issue before submitting PRs. It can save a lot of churn in review.

lvermeulen:
Certainly, I thought I had initiated that with the issue that I created. How should I do this correctly?

Tratcher:
The issue was a bit vague :grin:. Propose a scope for what you're after (which algorithms, why), sumerize  how you expect to achieve it, and give a rough outline of any new APIs. Once there's some agreement on the features' viability then submit a PR.

We can leave this PR for now and come back to it.

LoadBalancingMode.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// Select the first endpoint without considering load. This is useful for dual endpoint fail-over systems.
+        /// </summary>
+        First,
+        /// <summary>
+        /// Let a callback decide with endpoint to pick.
+        /// </summary>
+        Callback,
+        /// <summary>
+        /// Selects an endpoint by cycling through them in order, taking a quantum into account.
```
Tratcher, line 37: A quantum?

LoadBalancingMode.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        DeficitRoundRobin,
+        /// <summary>
+        /// Selects the same endpoint until it is unavailable.
+        /// </summary>
+        FailOver,
+        /// <summary>
+        /// Selects an endpoint based on traffic allocation.
```
Tratcher, line 45: ?

FailOverLoadBalancingStrategy.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace ReverseProxy.Core.Service.Proxy.LoadBalancingStrategies
+{
+    public class FailOverLoadBalancingStrategy : ILoadBalancingStrategy
+    {
+        private readonly EndpointInfo _preferredEndpoint;
+        private readonly Predicate<EndpointInfo> _isAvailable;
+        private readonly ILoadBalancingStrategy _fallBackLoadBalancingStrategy;
+
+        public FailOverLoadBalancingStrategy(EndpointInfo preferredEndpoint, Predicate<EndpointInfo> isAvailable, Func<ILoadBalancingStrategy> fallBackLoadBalancingStrategyFactory = null)
```
Tratcher, line 14: While we do consider failover a likely feature of the proxy, we don't expect it to be built directly into the load balancer portion. `isAvailable` is handled elsewhere via health checks and circuit breakers.

CallbackLoadBalancingStrategy.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Collections.Generic;
+using Microsoft.ReverseProxy.Core.RuntimeModel;
+
+namespace ReverseProxy.Core.Service.Proxy.LoadBalancingStrategies
+{
+    public class CallbackLoadBalancingStrategy : ILoadBalancingStrategy
+    {
+        private readonly Func<IEnumerable<EndpointInfo>, BackendConfig.BackendLoadBalancingOptions, EndpointInfo> _callback;
+
+        public CallbackLoadBalancingStrategy(Func<IEnumerable<EndpointInfo>, BackendConfig.BackendLoadBalancingOptions, EndpointInfo> callback)
```
Tratcher, line 11: This just bypasses load balancing. That's likely better accomplished via a custom middleware or replacing ILoadBalancer. Then you wouldn't have to deal with the issue of how to flow the callback.

LoadBalancer.cs
```diff
@@ -0,0 +1,62 @@
...
+        public LoadBalancer(IRandomFactory randomFactory)
+        {
+            _randomFactory = randomFactory;
+        }
+
+        private ILoadBalancingStrategy GetLoadBalancingStrategy(LoadBalancingMode loadBalancingMode, BackendConfig.BackendLoadBalancingOptions loadBalancingOptions)
+        {
+            return loadBalancingMode switch
+            {
+                LoadBalancingMode.PowerOfTwoChoices => new PowerOfTwoChoicesLoadBalancingStrategy(_randomFactory),
```
Tratcher, line 29: We would not want to allocate a new strategy per request like this.

LoadBalancingMode.cs
lvermeulen: A quantum = a number :-)

LoadBalancer.cs
lvermeulen: This is just a naive implementation so that I could implement load balancing strategies. Obviously you can do what you like with this.

lvermeulen:
> This doesn't actually compile or run, none of the files are in project directories? And several existing files have been copied out of their existing directories?

The checks passed, so I thought this did compile. My commit was special though - I had to upload files to GitHub instead of doing a normal push. Maybe that's what threw it off.

> I have enough concerns with the design and implementation that I think it's best to close this PR and continue the discussion on the issue instead.

Absolutely.

LoadBalancingMode.cs
Tratcher: But what does it represent/accomplish?

LoadBalancer.cs
Tratcher: Since the algorithms are short and don't have any additional dependencies it would make sense to put them into their own methods on the LoadBalancer class. The interface and separate classes aren't needed.

Tratcher:
The checks ignored these files because they weren't in project directories.

Instead of doing a normal push?

LoadBalancingMode.cs
lvermeulen: A number that lets you direct the round robin load balancing. In the test you can see that I use the number to direct the number of times an endpoint gets hit, but you could do a lot of things.

FailOverLoadBalancingStrategy.cs
lvermeulen: Fair enough

LoadBalancer.cs
lvermeulen: I was thinking about extensibility, for which you would need an interface of sorts.

lvermeulen:
Yeah, I was doing this in a VM on a corporate laptop. The VM has a proxy, and the host has another proxy. I couldn't figure out git's proxy settings, so I had to upload the files with GitHub, instead of doing a command-line or Visual Studio git push.

PR #117, opened 2020-04-27
Reduce allocation when adding request headers

Kahbazi:


src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
```diff
@@ -353,9 +353,9 @@ private void CopyHeadersToUpstream(IHeaderDictionary source, HttpRequestMessage
                 // HttpRequestMessage.Headers and HttpRequestMessage.Content.Headers.
                 // We don't really care where the proxied headers appear among those 2,
                 // as long as they appear in one (and only one, otherwise they would be duplicated).
-                if (!destination.Headers.TryAddWithoutValidation(header.Key, header.Value.ToArray()))
+                if (!destination.Headers.TryAddWithoutValidation(header.Key, values: header.Value))
```
Tratcher, line 356: Do we have any multi-value header tests that show this calls the right overload of TryAddWithoutValidation? There's `IEnumerable<string>` and `string`, and StringValues is implicitly convertible to both. Converting multiple values to a single string would be incorrect in this case, not to mention allocating.

Rodrigo-Andrade: these are always arrays, is this trying not to violate the IEnumerable contract?

Tratcher: > these are always arrays, is this trying not to violate the IEnumerable contract?

They're `StringValues`.

Kahbazi: > Converting multiple values to a single string would be incorrect in this case, 

I think the case here is opposite, A single value might converts to an array. But I made a mistake, the extra allocation happens when a `ToArray()` is called for a single value. What I did changes nothing since both `ToArray()` and implicit conversion using the same [method](https://github.com/dotnet/runtime/blob/93e96a8468ded31928bb6c3e90ed2b5f7f5914d1/src/libraries/Microsoft.Extensions.Primitives/src/StringValues.cs#L285).

How about this? we check the `Count` property

- If it is 0, we do nothing

- If it is 1, convert it to `string`

- - If it is more than 1, convert it to `string[]`

davidfowl: That’s what I expected to see

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
```diff
@@ -337,6 +337,12 @@ private void CopyHeadersToUpstream(IHeaderDictionary source, HttpRequestMessage
         {
             foreach (var header in source)
             {
+                var headerValueCount = header.Value.Count;
+                if (headerValueCount == 0)
+                {
+                    continue;
+                }
+
                 if (header.Key.Length > 0 && header.Key[0] == ':')
```
Tratcher, line 346: Not related to your change, but this line should have a comment.
```suggestion
                // Filter out HTTP/2 pseudo headers like ":method" and ":path", those go into other fields.
                if (header.Key.Length > 0 && header.Key[0] == ':')
```

Tratcher:
Thanks

PR #119, opened 2020-04-28
minor clean code

Marthyi:
- reduce verbosity of HttpUtilities

### DIFF
--- a/src/ReverseProxy.Core/Service/Proxy/HttpUtilities.cs
+++ b/src/ReverseProxy.Core/Service/Proxy/HttpUtilities.cs
@@ -13,55 +13,18 @@ namespace Microsoft.ReverseProxy.Core.Service.Proxy
         /// Converts the given HTTP method (usually obtained from <see cref="HttpRequest.Method"/>)
         /// into the corresponding <see cref="HttpMethod"/> static instance.
         /// </summary>
-        public static HttpMethod GetHttpMethod(string method)
+        public static HttpMethod GetHttpMethod(string method) => method switch
         {
-            if (HttpMethods.IsGet(method))
-            {
-                return HttpMethod.Get;
-            }
-
-            if (HttpMethods.IsPost(method))
-            {
-                return HttpMethod.Post;
-            }
-
-            if (HttpMethods.IsPut(method))
-            {
-                return HttpMethod.Put;
-            }
-
-            if (HttpMethods.IsDelete(method))
-            {
-                return HttpMethod.Delete;
-            }
-
-            if (HttpMethods.IsOptions(method))
-            {
-                return HttpMethod.Options;
-            }
-
-            if (HttpMethods.IsHead(method))
-            {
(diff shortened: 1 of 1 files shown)


src/ReverseProxy.Core/Service/Proxy/HttpUtilities.cs
```diff
@@ -13,55 +13,17 @@ internal static class HttpUtilities
...
-        }
+            string mth when HttpMethods.IsGet(mth) => HttpMethod.Get,
+            string mth when HttpMethods.IsPost(mth) => HttpMethod.Post,
+            string mth when HttpMethods.IsPut(mth) => HttpMethod.Put,
+            string mth when HttpMethods.IsDelete(mth) => HttpMethod.Delete,
+            string mth when HttpMethods.IsOptions(mth) => HttpMethod.Options,
+            string mth when HttpMethods.IsHead(mth) => HttpMethod.Head,
+            string mth when HttpMethods.IsPatch(mth) => HttpMethod.Patch,
+            string mth when HttpMethods.IsTrace(mth) => HttpMethod.Trace,
+            _ => throw new InvalidOperationException($"Unsupported request method '{method}'.")
```
Tratcher, line 26:
```suggestion
            // NOTE: Proxying "CONNECT" is not supported (by design!)
            _ => throw new InvalidOperationException($"Unsupported request method '{method}'.")
```

Marthyi: > 

- Good point, I 've re-inserted the comment.
- No issue detected on AtomicCounter. Here I just wanted to "force" the volatile access

halter73:
> What was wrong with AtomicCounter? It doesn't look like you've changed the behavior, just the pattern?

I think I prefer the old way. The `volatile` keyword is too easy to miss.

Tratcher:
Thanks

src/ReverseProxy.Core/Service/Proxy/HttpUtilities.cs
```diff
@@ -13,55 +13,18 @@ internal static class HttpUtilities
...
+            string mth when HttpMethods.IsTrace(mth) => HttpMethod.Trace,
             // NOTE: Proxying "CONNECT" is not supported (by design!)
-            //if (HttpMethods.IsConnect(method))
-            //{
-            //    return new HttpMethod("CONNECT");
-            //}
-
-            throw new InvalidOperationException($"Unsupported request method '{method}'.");
-        }
+            _ => throw new InvalidOperationException($"Unsupported request method '{method}'.")
```
Kahbazi, line 27: @Tratcher Is there a reason YARP is not proxying custom http methods?

Tratcher: Probably? At the very least you need to consider the semantics of a method. E.g. we wouldn't want to proxy CONNECT using default logic, that would end badly.

PR #132, opened 2020-04-30
add docs for branching for release

anurse:
Added docs on how to manage branching for releases. The flow described here is what we've previously discussed

* .NET 5 snaps their release branch, and at that time we change our `master` to pull from that release branch
* We continue open development in `master`, but won't get new .NET 5 builds
* At some point prior to release, we branch `release/1.0.0-previewX` and move `master` on to the next preview.

An alternate plan is to just branch when .NET branches. It might make things smoother. I'd be happy to write up the docs for that strategy if it seems like it would be easier. We'd still be able to decide for ourselves when to stop coding, but we might have to switch PRs between branches more.

Thoughts?

### DIFF
--- a/docs/operations/BranchingAndRelease.md
+++ b/docs/operations/BranchingAndRelease.md
@@ -6,6 +6,10 @@ We are aiming to ship YARP previews aligned with .NET 5, since we are working on
 2. Open development, .NET 5 builds frozen - After .NET 5 branches for a release, we will switch `master` to receive builds from that branch. Our development will continue.
 3. Branch and prepare for release - Prior to the release date, we'll branch and prepare for release.
 
+## Scheduling
+
+We need to ship versions of our package that depend upon the **released** preview builds. In order to ensure we do that, we try to ship same-date or shortly after .NET 5 previews (see the [.NET Wiki (Internal)](https://aka.ms/dotnet-wiki) for schedule details).
+
 ## Prerequisites
 
 * Properly configured `darc` global tool:
@@ -23,6 +27,7 @@ To do this, run the following commands:
 2. Copy the `Id` value
 3. Run `darc update-subscription --id [Id]`. An editor will open.
 4. Change the value of the `Channel` field to `.NET 5 Preview X` (where `X` is the preview we are remaining on)
+4. Change the value of the `Update Frequency` field to `EveryBuild` (since there should be relatively few builds and we want to be sure we're on the latest).
 5. Save and close the file in your editor, `darc` will proceed to make the update.
 6. Answer `y` to `Trigger this subscription immediately?` and a PR will be opened to update versions to the latest ones in that channel.
 7. Merge the PR as soon as feasible.
@@ -60,6 +65,13 @@ Finally, restore the `master` branch to pulling the latest bits from .NET 5:
 2. Copy the `Id` value
 3. Run `darc update-subscription --id [Id]`. An editor will open.
 4. Change the value of the `Channel` field to `.NET 5 Dev`
+4. Change the value of the `Update Frequency` field back to `EveryWeek` to reduce PR noise.
 5. Save and close the file in your editor, `darc` will proceed to make the update.
 6. Answer `y` to `Trigger this subscription immediately?` and a PR will be opened to update versions to the latest ones in that channel.
 7. Merge the PR as soon as feasible.
+
+## Releasing Packages
+
+In order to release packages, we first need to identify the final build. Assuming we've branched, the latest build of the Azure Pipeline [`microsoft-reverse-proxy-official`](https://dev.azure.com/dnceng/internal/_build?definitionId=809&_a=summary) should contain the final bits.
+
+Remaining docs TBD.
--- a/docs/operations/BranchingAndRelease.md
+++ b/docs/operations/BranchingAndRelease.md
@@ -1,0 +1,65 @@
+# Branching Tasks
+
+We are aiming to ship YARP previews aligned with .NET 5, since we are working on changes to the runtime that will improve the experience for YARP. As part of that, our schedule for a preview milestone has three phases:
+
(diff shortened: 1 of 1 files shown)


docs/operations/BranchingAndRelease.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+3. Save and close the editor window.
+
+Finally, restore the `master` branch to pulling the latest bits from .NET 5:
+
+1. Run `darc get-subscriptions --target-repo microsoft/reverse-proxy --source-repo dotnet/runtime` to get the subscription from `dotnet/runtime` to `microsoft/reverse-proxy`
+2. Copy the `Id` value
+3. Run `darc update-subscription --id [Id]`. An editor will open.
+4. Change the value of the `Channel` field to `.NET 5 Dev`
```
halter73, line 62: darc is showing that me we're subscribed to the ".NET Eng - Latest" channel? Is it supposed to be ".NET 5 Dev"? What's the difference?

```
> darc get-subscriptions --source-repo arcade --target-repo reverse-proxy
https://github.com/dotnet/arcade (.NET Eng - Latest) ==> 'https://github.com/microsoft/reverse-proxy' ('master')
...
```

docs/operations/BranchingAndRelease.md
```diff
@@ -0,0 +1,65 @@
...
+## When .NET 5 branches for a preview
+
+When .NET 5 branches for a preview, we need to switch the "channel" from which we are getting builds. See [Dependency Flow Onboarding](https://github.com/dotnet/arcade/blob/master/Documentation/DependencyFlowOnboarding.md) for more information on channels.
+
+To do this, run the following commands:
+
+1. Run `darc get-subscriptions --target-repo microsoft/reverse-proxy --source-repo dotnet/runtime` to get the subscription from `dotnet/runtime` to `microsoft/reverse-proxy`
+2. Copy the `Id` value
+3. Run `darc update-subscription --id [Id]`. An editor will open.
+4. Change the value of the `Channel` field to `.NET 5 Preview X` (where `X` is the preview we are remaining on)
```
halter73, line 25: Why not keep master on the "Dev"/"Latest" channel and only have release branches subscribe to "Preview X" channels for a specific .NET 5 preview?

docs/operations/BranchingAndRelease.md
anurse: For arcade that's correct. For `dotnet/runtime` it's `.NET 5 Dev`. The arcade repo publishes to a different channel because it's tools, not product code.

docs/operations/BranchingAndRelease.md
anurse: That's basically the alternate proposal. We always have `master` on `.NET 5 Dev` and only switch `release/1.0.0-previewX` builds to `.NET 5 Preview Y`. The problem is that means we have to branch when .NET 5 branches which can be up to 4 weeks before release. We could continue to develop during that time, but we have to swich PRs to the `release/1.0.0-previewX` branch.

docs/operations/BranchingAndRelease.md
```diff
@@ -0,0 +1,65 @@
+# Branching Tasks
+
+We are aiming to ship YARP previews aligned with .NET 5, since we are working on changes to the runtime that will improve the experience for YARP. As part of that, our schedule for a preview milestone has three phases:
+
+1. Open development and receiving new builds of .NET 5 - Commits are going to `master` and new builds of .NET 5 are coming regularly
+2. Open development, .NET 5 builds frozen - After .NET 5 branches for a release, we will switch `master` to receive builds from that branch. Our development will continue.
```
Tratcher, line 6:
```suggestion
2. Open development, .NET 5 builds frozen - .NET 5 branches about a month before release, so after they branch we will switch `master` to receive builds from that release branch. Our development will continue until approximately a week before release.
```

docs/operations/BranchingAndRelease.md
halter73: > The problem is that means we have to branch when .NET 5 branches which can be up to 4 weeks before release. We could continue to develop during that time, but we have to swich PRs to the release/1.0.0-previewX branch.

What about auto-merging master into the pending release branches? I imagine most commits are not going to be using new runtime APIs, and those that are could be skipped when we see a merge PR fails because of a compilation error. That way we can continue working in master without our release branches getting too stale.

Tratcher: Auto-merging forward might make sense, but merging backwards from master to release seems problematic.

halter73: I'd prefer to keep development in master as much as possible though. I get the current proposal is one way of achieving that, but not having the latest runtime in master is also annoying.

What's so bad about merging backwards? I doubt the branches will diverge much. We could not merge backwards if we're okay with staler releases, but I prefer merging so people can more easily try out the latest stuff.

Tratcher: Code developed on old dependencies is more likely to work on new ones than the other way around. You're right that there's only a few weeks involved so the underlying system shouldn't diverge too much, but that also means the value of moving the the latest runtime dependencies is minimal.

As a rule we should dev and test against the dependencies we plan to ship with, not later ones. @anurse's proposal seems like the simplest way to do that.

anurse:
> Needs a section on checking the current state of feeds

What do you mean by this? NuGet feeds?

Tratcher:
> > Needs a section on checking the current state of feeds
> 
> What do you mean by this? NuGet feeds? Darc Channels?

These docs tell me how to change our Darc channel subscriptions, but they don't say how to check which one we're currently subscribed to. E.g. How do I check if Andrew switched us to the correct subscriptions?

anurse:
Ah gotcha. I'll put some darc basics in :). Thanks!

anurse:
@Tratcher I added an overview of Maestro. Does that help?

PR #133, opened 2020-05-01
Basic x-fowarded-* headers

Tratcher:
#13 This is a stopgap for preview1 that sets the values without providing any configurability. CopyHeadersToUpstream is the pace I'd expect this transformation to happen long term, but I expect it will be built into a larger request transformation infrastructure such as being discussed in #21.

I changed the Startup sample to reflect that we don't recommend modifying the HttpContext for these kinds of transformations, it would break retries.

### DIFF
--- a/samples/ReverseProxy.Sample/Startup.cs
+++ b/samples/ReverseProxy.Sample/Startup.cs
@@ -6,6 +6,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.ReverseProxy.Core.Middleware;
 
 namespace Microsoft.ReverseProxy.Sample
 {
@@ -47,15 +48,21 @@ namespace Microsoft.ReverseProxy.Sample
                 endpoints.MapControllers();
                 endpoints.MapReverseProxy(proxyPipeline =>
                 {
-                    proxyPipeline.UseProxyLoadBalancing();
-                    // Customize the request before forwarding
+                    // Custom endpoint selection
                     proxyPipeline.Use((context, next) =>
                     {
-                        var connection = context.Connection;
-                        context.Request.Headers.AppendCommaSeparatedValues("X-Forwarded-For",
-                            new IPEndPoint(connection.RemoteIpAddress, connection.RemotePort).ToString());
+                        var someCriteria = false; // MeetsCriteria(context);
+                        if (someCriteria)
+                        {
+                            var availableEndpointsFeature = context.Features.Get<IAvailableBackendEndpointsFeature>();
+                            var endpoint = availableEndpointsFeature.Endpoints[0]; // PickEndpoint(availableEndpointsFeature.Endpoints);
+                            // Load balancing will no-op if we've already reduced the list of available endpoints to 1.
+                            availableEndpointsFeature.Endpoints = new[] { endpoint };
+                        }
+
                         return next();
                     });
+                    proxyPipeline.UseProxyLoadBalancing();
                 });
             });
         }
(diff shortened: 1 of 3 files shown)


samples/ReverseProxy.Sample/Startup.cs
```diff
@@ -47,15 +48,21 @@ public void Configure(IApplicationBuilder app)
...
-                    proxyPipeline.UseProxyLoadBalancing();
-                    // Customize the request before forwarding
+                    // Custom endpoint selection
                     proxyPipeline.Use((context, next) =>
                     {
-                        var connection = context.Connection;
-                        context.Request.Headers.AppendCommaSeparatedValues("X-Forwarded-For",
-                            new IPEndPoint(connection.RemoteIpAddress, connection.RemotePort).ToString());
+                        var someCriteria = false; // MeetsCriteria(context);
+                        if (someCriteria)
```
halter73, line 55: Did you mean to leave this?

test/ReverseProxy.Core.Tests/Service/Proxy/HttpProxyTests.cs
```diff
@@ -104,19 +111,76 @@ public async Task ProxyAsync_NormalRequest_Works()
...
         // Tests proxying an upgradable request.
         [Fact]
         public async Task ProxyAsync_UpgradableRequest_Works()
         {
             // Arrange
             var httpContext = new DefaultHttpContext();
             httpContext.Request.Method = "GET";
-            httpContext.Request.Scheme = "https";
-            httpContext.Request.Host = new HostString("example.com");
+            httpContext.Request.Scheme = "http";
```
halter73, line 177: Why did you change the scheme in this test? Should the targetUri be updated?

samples/ReverseProxy.Sample/Startup.cs
Tratcher: Yeah 😁

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
```diff
@@ -417,6 +420,17 @@ private void CopyHeadersToUpstream(IHeaderDictionary source, HttpRequestMessage
...
+
+            // Add common forwarders
+            // TODO: these need to be customizable
+            // https://github.com/microsoft/reverse-proxy/issues/13
+            // https://github.com/microsoft/reverse-proxy/issues/21
+            destination.Headers.TryAddWithoutValidation(ForwardedHeadersDefaults.XForwardedProtoHeaderName, context.Request.Scheme);
+            destination.Headers.TryAddWithoutValidation(ForwardedHeadersDefaults.XForwardedHostHeaderName, context.Request.Host.ToString());
+            if (context.Connection.RemoteIpAddress != null)
+            {
+                destination.Headers.TryAddWithoutValidation(ForwardedHeadersDefaults.XForwardedForHeaderName, context.Connection.RemoteIpAddress.ToString());
```
Kahbazi, line 432: Is there any plan to support X-Forwarded-For header for multiple proxies like 
X-Forwarded-For: client, proxy1, proxy2

Tratcher: This appends, so that should already work. I even added a test for it.

test/ReverseProxy.Core.Tests/Service/Proxy/HttpProxyTests.cs
Tratcher: > Why did you change the scheme in this test? 

To make it clear this was the value put into x-forwarded-proto.

> Should the targetUri be updated?

Nope, that's the target of the backend and the external and internal schemes are unrelated, that's why we needed to add x-forwarded-proto.

src/ReverseProxy.Core/Service/Proxy/HttpProxy.cs
Kahbazi: Oh I didn't know it appends! I have another question, if it appends, what should be done to prevent x-forwarded-for spoofing?
Is using the `ForwardedHeadersMiddleware` the answer? if so isn't this middleware override `context.Connection.RemoteIpAddress`?

Tratcher: Yes the validation steps performed by ForwardedHeadersMiddleware are designed to prevent spoofing. We may need a non-destructive variant for the proxy to use. Or we just let it do its thing because it also removes validated entries from the request header when moving them to RemoteIpAddress, so re-appending it shouldn't hurt.

PR #148, opened 2020-05-08
Add sample app for benchmarking

halter73:
This is the target of https://github.com/aspnet/Benchmarks/pull/1507/

### DIFF
--- a/samples/BenchmarkApp/Program.cs
+++ b/samples/BenchmarkApp/Program.cs
@@ -1,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Hosting;
+
+namespace BenchmarkApp
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            CreateHostBuilder(args).Build().Run();
+        }
+
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+                    webBuilder.UseStartup<Startup>();
+                });
+    }
+}
--- a/samples/BenchmarkApp/Program.cs
+++ b/samples/BenchmarkApp/Program.cs
@@ -17,6 +17,10 @@ namespace BenchmarkApp
             Host.CreateDefaultBuilder(args)
                 .ConfigureWebHostDefaults(webBuilder =>
                 {
+#if !DEBUG
+                    // The benchmark infrastructure seems to default to a development environment, but it always uses Release builds.
+                    webBuilder.UseEnvironment(Environments.Production);
+#endif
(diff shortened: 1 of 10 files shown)


samples/BenchmarkApp/Program.cs
```diff
@@ -0,0 +1,27 @@
...
+            CreateHostBuilder(args).Build().Run();
+        }
+
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+#if !DEBUG
+                    // The benchmark infrastructure seems to default to a development environment, but it always uses Release builds.
+                    webBuilder.UseEnvironment(Environments.Production);
```
Tratcher, line 22: Why do we care what the environment is set to? Logging?

samples/BenchmarkApp/Program.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            CreateHostBuilder(args).Build().Run();
+        }
+
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
```
sebastienros, line 17: You could also use conditional logging using an argument, that would allow to benchmark when it's off. It has a significant impact.

halter73: I made sure it doesn't log anything after startup when it's in a production environment. The logs are helpful when debugging locally.

samples/BenchmarkApp/Program.cs
halter73: Yep. Just logging. @sebastienros, do you know why BenchmarksServer seems to default to a dev environment?

sebastienros: I don't believe it does, it's even displayed in the output of the Benchmarks app ... I will check again.

samples/BenchmarkApp/Program.cs
sebastienros: Just having Console logger registered impacts perf. Hence why I call ClearProviders in Benchmarks.csproj. There is a benchmark for that in the Custom page. FYI.

samples/BenchmarkApp/Program.cs
halter73: Maybe I was running BenchmarksServer in a dev environment and that's what caused it. I'll check again.

samples/BenchmarkApp/Program.cs
halter73: I do want the benchmark to be be somewhat realistic. I think most customers will want warning logs enabled.

samples/BenchmarkApp/Program.cs
halter73: I was running BenchmarksServer in a dev environment. VS "helpfully" added a launchSettings.json for me that `dotnet run` then respected.

Issue #128, opened 2020-04-29
Rename Endpoint, Backend


Kahbazi:
Since Endpoint is already a concept in ASP.NET and right now Routes (from config) are being mapped to an Endpoint (ASP.NET) and Endpoints (from config) are referred to backend servers. 

I suggest renaming _Endpoint_ to _Server_ or _Resource_ and renaming _Route_ to _Endpoint_.

So this
```json
"ReverseProxy": {
  "Backends": {
    "backend1": {
      "Endpoints": {
        "Green": {
          "Address": "https://localhost:10001/"
        }
      }
    }
  },
  "Routes": [
    {
      "RouteId": "backend1/route1",
      "BackendId": "backend1",
      "Match": {
        "Methods": [ "GET", "POST" ],
        "Host": "localhost",
        "Path": "/{**catchall}"
      }
    }
  ]
}
```

would change to 
```json
"ReverseProxy": {
  "Backends": {
    "backend1": {
      "Servers": {
        "Green": {
          "Address": "https://localhost:10001/"
        }
      }
    }
  },
  "Endpoints": [
    {
      "EndpointId": "backend1/route1",
      "BackendId": "backend1",
      "Match": {
        "Methods": [ "GET", "POST" ],
        "Host": "localhost",
        "Path": "/{**catchall}"
      }
    }
  ]
}
```

Tratcher:
Yeah, the terminology could be improved, many of these terms are used in too many different contexts.

That said, endpoints is already an overloaded term (used by kestrel, routing, and the proxy). At least routes are a unique term. I don't see renaming routes to endpoints as an improvement.

I would like to rename backend endpoints to something to limit the overloading. Servers is also an overloaded term, we already have a server in the app at the front end. `Destination(s)`? `Backend(s)` could also work if we found a better group name for what we currently call backends (e.g. a group of endpoints).

davidni:
Comparing terminology with other proxies may help:

Product|YARP `Endpoint`|YARP `Backend`|YARP `Route`
-|:-|:-|:-
[Envoy](https://jvns.ca/blog/2018/10/27/envoy-basics/)|`host`|`cluster`|`route`
[Traefik v2](https://docs.traefik.io/routing/services/)|`server`|`service`|`router`
[Ocelot](https://ocelot.readthedocs.io/en/latest/features/routing.html)|`ReRoute.DownstreamHostAndPorts`|--|`ReRoute`

guibirow:
[...]

PR #150, opened 2020-05-09
Rename Endpoint to Destination

Tratcher:
#128 There's some churn, but I do think the end result is better. Maybe good to get this in for preview1...


Kahbazi:
There are some [README.md ](https://github.com/microsoft/reverse-proxy/blob/5c54f00e92b7f00592382c32ae12cc42c547a4cd/src/ReverseProxy.Core/Service/RuntimeModel/README.md)files in repo. Do they also need to be updated?

Tratcher:
> There are some [README.md ](https://github.com/microsoft/reverse-proxy/blob/5c54f00e92b7f00592382c32ae12cc42c547a4cd/src/ReverseProxy.Core/Service/RuntimeModel/README.md)files in repo. Do they also need to be updated?

Thanks @Kahbazi, updated.

PR #156, opened 2020-05-12
Update branding to preview.1

anurse:


eng/Versions.props
```diff
@@ -1,8 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <VersionPrefix>0.1.0</VersionPrefix>
-    <PreReleaseVersionLabel>alpha1</PreReleaseVersionLabel>
+    <VersionPrefix>1.0.0</VersionPrefix>
+    <!-- 
+      For sorting purposes, it is very important there be a '.' between the word 'preview' and the preview number.
+      That ensures that if we get up to preview 10, it doesn't sort as between preview 1 and preview 2!
```
Tratcher, line 7:
```suggestion
      That ensures that if we get up to preview 10, it doesn't sort as between preview 1 and preview 2!
```

I aspire to 10 previews.

anurse:
> The final package we ship will still have build numbers appended to this, right?

Yes, the build numbers are tacked on automatically. That **does** mean that when it's time to ship `1.0.0` someone will need to work out whatever Arcade incantations are necessary to produce "Stable versioned" bits. The .NET Engineering Team can help out here I'm sure.

anurse:
I suppose I could've just merged it... Oh well, now I can blame the bot if something goes wrong 😆

Issue #159, opened 2020-05-12
Cross target .NET 5.0 and 3.1


Tratcher:
Multiple partners have said that they will be able to adopt YARP sooner if we provide a netcoreapp3.1 target. We can cross target the package and light up new functionality. We'll want to have a doc tracking things that won't work on 3.1.

kasini3000:
mypoint:
Q: Will YARP provide compatibility with .net core 3.1?
A:
will not. You can use the .net core3.1 client to access YARP through the command line, after ssh, there is a powershell command line interface. This command line interface is compatible with any external calls such as win, linux, python, etc.

user app want add back end to YARP--->commandline--->ssh client---> ssh server--->
powershell YARP module (powershell 7.1 preview base .net core 5)
--->  the static method of .net core 5 of YARP

Tratcher:
> If there is a lot of work for compatibility with .net core 3.1, I suggest not supporting it first.

So far almost every works, the main 5.0 improvements have been in performance. We'll see if we run into any roadblocks later on.

kasini3000:
ok,3q

PR #168, opened 2020-05-15
Cross target .NET 5.0 and 3.1

Tratcher:
#159 We were calling one new 5.0 API, Microsoft.AspNetCore.Http.HttpProtocol.IsHttp2. We don't strictly need it, but it does make a good example for how to #if around differences.

### DIFF
--- a/src/ReverseProxy/Utilities/ProtocolHelper.cs
+++ b/src/ReverseProxy/Utilities/ProtocolHelper.cs
@@ -5,9 +5,20 @@ using System;
 
 namespace Microsoft.ReverseProxy
 {
-    internal static class GrpcProtocolHelper
+    internal static class ProtocolHelper
     {
         internal const string GrpcContentType = "application/grpc";
+
+        public static bool IsHttp2(string protocol)
+        {
+#if NETCOREAPP5_0
+            return Microsoft.AspNetCore.Http.HttpProtocol.IsHttp2(protocol);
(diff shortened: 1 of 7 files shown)


src/ReverseProxy/Utilities/ProtocolHelper.cs
```diff
@@ -5,10 +5,21 @@
...
         internal const string GrpcContentType = "application/grpc";
 
+        public static bool IsHttp2(string protocol)
+        {
+#if NETCOREAPP5_0
+            return Microsoft.AspNetCore.Http.HttpProtocol.IsHttp2(protocol);
+#elif NETCOREAPP3_1
+            return StringComparer.OrdinalIgnoreCase.Equals("HTTP/2", protocol);
+#else
+            TFMs need to be updated
```
davidni, line 19: `#error` directive?

anurse, line 19: One: use “#error [message]” and it will be a real computer error instead of a syntax error you have to track down to understand

Two: I like this pattern since it means we have to explicitly list the TFMs and each alternative, but suggest a clearer message: “A target framework was added to the project and needs to be added to this condition”

Tratcher:
Need to figure out how to get 3.1 installed as part of the build setup.

Tratcher:
It's alive!

PR #169, opened 2020-05-15
Upload test logs

Tratcher:
I'm trying to debug a test failure in https://github.com/microsoft/reverse-proxy/pull/168 and found that the test logs are not being uploaded.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -88,11 +88,11 @@ stages:
             artifactType: Container
             parallel: true
         - task: PublishBuildArtifacts@1
-          displayName: Upload Test Logs
+          displayName: Upload Logs
           condition: always()
           continueOnError: true
           inputs:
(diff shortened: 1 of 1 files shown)


Tratcher:
Nevermind, I finally found them.

halter73:
Were you looking for Microsoft.ReverseProxy.Tests_netcoreapp3.1_x64.log from https://dev.azure.com/dnceng/public/_build/results?buildId=647030&view=results? If so, where did you find it?

Also, it looks like the following line in our job template is broken:

https://github.com/microsoft/reverse-proxy/blob/f12c9eacf954f3d5d5cd15d8912841ead06ceabf/eng/common/templates/job/job.yml#L29

image

Not sure why or what the fix is though.

Tratcher:
image

anurse:
> Not sure why or what the fix is though.

https://github.com/dotnet/arcade/pull/5171

halter73:
Approved!

PR #170, opened 2020-05-16
Output runtime versions during benchmark startup

halter73:
As requested by @sebastienros. 

https://github.com/aspnet/Benchmarks/pull/1517#issuecomment-629568200

### DIFF
--- a/samples/BenchmarkApp/Program.cs
+++ b/samples/BenchmarkApp/Program.cs
@@ -1,6 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
+using System.Reflection;
+using System.Text.Json;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Hosting;
 
@@ -10,6 +13,7 @@ namespace BenchmarkApp
     {
         public static void Main(string[] args)
         {
+            WriteStatistics();
             CreateHostBuilder(args).Build().Run();
         }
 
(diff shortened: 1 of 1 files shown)


samples/BenchmarkApp/Program.cs
```diff
@@ -19,5 +23,25 @@ public static void Main(string[] args)
...
+
+            Console.WriteLine(JsonSerializer.Serialize(new {
+                Metadata = new object[]
+                {
+                    new { Source= "Benchmarks", Name= "AspNetCoreVersion", ShortDescription = "ASP.NET Core Version", LongDescription = "ASP.NET Core Version" },
+                    new { Source= "Benchmarks", Name= "NetCoreAppVersion", ShortDescription = ".NET Runtime Version", LongDescription = ".NET Runtime Version" },
+                },
+                Measurements = new object[]
+                {
+                    new { Timestamp = DateTime.UtcNow, Name = "AspNetCoreVersion", Value = typeof(WebHostBuilder).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion },
```
Tratcher, line 39:
```suggestion
                    new { Timestamp = DateTime.UtcNow, Name = "AspNetCoreVersion", Value = typeof(IWebHostBuilder).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion },
```
We plan to obsolete WebHostBuilder.

samples/BenchmarkApp/Program.cs
```diff
@@ -19,5 +23,25 @@ public static void Main(string[] args)
                 {
                     webBuilder.UseStartup<Startup>();
                 });
+
+        private static void WriteStatistics()
+        {
+            Console.WriteLine("#StartJobStatistics");
```
sebastienros, line 29: I think we should also use a single call to `Console.WriteLine` or other messages could be interleaved. Pranav had that issue. I will change it on my side too.

Issue #172, opened 2020-05-18
Initial preview1 docs


Tratcher:
We should be able to seed the docs from the readme and preview1 blog post content.

- [x] Introduction
- [ ] Roadmap
- [x] Getting started sample
- [ ] Config based routes
- [ ] Supported runtimes (3.1 with caveats, 5.0-previews)

karelz:
Fixed in #178

PR #178, opened 2020-05-19
Getting started doc content

Tratcher:
Contributes to #172. This is largely the same as the preview1 blog content.

I can't tell if I've broken the links. The local build output does not quite match the CI build output.

### DIFF
--- a/docs/docfx/articles/getting_started.md
+++ b/docs/docfx/articles/getting_started.md
@@ -84,11 +84,13 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
 The configuration for YARP is defined in the appsettings.json file. It defines a set of:
 
-Routes - which map incoming requests to the backend clusters based on aspects of the request such as host name, path, method, request headers etc. Specifying a Host is the only required field. Routes are ordered, so the "app1" route needs to be defined first as  "route2" will act as a catchall for all paths that have not already been matched. 
+Routes - which map incoming requests to the backend clusters based on aspects of the request such as host name, path, method, request headers etc. Specifying a Host is the only required field. Routes are ordered, so the "app1" route needs to be defined first as "route2" will act as a catchall for all paths that have not already been matched. 
 
 Backends - which are the clusters of destination servers that requests can be routed to and load balanced across.
 
-Address is the URI prefix that will have the original request path appended to it 
+Address is the URI prefix that will have the original request path appended to it.
+
+You can find out more about the available configuration options by looking at [ProxyConfigOptions](xref:Microsoft.ReverseProxy.Configuration.ProxyConfigOptions).
  
  ```
  {
--- a/docs/docfx/articles/getting_started.md
+++ b/docs/docfx/articles/getting_started.md
@@ -43,6 +43,23 @@ And then add:
 
 YARP is implemented as a ASP.NET Core component, and so the majority of the sample code is in Startup.cs. 
 
+YARP currently uses configuration files to define the routes and endpoints for the proxy. That is loaded in the `ConfigureServices` method. 
+
+```
+public IConfiguration Configuration { get; }
+
+public Startup(IConfiguration configuration)
+{
+    Configuration = configuration;
+}
+
+public void ConfigureServices(IServiceCollection services) 
+{ 
+    services.AddReverseProxy() 
+        .LoadFromConfig(Configuration.GetSection("ReverseProxy")); 
+} 
+```
+
 The `Configure` method defines the ASP.NET pipeline for processing requests. The reverse proxy is plugged in to ASP.NET endpoint routing, and then has its own sub pipeline for the proxy. Here proxy pipeline modules, such as load balancing can be added to customize the handling of the request. 
 ```
 public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
@@ -60,23 +77,6 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             proxyPipeline.UseProxyLoadBalancing(); 
         }); 
     }); 
-} 
-```
-
-YARP currently uses configuration files to define the routes and endpoints for the proxy. That is loaded in the `ConfigureServices` method. 
-
-```
-public IConfiguration Configuration { get; }
-
-public Startup(IConfiguration configuration)
-{
-    Configuration = configuration;
-}
-
-public void ConfigureServices(IServiceCollection services) 
-{ 
-    services.AddReverseProxy() 
-        .LoadFromConfig(Configuration.GetSection("ReverseProxy")); 
 } 
 ```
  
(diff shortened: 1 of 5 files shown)


docs/docfx/articles/getting_started.md
```diff
@@ -0,0 +1,146 @@
...
+
+Or create a new ASP.NET Core web application in Visual Studio, and choose "Empty" for the project template. 
+
+### Update the project file
+
+Open the Project and make sure it includes the appropriate target framework: 
+ 
+ ```
+<PropertyGroup>
+  <TargetFramework>netcoreapp5.0</TargetFramework>
```
BrennanConroy, line 30: nit: net5.0

Tratcher: The preview4 templates still use netcoreapp5.0.

docs/docfx/articles/getting_started.md
```diff
@@ -0,0 +1,148 @@
+---
+uid: getting_started
+title: Getting Started with YARP
+---
+
+# Getting Started with YARP
+
+YARP is designed as a library that provides the core proxy functionality which you can then customize by adding or replacing modules. YARP is currently supplied as a nuget package and code snippets. We plan on having a project template, and pre-built exe in the future.
```
halter73, line 8: Nit:

```suggestion
YARP is designed as a library that provides the core proxy functionality which you can then customize by adding or replacing modules. YARP is currently provided as a NuGet package and code snippets. We plan on providing a project template and pre-built exe in the future. 
```

docs/docfx/articles/getting_started.md
```diff
@@ -0,0 +1,148 @@
...
+}
+
+public void ConfigureServices(IServiceCollection services) 
+{ 
+    services.AddReverseProxy() 
+        .LoadFromConfig(Configuration.GetSection("ReverseProxy")); 
+} 
+```
+
+The `Configure` method defines the ASP.NET pipeline for processing requests. The reverse proxy is plugged in to ASP.NET endpoint routing, and then has its own sub pipeline for the proxy. Here proxy pipeline modules, such as load balancing can be added to customize the handling of the request.
```
halter73, line 63: Nit:

```suggestion
The `Configure` method defines the ASP.NET pipeline for processing requests. The reverse proxy is plugged into ASP.NET endpoint routing and then has its own sub-pipeline for the proxy. Here proxy pipeline modules, such as load balancing, can be added to customize the handling of the request. 
```

docs/docfx/articles/getting_started.md
```diff
@@ -0,0 +1,148 @@
...
+ 
+### Configuration 
+
+The configuration for YARP is defined in the appsettings.json file. It defines a set of:
+
+Routes - which map incoming requests to the backend clusters based on aspects of the request such as host name, path, method, request headers etc. Specifying a Host is the only required field. Routes are ordered, so the "app1" route needs to be defined first as "route2" will act as a catchall for all paths that have not already been matched. 
+
+Backends - which are the clusters of destination servers that requests can be routed to and load balanced across.
+
+Address is the URI prefix that will have the original request path appended to it.
```
halter73, line 91: Should we also call out the query string here?

```suggestion
Address is the URI prefix that will have the original request path and query string appended to it.
```

docs/docfx/index.md
```diff
@@ -5,6 +5,12 @@ title: YARP Documentation
 
 # YARP: A Reverse Proxy
 
-Welcome to the documentation for YARP! YARP is a high-performance, production-ready, reverse proxy written for .NET 5. Right now it's still in preview, but please provide us your feedback by going to [the GitHub repository](https://github.com/microsoft/reverse-proxy).
+Welcome to the documentation for YARP! YARP is a project and toolkit to help create reverse proxy servers that are high-performance, production-ready, and highly customizable. Right now it's still in preview, but please provide us your feedback by going to [the GitHub repository](https://github.com/microsoft/reverse-proxy).
```
halter73, line 8: What's the difference between a project and a toolkit? Is YARP a toolkit because we plan to ship a project template? Do we have .NET CLI tools planned?

Should we just call it a library like we do in getting_started.md?

docs/docfx/index.md
```diff
@@ -5,6 +5,12 @@ title: YARP Documentation
...
\ No newline at end of file
+We found a bunch of internal teams at Microsoft who were either building a reverse proxy for their service or had been asking about APIs and tech for building one, so we decided to get them all together to work on a common solution, this project.
+
+YARP is built on .NET using the infrastructure from ASP.NET and .NET (.NET Core 3.1 and .NET 5.0). The key differentiator for YARP is that it's been designed to be easily customized and tweaked to match the specific needs of each deployment scenario. 
+
+We expect YARP to ship as a library and project template that together provide a robust, performant proxy server. Its pipeline and modules are designed so that you can then customize the functionality for your needs. For example, while YARP supports configuration files, we expect that many users will want to manage the configuration programmatically based on their own backend configuration management system, YARP will provide a configuration API to enable that customization in-proc.  YARP is designed with customizability as a primary scenario, rather than requiring you to break out to script or having to rebuild from source.
```
halter73, line 14: Nit: comma splice

```suggestion
We expect YARP to ship as a library and project template that together provide a robust, performant proxy server. Its pipeline and modules are designed so that you can then customize the functionality for your needs. For example, while YARP supports configuration files, we expect that many users will want to manage the configuration programmatically based on their own backend configuration management system. YARP will provide a configuration API to enable that customization in-proc. YARP is designed with customizability as a primary scenario rather than requiring you to break out to script or having to rebuild from source.
```

docs/docfx/articles/getting_started.md
```diff
@@ -0,0 +1,148 @@
+---
+uid: getting_started
+title: Getting Started with YARP
+---
+
+# Getting Started with YARP
+
+YARP is designed as a library that provides the core proxy functionality which you can then customize by adding or replacing modules. YARP is currently supplied as a nuget package and code snippets. We plan on having a project template, and pre-built exe in the future. 
+
+YARP supports ASP.NET Core 3.1 or 5.0 preview4 or later. You can download preview 4 of .NET 5 SDK from https://dotnet.microsoft.com/download/dotnet/5.0.
```
halter73, line 10: Is and better than or here?

```suggestion
YARP supports ASP.NET Core 3.1 and 5.0.0-preview.4 or later. You can download preview 4 of .NET 5 SDK from https://dotnet.microsoft.com/download/dotnet/5.0.
```

Tratcher: I've applied all of the feedback except the 5.0.0-preview.4 name change. 5.0.0-preview.4 is the format used in file names, not the name used when discussing the release. 
https://devblogs.microsoft.com/aspnet/asp-net-core-updates-in-net-5-preview-4/

PR #181, opened 2020-05-21
[Getting Started] Clarifying word usage

rwkarg:
I though it was saying that `"app1" needs to be defined as the value "route2"`


Tratcher:
Thanks

PR #182, opened 2020-05-21
Document supported runtimes

Tratcher:
#172 Starting this page to track what is and isn't supported across runtime versions.

### DIFF
--- a/docs/docfx/articles/runtimes.md
+++ b/docs/docfx/articles/runtimes.md
@@ -16,4 +16,6 @@ These are related improvements in .NET or ASP.NET Core 5.0.0 that YARP is able t
 - Kestrel HTTP/2 performance improvements.
   - [HPACK static compression](https://github.com/dotnet/aspnetcore/pull/20058)
   - [HPACK dynamic compression](https://github.com/dotnet/aspnetcore/pull/19521).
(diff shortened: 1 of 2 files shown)


docs/docfx/articles/runtimes.md
```diff
@@ -0,0 +1,19 @@
...
+
+YARP will be taking advantage of 5.0.0 features and optimizations as they become available. This does mean that some features may not be available if you're running on 3.1.
+
+## Related 5.0.0 Runtime Improvements
+
+These are related improvements in .NET or ASP.NET Core 5.0.0 that YARP is able to take advantage of. We expect to add more as they become available.
+- Kestrel [reloadable config](https://github.com/dotnet/aspnetcore/issues/19376).
+- Kestrel HTTP/2 performance improvements.
+  - [HPACK static compression](https://github.com/dotnet/aspnetcore/pull/20058)
+  - [HPACK dynamic compression](https://github.com/dotnet/aspnetcore/pull/19521).
```
halter73, line 18: We should also call out that we allocate less for HTTP/2 than before (amortized) because we pool stream objects now.

https://github.com/dotnet/aspnetcore/pull/18601
https://github.com/dotnet/aspnetcore/pull/19356

Issue #189, opened 2020-05-22
Change framework to net5.0


TeddyAlbina:
### Follow dotnet 5 framework identifier evolution

I have Visual Studio 2019 16.6 the installed dotnet framework 5.0.100-preview.4.20258.7 wich have the framework identifier .net5 instead of netcoreapp5.0 .

So it cannot loads the projects within the solution,

Tratcher:
Note we have some scripts in the repo to address this like restore.cmd that downloads a specific version of the SDK to a local .dotnet directory and startvs.cmd that starts VS using that local SDK. With these local versions I'm able to open the repo solution in VS 16.7.0-p1.

But yes, we do still need to move everything to net5.0.

PR #192, opened 2020-05-22
[Getting Started] Updating TFM for .NET 5

rwkarg:
https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md#design

I've only updated this documentation for consuming the package. 

It still appears that the projects in the repo need to be updated as #189 but that's a separate concern from this documentation.

### DIFF
--- a/docs/docfx/articles/getting_started.md
+++ b/docs/docfx/articles/getting_started.md
@@ -17,7 +17,7 @@ Start by creating an "Empty" ASP.NET Core application using the command line:
 dotnet new web -n MyProxy 
 ```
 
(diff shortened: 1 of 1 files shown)


rwkarg:
I'm using VS 16.7.0 Preview 1.0 and the 5.0.100-preview.4.20258.7 SDK. 

I haven't updated my non-Preview VS in a while...

Tratcher:
So both TargetFramework net5.0 and netcoreapp5.0 work fine in VS 16.6. However, `-f net5.0` doesn't work with dotnet new. Lets hold off on updating the docs until dotnet new is fixed in the next preview.

PR #201, opened 2020-05-23
Fix capitalization of activate.sh error message command suggestion

nemec:
`activate.sh` prompts to install .NET Core with a shell script, however the `Restore.sh` file on disk begins with a capital letter. Since .sh shell scripts are usually run in Linuxy environments and Linuxy environments generally have case-sensitive filesystems, the prompt should also match the file capitalization.

It's a very minor change, but hopefully it will reduce friction for devs trying to get this project to build for the first time.

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -88,7 +88,7 @@ fi
 printf "${_MAGENTA}Enabled the .NET Core environment. Execute 'deactivate' to exit.${_RESET}\n"
 
 if [ ! -f "$DOTNET_ROOT/dotnet" ]; then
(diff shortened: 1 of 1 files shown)


Tratcher:
> I'd also be OK with just switching the scripts to lowercase in repo.

Yeah, all of the scripts in the repo root should be renamed to lower case, cmd and sh.

halter73:
Thanks for the contribution, but I think we're going just change the scripts names to be lowercase.

halter73:
https://github.com/microsoft/reverse-proxy/pull/208

nemec:
All lowercase is my preference, too, but I didn't want to rock the boat 😆
Thanks @halter73!

PR #217, opened 2020-06-01
Moved to use of IRandomFactory in ProxyInvokerMiddleware

### LINKED ISSUE #186: Non thread safe Random is used in ProxyInvokerMiddleware
### Describe the bug
Non thread safe Random is used in ProxyInvokerMiddleware. Is it some kind of intentional? Should IRandomFactory be used instead of direct Random class instantiation?

### To Reproduce
Look inside code base :)

### Further technical details
None

Insomniak47:
closes #186 

Saw this one show up in my email and figured it was an easy fix. Migrated the initialization to use a RandomFactory and then grab a new instance of System.Random from there.

### DIFF
--- a/src/ReverseProxy/Middleware/ProxyInvokerMiddleware.cs
+++ b/src/ReverseProxy/Middleware/ProxyInvokerMiddleware.cs
@@ -19,7 +19,7 @@ namespace Microsoft.ReverseProxy.Middleware
     /// </summary>
     internal class ProxyInvokerMiddleware
     {
-        private readonly IRandomFactory _randomFactory = new RandomFactory();
+        private readonly IRandomFactory _randomFactory;
         private readonly RequestDelegate _next; // Unused, this middleware is always terminal
         private readonly ILogger _logger;
         private readonly IOperationLogger<ProxyInvokerMiddleware> _operationLogger;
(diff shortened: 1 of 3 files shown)


src/ReverseProxy/Middleware/ProxyInvokerMiddleware.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.ReverseProxy.Middleware
     /// </summary>
     internal class ProxyInvokerMiddleware
     {
-        private readonly Random _random = new Random();
+        private readonly IRandomFactory _randomFactory = new RandomFactory();
```
Kahbazi, line 22: How about injecting `IRandomFactory` in the constructor?

Insomniak47: I was thinking that as well but I noticed it was inconsistent in the codebase. I'll make the fix, would you like me to also standardize that in:

```
src\ReverseProxy\Service\HealthProbe\BackendProberFactory.cs
```

as well?

Tratcher: Yes, please update BackendProberFactory as well.

Insomniak47: Completed in 3e1762d

Issue #60, opened 2020-04-15
Custom request/response modifications via code transforms


samsp-msft:
Ability to customize headers (beyond x-forward or affinity) from code extensibility for both request to back end, and response to the the client.

Tratcher:
See feedback from https://github.com/microsoft/reverse-proxy/pull/225#discussion_r441143683 about config formats and how that may affect extensibility.

AJAnderhub:
I would second the need for custom header modifications. Prepending to response headers would the use case for my scenario.

robbieknuth:
+1, I have a need to modify headers in an ad-hoc fashion. Concrete scenario is ripping an auth token from a cookie and setting that as the Authorization header. This would eliminate my need for CORS entirely and simplify my api server to be jwt only, resulting in all client-ish things on the client server.

Tratcher:
Requirements summary:
- Someone can write code to modify request and response parameters/headers.
- That code can be associated with specific routes via config, including parameters.
- Config validation can validate parameters for transforms.
- Transforms compose so you don't have to write one uber transform.
- Transformations need to allow async code.

Example scenarios:
- Swap auth format: https://github.com/microsoft/reverse-proxy/issues/416

PR #225, opened 2020-06-03
Request and response transforms

Tratcher:
https://github.com/microsoft/reverse-proxy/issues/21, https://github.com/microsoft/reverse-proxy/issues/13

I'm posting this partial work for some early feedback on the design. I have most of the infrastructure set up and some common scenarios implemented.

Principals:
- Keep the config as simple as possible, use default config bindings.
- Transforms are configured per route. They're built once when setting up the route and then run per-request.
- Avoid mutating the HttpRequest, we may need to retry the request.
- Apply transforms before creating the HttpRequestMessage, it does not provide an efficient/granular object model for manipulating the url, and the headers are awkwardly split between request and content.
- Avoid making extra copies of collections like Headers. Instead, transform them in flight while copying from HttpRequest to HttpRequest message.

Working:
- Request paths, simple and route pattern based
- Request headers
- Option to allow/suppress all request headers.
- Option to allow/exclude the original host.
- Response headers, trailers
- Moving the x-forwarded-headers to the new infrastructure and making them configurable.
- Granular tests
- Forwarded rfc7239
- Documentation

Not started:
- Factory support for custom transforms. https://github.com/microsoft/reverse-proxy/issues/60 I'll break this out into another PR. That work will focus around the TransformBuilder and adding extensible factories.

Open questions:
- Do we need more than one transform per request/response header name? I couldn't come up with an example and it would complicate the object model.


src/ReverseProxy/Service/Config/ITransformBuilder.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using Microsoft.ReverseProxy.Service.RuntimeModel.Transforms;
+
+namespace Microsoft.ReverseProxy.Service.Config
+{
+    public interface ITransformBuilder
+    {
+        void Build(IList<IDictionary<string, string>> rawTransforms, out Transforms transforms);
```
davidfowl, line 11: `List<IDictionary<string,string>>` because order matters? Why not `IList<KeyValuePair<..>>`

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -430,6 +471,30 @@ private void CopyHeadersToUpstream(HttpContext context, HttpRequestMessage desti
             {
                 destination.Headers.TryAddWithoutValidation(ForwardedHeadersDefaults.XForwardedForHeaderName, context.Connection.RemoteIpAddress.ToString());
             }
+
+            // Note: HttpClient.SendAsync will end up sending the union of
+            // HttpRequestMessage.Headers and HttpRequestMessage.Content.Headers.
+            // We don't really care where the proxied headers appear among those 2,
+            // as long as they appear in one (and only one, otherwise they would be duplicated).
+            void AddHeader(HttpRequestMessage request, string headerName, StringValues value)
```
davidfowl, line 479: static

src/ReverseProxy/Service/Config/ITransformBuilder.cs
Tratcher: I started with IList of KeyValuePair but the config binder didn't support it, it only supports IList of Dictionary. Using Dictionary also ended up being useful for some of the later transforms like response headers that have a more optional parameters. If you only had KeyValuePair then you have to encode more information in the names, and possibly parse the values.

Order may matter in theory, I'm not sure how much it will matter in practice. E.g. if you want to both remove a path prefix and add another one there might be an ordering conflict, but there are probably other ways to handle that kind of transformation with a single rule.

src/ReverseProxy/Service/Config/DynamicConfigBuilder.cs
```diff
@@ -124,8 +127,11 @@ private async Task<IList<ParsedRoute>> GetRoutesAsync(IConfigErrorReporter error
                     Priority = route.Priority,
                     BackendId = route.BackendId,
                     Metadata = route.Metadata,
+                    Transforms = route.Transforms,
                 };
 
+                // TODO: validate transforms? What about custom ones?
```
halter73, line 133: How do developers register custom transforms? I suppose they could replace IHttpProxy or even ITransform builder and inject custom transforms before calling the base implementation, but that's might inconvenient. Even customizing the set of custom transforms by route id would be nontrivial this way.

I think we should provide a first-class way register custom transforms that integrates nicely with config similar to what we've talked about doing for custom load balancers (#106).

I think the way to make custom transforms and config play nicely together is to allow developers to configure their custom transforms in "Routes[].Transforms" subsections just like default transforms. This could be done by using a special top-level property in the custom subsections like "CustomTransformName" to distinguish them from built-in transforms.

Developers could then register their custom transforms as a service that looks something like the following:

```C#
public interface ITransformProvider
{
     // transformName == configSection["CustomTransformName"], but I think it's convenient.
     Transforms GetTransforms(string transformName, IConfigurationSection configSection);
}
```

We'd have to write logic to properly merge Transforms, but I don't think that would be too difficult and it would only need to be done once per config change.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -340,6 +334,47 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
...
+            if (transforms == null || transforms.Count == 0)
+            {
+                // TODO: destination vs request PathBase policy?
+                var url = UriHelper.BuildAbsolute(destinationScheme, destinationHost, destinationPathBase, request.Path, request.QueryString);
+                Log.Proxying(_logger, url);
+                var uri = new Uri(url, UriKind.Absolute);
+                return new HttpRequestMessage(HttpUtilities.GetHttpMethod(context.Request.Method), uri);
+            }
+
+            var transformContext = new RequestParametersTransformContext()
```
halter73, line 358: I wonder if this wouldn't be better as a struct passed by reference.

src/ReverseProxy/Service/Config/DynamicConfigBuilder.cs
Tratcher: Yeah, I'll be working on this extensibility as part of the Forwarders support.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
```diff
@@ -45,6 +48,7 @@ ProxyRoute IDeepCloneable<ProxyRoute>.DeepClone()
                 Priority = Priority,
                 BackendId = BackendId,
                 Metadata = Metadata?.DeepClone(StringComparer.Ordinal),
```
alnikola, line 50: Do we really want to keep StringComparer.Ordinal here since on the next line we already use OrdinalIgnoreCase?

src/ReverseProxy/Service/Config/TransformBuilder.cs
```diff
@@ -0,0 +1,137 @@
...
+            bool? useOriginalHost = null;
+            var requestTransforms = new List<RequestParametersTransform>();
+            var requestHeaderTransforms = new Dictionary<string, RequestHeaderTransform>();
+            var responseHeaderTransforms = new Dictionary<string, ResponseHeaderTransform>();
+            var responseTrailerTransforms = new Dictionary<string, ResponseHeaderTransform>();
+
+            foreach (var rawTransform in rawTransforms)
+            {
+                // TODO: Ensure path string formats like starts with /
+                if (rawTransform.TryGetValue("PathPrefix", out var pathPrefix))
```
alnikola, line 40: It seems there is no check that a rawTransform contains only one transformation definition, so the developer can by mistake put several transform definitions into one JSON object (see  below) which will lead to unpredictable behavior because the picked up transformation will depend on the order of if statements here. Does it make sense to add validation that each rawTransform contains only 1 transformation?

In example below, RequestHeader transformation will be **ignored** in the following case because `rawTransform.TryGetValue("PathRemovePrefix", out var pathRemovePrefix)` called first.
```JSON
"Transforms": [
    {
        "RequestHeader": "foo0",
        "Append": "bar",
        "PathRemovePrefix": "/apis"
    }
]
```

src/ReverseProxy/Service/DynamicEndpoints/RuntimeRouteBuilder.cs
```diff
@@ -16,8 +18,14 @@ namespace Microsoft.ReverseProxy.Service
     /// </summary>
     internal class RuntimeRouteBuilder : IRuntimeRouteBuilder
     {
+        private readonly ITransformBuilder _transformBuilder;
         private RequestDelegate _pipeline;
 
+        public RuntimeRouteBuilder(ITransformBuilder transformBuilder)
+        {
+            _transformBuilder = transformBuilder;
```
alnikola, line 26: Ctor parameter needs to be checked for null.

src/ReverseProxy/Service/RuntimeModel/Transforms/PathRouteValueTransform.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// Generates a new request path by plugging matched route parameters into the given pattern.
+    /// </summary>
+    public class PathRouteValueTransform : RequestParametersTransform
+    {
+        private readonly TemplateBinder _binder;
+
+        public PathRouteValueTransform(string pattern, TemplateBinderFactory binderFactory)
+        {
+            // TODO: Config validation
+            _binder = binderFactory.Create(RoutePatternFactory.Parse(pattern));
```
alnikola, line 19: Ctor parameter validation is missing.

src/ReverseProxy/Service/RuntimeModel/Transforms/Transforms.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Microsoft.ReverseProxy.Service.RuntimeModel.Transforms
+{
+    public class Transforms
+    {
+        public Transforms(IReadOnlyList<RequestParametersTransform> requestTransforms, bool? copyRequestHeaders,
+            IReadOnlyDictionary<string, RequestHeaderTransform> requestHeaderTransforms,
+            IReadOnlyDictionary<string, ResponseHeaderTransform> responseHeaderTransforms,
+            IReadOnlyDictionary<string, ResponseHeaderTransform> responseTrailerTransforms)
+        {
+            CopyRequestHeaders = copyRequestHeaders;
```
alnikola, line 15: Ctor parameters validation is missing.

src/ReverseProxy/Service/Config/TransformBuilder.cs
Tratcher: Right. I do plan to add a validation code path that runs during config route validation to catch errors like that.

samples/ReverseProxy.Sample/appsettings.json
```diff
@@ -47,8 +50,51 @@
         "Match": {
           "Methods": [ "GET", "POST" ],
           "Host": "localhost",
-          "Path": "/{**catchall}"
-        }
+          "Path": "/api/{plugin}/stuff/{*remainder}"
+        },
+        "Transforms": [
+          { "PathPattern": "/foo/{plugin}/bar/{remainder}" }
+          // { "RequestHeader": "foo3", "SetRouteValue": "plugin" }, // Binds to route parameter
```
samsp-msft, line 57: IMHO, this is where we are trying to produce our own language for the header transforms, and I would rather spend the effort on enabling a code callback instead that enables the dev to mangle headers as they need to in code which is going to be more flexible and powerful.

The main problem is how to hook that up via DI or reflection. Ideally I'd be able to create my own methods that match a delegate signature or an interface, and have that called being passed the header collections, with the ability to write to the outgoing one. 
Probably a separate method for each direction.

samples/ReverseProxy.Sample/appsettings.json
```diff
@@ -47,8 +50,51 @@
...
+          "Host": "localhost"
+        },
+        "Transforms": [
+          {
+            "X-Forwarded": "proto,host,for,pathbase",
+            "Append": "true",
+            "Prefix": "X-Forwarded-"
+          }, // TODO: Mode: Append/AppendLine/Set
+          /*{ "Forwarded": "by,host,for,proto" },*/ // https://tools.ietf.org/html/rfc7239
+          { "ClientCert": "X-Client-Cert" }
```
samsp-msft, line 73: Is this setting the x-client-cert to be something from the request?

Tratcher: Yes, the base64 encoded client certificate (if any).

davidfowl: I think we should do a bigger review of the entire configuration once we have most of the core functionality in place.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
WeihanLi: Could we have a config for a backend to use case-sensitive or case-insensitive route?

samples/ReverseProxy.Sample/appsettings.json
Tratcher: Indeed. The question for now is if these are even remotely the right shape? Are they understandable?

ClientCert is a bit vague. Maybe "ClientCertForwardedHeader", or something more verbose like this?
```
{ "ForwardClientCertificate": "true",
  "HeaderName": "X-Client-Cert" }
```

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
Tratcher: @WeihanLi Do you mean a case-sensitive/insensitive path matching for the route? We're using AspNetCore's routing which does [case-insensitive](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.1#route-template-reference) path matching. Please open an issue in https://github.com/dotnet/aspnetcore if you want case sensitive support.

Tratcher:
After reviewing the schedule I'm going to add the tests and wrap this up. The work we've done so far is important to land in preview2 and fulfills many customer asks. I'll come back to the extensibility model for this after a few other high priority items like auth.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
Tratcher: @alnikola I'm going to leave it since it's not part of this PR. I needed transforms to be different because Ordinal was actively breaking the transforms.

samples/ReverseProxy.Sample/appsettings.json
Tratcher: I agree that the code extensibility will be critical, but there's a certain minimum functionality that's expected to be available standard via config. I want to make sure we have those basics covered early since those are the common cases our early customers are asking about. E.g. Mess with the path, add and remove headers, and x-forwarded-*. While we may add a few more built in rules, I expect all more complicated transforms to be added via extensibility and then referenced by name here in this list.

src/ReverseProxy/Service/Config/TransformBuilder.cs
```diff
@@ -0,0 +1,419 @@
...
+using Microsoft.AspNetCore.HttpOverrides;
+using Microsoft.AspNetCore.Routing.Template;
+using Microsoft.Net.Http.Headers;
+using Microsoft.ReverseProxy.Abstractions;
+using Microsoft.ReverseProxy.Service.RuntimeModel.Transforms;
+
+namespace Microsoft.ReverseProxy.Service.Config
+{
+    /// <inheritdoc/>
+    public class TransformBuilder : ITransformBuilder
```
Tratcher, line 16: Extensibility plan: TransformBuilder will be broken down into a factory for each unique rule type. Each factory will own its own Validate and Build APIs. The TransformBuilder will primarily be responsible for looping through the factories and transform inputs.

Tratcher:
I've completed the work for the Forwarded RFC. That's everything I'd hoped to accomplish in this PR, with the extensibility bumped to next milestone.

@halter please review.

docs/docfx/articles/transforms.md
```diff
@@ -0,0 +1,341 @@
...
+          },
+          { "ClientCert": "X-Client-Cert" },
+          { "RequestHeadersCopy": "true" },
+          { "RequestHeaderOriginalHost": "true" },
+          {
+            "X-Forwarded": "proto,host,for,pathbase",
+            "Append": "true",
+            "Prefix": "X-Forwarded-"
+          }
+        ]
```
halter73, line 49: How important is order to the transforms? This could read a lot better if we went from a top-level array to an object. We could still use arrays in subsections where needed. E.g:

```json
"Transforms": {
  "PathPrefix": "/apis",
  "RequestHeaders": {
    "Copy": true,
    "CopyHost": true,
    "ClientCert": "X-Client-Cert",
    "X-Forwarded": {
        "Prefix": "X-Forwarded-",
        "List": "proto,host,for,pathbase",
        "Append": true
    },
    "Append": {
      "header1": "bar"
    },
    "Set": {
      "header3": "baz"
    }
  },
  "ReponseHeaders": {
    "Append": {
      "header2": {
        "Value": "bar",
        "When": "Always"
      }
    }
  }
}
```

Think about how much more self-documenting the json schema for transform would be if we moved from an array to a top level object with defined keys.

The way it's designed now, it's super hard to determine the "type" of any transformation by looking at the transformation object especially while it's still being authored. Trying to make autocomplete suggestions in an editor as you start to write a transform wouldn't be very helpful because there are so many valid keys coming from about a dozen transform types available.

docs/docfx/articles/transforms.md
```diff
@@ -0,0 +1,341 @@
...
+
+This transform is enabled by default even if not specified in the route config.
+
+Set the `X-Forwarded` value to a comma separated list containing the headers you need to enable. All for headers are enabled by default. All can be disabled by specifying an empty value `""`.
+
+The Prefix specifies the header name prefix to use for each header. With the default `X-Forwarded-` prefix the resulting headers will be `X-Forwarded-For`, `X-Forwarded-Proto`, `X-Forwarded-Host`, and `X-Forwarded-PathBase`.
+
+Append specifies if each header should append to or replace an existing header of the same name. A request traversing multiple proxies may accumulate a list of such headers and the destination server will need to evaluate the list to determine the original value. If append is false and the associated value is not available on the request (e.g. RemoteIpAddress is null), any existing header is still removed to prevent spoofing.
+
+The {Prefix}For header value is taken from `HttpContext.Connection.RemoteIpAddress` representing the prior caller's IP address. The port is not included. IPv6 addresses do not include the bounding `[]` brackets.
```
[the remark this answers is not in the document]
Tratcher: I opted not to include the port or provide config for that yet. It was not common in other implementations I looked at. We can see what feedback we get.

The Forwarded header spec explicitly allows it so I made it configurable there.

docs/docfx/articles/transforms.md
```diff
@@ -0,0 +1,341 @@
...
+
+ByFormat and ForFormat:
+
+The RFC allows a [variety of formats](https://tools.ietf.org/html/rfc7239#section-6) for the By and For fields. It requires that the default format uses an obfuscated identifier identified here as Random.
+
+| Format | Description | Example |
+|--------|-------------|---------|
+| Random | An obfuscated identifier that is generated randomly per request. This allows for diagnostic tracing scenarios while limiting the flow of uniquely identifying information for privacy reasons. | `by=_YQuN68tm6` |
+| RandomAndPort | The Random identifier plus the port. | `by="_YQuN68tm6:80"` |
+| Unknown | This can be used when the identity of the preceding entity is not known, but the proxy server still wants to signal that the request was forwarded. | `by=unknown` |
```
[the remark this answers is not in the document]
Tratcher: No, `unknown` is the literal value specified in the spec. https://tools.ietf.org/html/rfc7239#section-6.2
https://tools.ietf.org/html/rfc7239#section-6
       nodename = IPv4address / "[" IPv6address "]" /
                   "unknown" / obfnode

docs/docfx/articles/transforms.md
```diff
@@ -0,0 +1,341 @@
...
+{ "ClientCert": "X-Client-Cert" }
+```
+
+This transform causes the client certificate taken from `HttpContext.Connection.ClientCertificate` to be Base64 encoded and set as the value for the given header name. This is needed because client certificates from incoming connections are not used when making connections to the destination server. The destination server may need that certificate to authenticate the client. There is no standard that defines this header and implementations vary, check your destination server for support.
+
+### Response Headers and Trailers
+
+All response headers and trailers are copied from the proxied response to the outgoing response. Response header and trailer transforms may specify if they should be applied only for successful responses of for all responses.
+
+In code these are implemented as derivations of the abstract class [ResponseHeaderTransform](xref:Microsoft.ReverseProxy.Service.RuntimeModel.Transforms.ResponseHeaderTransform).
```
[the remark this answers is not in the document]
Tratcher: Nope, that's what I'll tackle next in #60.

docs/docfx/articles/transforms.md
```diff
@@ -0,0 +1,341 @@
...
+
+#### ResponseHeader
+
+| Key | Value | Default | Required |
+|-----|-------|---------|----------|
+| ResponseHeader | The header name | (none) | yes |
+| Set/Append | The header value | (none) | yes |
+| When | Success/Always | Success | no |
```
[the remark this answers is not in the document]
Tratcher: I mirrored the nginx capability here and they only allowed the two cases. This is certainly the type of thing I'd be curious to hear customer feedback on.

src/ReverseProxy/Service/Config/TransformBuilder.cs
```diff
@@ -0,0 +1,552 @@
...
+                    if (!string.Equals("True", originalHost, StringComparison.OrdinalIgnoreCase) && !string.Equals("False", originalHost, StringComparison.OrdinalIgnoreCase))
+                    {
+                        errorReporter.ReportError(ConfigErrors.TransformInvalid, routeId, $"Unexpected value for RequestHeaderOriginalHost: {originalHost}. Expected 'true' or 'false'");
+                        success = false;
+                    }
+                }
+                else if (rawTransform.TryGetValue("RequestHeader", out var headerName))
+                {
+                    success &= TryCheckTooManyParameters(rawTransform, routeId, expected: 2, errorReporter);
+                    if (!rawTransform.TryGetValue("Set", out var _) && !rawTransform.TryGetValue("Append", out var _))
```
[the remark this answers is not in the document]
Tratcher: That would fail the parameter count check (TryCheckTooManyParameters).

samples/ReverseProxy.Sample/appsettings.json
```diff
@@ -51,8 +54,61 @@
...
+        "BackendId": "backend1",
+        "Match": {
+          "Host": "localhost"
+        },
+        "Transforms": [
+          {
+            "X-Forwarded": "proto,host,for,pathbase",
+            "Append": "true",
+            "Prefix": "X-Forwarded-"
+          }, // TODO: Mode: Append/AppendLine/Set
```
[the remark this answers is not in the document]
Tratcher: Right now all of the append operations use StringValues.Concat which effectively gets you an additional header. I don't know if people need the granularity here to distinguish between adding a new line vs adding a comma separated value. It's functionally equivalent, but some clients are picky about that kind of formatting.

docs/docfx/articles/transforms.md
Tratcher: Order matters for things like PathRemovePrefix and PathPrefix. To avoid that ordering dependency you'd have to create combine rule for PathReplacePrefix.

For something like QueryStringAppend you'd want those to be strictly ordered.

The header transforms aren't ordered.

The other major limiter here is config binding. The most generic structure I could get config to bind was `IList<IDictionary<string, string>>`. I don't think there's a generic, extensible structure that can bind top level objects like that from config. Note we also want the same generic structure to be representable in other stores like databases.

docs/docfx/articles/transforms.md
```diff
@@ -0,0 +1,341 @@
...
+      "backend1": {
+        "Destinations": {
+          "backend1/destination1": {
+            "Address": "https://localhost:10001/Path/Base"
+          }
+        }
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
Tratcher: Examples are good. I'd added one for Forwarded, I'll add a few more.

docs/docfx/articles/transforms.md
Tratcher: Offline discussion: This has some readability advantages but complicates extensibility and config binding. It's too big a design change to accommodate in preview2. I expect we'll revisit this when working on the extensibility (#60).

PR #234, opened 2020-06-08
Making hostname optional if request path is specified for the route

### LINKED ISSUE #163: Remove the restriction that every route specify a Host
https://github.com/microsoft/reverse-proxy/blob/20d50978e3aea16b0e5e56f9fd1bdb4f57a28c03/src/ReverseProxy/Service/Config/RuleParsing/RouteValidator.cs#L67-L72

The host is not needed when you want to proxy all requests, all remaining requests, or all requests based on path.

ihvo:
Resolves #163

### DIFF
--- a/src/ReverseProxy/Service/Config/ConfigErrors.cs
+++ b/src/ReverseProxy/Service/Config/ConfigErrors.cs
@@ -16,6 +16,7 @@ namespace Microsoft.ReverseProxy.Service
         internal const string RouteUnsupportedAction = "Route_UnsupportedAction";
 
         internal const string ParsedRouteMissingId = "ParsedRoute_MissingId";
+        internal const string ParsedRouteMissingHostAndPath = "ParsedRoute_MissingHostAndPath";
         internal const string ParsedRouteRuleHasNoMatchers = "ParsedRoute_RuleHasNoMatchers";
         internal const string ParsedRouteRuleMissingHostMatcher = "ParsedRoute_RuleMissingHostMatcher";
         internal const string ParsedRouteRuleMultipleHostMatchers = "ParsedRoute_RuleMultipleHostMatchers";
--- a/src/ReverseProxy/Service/Config/ConfigErrors.cs
+++ b/src/ReverseProxy/Service/Config/ConfigErrors.cs
@@ -16,11 +16,7 @@ namespace Microsoft.ReverseProxy.Service
         internal const string RouteUnsupportedAction = "Route_UnsupportedAction";
 
         internal const string ParsedRouteMissingId = "ParsedRoute_MissingId";
-        internal const string ParsedRouteMissingHostAndPath = "ParsedRoute_MissingHostAndPath";
         internal const string ParsedRouteRuleHasNoMatchers = "ParsedRoute_RuleHasNoMatchers";
-        internal const string ParsedRouteRuleMissingHostMatcher = "ParsedRoute_RuleMissingHostMatcher";
-        internal const string ParsedRouteRuleMultipleHostMatchers = "ParsedRoute_RuleMultipleHostMatchers";
-        internal const string ParsedRouteRuleMultiplePathMatchers = "ParsedRoute_RuleMultiplePathMatchers";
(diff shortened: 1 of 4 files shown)


src/ReverseProxy/Service/Config/ConfigErrors.cs
```diff
@@ -16,6 +16,7 @@ internal static class ConfigErrors
         internal const string RouteUnsupportedAction = "Route_UnsupportedAction";
 
         internal const string ParsedRouteMissingId = "ParsedRoute_MissingId";
+        internal const string ParsedRouteMissingHostAndPath = "ParsedRoute_MissingHostAndPath";
         internal const string ParsedRouteRuleHasNoMatchers = "ParsedRoute_RuleHasNoMatchers";
         internal const string ParsedRouteRuleMissingHostMatcher = "ParsedRoute_RuleMissingHostMatcher";
```
Tratcher, line 21: Can ParsedRouteRuleMissingHostMatcher be removed now?

ihvo: Actually, I just realized that "no matchers" was probably intended to be used for the case when neither of host and path are specified. Removing the error code I added.

halter73:
Why not allow routes with both host and path unspecified?

Tratcher:
> Why not allow routes with both host and path unspecified?

As in it matches everything? You could only have one such route in the app. It seems like you should have to call out such a route explicitly.

AspNetCore Routing requires us to at least specify a path, but we default that path to a catch-all wildcard if you don't specify one.

halter73:
If I didn't know the a catch-all wildcard syntax, I would probably just try leaving path unspecified. I agree that a catch-all wildcard is more clear though.

Maybe the best option is to have an error that clearly recommends using a catch-all wildcard for the path when a rule has no matchers.

src/ReverseProxy/Service/Config/RuleParsing/RouteValidator.cs
```diff
@@ -55,6 +55,12 @@ public bool ValidateRoute(ParsedRoute route, IConfigErrorReporter errorReporter)
                 success = false;
             }
 
+            if (string.IsNullOrEmpty(route.Host) && string.IsNullOrEmpty(route.Path))
+            {
+                errorReporter.ReportError(ConfigErrors.ParsedRouteRuleHasNoMatchers, route.RouteId, $"Route requires {nameof(route.Host)} or {nameof(route.Path)} specified.");
```
Tratcher, line 60:
```suggestion
                errorReporter.ReportError(ConfigErrors.ParsedRouteRuleHasNoMatchers, route.RouteId, $"Route requires {nameof(route.Host)} or {nameof(route.Path)} specified. Set the Path to `/{**catchall}` to match all requests.");
```

src/ReverseProxy/Service/Config/RuleParsing/RouteValidator.cs
```diff
@@ -64,11 +70,10 @@ public bool ValidateRoute(ParsedRoute route, IConfigErrorReporter errorReporter)
 
         private static bool ValidateHost(string host, string routeId, IConfigErrorReporter errorReporter)
         {
-            // TODO: Why is Host required? I'd only expect Host OR Path to be required, with Path being the more common usage.
+            // Host is optional when Path is specified
             if (string.IsNullOrEmpty(host))
```
Tratcher, line 74: Hmm, the validation is inconsistent:
https://github.com/microsoft/reverse-proxy/blob/8969b63dbb67466231a2bcaf67f864982931ce6a/src/ReverseProxy/Service/DynamicEndpoints/RuntimeRouteBuilder.cs#L58

Can you update RuntimeRouteBuilder to be consistent?

src/ReverseProxy/Service/Config/RuleParsing/RouteValidator.cs
ihvo: Updated error message.

Tratcher:
Thanks

PR #237, opened 2020-06-10
CustomHeaderSessionAffinityProviderOptions replaces Settings collection

### LINKED ISSUE #236: Asymmetric affinity provider settings representation
There are currently 2 approaches to store session affinity provider settings. `CookieSessionAffinityProvider` uses the global instance `CookieSessionAffinityProviderOptions` as a source for affinity cookie parameters whereas `CustomHeaderSessionAffinityProvider` allows to define different custom affinity header names per backend in the configuration file.

It's suggested to make affinity cookie configurable per backend as well, move it to the configuration file, and get rid of the global `CookieSessionAffinityProviderOptions`.

alnikola:
`CustomHeaderSessionAffinityProvider` uses a new option type to retrieve the custom header name instead of an entry in `SessionAffinityOptions.Settings` collection to be consistent with `CookieSessionAffinityProvider`.

Fixes #236

### DIFF
--- a/src/ReverseProxy/Abstractions/BackendDiscovery/Contract/CustomHeaderSessionAffinityProviderOptions.cs
+++ b/src/ReverseProxy/Abstractions/BackendDiscovery/Contract/CustomHeaderSessionAffinityProviderOptions.cs
@@ -1,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Abstractions.BackendDiscovery.Contract
+{
+    /// <summary>
+    /// Defines customer header specific affinity provider options.
+    /// </summary>
+    public class CustomHeaderSessionAffinityProviderOptions
+    {
+        public static readonly string DefaultCustomHeaderName = "X-Microsoft-Proxy-Affinity";
+
+        /// <summary>
+        /// Name of a custom header storing an affinity key.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


alnikola:
@halter73 Yes, that's correct. Since `CookieSessionAffinityProvider` can be only configured via the corresponding `CookieSessionAffinityProviderOptions` and not via the config, it seems reasonable at this moment to use the same approach for `CustomHeaderSessionAffinityProvider`. If later we see a need for more flexibility, it can be revisited.

Tratcher:
The key design question here is if the affinity settings for either provider should be global or per backend. Do other implementations set these globaly?

Either way we can work out how to bind them to config if needed.

alnikola:
> Do other implementations set these globaly?

Yes, `CookieSessionAffinityProviderOptions` is set globally currently.
Theoretically, we can introduce a new API to tie them to backends. But, I'd suggest to firstly unify the approaches (i.e. complete this PR) and the listen to the customers feedback.

Tratcher:
I meant please check nginx and other implementations to see what level they allow this to be configured at.

Tratcher:
I took a quick look at nginx.
https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/#sticky

It does allow the cookie to be configured per backend. That makes some sense even if it is more work.

Let's close this PR for now, update #236, and we can re-triage it next milestone.

alnikola:
Closing due to a change in the design and the need for a reprioritization.
#236 updated according to the new proposal.

PR #244, opened 2020-06-15
TargetFramework set to .net5.0

### LINKED ISSUE #189: Change framework to net5.0

alnikola:
TFM set to `.net5.0` in all project. Additionally, `rollForward` is enabled and set to `latestMinor`

**Note.** `dotnet new` does **not** support `.net5.0` TFM, yet.

Fixes #189

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,7 @@
 {
   "sdk": {
-    "version": "5.0.100-preview.6.20266.3"
+    "version": "5.0",
+    "rollForward": "latestMinor"
   },
   "tools": {
     "dotnet": "5.0.100-preview.6.20266.3",
--- a/global.json
+++ b/global.json
@@ -1,7 +1,6 @@
 {
   "sdk": {
-    "version": "5.0",
-    "rollForward": "latestMinor"
+    "version": "5.0.100-preview.6.20266.3"
   },
   "tools": {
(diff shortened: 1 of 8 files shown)


global.json
```diff
@@ -1,6 +1,7 @@
 {
   "sdk": {
-    "version": "5.0.100-preview.6.20266.3"
+    "version": "5.0",
+    "rollForward": "latestMinor"
```
Tratcher, line 4: Why did you change this?

alnikola: Opening a solution directly without running our `startvs.cmd` only works if you have a proper SDK version in your shared `dotnet` folder (usually, inside Program Files). Previously, our `global.json` required the exact preview version like preview-6 and if you had only preview-5 it didn't work. Setting `rollForward` policy enables you to use any .NET 5 preview available on the machine.

Tratcher: startvs.cmd is required, there's no expectation that a developer have any version of the 5.0 SDK installed, let alone the right version.

Please revert.

alnikola: I see. Reverted.

PR #248, opened 2020-06-17
Remove public API

davidfowl:
- Make AutomicCounter internal
- Remove IRecurrentActivityManager
- Make IRandomFactory internal
- Remove unused APIs from AtomicCounter
- Made ConcurrencyCounter on DestinationInfo internal


src/ReverseProxy/Service/RuntimeModel/DestinationInfo.cs
```diff
@@ -44,7 +44,7 @@ public DestinationInfo(string destinationId)
         /// <summary>
         /// Keeps track of the total number of concurrent requests on this endpoint.
         /// </summary>
-        public AtomicCounter ConcurrencyCounter { get; } = new AtomicCounter();
+        internal AtomicCounter ConcurrencyCounter { get; } = new AtomicCounter();
```
Tratcher, line 47: This is used for load balancing. 3rd party algorithms will need this info publicly. Add a public read-only property to read the value?

davidfowl: Can do

halter73:
```
##[error]Service/Config/TransformBuilder.cs(28,16): error CS0051: (NETCORE_ENGINEERING_TELEMETRY=Build) Inconsistent accessibility: parameter type 'IRandomFactory' is less accessible than method 'TransformBuilder.TransformBuilder(TemplateBinderFactory, IRandomFactory)'
```

I think we just make the TransformBuilder class internal. It's registered with DI, but I don't think there's a problem with internal types, just internal constructors.

davidfowl:
Ah this is the rebase I just did

PR #250, opened 2020-06-18
Removed more APIs

davidfowl:
- Removed Result<T>
- Removed AsyncSemaphore
- Moved TaskExtensions to test project
- Did some random clean up of argument validaton


Tratcher:
Hold off for a day, I'm in the middle of renaming Backends and this is going to conflict.

src/ReverseProxy/Service/HealthProbe/BackendProber.cs
```diff
@@ -51,13 +51,13 @@ public BackendProber(string backendId, BackendConfig config, IDestinationManager
             Contracts.CheckValue(httpClient, nameof(httpClient));
             Contracts.CheckValue(randomFactory, nameof(randomFactory));
```
Tratcher, line 52: Remove Contracts?

davidfowl: I'll remove the ones in the classes that I touch.

Kahbazi: Do you accept a PR which removes all `Contracts`?

Tratcher: Yes please. That's style issue we haven't addressed yet.

davidfowl:
Rebased!

src/ReverseProxy/Service/HealthProbe/BackendProber.cs
davidfowl: Yep. I'm going to do what I touched in this PR

davidfowl:
@Tratcher can you review please? 😄

src/ReverseProxy/Service/Config/DynamicConfigBuilder.cs
```diff
@@ -30,25 +30,16 @@ internal class DynamicConfigBuilder : IDynamicConfigBuilder
...
+            _filters = filters ?? throw new ArgumentNullException(nameof(filters));
+            _clustersRepo = clustersRepo ?? throw new ArgumentNullException(nameof(clustersRepo));
+            _routesRepo = routesRepo ?? throw new ArgumentNullException(nameof(routesRepo));
+            _parsedRouteValidator = parsedRouteValidator ?? throw new ArgumentNullException(nameof(parsedRouteValidator));
+            _sessionAffinityProviders = sessionAffinityProviders?.ToProviderDictionary() ?? throw new ArgumentNullException(nameof(sessionAffinityProviders));
+            _affinityFailurePolicies = affinityFailurePolicies?.ToPolicyDictionary() ?? throw new ArgumentNullException(nameof(affinityFailurePolicies));
         }
 
-        public async Task<Result<DynamicConfigRoot>> BuildConfigAsync(IConfigErrorReporter errorReporter, CancellationToken cancellation)
+        public async Task<DynamicConfigRoot> BuildConfigAsync(IConfigErrorReporter errorReporter, CancellationToken cancellation)
```
alnikola, line 41: Do we really need to remove `Result<>` wrapper? Later, we will probably implement a more sophisticated handling of config reloading errors, so it might be necessary to return the reload completion status to the caller. Furthermore, having the `Result<>` wrapper over `DynamicConfigRoot` can also be handy for returning extra error details like the name of config section failed to load.
Or do you believe it would be better to implement all the error handling logic inside DynamicConfigBuilder (or on some services it depends on)?

davidfowl: You can use a tuple instead. I don't see evidence of needing a result type at all.

PR #251, opened 2020-06-18
Rename Backend to Cluster

### LINKED ISSUE #128: Rename Endpoint, Backend

Tratcher:
Renamed to address confusion about what the various config entities are for.
Fixes #128

This change is pretty mechanical, there shouldn't be any behavioral changes.


docs/docfx/articles/getting_started.md
```diff
@@ -84,13 +84,13 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
 The configuration for YARP is defined in the appsettings.json file. It defines a set of:
 
-Routes - which map incoming requests to the backend clusters based on aspects of the request such as host name, path, method, request headers etc. Specifying a Host is the only required field. Routes are ordered, so the "app1" route needs to be defined first since "route2" will act as a catchall for all paths that have not already been matched. 
+Routes - which map incoming requests to the clusters based on aspects of the request such as host name, path, method, request headers etc. A route must specify at least a host or path. Routes are ordered, so the "app1" route needs to be defined first since "route2" will act as a catchall for all paths that have not already been matched.
```
halter73, line 87:
```suggestion
Routes - which map incoming requests to the clusters based on aspects of the request such as host name, path, method, request headers etc. A route must specify at least a host or path, but the path can be a [catch-all route](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.1). Routes are ordered, so the "app1" route needs to be defined first since "route2" will act as a catchall for all paths that have not already been matched. 
```

src/ReverseProxy/Service/DynamicEndpoints/IRuntimeRouteBuilder.cs
```diff
@@ -22,13 +22,13 @@ internal interface IRuntimeRouteBuilder
...
-        /// <param name="backendOrNull">Backend that this route maps to.</param>
+        /// <param name="clusterOrNull">Cluster that this route maps to.</param>
         /// <param name="runtimeRoute">
         /// Representation of the route during runtime,
         /// whose <see cref="RouteInfo.Config"/> property is updated with a new objected computed from
-        /// <paramref name="source"/> and <paramref name="backendOrNull"/>.
+        /// <paramref name="source"/> and <paramref name="clusterOrNull"/>.
         /// </param>
-        RouteConfig Build(ParsedRoute source, BackendInfo backendOrNull, RouteInfo runtimeRoute);
+        RouteConfig Build(ParsedRoute source, ClusterInfo clusterOrNull, RouteInfo runtimeRoute);
```
halter73, line 31: I noticed that you've changed most instances of "backendOrNull" to "cluster" instead of "clusterOrNull"?. I'm fine with this, but did you mean to change it everywhere? There are a couple places that were missed.

Tratcher: Ah, yeah I'd caught a few more of those that I hadn't pushed yet.

PR #257, opened 2020-06-19
Proxy has Service Fabric integration for dynamic discovery of routes

davidni:
## Introduction

Adds `Microsoft.ReverseProxy.ServiceFabric` which implements dynamic discovery of Service Fabric services. The general architecture takes inspiration from [Traefik's Service Fabric integration](https://docs.traefik.io/v1.7/configuration/backends/servicefabric/), and is designed for easy and friction-free onboarding of multiple Service Fabric services operating behind YARP. Running multiple instances of YARP in the same cluster is possible, and an eventual consistency model helps avoid the need for any external storage except for the runtime Service Fabric queries used to dynamically discover services in the cluster.

A Service Fabric service that wants to leverage YARP simply needs to add an `<Extension>` element in their `ServiceManifest`, for example (see complete example further below):

```xml
<Extensions>
  <Extension Name="YARP-preview">
    <Labels xmlns="http://schemas.microsoft.com/2015/03/fabact-no-schema">
      <Label Key="YARP.Enable">true</Label>
      <Label Key="YARP.Routes.route1.Hosts">example.com</Label>
    </Labels>
  </Extension>
</Extensions>
```

## Functional overview
This works by periodically querying Service Fabric for all applications in a Service Fabric cluster. For each application, we enumerate all services. For each service, we extract its _labels_ and, if the service has opted in to use YARP (`YARP.Enable=true`). If the service that has opted in to use YARP, we then enumerate all partitions, and all instances/replicas within each partition.

### Service Fabric concepts mappings to YARP

Service Fabric terminology|YARP terminology
-|-
Cluster|--
Application|--
Service (e.g. `fabric:/App1/Svc2`)|Cluster (ClusterId=ServiceName)
Instance / replica|Destination (Address=instance' or replica's endpoint)
`YARP.Routes.<routeName>.*` in ServiceManifest|ProxyRoute (id=ServiceName+routeName, Match=Hosts,Path extracted from the labels)

### Dynamic configurability
Labels defined in the Service Manifest support the following scenarios:
* Arbitrary key / value pair. Only keys understood by us are honored, and unknown keys are ignored
* Application parameter reference. If the label value is of the form `[appParamName]`, its value will be replaced at runtime with the value of the application parameter with the same name defined for the application the service belongs to
* Naming Service overrides. If label `YARP.EnableDynamicOverrides` is specified as `true`, we will also query Service Fabric Naming Service to discover any [named properties](https://docs.microsoft.com/en-us/rest/api/servicefabric/sfclient-index-property-management). Any _properties_ found for the service being enumerated take precedence over values specified in ServiceManifest. This functionality is opt-in, and disabled by default.

### Designed for high availability

We keep a cache of the last discovery results obtained from Service Fabric. If at any point, for whatever reason, we are not able to retrieve an updated view of the cluster, we will continue to operate with last-known-good data. The cache is kept in memory, but could be extended to support permanent storage on disk to also survive reboots.

### Designed for use by large teams
This integration is designed to support hundreds of Service Fabric services hosted behind YARP, and it provides additional features to help owners of those teams self-service their devops requirements relating to YARP. When we successfully discover a service, we emit a [Service Fabric health report](https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-health-introduction) against the discovered service. If any config errors are encountered, a Warning health report is issued, which makes it easy for the service owners to observe the errors on Service Fabric Explorer and / or telemetry. This closes the loop and makes it easy for a service in Service Fabric to confirm that its configurations are being honored and to detect any issues.

### Complete ServiceManifest example:

```xml
<ServiceManifest Name="Service1Pkg" Version="1.0.0" xmlns="http://schemas.microsoft.com/2011/01/fabric">
  <ServiceTypes>
    <StatelessServiceType ServiceTypeName="Service1Type" >
      <Extensions>
        <Extension Name="YARP-preview">
          <Labels xmlns="http://schemas.microsoft.com/2015/03/fabact-no-schema">
            <Label Key="YARP.Enable">true</Label>
            <Label Key="YARP.Routes.route1.Hosts">example.com</Label>
            <!-- Optional: enable active health probes -->
            <Label Key='YARP.Backend.Healthcheck.Active.Enabled'>true</Label>
            <Label Key='YARP.Backend.Healthcheck.Active.Path'>api/health</Label>
            <Label Key='YARP.Backend.Healthcheck.Active.Timeout'>30</Label>
            <Label Key='YARP.Backend.Healthcheck.Active.Interval'>10</Label>
          </Labels>
        </Extension>
      </Extensions>
    </StatelessServiceType>
  </ServiceTypes>

  <!-- ... -->
</ServiceManifest>
```


src/Integrations/ServiceFabric/ServiceDiscovery/Worker/ServiceFabricExtensionConfigProvider.cs
```diff
@@ -0,0 +1,233 @@
...
+            using (var reader = XmlReader.Create(new StringReader(rawServiceManifest), CreateSafeXmlSetting(1024 * 1024, 1024)))
+            {
+                XDocument parsedManifest;
+                try
+                {
+                    parsedManifest = await XDocument.LoadAsync(reader, LoadOptions.None, cancellationToken);
+                }
+                catch (System.Xml.XmlException ex)
+                {
+                    // TODO: we don't know if the service wants to use the gateway yet, so not sure if this classifies as config error (considering it will escalate into a bad health report)
```
alnikola, line 207: Could you please clarify why getting an invalid XML can mean that the service is not yet ready to use the reverse proxy? I'd expect that in this case either the label `IslandGateway.Enable` is completely missing or set to `false`.

src/Integrations/ServiceFabric/ServiceDiscovery/Worker/ServiceFabricExtensionConfigProvider.cs
```diff
@@ -0,0 +1,233 @@
...
+                    .Elements(XNSServiceManifest + "Extension").Where(s => (string)s.Attribute("Name") == "IslandGateway")
+                    .Elements(XNameLabels)
+                    .Elements(XNameLabel);
+
+                foreach (var label in labels)
+                {
+                    if (!result.TryAdd(label.Attribute("Key").Value, label.Value))
+                    {
+                        // TODO: we don't know if the service wants to use the gateway yet, so not sure if this classifies as config error (considering it will escalate into a bad health report)
+                        throw new ConfigException($"Duplicate label key {label.Attribute("Key").Value} in service manifest extensions.");
```
alnikola, line 226: The same question as above. Why do we treat duplicate label keys as some valid intermediate config state?

src/ReverseProxy/Service/Config/ServiceDiscoveryConfigApplier.cs
```diff
@@ -0,0 +1,149 @@
...
+{
+    /// <summary>
+    /// Reacts to configuration changes for type <see cref="ServiceDiscoveryConfig"/>
+    /// via <see cref="IOptionsMonitor{TOptions}"/>, and applies configurations
+    /// to the Island Gateway core.
+    /// When configs are loaded from appsettings.json, this takes care of hot updates
+    /// when appsettings.json is modified on disk.
+    /// </summary>
+    // TODO: This is needs to be wired up with the Reverse Proxy core, and should probably be re-written from scratch.
+    internal class ServiceDiscoveryConfigApplier : IHostedService, IDisposable
```
alnikola, line 25: Wouldn't it cause a conflict with YARP's default [ProxyConfigLoader](https://github.com/microsoft/reverse-proxy/blob/77d035e2b0c5e978d28551ebf03c0ca6cbba777b/src/ReverseProxy/Configuration/ProxyConfigLoader.cs#L24)?
Do I get it right that these two config loaders are mutually exclusive meaning we might have only one of them registered? As far as I remember, we planned to support a fallback scenario where if the dynamic routing couldn't found a destination, it falls back to the default routing logic.

src/Integrations/ServiceFabric/FabricWrapper/QueryClientWrapper.cs
```diff
@@ -0,0 +1,261 @@
...
+                // NOTE: App Params in Service Fabric are case insensitive (verified on version 7.0.457.9590).
+                // Since this is not documented behavior, the code below tries to play it safe by ignoring
+                // duplicated app params instead of throwing and preventing such service from working at all
+                // behind Island Gateway.
+                var result = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                foreach (var param in app.ApplicationParameters)
+                {
+                    if (!result.TryAdd(param.Name, param.Value))
+                    {
+                        _logger.LogInformation($"Duplicate app parameter '{param.Name}' for application '{app.ApplicationName}'");
```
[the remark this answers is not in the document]
davidni: It's a warning for the target service, but we did not malfunction, so it's not a warning from YARP's perspective. There is some work left to raise these kinds of issues consistently against the target services.

src/Integrations/ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -0,0 +1,174 @@
...
+                    Average = GetLabel(labels, "IslandGateway.Backend.Quota.Average", DefaultQuotaAverage),
+                    Burst = GetLabel(labels, "IslandGateway.Backend.Quota.Burst", DefaultQuotaBurst),
+                },
+                PartitioningOptions = new ClusterPartitioningOptions
+                {
+                    PartitionCount = GetLabel(labels, "IslandGateway.Backend.Partitioning.Count", DefaultPartitionCount),
+                    PartitionKeyExtractor = GetLabel(labels, "IslandGateway.Backend.Partitioning.KeyExtractor", DefaultPartitionKeyExtractor),
+                    PartitioningAlgorithm = GetLabel(labels, "IslandGateway.Backend.Partitioning.Algorithm", DefaultPartitioningAlgorithm),
+                },
+                LoadBalancing = new LoadBalancingOptions(), // TODO
```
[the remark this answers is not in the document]
davidni: We should add. SF itself is oblivious to this, but it of course makes sense in the context of YARP. I'd like to have that as a separate PR though. Might be a good idea to file an issue to track if you'd like.

src/Integrations/ServiceFabric/ServiceDiscovery/Worker/ServiceFabricExtensionConfigProvider.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        /// <summary>
+        /// Gets the labels from the extensions of the provided raw service manifest.
+        /// </summary>
+        private async Task<Dictionary<string, string>> ExtractLabelsAsync(
+           string rawServiceManifest,
+           string targetServiceTypeName,
+           CancellationToken cancellationToken)
+        {
+            var result = new Dictionary<string, string>(StringComparer.Ordinal);
```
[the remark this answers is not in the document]
davidni: At the moment yes. It is worth revisiting.

src/Integrations/ServiceFabric/ServiceDiscovery/Worker/ServiceFabricExtensionConfigProvider.cs
davidni: The XML contents inform us whether to include a given service or not. If we cannot parse the XML, we don't know if the service intended to say Enabled=true or otherwise. Note that this is not expected to ever happen, since SF should never allow malformed XML to be deployed -- a lot of other things would have already failed by this point.

src/ReverseProxy/Service/Config/ServiceDiscoveryConfigApplier.cs
davidni: This is precisely why I didn't integrate this into YARP yet. It needs review and I'd defer to @Tratcher and @davidfowl to own it as part of their other ongoing work with improving the config story.

For reference, here is what we did internally:
* Modified config structure to take a `ServiceDiscoveryName` parameter specifying where configs come from. Valid values are `static`, `servicefabric`.
* Each provider's configs go in a new section of configs. `static` configs appear under `ReverseProxy:static`; SF configs appear under `ReverseProxy:servicefabric`. Example:

```js
{
  "ReverseProxy": {
    "ServiceDiscoveryName": "static", // valid options: `static`, `servicefabric`
    "ServiceDiscoveryConfigs": {
      "servicefabric": { // only takes effect when ServiceDiscoveryName==`servicefabric`
        "DiscoveryPeriodSeconds": 30
      },
      "static": { // only takes effect when ServiceDiscoveryName==`static`
        "Clusters": {
          "backend1": {
            "Destinations": {
              "destination1": {
                "Address": "https://localhost:10000/"
              }
            }
          }
        },
        "Routes": [
          {
            "RouteId": "backend1:route1",
            "ClusterId": "backend1",
            "Host": "localhost"
          }
        ]
      }
    }
  }
}
```

I do not like this, however:
* It is limiting -- I cannot have more than one provider enabled at each time. It also makes no attempt to segregate data from different providers (e.g. how would it handle 2 providers specifying routes with the same RouteId).
* It is too verbose in typical simple cases -- if i don't want dynamic discovery, i still need to specify config keys ServiceDiscoveryName and ServiceDiscoveryConfigs

src/Integrations/ServiceFabric/DependencyInjection/ServiceFabricIntegrationIslandGatewayBuilderExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// used to register Service Fabric integration components.
+    /// </summary>
+    public static class ServiceFabricIntegrationIslandGatewayBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the services needed to integrate Service Fabric with the Island Gateway to Dependency Injection.
+        /// </summary>
+        public static IReverseProxyBuilder AddServiceFabricServiceDiscovery(this IReverseProxyBuilder builder)
+        {
+            builder.Services.AddSingleton<IQueryClientWrapper, QueryClientWrapper>();
```
Kahbazi, line 21: Is it better to use `TryAddSingleton` to match the other service registeration?

samsp-msft:
Do we need to use the IslandGateway name in the parameter keys Eg
 <Label Key='IslandGateway.Backend.Healthcheck.Enabled'>true</Label>

davidni:
> Do we need to use the IslandGateway name in the parameter keys

@samsp-msft this is covered by `Organize and rename labels` in the PR description under _Remaining work (outside of scope of this PR)_. The name IslandGateway doesn't make sense outside my team, I just don't have the bandwidth to change and verify the changes immediately.

src/Integrations/ServiceFabric/FabricWrapper/PropertyManagementClientWrapper.cs
```diff
@@ -0,0 +1,90 @@
...
+            return null;
+        }
+
+        /// <summary>
+        /// Enumerates all Service Fabric properties under a given name.
+        /// Also takes in timeout interval, which is the maximum of time the system will allow this operation to continue before returning.
+        /// </summary>
+        public async Task<IDictionary<string, string>> EnumeratePropertiesAsync(Uri parentName, TimeSpan timeout, CancellationToken cancellationToken)
+        {
+            var namedProperties = new Dictionary<string, string>(StringComparer.Ordinal);
```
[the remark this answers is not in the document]
davidni: Good point, I have reached out to SF folks for comments.

src/Integrations/ServiceFabric/DependencyInjection/ServiceFabricIntegrationIslandGatewayBuilderExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Extensions for <see cref="IIslandGatewayBuilder"/>
+    /// used to register Service Fabric integration components.
+    /// </summary>
+    public static class ServiceFabricIntegrationIslandGatewayBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the services needed to integrate Service Fabric with the Island Gateway to Dependency Injection.
+        /// </summary>
+        public static IReverseProxyBuilder AddServiceFabricServiceDiscovery(this IReverseProxyBuilder builder)
```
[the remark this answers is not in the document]
davidni: I expect a common customer scenario will be to deploy a pre-built yarp.exe that supports multiple dynamic discovery providers (e.g. Service Fabric, k8s, Consul, ...) and to select their dynamic provider in config. The intent of this method is to register a specific provider, which would then be activated separately. So this wouldn't be loading any configs at all, rather it just registers the components necessary to support SF as a provider.

src/Integrations/ServiceFabric/Exceptions/ServiceFabricIntegrationException.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.ReverseProxy.ServiceFabricIntegration
+{
+    /// <summary>
+    /// Represents errors related to Service Fabric integration with the gateway.
+    /// </summary>
+    public sealed class ServiceFabricIntegrationException : Exception
```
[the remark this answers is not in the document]
davidni: Agreed this is of limited use. I prefer this to the alternative of throwing unqualified `new Exception(...)` in cases where we just want to enrich an inner exception with extra context about what we were doing when an unhandled exception occurred.

Example use today:
```cs
// ...
catch (Exception ex) when (!ex.IsFatal())
{
    throw new ServiceFabricIntegrationException($"Failed to get service manifest name for service type {service.ServiceTypeName} of application type {application.ApplicationTypeName} {application.ApplicationTypeVersion} from Service Fabric.");
}
```

Would you prefer this?
```cs
// ...
catch (Exception ex) when (!ex.IsFatal())
{
    throw new Exception($"Failed to get service manifest name for service type {service.ServiceTypeName} of application type {application.ApplicationTypeName} {application.ApplicationTypeVersion} from Service Fabric.");
}
```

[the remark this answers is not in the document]
Tratcher: I appreciate not wanting to throw a base `Exception`, but the framework provides a variety of other ones to choose from such as InvalidOperationException.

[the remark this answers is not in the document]
davidni: This exception will usually represent an external error such as that Service Fabric misbehaved (e.g. an SF api failed, or we got inconsistent data). `InvalidOperationException` would imply a problem in our code, which seems inaccurate. I can change to InvalidOperationException, but IMHO it would be misleading. Thoughts?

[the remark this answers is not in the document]
Tratcher: You can leave it for now, we'll sort it out later.

davidfowl:
I'm not a fan of how this plugs into the proxy. It seems like the `IServiceDiscovery` is the interface being plugged in here. Is the intent to replace the configuration based model completely? This comment seems like it's not resolved https://github.com/microsoft/reverse-proxy/pull/257#discussion_r443667820.

> We keep a cache of the last discovery results obtained from Service Fabric. If at any point, for whatever reason, we are not able to retrieve an updated view of the cluster, we will continue to operate with last-known-good data. The cache is kept in memory, but could be extended to support permanent storage on disk to also survive reboots.

What happens when multiple instances have an inconsistent view of the configuration because of network issues between SF and the node?

src/ReverseProxy.ServiceFabric/FabricWrapper/QueryClientWrapper.cs
```diff
@@ -0,0 +1,259 @@
...
+        private readonly FabricClient.QueryClient _queryClient;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="QueryClientWrapper"/> class.
+        /// </summary>
+        public QueryClientWrapper(ILogger<QueryClientWrapper> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            _queryClient = new FabricClient().QueryManager;
```
davidfowl, line 32: Does this need to get disposed?

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/FabricServiceEndpointSelector.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            endpointUri = null;
+            string endpointAddress = null;
+
+            foreach (var listenerName in fabricServiceEndpoint.ListenerNames)
+            {
+                // SF Reverse Proxy endpoint selection logic: https://msazure.visualstudio.com/One/_git/WindowsFabric?path=%2Fsrc%2Fprod%2Fsrc%2FManagement%2FApplicationGateway%2FHttp%2FServiceEndpointsList.cpp&version=GBrelease_6.5&line=52&lineStyle=plain&lineEnd=53&lineStartColumn=1&lineEndColumn=1
+                if (listenerName == string.Empty && fabricServiceEndpoint.EmptyStringMatchesAnyListener)
+                {
+                    endpointUri = endpoints.ToReadOnlyDictionary()
```
davidfowl, line 39: Seems expensive to do this every time this is called. How often is it called?

davidni:
> I'm not a fan of how this plugs into the proxy.

@davidfowl ditto. It needs additional work and design, which is why I am leaving the last-mile open for you and the team to decide what fits best.

---

> This comment seems like it's not resolved #257 (comment).

@davidfowl I am suggesting this should be tackled as a separate PR, which would also address how dynamic discovery impacts the configuration model (see next point below), I would rather defer to you and the team to decide how best to take it, hence I am not attempting to address it in this PR.

---

> Is the intent to replace the configuration based model completely?

@davidfowl Not necessarily, but possibly. Examples of what others are doing:

* Envoy lets you specify a [config source](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/config_source.proto#envoy-api-msg-core-configsource) at the top level. We ended up with something similar in our implementation -- a top level config field `ServiceDiscoveryName`, which can be `servicefabric` or `static` for us. The latter means we will read Routes, Clusters, Destinations from ASP .NET Core configs in a section `"static"`. **But IMHO this should NOT be the north star. It would be nice to allow N >= 1 config sources to work together, so that YARP could have static configs, which can be augmented by any number of dynamic config providers (such as Service Fabric)**. That aligns with Traefik (see next line)
* Traefik lets you specify [providers](https://docs.traefik.io/providers/overview/), all of which provide dynamic configurations and which can use configs from one another. See also [this PR](https://github.com/containous/traefik/pull/4372).

---

> What happens when multiple instances have an inconsistent view of the configuration because of network issues between SF and the node?

@davidfowl At the risk of preaching to the choir, I'll say there's no free lunch. Eventual Consistency models offer simplicity but it comes at a cost. We believe this is the right choice for a component like YARP in distributed hosting scenarios, and there is precedent for this choice. Suggested reading: [Embracing eventual consistency in SoA networking](https://blog.envoyproxy.io/embracing-eventual-consistency-in-soa-networking-32a5ee5d443d) re. Envoy's design choice, by Envoy's lead author. Orchestration layers can be built on top of our eventually consistent model to provide strong guarantees when needed, and separately from YARP's core.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/FabricServiceEndpointSelector.cs
davidni: Up to twice per Service Fabric service replica per polling cycle. Since usually each replica exposes a single endpoint, we could add an optimization for the trivial case, but this is unlikely to become a hotspot -- all the XML parsing happening elsewhere and the SF API's themselves are more worrisome than this.

src/ReverseProxy.ServiceFabric/FabricWrapper/QueryClientWrapper.cs
davidni: Ideally yes, but since it is registered as a singleton and in our current scenarios we only teardown when we also kill the process, dispose logic adds no value.

alnikola:
>Eventual Consistency models offer simplicity but it comes at a cost. We believe this is the right choice for a component like YARP in distributed hosting scenarios, and there is precedent for this choice. Suggested reading: Embracing eventual consistency in SoA networking

In case of YARP, EC simplifies the code and consequently **our** work as proxy developers, but it will likely make user's experience worse because, as it was pointed out in the above article, eventual consistency is counterintuitive.

Moreover, I would challenge the assumption that in real-world changes are **always** rolled-out gradually with enough time between change so that nodes can reach consensus before the next change is applied and users are shielded from observing inconsistent state. In my experience, service changes might also be applied immediately to a large share of nodes (up to 100%) in special, but **not so uncommon** cases. Examples can be rolling out a security hot fix or deploying a change to relatively small geo region. Furthermore, even in case of gradual deployment, percentage can be changed in incrementally increasing steps (e.g. 1%, 2%, 4%, 8%, 16%, 32%, etc.) to speed-up roll-out, thus at the later stages a substantial share of system nodes get affected by each step which can expose to users inconsistencies inherent to EC approach.

All in all, I realize than network is inherently unreliable and CAP theorem imposes hard limit on what we can achieve, but I strongly believe we should always aim to firstly provide strong consistency and relax the model only if it looks unreasonably expensive to implement or maintain.

Having that said, implementing mechanism synchronizing SF configuration among YARP nodes seems to be outside this PR scope.

src/ReverseProxy.ServiceFabric/FabricWrapper/QueryClientWrapper.cs
davidfowl: Feels bad, I would dispose it (by making this disposable)

mcpride:
Hello, 
in my opinion your currently implemented Service Fabric concept mappings to YARP doesn't work very well because a Service Fabric service can provide more than one endpoint. So the YARP cluster should be mapped from a combination of service and endpoint name.

Tratcher:
We have other teams asking about this. At the very least it needs to be rebased to resolve merge conflicts. I'll give it try.

Tratcher:
@davidni has offered to rebase and update this PR in the next week or two.

Tratcher:
Triage notes: We'd like to merge this PR (once it's updated) so that it doesn't keep getting out of date. We can mark it as experimental, skip shipping the package to nuget.org, etc., but it will be easier to iterate and maintain if we get the broader structure merged into master sooner than later.

davidfowl:
> Triage notes: We'd like to merge this PR (once it's updated) so that it doesn't keep getting out of date. We can mark it as experimental, skip shipping the package to nuget.org, etc., but it will be easier to iterate and maintain if we get the broader structure merged into master sooner than later.

Who's going to make sure it's not going to bitrot?

Tratcher:
> Who's going to make sure it's not going to bitrot?

If we get it checked in then we can at least make sure it still compiles and avoid that class of bitrot.

Alexander will start ramping up on SF after he wraps up the health checks work so we can take real ownership soon.

akanieski:
I was wondering what this thread's general opinion would be on receiving service updates proactively from SF instead of (or in addition to) polling. I think you could use the [ServiceManagementClient.ServiceNotificationFilterMatched](https://docs.microsoft.com/en-us/dotnet/api/system.fabric.fabricclient.servicemanagementclient.servicenotificationfiltermatched?view=azure-dotnet) Event to be notified of such changes.

Tratcher:
@akanieski that is something we're looking into to see if it scales better.

akanieski:
> @akanieski that is something we're looking into to see if it scales better.

I could add this if you need an extra set of hands.

Tratcher:
Thanks for the offer. Let's get this beast merged first. Then you can sync with @davidni and @alnikola about an event based pattern. Can you capture the details for an event pattern in a new issue for now?

mcpride:
Event based sounds interesting for app changes, but how you get new applications signaled?

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -0,0 +1,180 @@
...
+                {
+                    if (kvp.Key.StartsWith($"{thisRoutePrefix}.Metadata.", StringComparison.Ordinal))
+                    {
+                        metadata.Add(kvp.Key.Substring($"{thisRoutePrefix}.Metadata.".Length), kvp.Value);
+                    }
+                }
+
+                if (!labels.TryGetValue($"{thisRoutePrefix}.Hosts", out var hosts))
+                {
+                    throw new ConfigException($"Missing '{thisRoutePrefix}.Hosts'.");
```
akanieski, line 96: I notice in [our docs](https://microsoft.github.io/reverse-proxy/articles/configfiles.html#configuration-structure) that `Hosts` is not required for ProxyRoute to function. What happens if you want to support any host (I guess by leaving it blank) but put each application behind it's own respective path.

Tratcher: Hosts used to be required by YARP but aren't any more. This code is likely out of date.

davidni:
I'm close to having a new iteration on top of latest master, expect this to land this week.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/ServiceFabricConfigProvider.cs
```diff
@@ -0,0 +1,185 @@
...
+                try
+                {
+                    _backgroundCts.Token.ThrowIfCancellationRequested();
+                    if (!first)
+                    {
+                        await _timer.Delay(_optionsMonitor.CurrentValue.DiscoveryPeriod, _backgroundCts.Token);
+                    }
+
+                    var result = await _discoverer.DiscoverAsync(_backgroundCts.Token);
+                    UpdateSnapshot(result.Routes, result.Clusters);
```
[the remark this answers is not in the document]
davidni: Spot on, we must throttle the exception case too, else we may end up with a very tight CPU loop. Re. removing `first`, I could, but then I'd need two `try` blocks to handle graceful cancellations properly. Please reactivate if you feel strongly about this.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/FabricServiceEndpointSelector.cs
```diff
@@ -0,0 +1,100 @@
...
+
+                        // NOTE: Ordinal comparison used to match sort order of Service Fabric Reverse Proxy
+                        .OrderBy(listenerAddressPair => listenerAddressPair.Key, StringComparer.Ordinal)
+
+                        // From the endpoints above, select endpoints with valid URIs
```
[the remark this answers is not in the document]
davidni: Additional telemetry is not critical and risks causing too much noise, I suggest we punt and re-evaluate in the future.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -0,0 +1,180 @@
...
+            {
+                backendId = serviceName.ToString();
+            }
+
+            return backendId;
+        }
+
+        private static IReadOnlyList<string> SplitHosts(string hosts)
+        {
+            return hosts.Split(',').Select(h => h.Trim()).Where(h => h.Length > 0).ToList();
```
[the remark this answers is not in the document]
davidni: Good to know! This seems fine for now, please reactivate if you feel strongly.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Discoverer.cs
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        private bool HttpsSchemeSelector(string urlScheme)
+        {
+            if (_optionsMonitor.CurrentValue.DiscoverInsecureHttpDestinations)
+            {
+                return urlScheme == "https" || urlScheme == "http";
+            }
+
+            return urlScheme == "https";
```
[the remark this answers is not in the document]
davidni: Yep, `Uri.Scheme` normalizes it to lowercase.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Discoverer.cs
```diff
@@ -0,0 +1,432 @@
...
+                            // Skip this service
+                            continue;
+                        }
+
+                        var cluster = LabelsParser.BuildCluster(service.ServiceName, serviceExtensionLabels);
+                        await DiscoverDestinationsAsync(cluster, options, service, serviceExtensionLabels, cancellation);
+                        var clusterValidationErrors = await _configValidator.ValidateClusterAsync(cluster);
+                        if (clusterValidationErrors.Count > 0)
+                        {
+                            throw new ConfigException($"Skipping cluster id '{cluster.Id} due to validation errors.", new AggregateException(clusterValidationErrors));
```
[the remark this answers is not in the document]
davidni: Agreed, but let's take it in a separate PR.

src/ReverseProxy.ServiceFabric/FabricWrapper/PropertyManagementClientWrapper.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// Get the specified NamedProperty.
+        /// Also takes in timeout interval, which is the maximum of time the system will allow this operation to continue before returning.
+        /// </summary>
+        public async Task<string> GetPropertyAsync(Uri parentName, string propertyName, TimeSpan timeout, CancellationToken cancellationToken)
+        {
+            var result = await ExceptionsHelper.TranslateCancellations(
+                () => _propertyManagementClient.GetPropertyAsync(parentName, propertyName, timeout, cancellationToken),
+                cancellationToken);
```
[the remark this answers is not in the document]
davidni: Agreed, but let's take it in a separate PR.

src/ReverseProxy.ServiceFabric/Exceptions/ConfigException.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.ReverseProxy.ServiceFabric
+{
+    /// <summary>
+    /// Represents errors related to a user's configuration for the Island Gateway to use.
+    /// </summary>
+    public sealed class ConfigException : Exception
```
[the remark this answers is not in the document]
davidni: Agreed, but let's take it in a separate PR.

samples/ReverseProxy.ServiceFabric.Sample/Startup.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <summary>
+        /// This method gets called by the runtime. Use this method to add services to the container.
+        /// </summary>
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddControllers();
+            services.AddReverseProxy()
+                .AddServiceFabricDiscovery();
+
+            services.Configure<ServiceFabricDiscoveryOptions>(_configuration.GetSection("ServiceFabricDiscovery"));
```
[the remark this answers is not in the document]
davidni: Agreed, but let's take it in a separate PR.

davidni:
> in my opinion your currently implemented Service Fabric concept mappings to YARP doesn't work very well because a Service Fabric service can provide more than one endpoint.

That's true. You **can** specify a custom ListenerName which will influence how we choose each replica's endpoint, but you're right that, as it is, we do not provide a mechanism for you to expose more than one endpoint at the same time from each service. Can you file an issue to track and link to this discussion? Please include there some of your motivation why this is an important scenario, it will help us prioritize accordingly.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -44,16 +44,16 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
 
             try
             {
-                long lastTime = 0;
-                long nextTransferringEvent = 0;
-                long stopwatchTicksBetweenTransferringEvents = 0;
+                var lastTime = TimeSpan.Zero;
+                var nextTransferringEvent = TimeSpan.Zero;
+                var stopwatchTicksBetweenTransferringEvents = TimeSpan.Zero;
```
MihaZupan, line 49: If we're changing the `IClock` API away from raw ticks, this local isn't needed

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -77,8 +77,8 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
                             contentLength += read;
                             iops++;
 
-                            var readStop = clock.GetStopwatchTimestamp();
-                            var currentReadTime = readStop - lastTime;
+                            var readStop = clock.GetStopwatchTime();
+                            var currentReadTime = (readStop - lastTime).Ticks;
```
MihaZupan, line 81: I think it's better if we keep all the time locals in the same format (either long or TimeSpan) instead of mixing

src/ReverseProxy/Service/Proxy/StreamCopier.cs
MihaZupan, line 49: If we're changing the `IClock` API away from raw ticks, this local isn't needed.

It's better if we keep all the time locals in the same format (either long or TimeSpan) instead of mixing.

Since `readTime` and `writeTime` are already `DateTime` ticks (100 ns), the `StopwatchTicksToDateTimeTicks` has to be removed.

mcpride:
> > in my opinion your currently implemented Service Fabric concept mappings to YARP doesn't work very well because a Service Fabric service can provide more than one endpoint.
> 
> That's true. You **can** specify a custom ListenerName which will influence how we choose each replica's endpoint, but you're right that, as it is, we do not provide a mechanism for you to expose more than one endpoint at the same time from each service. Can you file an issue to track and link to this discussion? Please include there some of your motivation why this is an important scenario, it will help us prioritize accordingly.

Yes, for example if you host a 3rd party guest app (maybe a tomcat app) and these app provides more than one api on different ports (e.g. an administration or control api on a different port instead of a different route). Or you write your own hosting service for such a 3rd party process which provides additional endpoints (e.g. see my hashi vault hosting example, which provides an additional S3 storage endpoint for hashicorp's vault tool: https://github.com/mcpride/servicefabric-vault-integration/blob/master/src/VaultService/PackageRoot/ServiceManifest.xml)

I modified your concepts with the service manifist extensions a bit and i used a schemaless xml which will then be parsed to the labels. As you can see here I defined "Endpoints" per service extension:

```xml
...
<Extensions>
  <Extension Name="Proxy">
    <Service xmlns="http://schemas.microsoft.com/2015/03/fabact-no-schema" enableDynamicOverrides="false">
      <Endpoints>
        <Endpoint id="WebApiEndpoint" enable="true">
          <Routes>
            <Route id="RouteWebApi" order="1">
              <Match path="/WebStateless/api/{**catch-all}" />
            </Route>
            <Route id="RouteWebUI" order="1">
              <Match path="/WebStateless/ui/{**catch-all}" />
            </Route>
          </Routes>
        </Endpoint>
        <Endpoint id="ControlApiEndpoint" enable="true">
          <Routes>
            <Route id="RouteWebApi" order="1">
              <Match path="/WebStateless/admin/{**catch-all}" />
            </Route>
          </Routes>
        </Endpoint>
      </Endpoints>
    </Service>
  </Extension>
</Extensions>
...
```
I'm currently not sure how I should handle endpoints without a name. ... Also should healthcare routes be a child of the service or of an endpoint (which at the end builds the cluster)?

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/CachedServiceFabricCaller.cs
```diff
@@ -0,0 +1,189 @@
...
+            _replicaListCache = new Cache<IEnumerable<ReplicaWrapper>>(_clock, CacheExpirationOffset);
+            _serviceManifestCache = new Cache<string>(_clock, CacheExpirationOffset);
+            _serviceManifestNameCache = new Cache<string>(_clock, CacheExpirationOffset);
+            _propertiesCache = new Cache<IDictionary<string, string>>(_clock, CacheExpirationOffset);
+        }
+
+        public async Task<IEnumerable<ApplicationWrapper>> GetApplicationListAsync(CancellationToken cancellation)
+        {
+            return await TryWithCacheFallbackAsync(
+                operationName: "IslandGateway.ServiceFabric.GetApplicationList",
```
mcpride, line 61: All refererences to "Island Gateway" should be removed for public release.

mcpride (commented):
Code (CachedServiceFabricCaller.cs) and comment refererences to "Island Gateway" should be renamed for public release.

src/Integrations/ServiceFabric/ServiceDiscovery/Worker/ServiceFabricExtensionConfigProvider.cs
davidni: As discussed today, this needs further investigation which we can track outside of this PR.

src/ReverseProxy.ServiceFabric/FabricWrapper/QueryClientWrapper.cs
davidni: Let's take it separately from this PR. I am concerned that attempting a quick fix for this could make things worse without careful testing, and it wouldn't provide any tangible benefit (other than paying engineering debt, which I reckon does feel good ;))

samples/ReverseProxy.ServiceFabric.Sample/ReverseProxy.ServiceFabric.Sample.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFrameworks>net5.0;netcoreapp3.1</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+    <RootNamespace>Microsoft.ReverseProxy.ServiceFabric.Sample</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\ReverseProxy\Microsoft.ReverseProxy.csproj" />
```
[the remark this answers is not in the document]
davidni: As a sample I think being explicit helps folks understand what is actually being pulled in, but I agree your suggestion would work, Please reactivate if you feel strongly, I don't mind taking your suggestion, though it isn't my preference.

src/ReverseProxy.ServiceFabric/Microsoft.ReverseProxy.ServiceFabric.csproj
```diff
@@ -0,0 +1,34 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Service Fabric integration for Microsoft.ReverseProxy</Description>
+    <TargetFrameworks>net5.0;netcoreapp3.1</TargetFrameworks>
+    <OutputType>Library</OutputType>
+    <RootNamespace>Microsoft.ReverseProxy.ServiceFabric</RootNamespace>
+    <PlatformTarget>x64</PlatformTarget>
```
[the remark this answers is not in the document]
davidni: SF nugets use native libs for their heavy lifting

src/ReverseProxy/Utilities/IClock.cs
```diff
@@ -1,15 +1,24 @@
...
+
 namespace Microsoft.ReverseProxy.Utilities
 {
     /// <summary>
-    /// Abstraction over time providers (Environment.TickCount64, Stopwatch.GetTimestamp)
+    /// Abstraction over monotonic time providers
+    /// (Environment.TickCount64, Stopwatch.GetTimestamp, as opposed to DateTime).
     /// </summary>
-    internal interface IClock
+    public interface IClock
```
Tratcher, line 14: Do we really want this to be public? Or would it be better for SF to have its own internal test abstraction? SF doesn't rely on sharing an implementation anywhere does it?

davidni: Ideally these would be common enough to be in neither of these projects, perhaps dotnet-extensions... It sure would be a shame to have to duplicate the `ManualClock` test logic in case the 2 projects cannot share the same abstractions.

Tratcher: 😁 https://github.com/dotnet/runtime/issues/36617

There's not much duplication is there? Aren't RP and SF using different APIs on IClock?

davidni: > SF doesn't rely on sharing an implementation anywhere does it?

It does not, but it benefits from being able to reuse the same `ManualClock` test implementation. Another reason to make iclock public is to help others write E2E integration tests with YARP where they may need to manipulate time. I'll leave public unless there's strong pushback against it.

src/ReverseProxy.ServiceFabric/Configuration/ServiceFabricServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Extensions for <see cref="IServiceCollection"/>
+    /// used to register Service Fabric integration components.
+    /// </summary>
+    public static class ServiceFabricServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Adds the services needed to integrate Service Fabric with the Island Gateway to Dependency Injection.
+        /// </summary>
+        public static IReverseProxyBuilder AddServiceFabricDiscovery(this IReverseProxyBuilder builder)
```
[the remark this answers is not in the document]
davidni: Done as requested. FWIW, I find those terms rather misleading as `Load` generally indicates a one-off process that starts and finishes soon thereafter, whereas Service Discovery will keep executing in the background. My comment also applies to loading from config, as YARP doesn't just load startup configs, but continues to react to future changes. We may want to revisit naming in the future.

Tratcher (commented):
Follow-up
- Custom exception types
- IClock
- LoadFromServiceFabric should take options.
- What else?

This looks OK to merge, but I'll let Alexander handle the final sign-off and merge.

src/ReverseProxy/Utilities/IClock.cs
alnikola: @davidni Could you please xmldocs to all `IClock` members if you want to make it public?

@davidfowl Are you OK with making `IClock` public to avoid test code duplication as discussed above?

alnikola (commented):
It's almost ready to be merged, there is only a few items left:
- Confirm making IClock public
- Close on "Backend" vs "Cluster" in labels questions
- Submit at least one issue capturing all follow-up work items:
  - Review error handling logic and harden it where necessary
  - Label name casing
  - Custom exception types
  - LoadFromServiceFabric should take options
  - Get rid of unnecessary closure (e.g. in `PropertyManagementClientWrapper.GetPropertyAsync`)
  - Parse all supported `*Options` (e.g. `LoadBalancingOptions`, `SessioAffinityOptions`, etc.) in `LabelsParser`
  - Convert static classes to interfaces + implementations where it's appropriate (e.g. `LabelsParser`)
  - Apply Logging Extension pattern instead of direct `ILogger` calls.
  - Finalize the extension's name discussion

alnikola:
@davidni I submitted all necessary follow-up items listed below. Please, only confirm `IClock` API and address the two remaining comment, Then, the PR will be merged.

Follow-up items:
  - Review error handling logic and harden it where necessary #552
  - Label name casing #553
  - Custom exception types #554
  - LoadFromServiceFabric should take options #555
  - Get rid of unnecessary closure (e.g. in `PropertyManagementClientWrapper.GetPropertyAsync`) #556
  - Parse all supported `*Options` (e.g. `LoadBalancingOptions`, `SessioAffinityOptions`, etc.) in `LabelsParser` #557
  - Convert static classes to interfaces + implementations where it's appropriate (e.g. `LabelsParser`) #558
  - Apply Logging Extension pattern instead of direct `ILogger` calls. #559
  - Finalize the extension's name discussion #560

cc: @Tratcher

samples/ReverseProxy.ServiceFabric.Sample/README.md
```diff
@@ -0,0 +1,102 @@
...
+ <ServiceManifest Name="Service1Pkg" Version="1.0.0" xmlns="http://schemas.microsoft.com/2011/01/fabric">
+   <ServiceTypes>
+     <StatelessServiceType ServiceTypeName="Service1Type" >
+       <Extensions>
++        <Extension Name="YARP-preview">
++          <Labels xmlns="http://schemas.microsoft.com/2015/03/fabact-no-schema">
++            <Label Key="YARP.Enable">true</Label>
++            <Label Key="YARP.Routes.route1.Hosts">example.com</Label>
++            <!-- Optional: enable active health probes -->
++            <Label Key='YARP.Backend.Healthcheck.Active.Enabled'>true</Label>
```
alnikola, line 79: Please, add the mandatory 'Policy' parameter. In example
```XML
<Label Key='YARP.Backend.Healthcheck.Active.Policy'>ConsecutiveFailures</Label>
```

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            var clusterId = GetClusterId(serviceName, labels);
+
+            var cluster = new Cluster
+            {
+                Id = clusterId,
+                LoadBalancing = new LoadBalancingOptions(), // TODO
+                HealthCheck = new HealthCheckOptions
+                {
+                    Active = new ActiveHealthCheckOptions
```
alnikola, line 134: The mandatory 'Policy' parameter is not parsed, thus the configuration will be **always invalid** if the active health checks is enabled. Please, fix.

alnikola (changes requested):
Found at least one blocking bug.

src/ReverseProxy.ServiceFabric/FabricWrapper/HealthClientWrapper.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Microsoft.ReverseProxy.ServiceFabric
+{
+    /// <inheritdoc/>
+    internal class HealthClientWrapper : IHealthClientWrapper
+    {
+        private readonly FabricClient.HealthClient _healthClient;
+
+        public HealthClientWrapper()
+        {
+            _healthClient = new FabricClient().HealthManager;
```
Kahbazi, line 16: Does `FabricClient` need to be disposed?

alnikola:
Hmm, it seems strange. The last comment on this PR has just disappeared.

Kahbazi:
> Hmm, it seems strange. The last comment on this PR has just disappeared.

@alnikola Sorry, it was me :grin: , I added a comment about disposing `FabricClient` and then find out that it's already been discussed so I delete my comment.

samples/ReverseProxy.ServiceFabric.Sample/README.md
```diff
@@ -0,0 +1,102 @@
...
++            <Label Key='YARP.Backend.Healthcheck.Active.Path'>api/health</Label>
```
alnikola, line 80: Please, remove this since it won't work for a sample service.

samples/ReverseProxy.ServiceFabric.Sample/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+```diff
+ <ServiceManifest Name="Service1Pkg" Version="1.0.0" xmlns="http://schemas.microsoft.com/2011/01/fabric">
+   <ServiceTypes>
+     <StatelessServiceType ServiceTypeName="Service1Type" >
+       <Extensions>
++        <Extension Name="YARP-preview">
++          <Labels xmlns="http://schemas.microsoft.com/2015/03/fabact-no-schema">
++            <Label Key="YARP.Enable">true</Label>
++            <Label Key="YARP.Routes.route1.Hosts">example.com</Label>
```
alnikola, line 77: Host-based matching is not an ideal choice for a sample. It should have been changed to path matching, but by default YARP will add the specified path pattern to the proxied request URI which will disrupt the internal SF routing. It can be later mitigated by applying a transform, once they are supported here. For now, let's add simple `/{**catch-all} to make it work.

```
<Label Key="YARP.Routes.route1.Path">/{**catch-all}</Label>
```

alnikola:
> Hello,
in my opinion your currently implemented Service Fabric concept mappings to YARP doesn't work very well because a Service Fabric service can provide more than one endpoint. So the YARP cluster should be mapped from a combination of service and endpoint name.

Simply producing a new cluster for each service-endpoint pair will not be compatible with the current load balancing implementation, because YARP will handle each of such clusters independently.

mcpride:
> > Hello,
> > in my opinion your currently implemented Service Fabric concept mappings to YARP doesn't work very well because a Service Fabric service can provide more than one endpoint. So the YARP cluster should be mapped from a combination of service and endpoint name.
> 
> Simply producing a new cluster for each service-endpoint pair will not be compatible with the current load balancing implementation, because YARP will handle each of such clusters independently.

Why do you think that clustering of endpoints isn't compatible? Healthcheck might be a bit redundant on endpoint level instead of checking the health of the service itself but i can't see a problem with this.

alnikola:
>Why do you think that clustering of endpoints isn't compatible? Healthcheck might be a bit redundant on endpoint level instead of checking the health of the service itself but i can't see a problem with this.

The main concern is load balancing (in case it's enabled of course). Currently, load balancing considers only destinations belonging to one cluster. Meaning, it's based on the assumption that clusters are independent. However, in the above approach multiple YARP cluster will look **at the same SF service**. Thus, that, what looks like a balanced load from YARP perspective, might not be actually balanced from SF service perspective.

alnikola:
Note. Load balancing config is not yet supported here, but it will be soon.

src/ReverseProxy/Utilities/IClock.cs
davidni: Added xmldocs in new iteration (submitting soon)

src/ReverseProxy.ServiceFabric/ServiceDiscovery/IServiceExtensionLabelsProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <remarks>
+    /// The key-value labels to configure the Proxy are first read from the YARP extension
+    /// in the "Extensions" section of a service in the ServiceManifest.xml. Example:
+    /// <![CDATA[
+    /// <StatelessServiceType ServiceTypeName="ExampleServiceTypeName">
+    ///   <Extensions>
+    ///     <Extension Name="YARP-preview">
+    ///       <Labels xmlns="http://schemas.microsoft.com/2015/03/fabact-no-schema">
+    ///         <Label Key="YARP.Enable">true</Label>
+    ///         <Label Key="YARP.Backend.BackendId">exampleId</Label>
```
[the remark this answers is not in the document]
davidni: Using the term `cluster` in SF config will be extremely confusing to SF users, for whom `cluster` has a very specific meaning. Keeping as is for now, keeping in mind that we can still rename all labels prior to announcing a v1.0 contract.

src/ReverseProxy.ServiceFabric/Exceptions/ConfigException.cs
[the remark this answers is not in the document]
davidni: @alnikola was kind enough to file follow up issues and he added the link to them in the discussion below. Thanks Alex!

samples/ReverseProxy.ServiceFabric.Sample/README.md
```diff
@@ -0,0 +1,104 @@
...
+   <ServiceTypes>
+     <StatelessServiceType ServiceTypeName="Service1Type" >
+       <Extensions>
++        <Extension Name="YARP-preview">
++          <Labels xmlns="http://schemas.microsoft.com/2015/03/fabact-no-schema">
++            <Label Key="YARP.Enable">true</Label>
++            <Label Key="YARP.Routes.route1.Path">{**catchall}</Label>
++            <!-- Optional: enable active health probes -->
++            <Label Key='YARP.Backend.Healthcheck.Active.Enabled'>true</Label>
++            <Label Key='YARP.Backend.Healthcheck.Active.Path'>api/health</Label>
```
alnikola, line 80: `api/health` health probing path is not supported by a basic Asp.Net Core SF service, so I'd suggest to removed it.
```suggestion
```

alnikola (approved):
LTGM

mcpride:
> > Why do you think that clustering of endpoints isn't compatible? Healthcheck might be a bit redundant on endpoint level instead of checking the health of the service itself but i can't see a problem with this.
> 
> The main concern is load balancing (in case it's enabled of course). Currently, load balancing considers only destinations belonging to one cluster. Meaning, it's based on the assumption that clusters are independent. However, in the above approach multiple YARP cluster will look **at the same SF service**. Thus, that, what looks like a balanced load from YARP perspective, might not be actually balanced from SF service perspective.

At first ... thanx for your answer! Sorry, but I didn't see the problem why endpoints of a service shouldn't be handled indepentently (including healthcheck). If a service has for example a control-api endpoint and a data-api endpoint defined and in the "service fabric cluster" are parallel instances of the service hosted on different nodes, why can't all control-api endpoints build a "proxy cluster" and all data-api endpoints another? The apis operate independently in context of proxy routing and this maybe also includes optional health check routes per endpoint. If a service instance becomes busy then the health checks of both endpoints should report this status independently. I think that health checks are not directly tied to services. Also metrics like CPU load is not tied to service, it is a property of the node (the machine) on which the service instance lives. 

So my question is: What are the specific blockers for health checks in this context? 

My SF mapping here is that SF services are a kind of hosting infrastructure (application nodes) for optional state persistence and independent listeners/endpoints which are the finer grained "microservices" (yes, i know this is a bit pragmatic and doesn't follow the microservice pattern) ;-)

mcpride:
Another question is: Shouldn't the work of collecting healthcare metrics for nodes, services or endpoints  be delegated to another service which is much more specialized? 
There exists the project [FabricObserver](https://github.com/microsoft/service-fabric-observer) which is maybe a good candidate.

alnikola:
My concern was only about **load balancing** operating in the given case of service-endpoint pair to cluster mapping. Health checks should work fine.

alnikola:
> Shouldn't the work of collecting healthcare metrics for nodes, services or endpoints be delegated to another service which is much more specialized?

YARP's health check mechanism is aimed to support the main YARP scenarios. However, its health checks subsystem is quite extensible so you can integrate it with some SF-specific monitoring tool if that's necessary in your case.

PR #260, opened 2020-06-23
Update docs for preview2

Tratcher:
- Getting started
- Supported runtimes

I don't plan to merge until we have the package available on nuget.

PR #263, opened 2020-06-24
Update getting_started.md

amweiss:
Fix flag in getting started for `dotnet new` to be `net5.0` instead of `netcoreapp5.0`


Tratcher:
Thanks for checking. We'd been holding off on this until we moved to support 5.0.0-preview6 which will be part of our preview2 release. We've already updated this as part of the preview2 docs update here: https://github.com/microsoft/reverse-proxy/pull/260

amweiss:
Sorry I missed that, I just saw that it had the new framework name in the csproj but not the command already. Thanks for keeping the docs updated!

Issue #122, opened 2020-04-28
Configure authorization for proxy routes


Tratcher:
When defining proxy routes we should be able to specify Authorization criteria for that route and avoid proxying requests until that criteria is met.

This will need some support to be specified from config (probably by specifying the name of an authz policy that's configured in code), but it should also be configurable when using other data sources.

Kahbazi:
> avoid proxying requests until that criteria is met

This would be done by `AuthorizationMiddleware`, YARP doesn't need to do anything, right? 

This could be the config.
```json
"Routes": [
  {
    "RouteId": "backend1/route1",
    "BackendId": "backend1",
    "Match": {
      "Methods": [ "GET", "POST" ],
      "Host": "localhost",
      "Path": "/{**catchall}"
    },
    "AllowAnonymous": false,
    "Authorize": [
      {
        "Policy": "Policy1",
        "Roles": "Role1,Role2",
        "AuthenticationSchemes": "jwt"
      },
      {
        "Policy": "Policy2",
        "Roles": "Role3",
        "AuthenticationSchemes": "jwt"
      }
    ]
  }
]
```
and we could create `IAuthorizeData`, `IAllowAnonymous` in
https://github.com/microsoft/reverse-proxy/blob/ea7164c689ea06343b99a30aa7feecef4eda215d/src/ReverseProxy.Core/Service/DynamicEndpoints/RuntimeRouteBuilder.cs#L51-L66

My only question is that do we need two identical class for Authorize? One for `ProxyRoute` and one for `ParsedRoute`?

Tratcher:
> > avoid proxying requests until that criteria is met
> 
> This would be done by `AuthorizationMiddleware`, YARP doesn't need to do anything, right?

Right. MVC does have an extra safety check in it to make sure the AuthorizationMiddleware ran, and we might copy that, but that should be all we need.

I hesitate to actually craft Authorize policies from scratch as part of the proxy config. If we really want to build those in config then that should be an AspNetCore feature outside of the proxy section that we reference by policy name.

Kahbazi:
> I hesitate to actually craft Authorize policies from scratch as part of the proxy config. If we really want to build those in config then that should be an AspNetCore feature outside of the proxy section that we reference by policy name.

I'm on the same page with you. `Policy1` and `Policy2` that I wrote in config would be a reference to ASP.NET Core policies. Just like the authentication scheme.

Do you think there should be also a config to set authorization for all routes?

[...]

PR #265, opened 2020-06-25
 Enable authorization for proxy routes

Tratcher:
#122 This enables proxy routes to integrate with the existing ASP.NET Core authentication and authorization infrastructure.

This should unblock most users. I expect some feedback on the scenario for flowing user information to the destination server, we'll need to wait and see what people's scenarios are.


src/ReverseProxy/Service/DynamicEndpoints/RuntimeRouteBuilder.cs
```diff
@@ -75,6 +79,19 @@ public RouteConfig Build(ParsedRoute source, ClusterInfo cluster, RouteInfo runt
...
 
+            if (string.Equals(AuthorizationContants.Default, source.Authorization, StringComparison.OrdinalIgnoreCase))
+            {
+                endpointBuilder.Metadata.Add(DefaultAuthorization);
+            }
+            else if (string.Equals(AuthorizationContants.Anonymous, source.Authorization, StringComparison.OrdinalIgnoreCase))
+            {
+                endpointBuilder.Metadata.Add(AnonymousAuthorization);
+            }
+            else if (!string.IsNullOrEmpty(source.Authorization))
```
Kahbazi, line 90: Is there a way to add multiple Policy to a route?

Kahbazi, line 90: Is there a way to set Authentication Scheme for AuthorizeAttribute?

Tratcher: No, what would that mean?

This is the same as adding an Authorize attribute on a controller.

Tratcher: The auth schemes can be set on the policy, is that enough?

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/AuthorizationContants.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Abstractions.RouteDiscovery.Contract
+{
+    internal static class AuthorizationContants
```
alnikola, line 6:
```suggestion
    internal static class AuthorizationConstants
```

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/AuthorizationContants.cs
```diff
@@ -0,0 +1,11 @@
...
+    {
+        internal const string Default = "Default";
```
alnikola, line 8: I believe it's better to set `public` visibility on type's member if the type itself is scoped `internal`.

docs/docfx/articles/authn-authz.md
```diff
@@ -0,0 +1,112 @@
...
+    app.UseAuthorization();
+
+    app.UseEndpoints(endpoints =>
+    {
+        endpoints.MapReverseProxy();
+    });
+}
+```
+
+See the [Authentication](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/) docs for setting up your preferred kind of authentication.
```
alnikola, line 72: Would it be useful to provide an example of creating and registering a custom policy?

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
```diff
@@ -31,6 +31,14 @@ public sealed class ProxyRoute : IDeepCloneable<ProxyRoute>
         /// </summary>
         public string ClusterId { get; set; }
 
+        /// <summary>
+        /// The name of the AuthorizationPolicy to apply to this route.
+        /// If not set then only the FallbackPolicy will apply.
+        /// Set to "Default" to enable authorization with the applications default policy.
+        /// Set to "Anonymous" to disable all authorization checks for this route.
+        /// </summary>
+        public string Authorization { get; set; }
```
HaoK, line 40: Seems clearer if this was called AuthorizationPolicyName maybe?

docs/docfx/articles/authn-authz.md
Tratcher: Like this? https://github.com/microsoft/reverse-proxy/pull/265/files#diff-e96ca1546d1b5586242bbfb46c19abb9R49-R50

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/AuthorizationContants.cs
Tratcher: Why? You get the same behavior either way.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
Tratcher: That's a bit verbose. AuthorizeAttribute just calls it Policy. Compromise at AuthorizationPolicy?

HaoK: Yeah that sounds good, adding Policy is good enough

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/AuthorizationConstants.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Abstractions.RouteDiscovery.Contract
+{
+    internal static class AuthorizationConstants
+    {
+        internal const string Default = "Default";
+        internal const string Anonymous = "Anonymous";
```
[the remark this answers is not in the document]
Tratcher: `AuthorizationPolicy: Everyone`? Everyone doesn't say what you're allowing everyone to do. Anonymous seems clearer about the intent. I can put back the Allow if we need it.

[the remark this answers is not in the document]
HaoK: Right but just to be precise, the actual behavior is it disables authorization entirely and lets everything through.  AllowAnonymous is sort of the carry over of the old attribute.  It doesn't make sense as a policy name really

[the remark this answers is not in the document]
HaoK: Its AllowEverything or AllowAll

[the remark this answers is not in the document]
HaoK: What i'm trying to get at, is anonymous or not that has no actual effect, it is Authorized = Always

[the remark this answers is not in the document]
HaoK: Maybe `AuthorizationPolicy: Always` is an apt name

docs/docfx/articles/authn-authz.md
alnikola: Ah, it's already in there. I missed that.

src/ReverseProxy/Service/Config/RuleParsing/RouteValidator.cs
```diff
@@ -141,5 +146,38 @@ private static bool ValidateMethods(IReadOnlyList<string> methods, string routeI
...
+                || string.Equals(AuthorizationConstants.Anonymous, authorizationPolicyName, StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            try
+            {
+                // The default implementation is sync.
+                // https://github.com/dotnet/aspnetcore/blob/5155e11120cf7ee2e07383225057f66512f00fde/src/Security/Authorization/Core/src/DefaultAuthorizationPolicyProvider.cs#L73
+                var policy = _authorizationPolicyProvider.GetPolicyAsync(authorizationPolicyName).GetAwaiter().GetResult();
```
davidfowl, line 167: Uhhhhh what?

samples/ReverseProxy.Auth.Sample/Startup.cs
```diff
@@ -0,0 +1,72 @@
...
+            _configuration = configuration;
+        }
+
+        /// <summary>
+        /// This method gets called by the runtime. Use this method to add services to the container.
+        /// </summary>
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddControllers();
+            services.AddMvc();
```
davidfowl, line 34: Why do you need AddMvc?

samples/ReverseProxy.Auth.Sample/Startup.cs
```diff
@@ -0,0 +1,72 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.ReverseProxy.Auth.Sample
```
davidfowl, line 11: Why is this sample doing auth with MVC?

Tratcher: Because I needed something visual and self-contained.

Or do you mean I should use razor pages instead?

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/AuthorizationConstants.cs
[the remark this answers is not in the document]
Tratcher: Alternatively I could remove this, it's only needed if someone sets FallbackPolicy. Those people can define their own anonymous policy if they need it.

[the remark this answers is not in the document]
HaoK: Yeah removing is even better

src/ReverseProxy/Service/ConfigModel/ParsedRoute.cs
```diff
@@ -101,6 +109,11 @@ internal int GetConfigHash()
                 hash ^= ClusterId.GetHashCode();
             }
 
+            if (!string.IsNullOrEmpty(AuthorizationPolicy))
```
davidfowl, line 112: I just looked at this hash function and realized how crazy expensive this is. How often is it used? Not very I hope.

Tratcher: Once per config reload.

Issue #268, opened 2020-06-26
Query string behaviour with PathPattern transform


jmezach:
I've been trying out the new preview and have been diving into what's possible with the new transforms feature. Essentially I'm trying to reproduce what I currently have using Ocelot but with YARP. I'm seeing some interesting behaviour with the PathPattern transform, that I couldn't find in the documentation. I'm not sure if it's a bug, or just me not understanding the feature, but here goes.

In my development environment I have setup an ASP.NET Core app targeting net5.0 that includes YARP. I also have an ASP.NET Core 3.1 app that serves as my backend service for testing. It has one API at `/api/values` which echoes the incoming headers back in the response, as well as the value of the `path` query string parameter.

For my configuration I have the following:

```json
"ReverseProxy": {
    "Routes": [
      {
        "RouteId": "realization",
        "ClusterId": "EchoService",
        "Match": {
          "Methods": [ "GET", "POST" ],
          "Path": "/realization/{**catchall}"
        },
        "Transforms": [
            { "PathSet": "/api/values" }
        ]
      },
      {
          "RouteId": "scheduling",
          "ClusterId": "EchoService",
          "Match": {
            "Methods": [ "GET", "POST" ],
            "Path": "/scheduling/{*path}"
          },
          "Transforms": [
              { "PathPattern": "/api/values" }
          ]
      }
    ],
    "Clusters": {
      "EchoService": {
        "LoadBalancing": {
          "Mode": "Random"
        },
        "Destinations": {
          "backend_destination1": {
            "Address": "https://localhost:5003/"
          }
        }
      }
    }
  },
```

If I do a request to `/realization` everything is fine and I get back the headers and no path because there is no path after /realization. But if I do a request to `/scheduling/shops/4/2019/52/versions/Actual` I see a request to `https://localhost:5003/api/values?path=shops/4/2019/52/versions/Actual` coming into the EchoService. First of all I wasn't necessarily expecting this since I couldn't find anything in the docs that it actually appends the rest as a query string parameter, although that is a nice feature to have.

However, I don't think it is passing the path correctly, because I believe it should be URL encoded at least. My EchoService currently replies with a 404 because of this I think.

### Some details

**How many backends are in your application?**

- [x] 1-2
- [ ] 3-5
- [ ] 6-10
- [ ] 10+

**How do you host your application?**

- [ ] Kubernetes
- [ ] Azure App Service
- [ ] Azure VMs
- [ ] Other Cloud Provider (please include details below)
[...]

PR #269, opened 2020-06-26
Drop unused paramters from PathPattern 

Tratcher:
#268 (Preview2 feedback)
The way we used the binder before it would append unused route parameters to the path as a query string.
A) We don't want the unused parameters.
B) That query string was being encoded inside the path.

@javiercn

### DIFF
--- a/src/ReverseProxy/Service/RuntimeModel/Transforms/PathRouteValuesTransform.cs
+++ b/src/ReverseProxy/Service/RuntimeModel/Transforms/PathRouteValuesTransform.cs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Microsoft.AspNetCore.Routing.Patterns;
+using System;
 using Microsoft.AspNetCore.Routing.Template;
 
 namespace Microsoft.ReverseProxy.Service.RuntimeModel.Transforms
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Service/RuntimeModel/Transforms/PathRouteValuesTransform.cs
```diff
@@ -11,21 +11,14 @@ namespace Microsoft.ReverseProxy.Service.RuntimeModel.Transforms
...
-            {
-                throw new System.ArgumentNullException(nameof(pattern));
-            }
-
-            if (binderFactory is null)
-            {
-                throw new System.ArgumentNullException(nameof(binderFactory));
-            }
-
-            _binder = binderFactory.Create(RoutePatternFactory.Parse(pattern));
```
halter73, line 28: What's the difference between `TemplateParser.Parse(pattern)` and `RoutePatternFactory.Parse(pattern)`?

Tratcher: Not much. TemplateParser.Parse calls RoutePatternFactory.Parse and then wraps the result in a RouteTemplate. RouteTemplate is what was needed for this other overload.
https://github.com/dotnet/aspnetcore/blob/e280884a94e5349bafebfed2073b1cd00d0ce1ad/src/Http/Routing/src/Template/TemplateParser.cs#L20-L21

PR #278, opened 2020-06-30
Make small fixes to configuration

johnazariah:
fix 'Endpoint' -> 'Destination' in config
fix 'cluster1' -> 'cluster2' in config

Paired with : @masenol


Tratcher:
Thanks

PR #280, opened 2020-07-01
Add instructions for using daily builds

### LINKED ISSUE #271: [Request] Nightly package builds

Now that Authorization was merged, we can fully recreate our current setup with YARP. However since preview 2 was just released I assume preview 3 is still a ways away. Are there any plans to make nightly packages available somewhere or should I make builds in our local feed?

This unblocks us for adopting YARP in the short term and evaluating it as development continues in the long term without having to create our own builds of it locally.

Tratcher:
We've started publishing daily builds to a feed so this adds instructions for accessing them. They're copied wholesale from [AspNetCore](https://github.com/dotnet/aspnetcore/blob/master/docs/DailyBuilds.md).

I'm also updating the readme to recommend previews and daily builds.

Fixes #271

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,8 @@ If you're having trouble building the project, or developing in Visual Studio, p
 
 # Getting started
 
+- See our [Getting Started](https://microsoft.github.io/reverse-proxy/articles/getting_started.html) docs.
 - Try our [previews](https://github.com/microsoft/reverse-proxy/releases).
-- Check out our [docs](https://microsoft.github.io/reverse-proxy/) and tutorials, we'll be publishing more as the project develops!
 - Try our latest [daily build](/docs/DailyBuilds.md).
 
 # Roadmap
--- a/README.md
+++ b/README.md
@@ -33,7 +33,9 @@ If you're having trouble building the project, or developing in Visual Studio, p
 
 # Getting started
 
-Take a look at the [sample apps](samples/), for some examples of how to use YARP. We'll be publishing more [docs](https://microsoft.github.io/reverse-proxy/) and tutorials as the project develops!
+- Try our [previews](https://github.com/microsoft/reverse-proxy/releases).
+- Check out our [docs](https://microsoft.github.io/reverse-proxy/) and tutorials, we'll be publishing more as the project develops!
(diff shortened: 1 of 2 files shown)


docs/DailyBuilds.md
```diff
@@ -0,0 +1,32 @@
...
+  <configuration>
+      <packageSources>
+          <clear />
+          <add key="net5" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
+          <add key="NuGet.org" value="https://api.nuget.org/v3/index.json" />
+      </packageSources>
+  </configuration>
+  ```
+
+  *NOTE: This NuGet.Config should be with your application unless you want nightly packages to potentially start being restored for other apps on the machine.*
```
halter73, line 22: After this, shouldn't we add a step like the following?

```
> dotnet add package Microsoft.ReverseProxy --version 1.0.0-*
```

README.md
```diff
@@ -33,7 +33,9 @@ If you're having trouble building the project, or developing in Visual Studio, p
 
 # Getting started
 
-Take a look at the [sample apps](samples/), for some examples of how to use YARP. We'll be publishing more [docs](https://microsoft.github.io/reverse-proxy/) and tutorials as the project develops!
+- Try our [previews](https://github.com/microsoft/reverse-proxy/releases).
```
halter73, line 36: Why link to the releases page with "Try our previews"? That just gives you a zipped version of what you could have cloned from github when we snapped a release.

Why not provide instructions for using the preview from NuGet.org similar to what we're doing for the daily builds?

docs/DailyBuilds.md
Tratcher: That's covered in the Getting Started guide. I could link to that from here.

README.md
Tratcher: Interesting point. The releases page already links to Getting Started instructions, but I've updated the releases to explicitly link to the package on nuget.org.

I should also make the first link here go to Getting Started, which includes the nuget instructions.

Tratcher:
How's that?

Issue #285, opened 2020-07-02
Configure CORS for proxy routes


Kahbazi:
### What should we add or change to make your life better?
Support adding CORS metadata for individual routes and group routes (https://github.com/microsoft/reverse-proxy/issues/164).

[CorsMiddleware](https://github.com/dotnet/aspnetcore/blob/master/src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs#L144-L173) supports three types: `IDisableCorsAttribute`, `ICorsPolicyMetadata` and `IEnableCorsAttribute`.
I suggest YARP should only supports `IDisableCorsAttribute` and  `IEnableCorsAttribute`. It would be the same as authorization that we can only set Policy name and not authentication schemes or roles.

### Why is this important to you?
I want to use YARP as a backend for a SPA and I want to check CORS before sending the request to the API Server.

Tratcher:
Forked from https://github.com/microsoft/reverse-proxy/issues/87#issuecomment-652933954

PR #294, opened 2020-07-02
Enable cors for for proxy routes

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/285

### DIFF
--- a/docs/docfx/articles/cors.md
+++ b/docs/docfx/articles/cors.md
@@ -74,8 +74,8 @@ public void Configure(IApplicationBuilder app)
 
 ### DefaultPolicy
 
-Specifying the value `default` in a route's CORS parameter means that route will use the policy defined in [CorsOptions.DefaultPolicy](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.cors.infrastructure.corsoptions.defaultpolicyname).
+Specifying the value `default` in a route's `CorsPolicy` parameter means that route will use the policy defined in [CorsOptions.DefaultPolicy](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.cors.infrastructure.corsoptions.defaultpolicyname).
 
 ### Disable CORS
 
-Specifying the value `disable` in a route's CORS parameter means the CORS middleware will refuse the CORS requests.
+Specifying the value `disable` in a route's `CorsPolicy` parameter means the CORS middleware will refuse the CORS requests.
--- a/docs/docfx/articles/cors.md
+++ b/docs/docfx/articles/cors.md
@@ -1,0 +1,81 @@
+# Cross-Origin Requests (CORS)
+
+Introduced: preview3
+
+## Introduction
+
+The reverse proxy can handle cross-origin requests before they are proxied to the destination servers. This can reduce load on the destination servers and ensure consistent policies are implemented across your applications.
+
+## Defaults
+The requests won't be automatically matched for cors preflight requests unless enabled in the route or application configuration.
+
+## Configuration
+CORS policies can be specified per route via [ProxyRoute.CorsPolicy](xref:Microsoft.ReverseProxy.Abstractions.ProxyRoute.CorsPolicy) and can be bound from the `Routes` sections of the config file. As with other route properties, this can be modified and reloaded without restarting the proxy. Policy names are case insensitive.
+
+Example:
+```JSON
+{
+  "ReverseProxy": {
+    "Routes": [
+      {
+        "RouteId": "route1",
+        "ClusterId": "cluster1",
+        "CorsPolicy": "customPolicy",
+        "Match": {
+          "Host": "localhost"
+        },
+      }
+    ],
+    "Clusters": {
+      "cluster1": {
+        "Destinations": {
+          "cluster1/destination1": {
+            "Address": "https://localhost:10001/"
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+[CORS policies](https://docs.microsoft.com/en-us/aspnet/core/security/cors?view=aspnetcore-3.1#cors-with-named-policy-and-middleware) are an ASP.NET Core concept that the proxy utilizes. The proxy provides the above configuration to specify a policy per route and the rest is handled by existing ASP.NET Core CORS Middleware.
+
+CORS policies can be configured in Startup.ConfigureServices as follows:
+```
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddCors(options =>
+    {
+        options.AddPolicy("customPolicy", builder =>
+        {
+            builder.AllowAnyOrigin();
+        });
+    });
+}
+```
+
+In Startup.Configure add the CORS middleware between Routing and Endpoints.
+
+```
+public void Configure(IApplicationBuilder app)
+{
+    app.UseRouting();
+
+    app.UseCors();
+
+    app.UseEndpoints(endpoints =>
+    {
+        endpoints.MapReverseProxy();
+    });
+}
+```
+
+
+### DefaultPolicy
+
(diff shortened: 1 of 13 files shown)


Tratcher:
That was fast 😉

src/ReverseProxy/Service/Config/RuleParsing/RouteValidator.cs
```diff
@@ -177,5 +184,40 @@ private async Task<bool> ValidateAuthorizationPolicyAsync(string authorizationPo
...
+            }
+
+            if (string.Equals(CorsConstants.Disable, corsPolicyName, StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            try
+            {
+                var policy = await _corsPolicyProvider.GetPolicyAsync(_dummyHttpContext, corsPolicyName);
```
Kahbazi, line 207: I'm not sure if I did the right thing here by passing a dummy `HttpContext` to `GetPolicyAsync`. The default implementation ([DefaultCorsPolicyProvider](https://github.com/dotnet/aspnetcore/blob/master/src/Middleware/CORS/src/Infrastructure/DefaultCorsPolicyProvider.cs)) is not using HttpContext at all.

src/ReverseProxy/Service/Config/RuleParsing/RouteValidator.cs
```diff
@@ -40,6 +43,7 @@ internal class RouteValidator : IRouteValidator
             DnsLabelRegexPattern +
             @"$";
         private static readonly Regex _hostNameRegex = new Regex(HostNameRegexPattern);
+        private static readonly HttpContext _dummyHttpContext = new DefaultHttpContext();
```
Tratcher, line 46: Don't make a static context, even if it's a dummy. The type isn't thread safe. Create a local instance in ValidateCorsPolicyAsync.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
```diff
@@ -39,6 +39,14 @@ public sealed class ProxyRoute : IDeepCloneable<ProxyRoute>
         /// </summary>
         public string AuthorizationPolicy { get; set; }
 
+        /// <summary>
+        /// The name of the CorsPolicy to apply to this route.
+        /// If not set then the route won't be matched for cors preflight request.
+        /// Set to "Default" to enable cors with the default policy.
+        /// Set to "Disable" to disable cors checks for this route.
+        /// </summary>
+        public string CorsPolicy { get; set; }
```
Kahbazi, line 48: If I understand the routing correctly, with `Disable`, the proxy endpoint would be selected and the `CorsMiddleware` would set `Status204NoContent`. But when `CorsPolicy` is empty, no endpoint would be selected and response would probably end up being a `Status404NotFound`.

Tratcher: Yeah, I was looking at that. `Disable` doesn't turn off Cors processing, it refuses Cors requests.
https://github.com/dotnet/aspnetcore/blob/a729c4230eefd4b5c3e21f9496de608cd25a72b8/src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs#L144-L153
Please clarify the comment.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
```diff
@@ -39,6 +39,14 @@ public sealed class ProxyRoute : IDeepCloneable<ProxyRoute>
         /// </summary>
         public string AuthorizationPolicy { get; set; }
 
+        /// <summary>
+        /// The name of the CorsPolicy to apply to this route.
+        /// If not set then the route won't be matched for cors preflight request.
```
Tratcher, line 44:
```suggestion
        /// If not set then the route won't be automatically matched for cors preflight requests.
        /// Routes that match the Options method, or all methods, will still be able to process cors 
        /// preflight requests at the destination server.
```

src/ReverseProxy/Service/Config/RuleParsing/RouteValidator.cs
Tratcher: Providing a dummy makes sense, it doesn't allow nulls.

Kahbazi: Why `CorsPolicyProvider.GetPolicyAsync` needs a `HttpContext` in the first place, but `authorizationPolicyProvider.GetPolicyAsync` doesn't? 
Is there any chance that `CorsPolicyProvider.GetPolicyAsync(HttpContext,string)` could be obsoleted and  `CorsPolicyProvider.GetPolicyAsync(string)` added instead?

Tratcher: > Why `CorsPolicyProvider.GetPolicyAsync` needs a `HttpContext` in the first place

I don't know.

> Is there any chance that `CorsPolicyProvider.GetPolicyAsync(HttpContext,string)` could be obsoleted and `CorsPolicyProvider.GetPolicyAsync(string)` added instead?

Updating interfaces is one of the harder kinds of breaking changes to make. We'll still need to support 3.1 regardless.

davidfowl:
This still requires the user to explicitly call UseCors right?

Kahbazi:
> This still requires the user to explicitly call UseCors right?

Yes.

Kahbazi:
When a route with cors is selected, is there any harm to send the cors headers to the cluster?

Tratcher:
> When a route with cors is selected, is there any harm to send the cors headers to the cluster?

You mean should they be stripped off before forwarding? No. If someone wants to do that they can handle it in the transforms.

docs/docfx/articles/cors.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+
+### DefaultPolicy
+
+Specifying the value `default` in a route's CORS parameter means that route will use the policy defined in [CorsOptions.DefaultPolicy](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.cors.infrastructure.corsoptions.defaultpolicyname).
+
+### Disable CORS
+
+Specifying the value `disable` in a route's CORS parameter means the CORS middleware will refuse the CORS requests.
```
halter73, line 81: Nit:

```suggestion
Specifying the value `disable` in a route's `CorsPolicy` parameter means the CORS middleware will refuse the CORS requests.
```

Unless I'm misunderstanding and this is meant to say something else.

src/ReverseProxy/Service/ConfigModel/ParsedRoute.cs
```diff
@@ -114,6 +122,11 @@ internal int GetConfigHash()
                 hash ^= AuthorizationPolicy.GetHashCode();
             }
 
+            if (!string.IsNullOrEmpty(CorsPolicy))
+            {
+                hash ^= CorsPolicy.GetHashCode();
```
pranavkm, line 127: Why not use the `HashCode` type?

Tratcher: That's new in 5.0, right? We still support 3.1.

pranavkm: It's in 3.1: https://docs.microsoft.com/en-us/dotnet/api/system.hashcode?view=netcore-3.1

davidfowl: It's not new in 5.0 it's in 3.1 but I was reserving that for when we rewrite all of this code.

Tratcher: Yeah, don't worry about it in this PR.

Tratcher:
Thanks

PR #301, opened 2020-07-04
Query string parameter transform

jmezach:
As discussed in #268 I would like to have the ability to transform parts of the request path into query string parameters to the downstream service. This is an initial stab at adding support for that with a new `QueryStringParameterTransform` so consider this a work-in-progress that I would like to get some early feedback on to see if I'm going in the right direction.

To configure this I'm using:

```json
{
  "QueryStringParameter": "path",
  "Append": "{remainder}"
}
```

I'm treating the value of Append here as a route template that I apply to the route values of the original request. I've inspired the design on the existing `PathRouteValuesTransform` and `RequestHeaderValueTransform` classes.


src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
```diff
@@ -0,0 +1,55 @@
...
+        private readonly string _key;
+        private readonly RouteTemplate _template;
+        private readonly TemplateBinderFactory _binderFactory;
+
+        public QueryStringParameterTransform(QueryStringTransformMode mode, string key, string value, TemplateBinderFactory binderFactory)
+        {
+            _mode = mode;
+            _key = key;
+            _binderFactory = binderFactory ?? throw new ArgumentNullException(nameof(binderFactory));
+            _template = TemplateParser.Parse(value);
```
Tratcher, line 23: I thought TemplateParser produced a path template that wasn't appropriate for query strings. Writing some tests would help explore the possibilities.

jmezach: I agree that it needed some tests, so I've added those now. I think it does what you would expect it to do, URL encoding the slashes in the path so that they can be part of the query string.

test/ReverseProxy.Tests/Service/RuntimeModel/Transforms/QueryStringParameterTransformTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.AspNetCore.Routing.Template;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.ReverseProxy.Service.RuntimeModel.Transforms
+{
+    public class QueryStringParameterTransformTests
+    {
+        [Theory]
+        [InlineData("/{a}/{b}/{c}", "?z=%2F6%2F7%2F8")]
```
Tratcher, line 14: This is actually over-escaped, `/` doesn't need to be escaped in the query string.

We'll want to test a wider variety of characters here. See https://tools.ietf.org/html/rfc3986#section-3.4.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            if (context == null)
+            {
+                throw new System.ArgumentNullException(nameof(context));
+            }
+
+            var routeValues = context.HttpContext.Request.RouteValues;
+            var binder = _binderFactory.Create(_template, defaults: routeValues);
+            var value = binder.BindValues(acceptedValues: routeValues);
+            var parsedQueryString = QueryHelpers.ParseQuery(context.Query.Value);
```
Tratcher, line 36: You shouldn't need to fully parse the query here, you should be able to use something like QueryString.Add to append to the end.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
```diff
@@ -0,0 +1,55 @@
...
+                    throw new NotImplementedException(_mode.ToString());
+            }
+
+            context.Query = QueryString.Create(parsedQueryString);
+        }
+    }
+
+    public enum QueryStringTransformMode
+    {
+        Append,
```
Tratcher, line 53: Any suggestions for alternate modes here? Otherwise this can be removed.

test/ReverseProxy.Tests/Service/RuntimeModel/Transforms/QueryStringParameterTransformTests.cs
Tratcher: Add a test where there's an existing query being appended to. And one with a conflicting key.

jmezach: Looks like the escaping is built into the `TemplateBinder` and as far as I can tell there's no way to disable it. I tried UrlDecoding the output using `WebUtility.UrlDecode` but that just keeps the escaped slashes for some reason. I could just do a `string.Replace()` I guess, but that just doesn't feel right. I'll have a look to see if there's something else we can use here.

I did find [this StackOverflow question](https://stackoverflow.com/questions/2322764/what-characters-must-be-escaped-in-an-http-query-string) about what valid characters are in a query string, so yeah, I think I'll need to add quite a bit more scenario's here. I'll also see if I can add additional tests for what happens with existing query string parameters etc.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
jmezach: You're right. I did this initially but it didn't work, but I think I forgot to do the assignment afterwards thinking that `.Add()` would add to the existing instance, but it returns a new instance of course which makes sense. Fixed that now.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
jmezach: I think having a Remove option would be nice, so that you can remove query string parameters on the incoming request before they are sent to the downstream service. Perhaps a Set as well, allowing you to override existing values.

test/ReverseProxy.Tests/Service/RuntimeModel/Transforms/QueryStringParameterTransformTests.cs
jmezach: Hmm, I've tried using `TemplateBinder.GetValues()` instead of `BindValues()` and then do the replacing myself. That works, although it is kinda ugly. However, if I pas the resulting value to `QueryString.Add()` it does the escaping thing again. Looks like that is [built into](https://github.com/dotnet/aspnetcore/blob/7a9c397b0d0c272afbf5d34f3626e384abdef704/src/Http/Http.Abstractions/src/QueryString.cs#L243) QueryString as well.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
Tratcher: Remove by key? That seems like it'd be implemented as a separate transform/class. It wouldn't need any of the binder infrastructure.

Set makes some sense.

test/ReverseProxy.Tests/Service/RuntimeModel/Transforms/QueryStringParameterTransformTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        [InlineData("/{a}/foo/{b}/{c}/{d}", "?z=%2F6%2Ffoo%2F7%2F8")] // Unknown value (d) dropped
```
Tratcher, line 15: You copied these patterns from the path tests, but is this really how you'd expect to generate a query string value?

If anything I'd expect to have matched a single segment from the path like a = 6, and insert it into the query string as "z=6". We wouldn't even need the binder for that, we could look up the parameters directly in the route value dictionary by name.

jmezach: You're right, I made this harder than it has to be. I've now simplified the implementation somewhat by allowing only a `routeValueKey` to map a query string parameter to a particular value in the original route. This works well, although I still have the issue of the QueryString class escaping `/` into `%2F`, causing the last scenario to fail.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
jmezach: Hmm, what would we expect `Append` to do if there's an existing query string parameter with the same key? My first thought was to simply ignore the transform in that case, but since we're using `QueryString.Add()` now it simply adds an additional x=y which also makes sense. Any ideas here?

Tratcher: That makes sense for append to not care about duplicates. That's what Set would be for.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public override void Apply(RequestParametersTransformContext context)
+        {
+            if (context == null)
+            {
+                throw new System.ArgumentNullException(nameof(context));
+            }
+
+            var routeValues = context.HttpContext.Request.RouteValues;
+            var value = routeValues.ContainsKey(_routeValueKey) ? routeValues[_routeValueKey] : null;
```
Tratcher, line 30: TryGetValue?

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            switch (_mode)
+            {
+                case QueryStringTransformMode.Append:
+                    context.Query = context.Query.Add(_key, value?.ToString());
```
Tratcher, line 35: It shouldn't get added if it doesn't have a value.

test/ReverseProxy.Tests/Service/RuntimeModel/Transforms/QueryStringParameterTransformTests.cs
Tratcher: Escaping the "/" in the query isn't wrong, it's just unnecessary. We could live with it if we needed to. The query parser will un-escape it.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
jmezach: Fixed

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
jmezach: I agree. Do we want to log a warning though in that scenario? I think that would help with debug-ability for users, but might be wasteful in terms of performance. What do you think?

test/ReverseProxy.Tests/Service/RuntimeModel/Transforms/QueryStringParameterTransformTests.cs
jmezach: Okay, I've changed the test to expect the %2F for now then.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
jmezach: I've implemented the Set mode now as well.

jmezach:
For some reason the build fails to compile, but it compiles just fine locally. Not sure what's going on there. Any ideas?

Kahbazi:
You need to rebase from master.

src/ReverseProxy/Service/Config/TransformBuilder.cs
```diff
@@ -284,6 +293,22 @@ public Transforms Build(IList<IDictionary<string, string>> rawTransforms)
...
+                        {
+                            requestTransforms.Add(new QueryStringParameterTransform(QueryStringTransformMode.Append, queryStringParameter, routeValueKeyAppend));
+                        }
+                        else if (rawTransform.TryGetValue("Set", out var routeValueKeySet))
+                        {
+                            requestTransforms.Add(new QueryStringParameterTransform(QueryStringTransformMode.Set, queryStringParameter, routeValueKeySet));
+                        }
+                        else
+                        {
+                            throw new NotSupportedException(string.Join(",", rawTransform.Keys));
```
Kahbazi, line 309:
```suggestion
                            throw new NotSupportedException(string.Join(';', rawTransform.Keys));
```

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
```diff
@@ -0,0 +1,68 @@
...
+        private static QueryString SetQueryStringParameter(QueryString input, string key, object value)
+        {
+            var queryStringParameters = QueryHelpers.ParseQuery(input.Value);
+            queryStringParameters[key] = value.ToString();
+
+#if NETCOREAPP3_1
+            var queryBuilder = new QueryBuilder(queryStringParameters.Select(pair => new System.Collections.Generic.KeyValuePair<string, string>(pair.Key, pair.Value.ToString())));
+#else
+            var queryBuilder = new QueryBuilder(queryStringParameters);
+#endif
```
Kahbazi, line 58:
```suggestion
#if NETCOREAPP3_1
            var queryBuilder = new QueryBuilder(queryStringParameters.Select(pair => new System.Collections.Generic.KeyValuePair<string, string>(pair.Key, pair.Value.ToString())));
#elif NETCOREAPP5_0 
            var queryBuilder = new QueryBuilder(queryStringParameters);
#else
#error A target framework was added to the project and needs to be added to this condition.
#endif
```

To be consistance, like here
https://github.com/microsoft/reverse-proxy/blob/60afde81b2c1c5180fd6ab30e1f6a44c57071b2a/src/ReverseProxy/Utilities/ProtocolHelper.cs#L25-L31

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.WebUtilities;
+
+namespace Microsoft.ReverseProxy.Service.RuntimeModel.Transforms
+{
+    internal class QueryStringParameterTransform : RequestParametersTransform
```
Kahbazi, line 12: Rename to `QueryStringFromRouteTransform`?

`QueryStringParameterTransform` could be used for removing, and adding a constant value.

jmezach: That makes sense. Do we want to reflect this into the configuration as well? It looks like this currently:

```json
{
  "QueryStringParameter": "<parameter-key>",
  // Either
  "Append": "<route-value>"
  // Or
  "Set": "<route-value>"
}
```

Perhaps we need an `AppendFromRoute` and `SetFromRoute` or something, instead of just `Append` and `Set`? Or do we want to introduce an additional key to indicate the source (like `"Source": "route"`)?

jmezach:
I've added the `Remove` option now and I've also added query string parameter transforms to the docs. I believe I've got what I need right now, although there are some open questions. Please let me know if there are any additional changes you'd like.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringRemoveTransform.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public override void Apply(RequestParametersTransformContext context)
+        {
+            if (context == null)
+            {
+                throw new System.ArgumentNullException(nameof(context));
+            }
+
+            var queryStringParameters = QueryHelpers.ParseQuery(context.Query.Value);
```
Tratcher, line 28: Logically this make sense, but the perf is going to be bad. We'll want to think about a more efficient way to delete a single parameter.

docs/docfx/articles/transforms.md
```diff
@@ -175,6 +179,34 @@ Example:
...
+| Key | Value | Required |
+| QueryStringParameter | Name of a query string parameter | yes |
+| Set/Append | Key of route value | yes |
+| Remove | No value | yes |
+
+Config:
+```JSON
+{
+  "QueryStringParameter": "foo",
+  "Append": "remainder"
```
Tratcher, line 192: Naming: If I was just reading this config I'd expect the result to be `?foo=remainder`. The naming doesn't indicate that `remainder` is a route parameter.

Alternative:
Make `Append/Set` add a static value and add a new entry `Append/SetRouteValue` for the current functionality. These could be implemented by the same or a different class internally.

docs/docfx/articles/transforms.md
```diff
@@ -175,6 +179,34 @@ Example:
 | PathPattern | `/my/{plugin}/api/{remainder}` |
 | Result | `/my/v1/api/more/stuff` |
 
+#### QueryStringParameter
+| Key | Value | Required |
+| QueryStringParameter | Name of a query string parameter | yes |
```
Tratcher, line 184:
```suggestion
| QueryParameter | Name of a query string parameter | yes |
```

jmezach: Renamed this and made that consistent in the class names as well.

docs/docfx/articles/transforms.md
```diff
@@ -175,6 +179,34 @@ Example:
...
+| Set/Append | Key of route value | yes |
+| Remove | No value | yes |
```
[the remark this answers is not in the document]
jmezach: I've refactored this to a separate transform and renamed it to `RemoveQueryParameter`.

docs/docfx/articles/transforms.md
jmezach: I've done some refactoring and introduced an abstract base class called `QueryParameterTransform` that contains the logic to Append/Set query parameters and then added the `QueryParameterFromRouteTransform` and `QueryParameterFromStaticTransform` derived classes. I can imagine having a `QueryParameterFromHeaderTransform` in the future as well.

Not sure about the configuration mechanism though. Currently I have `Append`/`Set` and `AppendRouteValue`/`SetRouteValue`, but that can get a bit unwieldy if we add the option to source from headers. Perhaps we need to have a `Source` configuration parameter as well?

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringRemoveTransform.cs
jmezach: I agree, but do you have any suggestions on I'd might go about it? The `QueryString` type from ASP.NET Core has an `Add` operation, but no `Delete`. I guess we could add it to ASP.NET Core, but I'm not sure if that is the right direction.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringParameterTransform.cs
Tratcher: I don't know. Not many route parameters are optional, it will either always or never work. Also, don't the route parameters themselves already get logged?

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringRemoveTransform.cs
Tratcher: If we think this is a common thing people need to do then yes, adding it to AspNetCore would be a good option. You can at least file a followup issue in AspNetCore and link back here as an example.

docs/docfx/articles/transforms.md
Tratcher: Or split them into separate configs entirely. QueryRouteParamter, QueryHeaderParamter, QueryValueParamter, etc.. Compare to some of the Path transforms.

jmezach: I've refactored this to separate transforms: `QueryValueParameter` and `QueryRouteParameter` now.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryStringRemoveTransform.cs
jmezach: I've posted dotnet/aspnetcore#23971 as a feature suggestion.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryParameterTransform.cs
```diff
@@ -0,0 +1,70 @@
...
+                        break;
+                    default:
+                        throw new NotImplementedException(_mode.ToString());
+                }
+            }
+        }
+
+        protected abstract string GetValue(RequestParametersTransformContext context);
+
+        private static QueryString SetQueryParameter(QueryString input, string key, object value)
```
davidfowl, line 49: Yea, this is super ineffiicent... 😄

halter73: Have you seen dotnet/aspnetcore#23971? @jmezach filed that issue for the efficient removal of QueryString parameters and moved it into this PR. We just backlogged it yesterday.

Even the QueryString.Add() methods are terribly inefficient when used repeatedly. It allocates a new StringBuilder, String and QueryString for every call. This cannot be easily fixed without drastically changing the API.

We should just copy an internal version of QueryString (or better yet, start from scratch) and reference IHttpRequestFeature.QueryString directly (which is a string not a QueryString) to avoid the implicit QueryString allocation. Then all Append/Set/Remove/Whatever operations should operate on a single mutable data structure (maybe StringBuilder) to avoid unnecessary allocations until we create the final QueryString.

If what we design internally is nice enough, maybe we can contribute it back to ASP.NET Core.

Tratcher: QueryString is a struct, it's not allocated or boxed in the common case.

The biggest trick with batching would be that the transforms operate independently, how would you know when transforms are done running?

Tratcher: Unless RequestParametersTransformContext used a custom query structure.
https://github.com/microsoft/reverse-proxy/blob/ab34da2c3f91cff0cb0b61172d0c9d510c21692e/src/ReverseProxy/Service/Proxy/HttpProxy.cs#L353-L364

jmezach:
@Tratcher No problem, I was away on holiday anyway ;). I made the suggested changes. Only thing remaining now I guess is how we're going to fix the performance for the transformation. I saw the suggestion of using a custom query string structure on the `RequestParametersTransformContext`, but I'm not entirely sure how I would go about it. Perhaps we could parse the query string on the request once and put the resulting dictionary on the context, then allow the transforms to manipulate that dictionary and then put it back into a query string before being sent downstream. Does that make sense?

Tratcher:
Yeah, that's similar to how we deal with it on the HttpRequest as well. See the QueryString and Query properties. I'd also want an optimization that reused the original QueryString if nothing changed.

https://github.com/dotnet/aspnetcore/blob/master/src/Http/Http/src/Features/QueryFeature.cs

https://github.com/dotnet/aspnetcore/blob/master/src/Http/Http/src/Internal/DefaultHttpRequest.cs

I wonder if routing always triggers the lazy HttpRequest.Query parse and if we can take advantage of that.

jmezach:
Okay, I've made some changes that should avoid parsing the query string unnecessarily and only once per request. I'm not too sure about the design and naming of things though, so I figured I'd push what I have right now to see if I'm going in the right direction. Also I still have some failing tests because now if we "append" a query parameter that already exists in the original request it comes out as `&z=1,6` instead of `&z=1&z=6`. I couldn't find an easy way to fix that with the current design and I'm also not sure if it is actually a problem. Feedback appreciated ;).

jmezach:
Just checking in here to see if there's still interest in this. I still think it is a nice and useful feature to have.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryParameterTransform.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public override void Apply(RequestParametersTransformContext context)
+        {
+            if (context == null)
+            {
+                throw new System.ArgumentNullException(nameof(context));
+            }
+
+            string value = GetValue(context);
+            if (!string.IsNullOrWhiteSpace(value))
```
Tratcher, line 27: The WhiteSpace checks have some odd language issues, avoid them.
```suggestion
            if (!string.IsNullOrEmpty(value))
```

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryTransformContext.cs
```diff
@@ -0,0 +1,85 @@
...
+                {
+#if NETCOREAPP3_1
+                    var queryBuilder = new QueryBuilder(_modifiedQueryParameters.Select(pair => new System.Collections.Generic.KeyValuePair<string, string>(pair.Key, pair.Value.ToString())));
+#elif NETCOREAPP5_0
+                    var queryBuilder = new QueryBuilder(_modifiedQueryParameters);
+#else
+#error A target framework was added to the project and needs to be added to this condition.
+#endif
+                    return queryBuilder.ToQueryString();
+                }
```
Tratcher, line 48:
```suggestion
#if NETCOREAPP3_1
                var queryBuilder = new QueryBuilder(_modifiedQueryParameters.Select(pair => new System.Collections.Generic.KeyValuePair<string, string>(pair.Key, pair.Value.ToString())));
#elif NETCOREAPP5_0
                var queryBuilder = new QueryBuilder(_modifiedQueryParameters);
#else
#error A target framework was added to the project and needs to be added to this condition.
#endif
                return queryBuilder.ToQueryString();
```

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryTransformContext.cs
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+
+        public void AppendQueryParameter(string name, StringValues value)
```
Tratcher, line 52: Rather than implementing these actions on the context (it doesn't scale), make the _modifiedQueryParameters collection directly accessible as a lazy property that the individual transforms can mutate.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryTransformContext.cs
```diff
@@ -0,0 +1,85 @@
...
+        private Dictionary<string, StringValues> _modifiedQueryParameters;
+
+        public QueryTransformContext(HttpRequest request)
+        {
+            _request = request ?? throw new ArgumentNullException(nameof(request));
+            _originalQueryString = request.QueryString;
+            _modifiedQueryParameters = null;
+        }
+
+        public QueryString QueryString
```
Tratcher, line 30: Consider how to make the entire query replacable / removable. Some possible inspiration:
https://github.com/dotnet/aspnetcore/blob/d14b62a4c38bec700125ea1eb0249d3fd3b7ba3d/src/Http/Http/src/Features/QueryFeature.cs#L42-L90

jmezach:
I'm now exposing a `ModifiedQueryParameters` collection on the `QueryTransformContext` which is lazily evaluated which the transforms can use. This does solve my problem with the transforms needing to now whether a query parameter has already been added.

For some reason the tests are failing during CI although they are green on my machine. The 2 failing tests both have the scenario that we append a query parameter that already exists on the original request. Locally I get the value appended to the existing parameter (ie. `?z=1,6`), but in CI it is appended as a new query parameter (ie. `?z=1&z=6`). Do you know if something changed downstream that causes this?

I still have to look into re-using the parsed collection on the HttpRequest. I looked at the code you've linked below but I'm not entirely sure how I can use it.

jmezach:
I just ran the tests using `build.sh` and `test.sh` locally and then they are failing locally as well. Interestingly changing the test to assert on `?z=1&z=6` makes them work on .NET 5.0 (preview 6 seems to be what's downloaded by build.sh currently), but not on .NET Core 3.1. Running the tests using my machine installed .NET SDK (5.0 preview 8) works if I assert on `?z=1,6`. This feels like a regression, but I'm not entirely sure what's causing it.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryTransformContext.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                if (_modifiedQueryParameters == null)
+                {
+                    return _originalQueryString;
+                }
+
+#if NETCOREAPP3_1
+                var queryBuilder = new QueryBuilder(_modifiedQueryParameters.Select(pair => new System.Collections.Generic.KeyValuePair<string, string>(pair.Key, pair.Value.ToString())));
+#elif NETCOREAPP5_0
+                var queryBuilder = new QueryBuilder(_modifiedQueryParameters);
```
Tratcher, line 42: I think there's an SDK bug. Try swapping the order, and switch to the new TFM:
```suggestion
#if NET5_0
                var queryBuilder = new QueryBuilder(_modifiedQueryParameters);
#elif NETCOREAPP3_1
                var queryBuilder = new QueryBuilder(_modifiedQueryParameters.Select(pair => new System.Collections.Generic.KeyValuePair<string, string>(pair.Key, pair.Value.ToString())));
```

Tratcher (commented):
> I just ran the tests using `build.sh` and `test.sh` locally and then they are failing locally as well. Interestingly changing the test to assert on `?z=1&z=6` makes them work on .NET 5.0 (preview 6 seems to be what's downloaded by build.sh currently), but not on .NET Core 3.1. Running the tests using my machine installed .NET SDK (5.0 preview 8) works if I assert on `?z=1,6`. This feels like a regression, but I'm not entirely sure what's causing it.

I rebased your changes onto https://github.com/microsoft/reverse-proxy/pull/382 and the tests stopped failing. But then I dug deeper...

`?z=1&z=6` is the result I'd expect from append, not `?z=1,6`. I see what's causing the latter in 3.1, I've comment on that directly in the code. The 5.0 scenario is much stranger though. I think I found a 5.0 SDK bug where the wrong `#if` block gets picked depending on ordering, meaning you were always using the NETCOREAP3_1 code with the bug I mentioned earlier. I've commented with a workaround.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryTransformContext.cs
Tratcher: https://github.com/dotnet/runtime/issues/41839

jmezach:
I made the suggested changes, but now it seems that things are no longer compiling. I'm guessing that's because of the SDK bug. Do we have to wait for #382 to land to make this work?

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryTransformContext.cs
```diff
@@ -36,18 +36,18 @@ public QueryString QueryString
                     return _originalQueryString;
                 }
 
-#if NETCOREAPP3_1
-                var queryBuilder = new QueryBuilder(_modifiedQueryParameters.Select(pair => new System.Collections.Generic.KeyValuePair<string, string>(pair.Key, pair.Value.ToString())));
-#elif NETCOREAPP5_0
+#if NET5_0
```
Tratcher, line 39: You probably need to use both defines until we update the SDK.
```suggestion
#if NET5_0 || NETCOREAPP5_0
```

jmezach: Yup, that makes sense. And it seems to fix the CI build.

jmezach:
So I think we now have the matter of re-using the parsed dictionary on the HttpContext if it is available remaining. I looked through the code you linked and I think I know how to get to the `IQueryFeature` (using `_request.HttpContext.Features.Get<IQueryFeature>();`) but I'm unsure how I would go about checking if it has already parsed the query string. Or would `Get<T>()` return null if the query string has not yet been parsed?

Tratcher:
FYI: 382 got merged so you can rebase on it.

As for the lazy query parsing, it occurs to me that it's already built into HttpRequest.Query. You just need to make a copy of that collection on-demand. If it hasn't been parsed yet then HttpRequest will take care of that.

jmezach:
I did the rebase and I've implemented the copying of the request's `Query` property. I can't use the `IQueryCollection` directly because it doesn't support a `Remove()` operation. So instead I'm copying the individual keys and values over. Not sure if there is a more efficient way to do this.

jmezach:
Never mind, I wasn't thinking straight. Of course I could just new up the dictionary and use the original `IQueryCollection` as the source. It probably does the same thing under the hood, but I guess that code path can be optimised separately.

jmezach:
@Tratcher So I guess we're about ready here, or do you want to give it another review?

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryParameterTransform.cs
```diff
@@ -0,0 +1,60 @@
...
+                    default:
+                        throw new NotImplementedException(_mode.ToString());
+                }
+            }
+        }
+
+        protected abstract string GetValue(RequestParametersTransformContext context);
+    }
+
+    public enum QueryStringTransformMode
```
Tratcher, line 55: Why is the enum public? It looks like all of the consumers are internal.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryParameterTransform.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            var value = GetValue(context);
+            if (!string.IsNullOrEmpty(value))
+            {
+                switch (_mode)
+                {
+                    case QueryStringTransformMode.Append:
+                        if (!context.Query.Query.TryGetValue(_key, out var currentValue))
+                        {
+                            context.Query.Query.Add(_key, value.ToString());
```
Tratcher, line 35: lol Query.Query...

Maybe Query.Entries? Query.Collection?

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryParameterTransform.cs
```diff
@@ -0,0 +1,60 @@
...
+                        }
+                        else
+                        {
+                            var newValue = StringValues.Concat(currentValue, value.ToString());
+                            context.Query.Query[_key] = newValue;
+                        }
```
Tratcher, line 41:
```suggestion
                        StringValue newValue = value;
                        if (context.Query.Query.TryGetValue(_key, out var currentValue))
                        {
                             newValue = StringValues.Concat(currentValue, value);
                        }
                        context.Query.Query[_key] = newValue;
```

Tratcher (approved):
That should about do it.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryParameterTransform.cs
jmezach: Yup, you're right. Made it internal.

src/ReverseProxy/Service/RuntimeModel/Transforms/QueryParameterTransform.cs
jmezach: Yeah, wasn't very happy about that either. I've changed it to Query.Collection now.

jmezach:
@Tratcher I've processed the comments.

Tratcher:
Thanks for sticking with it!

jmezach:
@Tratcher No problem. I learned a lot while working on this, so the pleasure was all mine. Glad to see this got merged though :). Looking forward to trying out the next preview.

Issue #308, opened 2020-07-07
Remove IConfigErrorReporter


Tratcher:
Design review: Remove IConfigErrorReporter. This was originally intended as an abstraction to aggregate errors, but it was only ever implemented to write to logging. The route validator and other components that use IConfigErrorReporter today can write directly to logging instead.

Kahbazi:
I will take this one.

PR #311, opened 2020-07-07
Remove IConfigErrorReporter

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/308


Kahbazi:
Some of the tests were dependent on `IConfigErrorReporter` and now the errors are not checked in them. Is that important?
Should I cover that with implementing a `TestLogger`?

Tratcher:
Yes, please use the test logger when the test is checking for specific errors.

Tratcher:
Thanks

Issue #317, opened 2020-07-09
Add Helper Methods for adding Transform at Runtime


Kahbazi:
### What should we add or change to make your life better?

Add helper methods for each `Transform`.
```csharp
public static class TransformHelper
{
    public static void AddPathPrefixTransform(this ProxyRoute proxyRoute, PathString prefix)
    {
        proxyRoute.Transforms.Add(new Dictionary<string, string>
        {
            ["PathPrefix"] = prefix,
        });
    }

    public static void AddPathRemovePrefixTransform(this ProxyRoute proxyRoute, PathString prefix)
    {
        proxyRoute.Transforms.Add(new Dictionary<string, string>
        {
            ["PathRemovePrefix"] = prefix,
        });
    }
}
```

### Why is this important to you?
This methods can be used in `IProxyConfigFilter`.

Tratcher:
This is right in line with https://github.com/microsoft/reverse-proxy/issues/60, though a big part of that one is to let you define new custom transforms.

samsp-msft:
If you want to contribute this - lets schedule an ad-hoc design meeting before working on the PR. Ping me to get that setup.

Kahbazi:
@samsp-msft yeah, I would like to contribute this.

Kahbazi:
For adding Transforms with parameters, we could either have two separate methods for each or use an Enum in input.
**1. Multiple methods** 
```csharp
public static void AddSetRequestHeaderTransform(this ProxyRoute proxyRoute, string headerName, string value)
{
    proxyRoute.Transforms.Add(new Dictionary<string, string>
    {
        ["RequestHeader"] = headerName,
        ["Set"] = value
    });
}

public static void AddAppendRequestHeaderTransform(this ProxyRoute proxyRoute, string headerName, string value)
{
    proxyRoute.Transforms.Add(new Dictionary<string, string>
    {
        ["RequestHeader"] = headerName,
        ["Append"] = value
    });
}
```

**2. Multiple parameters** 
```csharp
public static void AddSetRequestHeaderTransform(this ProxyRoute proxyRoute, string headerName, string value, TransformHeaderType transformHeaderType)
{
    var type = transformHeaderType switch
    {
        TransformHeaderType.Set => "Set",
        TransformHeaderType.Append => "Append",
        _ => throw new System.ArgumentException(nameof(transformHeaderType));
    };

    proxyRoute.Transforms.Add(new Dictionary<string, string>
    {
[...]

PR #324, opened 2020-07-13
Add Helper methods for Transforms

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/317

### DIFF
--- a/src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
+++ b/src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
@@ -6,47 +6,39 @@ namespace Microsoft.ReverseProxy.Service
 {
     public static class ProxyRouteTransformExtensions
     {
-        public static void AddPathSetTransform(this ProxyRoute proxyRoute, PathString path)
+        public static void AddTransformPathSet(this ProxyRoute proxyRoute, PathString path)
         {
             proxyRoute.Transforms.Add(new Dictionary<string, string>
             {
-                ["PathSet"] = path,
+                ["PathSet"] = path.Value,
             });
         }
 
-        public static void AddPathPrefixTransform(this ProxyRoute proxyRoute, PathString prefix)
+        public static void AddTransformPathPrefix(this ProxyRoute proxyRoute, PathString prefix)
         {
             proxyRoute.Transforms.Add(new Dictionary<string, string>
             {
-                ["PathPrefix"] = prefix,
+                ["PathPrefix"] = prefix.Value,
             });
         }
 
-        public static void AddPathRemovePrefixTransform(this ProxyRoute proxyRoute, PathString prefix)
+        public static void AddTransformPathRemovePrefix(this ProxyRoute proxyRoute, PathString prefix)
         {
             proxyRoute.Transforms.Add(new Dictionary<string, string>
             {
-                ["PathRemovePrefix"] = prefix,
+                ["PathRemovePrefix"] = prefix.Value,
             });
         }
 
-        public static void AddPathRouteValuesTransform(this ProxyRoute proxyRoute, string pattern)
+        public static void AddTransformPathRouteValues(this ProxyRoute proxyRoute, PathString pattern)
         {
             proxyRoute.Transforms.Add(new Dictionary<string, string>
             {
-                ["PathPattern"] = pattern,
+                ["PathPattern"] = pattern.Value,
             });
         }
 
-        public static void CopyRequestHeaders(this ProxyRoute proxyRoute)
-        {
-            proxyRoute.Transforms.Add(new Dictionary<string, string>
-            {
-                ["RequestHeadersCopy"] = "True",
-            });
-        }
-
-        public static void SuppressRequestHeaders(this ProxyRoute proxyRoute)
+        public static void AddTransformSuppressRequestHeaders(this ProxyRoute proxyRoute)
         {
             proxyRoute.Transforms.Add(new Dictionary<string, string>
             {
@@ -54,7 +46,7 @@ namespace Microsoft.ReverseProxy.Service
             });
         }
 
-        public static void AddRequestHeaderOriginalHost(this ProxyRoute proxyRoute)
+        public static void AddTransformUseOriginalHostHeader(this ProxyRoute proxyRoute)
         {
             proxyRoute.Transforms.Add(new Dictionary<string, string>
             {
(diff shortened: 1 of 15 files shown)


src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+
+            proxyRoute.Transforms.Add(new Dictionary<string, string>
+            {
+                ["X-Forwarded"] = string.Join(',', headers),
+                ["Append"] = append.ToString(),
+                ["Prefix"] = headerPrefix
+            });
+        }
+
+        public static void AddForwardedTransform(this ProxyRoute proxyRoute, bool useFor = true, bool useHost = true, bool useProto = true, bool useBy = true, bool append = true, string forFormat="None", string byFormat = "None")
```
Kahbazi, line 147: Should we expose the NodeFormat? It's internal now.

src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        public static void AddPathRemovePrefixTransform(this ProxyRoute proxyRoute, PathString prefix)
+        {
+            proxyRoute.Transforms.Add(new Dictionary<string, string>
+            {
+                ["PathRemovePrefix"] = prefix,
+            });
+        }
+
+        public static void AddPathRouteValuesTransform(this ProxyRoute proxyRoute, string pattern)
```
Tratcher, line 33: Try changing the naming scheme to `AddTransform...` to see if it's clearer and more consistent.
AddTransformPathRouteValues
AddTransformPathPrefix
AddTransformSuppressRequestHeaders

src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
Tratcher: Not yet. The default should be `Random`

src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+
+namespace Microsoft.ReverseProxy.Service
+{
+    public static class ProxyRouteTransformExtensions
+    {
+        public static void AddPathSetTransform(this ProxyRoute proxyRoute, PathString path)
+        {
+            proxyRoute.Transforms.Add(new Dictionary<string, string>
+            {
+                ["PathSet"] = path,
```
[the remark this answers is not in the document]
Kahbazi: Should I just then pass a string for input?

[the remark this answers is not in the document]
Tratcher: No, the PathString is good because it specifies what the encoding state is supposed to be. We just need to make sure that state flows correctly.

src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        public static void AddPathRouteValuesTransform(this ProxyRoute proxyRoute, string pattern)
+        {
+            proxyRoute.Transforms.Add(new Dictionary<string, string>
+            {
+                ["PathPattern"] = pattern,
+            });
+        }
+
+        public static void CopyRequestHeaders(this ProxyRoute proxyRoute)
```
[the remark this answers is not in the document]
Kahbazi: Maybe this could be useful when there are multiple filters with different logic, or when there's a global config for transform and a client suppress request headers for all routes but want to send request headers for one route.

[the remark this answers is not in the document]
Kahbazi: I removed the methods for now.

test/ReverseProxy.Tests/Service/Config/ProxyRouteTransformExtensionsTests.cs
```diff
@@ -0,0 +1,369 @@
...
+            var proxyRoute = CreateProxyRoute();
+
+            proxyRoute.AddTransformPathSet(new PathString("/path#"));
+
+            var transform = BuildTransform(proxyRoute);
+
+            var requestTransform = Assert.Single(transform.RequestTransforms);
+            var pathStringTransform = Assert.IsType<PathStringTransform>(requestTransform);
+            Assert.Equal(PathStringTransform.PathTransformMode.Set, pathStringTransform.Mode);
+            Assert.Equal("/path%23", pathStringTransform.Value);
```
Tratcher, line 31: This shouldn't have changed formats.
```suggestion
            Assert.Equal("/path#", pathStringTransform.Value.Value);
```

Same for the rest of these tests.

test/ReverseProxy.Tests/Service/Config/ProxyRouteTransformExtensionsTests.cs
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        public static IEnumerable<object[]> AllForwardedCombinations => GenerateForwardedCombinations();
+
+        private static IEnumerable<object[]> GenerateForwardedCombinations()
+        {
+            var boolCollection = new object[] { true, false };
+            var formatCollection = Enum.GetNames(typeof(RequestHeaderForwardedTransform.NodeFormat));
+
+            for (var x0 = 0; x0 < boolCollection.Length; x0++)
```
Tratcher, line 219: This isn't necessary. We only need to test a few representative combinations.
- All enabled
- All disabled
- Half enabled
- One with different format values for each IP.

test/ReverseProxy.Tests/Service/Config/ProxyRouteTransformExtensionsTests.cs
```diff
@@ -0,0 +1,369 @@
...
+            }
+        }
+
+        public static IEnumerable<object[]> AllXForwardedCombinations => GenerateXForwarderCombinations();
+
+        private static IEnumerable<object[]> GenerateXForwarderCombinations()
+        {
+            var boolCollection = new object[] { true, false };
+
+            for (var x0 = 0; x0 < boolCollection.Length; x0++)
```
Tratcher, line 313: Same

Tratcher:
Thanks

Issue #273, opened 2020-06-29
Support for websockets - Add tests


TropicalThunder:
### What should we add or change to make your life better?

Support for proxying websocket requests to the backend and/or proxy websockets to another backend other than where the http requests go.

### Why is this important to you?

It would broaden the scope of use cases for the application.

Tratcher:
WebSockets are already supported by default, have you tried it?

If you want them to go to a separate destination then you would assign them their own route.

samsp-msft:
Triage: Closing as its already supported. Please re-open if you find issues.

JensMagnus:
I'm getting

`core.b11c5824.js:1 WebSocket connection to 'wss://example.com/api/websocket' failed: Error during WebSocket handshake: 'Connection' header is missing`

Is it a YARP related issue?

Tratcher:
That's not enough information. A decrypted network trace or Trace level server logs with Kestrel's Connection logging enabled would help debug that.
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-3.1#connection-logging

samsp-msft:
Triage: Need to add a functional test for web sockets

PR #325, opened 2020-07-14
Add funtional test for web sockets

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/273


test/ReverseProxy.FunctionalTests/Common/TestUrlHelper.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Microsoft.ReverseProxy
+{
+    public static class TestUrlHelper
```
Kahbazi, line 3: Is there a nuget package that I could get these classes from? I couldn't find any.

test/ReverseProxy.FunctionalTests/Microsoft.ReverseProxy.FunctionalTests.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net5.0;netcoreapp3.1</TargetFrameworks>
+    <OutputType>Library</OutputType>
+    <RootNamespace>Microsoft.ReverseProxy</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Content Include="validSelfSignedClientEkuCertificate.cer">
```
halter73, line 10: Is this referenced anywhere?

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        private IHost CreateHost(Action<IServiceCollection> configureServices, Action<IApplicationBuilder> configureApp)
+        {
+            return new HostBuilder()
+               .ConfigureWebHost(webHostBuilder =>
+               {
+                   webHostBuilder
+                   .ConfigureServices(configureServices)
+                   .UseKestrel()
+                   .UseUrls(TestUrlHelper.GetTestUrl(ServerType.Kestrel))
```
halter73, line 234: I wouldn't bother with all the ServerTypes until we support testing with other servers.

test/ReverseProxy.FunctionalTests/Common/TestUrlHelper.cs
Tratcher: I don't think we publish these test helpers. Copying the few you need should be fine.

test/ReverseProxy.FunctionalTests/Common/TestPortHelper.cs
```diff
@@ -0,0 +1,56 @@
+using System.Net;
+using System.Net.Sockets;
+
+namespace Microsoft.ReverseProxy
+{
+    public static class TestPortHelper
```
Tratcher, line 6: Drop this class for now, we're not testing IIS.

test/ReverseProxy.FunctionalTests/Microsoft.ReverseProxy.FunctionalTests.csproj
```diff
@@ -0,0 +1,23 @@
...
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="coverlet.collector" Version="1.0.1" />
```
Tratcher, line 16: Is this used?

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,239 @@
...
+namespace Microsoft.ReverseProxy
+{
+    public class WebSocketTests
+    {
+        [Theory]
+        [InlineData(WebSocketMessageType.Binary)]
+        [InlineData(WebSocketMessageType.Text)]
+        public async Task WebSocketTest(WebSocketMessageType messageType)
+        {
+            var destinationHost = CreateWebSocketHost();
```
Tratcher, line 28: Stop and Dispose all the hosts, clients, sockets, streams, etc. by the end of the test.
```suggestion
            using var destinationHost = CreateWebSocketHost();
```

test/ReverseProxy.FunctionalTests/Common/Helpers.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.AspNetCore.Hosting.Server.Features;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using System.Linq;
+
+namespace Microsoft.ReverseProxy
+{
+    public static class Helpers
+    {
+        public static string GetAddress(IHost server)
```
Tratcher, line 11:
```suggestion
        public static string GetAddress(this IHost server)
```

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,239 @@
...
+            await destinationHost.StartAsync();
+            var destinationHostUrl = Helpers.GetAddress(destinationHost);
+
+            var proxyHost = CreateReverseProxyHost(destinationHostUrl);
+            await proxyHost.StartAsync();
+            var proxyHostUrl = Helpers.GetAddress(proxyHost);
+
+            var client = new ClientWebSocket();
```
Tratcher, line 36:
```suggestion
            using var client = new ClientWebSocket();
```

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,239 @@
...
+{
+    public class WebSocketTests
+    {
+        [Theory]
+        [InlineData(WebSocketMessageType.Binary)]
+        [InlineData(WebSocketMessageType.Text)]
+        public async Task WebSocketTest(WebSocketMessageType messageType)
+        {
+            var destinationHost = CreateWebSocketHost();
+            await destinationHost.StartAsync();
```
Tratcher, line 29: Create one 30 second cancellation token and pass it to every async operation in case something hangs. Dispose it at the end.

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,239 @@
...
+                endOfMessage: true,
+                CancellationToken.None);
+
+            var message = await client.ReceiveAsync(buffer, CancellationToken.None);
+
+            Assert.Equal(messageType, message.MessageType);
+            Assert.True(message.EndOfMessage);
+
+            var text = Encoding.UTF8.GetString(buffer.AsSpan(0, message.Count));
+            Assert.Equal(textToSend, text);
```
Tratcher, line 55: var result = await client.CloseAsync(...);
Assert the results.

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,239 @@
...
+                }
+
+                await using var stream = await upgradeFeature.UpgradeAsync();
+                var buffer = new byte[1];
+                int read;
+                while ((read = await stream.ReadAsync(buffer, httpContext.RequestAborted)) != 0)
+                {
+                    await stream.WriteAsync(buffer, 0, read);
+
+                    if (buffer[0] == 255)
```
Tratcher, line 217: You shouldn't need this check to break the loop, ReadAsync will eventually return 0 or throw.

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            var rawStream = await response.Content.ReadAsStreamAsync();
+
+            var buffer = new byte[1];
+            for (var i = 0; i <= 255; i++)
+            {
+                buffer[0] = (byte)i;
+                await rawStream.WriteAsync(buffer, 0, 1, CancellationToken.None);
+                var read = await rawStream.ReadAsync(buffer, CancellationToken.None);
+                if (i == 255)
```
Tratcher, line 95: You shouldn't need this special case.

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,239 @@
...
+            var targetUri = new Uri(new Uri(webSocketsTarget, UriKind.Absolute), "rawupgrade");
+            var request = new HttpRequestMessage(HttpMethod.Get, targetUri);
+            request.Headers.TryAddWithoutValidation("Connection", "upgrade");
+            request.Version = new Version(1, 1);
+
+            var response = await client.SendAsync(request, CancellationToken.None);
+
+            Assert.Equal(HttpStatusCode.SwitchingProtocols, response.StatusCode);
+
+            var rawStream = await response.Content.ReadAsStreamAsync();
```
Tratcher, line 87:
```suggestion
            using var rawStream = await response.Content.ReadAsStreamAsync();
```

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,241 @@
...
+
+            using var handler = new HttpClientHandler
+            {
+                AllowAutoRedirect = false,
+                AutomaticDecompression = DecompressionMethods.None,
+                UseCookies = false,
+                UseProxy = false
+            };
+            using var client = new HttpMessageInvoker(handler);
+            var webSocketsTarget = proxyHostUrl.Replace("https://", "wss://").Replace("http://", "ws://");
```
Tratcher, line 89: You don't need this line for a raw upgrade test, right?

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -0,0 +1,241 @@
...
+            for (var i = 0; i <= 255; i++)
+            {
+                buffer[0] = (byte)i;
+                await rawStream.WriteAsync(buffer, 0, 1, cts.Token);
+                var read = await rawStream.ReadAsync(buffer, cts.Token);
+
+                Assert.Equal(1, read);
+                Assert.Equal(i, buffer[0]);
+            }
+
```
Tratcher, line 117: The stream should get disposed before trying to stop the hosts, otherwise I think it would keep them from shutting down gracefully.
```suggestion

rawStream.Dispose();
```

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
Kahbazi: How about for the web socket? I copy this from client sample app.

Tratcher: Yes, it's required when using ClientWebSocket.

Kahbazi:
> How long do each of these tests take to run?

 less than 1 second each.

Tratcher:
Thanks

Issue #286, opened 2020-07-02
Custom routing endpoint metadata


Tratcher:
Forked from https://github.com/microsoft/reverse-proxy/issues/87#issuecomment-620896307

Endpoint.Metadata is a custom collection with generic search functionality (via downcasting).
https://github.com/dotnet/aspnetcore/blob/600d5ff4978c84b42aa15a11a0fdb0dbd1cc8692/src/Http/Http.Abstractions/src/Routing/EndpointMetadataCollection.cs#L75

We need some way to add custom endpoint metadata. ProxyRoute is too low level for that, it's a raw config representation, you're not going to be able to represent complex objects like IAuthorizeData there. We likely need an event in RuntimeRouteBuilder that lets you customize the routes as we build them. You'd take some config metadata and translate it to complex objects like IAuthorizeData.
https://github.com/microsoft/reverse-proxy/blob/ea7164c689ea06343b99a30aa7feecef4eda215d/src/ReverseProxy.Core/Service/DynamicEndpoints/RuntimeRouteBuilder.cs#L51-L66

Note we've added first class auth support in #122, and CORS is proposed in #281 

What other per-endpoint route metadata is there?

samsp-msft:
Triage: more flexible, but harder to use than just directly implementing Auth & CORS, but may be better in the long run.

Kahbazi:
I think `MapReverseProxy` should also return `IEndpointConventionBuilder`, so the client could add metadata for all routes.

Tratcher:
> I think `MapReverseProxy` should also return `IEndpointConventionBuilder`, so the client could add metadata for all routes.

What kind of metadata would you want to add for all routes? Hard coding authz or cors centrally?

Kahbazi:
> Hard coding authz or cors centrally?

Yes and also any other third-party middleware that might want to use endpoint metadata.
Almost every `Map` method on `IEndpointRouteBuilder` returns `IEndpointConventionBuilder`, and I think YARP should too.

Issue #277, opened 2020-06-30
Redesign IReverseProxyConfigManager


Tratcher:
Design review proposals:

IReverseProxyConfigManager.ApplyConfigurationsAsync as it currently exists is a push model. It also uses a side-car set of data stores IRouteRepo & IClusterRepo.

A) Remove IRouteRepo & IClusterRepo, they're not needed and could cause race conditions.

B) IReverseProxyConfigManager's push model gets replaced by a signal-and-pull model mimicking AspNet's endpoint routing data model. (Naming TBD)
```
interface IYarpConfigProvider
{
  IYarpConfig GetData();
  IChangeToken ChangeToken {get;}
}

interface IYarpConfig
{
  IReadOnlyList<RawRoute> Routes { get; }
  IReadOnlyList<RawCluster> Clusters { get; }
}
```

Startup sequence:
The new ReverseProxyConfigManager takes these providers (just one for now, we'll worry about multiple later) from DI.
- Call GetData() to retrieve a new data set. - If the provider needs to async load the first time it may either choose to block or it returns none and then fires the change token later.
  - If there are multiple providers, call them all and aggregate the raw data before continuing.
- Validate the data set. If any of the data is invalid then crash the app. We do not want to start with invalid config.
- Populate the internal object model (routes, clusters, etc.)
- It registers for change notifications.

On change:
- When notified, it calls GetData() to retrieve a new data set.
  - If there are multiple providers, call them all and aggregate the raw data before continuing.
- Validate the data set. If any of the data is invalid then do not apply the new change.
- Diff vs. the current object model. Add/Remove/Unchanged/Update as needed
  - Update config while preserving state
  - Preserve unchanged nodes because of state

ReverseProxyConfigManager is responsible for tracking change notifications and preventing overlapping updates.

Expose the validation logic publicly (IRouteValidator, etc.) so that config providers can pre-vet their data to either filter out bad routes or avoid signaling that changes are available. GetData should always return the immutable last known good dataset.

I expect we'll work on this as part of https://github.com/microsoft/reverse-proxy/issues/8.

PR #227, opened 2020-06-04
Fix casing of scripts in readme

danmosemsft:

Issue #179, opened 2020-05-20
Use freezable pattern for contract types instead of deep cloning


alnikola:
### What should we add or change to make your life better?
Currently, we are making deep clones of all `Abstractions.*.Contract` types' instances (e.g. `LoadBalancingOptions` or `QuotaOptions`) before adding them to repositories. On top of that, some of them even have duplicates in `RuntimeModel` namespace with exactly the same contract, but different names (e.g. `LoadBalancingOptions` and `BackendConfig.BackendLoadBalancingOptions`). This is done to ensure safety and consistency of concurrent configuration reads in requests handling as well as to avoid accidental modifications.

However, it leads to code duplication and extra allocations in runtime. Thus, it's proposed to get rid of deep cloning and make all `Abstractions.*.Contract` types freezable which means they will be mutable during configuration initialization or refresh, but will prohibit any state changes after a `Freeze()` method is called. Each top-most configuration type (e.g. `Backend` and `Route`) will freeze it's state on a `Freeze()` call and ensure that all nested instances do the same (i.e. `Backend.Freeze()` must also call` LoadBalancingOptions.Freeze()`). Further, `ProxyConfigLoader` will call `Freeze()` on all top-most instances before adding them into repositories.

### Why is this important to you?
It will eliminate code duplication, the reduce number of allocations and memory footprint.

Tratcher:
Prior art: AsReadOnly(), IsReadOnly, CopyAsReadOnly()
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.asreadonly?view=netcore-3.1
https://github.com/dotnet/aspnetcore/blob/0b90802e712a0dad12eebf2c7ad7abea9869ae54/src/Http/Headers/src/MediaTypeHeaderValue.cs#L421

Tratcher:
Another proposal is to use C# 9 record types so the config objects are always immutable after creation.
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#record-types

Known issues:
- Redesign IProxyConfigFilter (not used by anyone yet?).
- Some of the properties use mutable types like Cluster.Metadata is an IDictionary. These would need to become readonly collections. Would this complicate their creation?

Tratcher:
Note today we make some costly deep clones of many objects to avoid anyone mutating them while we're using them.

[...]

PR #328, opened 2020-07-15
Add support for adding metadata to proxy routes

Kahbazi:
Partially Fixes https://github.com/microsoft/reverse-proxy/issues/286

I changed the way endpoints are being built. I tried to mirror MVC, and now endpoints are being built in `ProxyDynamicEndpointDataSource` instead of `RuntimeRouteBuilder`. 

I think the main challenge is with `ProxyConfigLoader` which is a `IHostedService` and runs before `MapReverseProxy` has a chance to add metadata, so I add a `bool firstTime` parameter and pass in around , so I only update `RouteConfig` and not building the endpoints. 

I want feedback for design before I continue with unit tests.


Tratcher:
I hadn't realized this would be so complex a change. It's going to take me some time to go over this.

The firstTime parameter is concerning, it seems like a blunt way to address a race condition. The design changes proposed in https://github.com/microsoft/reverse-proxy/issues/277 should remove this race because they reverse the model from push to pull. I think we'll want to do #227 before continuing work on this PR.

Kahbazi:
I don't like the `firstTime` either, I just wanted to make it work.
Overall I didn't understand the need for the `IHostedService` for reloading the config. Can't it be done with just `IOptionsMonitor.OnChange(...)`?

Tratcher:
> I don't like the `firstTime` either, I just wanted to make it work.
> Overall I didn't understand the need for the `IHostedService` for reloading the config. Can't it be done with just `IOptionsMonitor.OnChange(...)`?

But where would you call OnChange?

Kahbazi:
> But where would you call OnChange?

Maybe in `LoadFromConfig()` but then somehow it should also run for the first time.

Tratcher:
> > But where would you call OnChange?
> 
> Maybe in `LoadFromConfig()` but then somehow it should also run for the first time.

That's the trick, LoadFromConfig doesn't have access to IOptionsMonitor, the DI container hasn't been built yet. Something has to resolve IOptionsMonitor from DI at startup in order to kick of the process. The IHostedService is one way of doing that, but #227 should be able to replace that with a direct config source resolved from DI.

Tratcher:
@Kahbazi can you please rebase this?

Kahbazi:
I did the rebase. I need some feedback and then I will fix and add tests.

Tratcher:
There's another merge conflict?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            _lock = @lock;
+            _conventions = conventions;
+        }
+
+        /// <summary>
+        /// Adds the specified convention to the builder. Conventions are used to customize <see cref="EndpointBuilder"/> instances.
+        /// </summary>
+        /// <param name="convention">The convention to add to the builder.</param>
+        public void Add(Action<EndpointBuilder> convention)
```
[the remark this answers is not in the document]
Tratcher: Should these return `this` for chaining?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            // The lock is shared with the data source. We want to lock here
+            // to avoid mutating this list while its read in the data source.
+            lock (_lock)
+            {
+                _conventions.Add(convention);
+            }
+        }
+
+        public void AddClusterId(string clusterId, Action<EndpointBuilder> convention)
```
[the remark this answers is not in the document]
Kahbazi: This could be used to add an authentication scheme to a cluster instead of adding it to each route.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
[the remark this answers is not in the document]
Kahbazi: > ConfigureRoutes?
> Should these return this for chaining?

This is the implementation of [`IEndpointConventionBuilder`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.iendpointconventionbuilder.add?view=aspnetcore-3.1#Microsoft_AspNetCore_Builder_IEndpointConventionBuilder_Add_System_Action_Microsoft_AspNetCore_Builder_EndpointBuilder__). Also there is a `WithMetadata()` extensions method which could be used to configure all routes,

> Should we have overloads for each of these that give you first class access to the route or cluster details?

That's a good idea, I think we should create an immutable version of `ProxyRoute` and `Cluster` to prevent user from changing them, because in this stage all the route and cluster configs are already created.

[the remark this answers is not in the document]
Kahbazi: Or we could use `DeepClone()`.

src/ReverseProxy/Configuration/DependencyInjection/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -32,6 +34,8 @@ public static IReverseProxyBuilder AddReverseProxy(this IServiceCollection servi
                 .AddProxy()
                 .AddBackgroundWorkers();
 
+            services.TryAddSingleton<ProxyEndpointFactory, ProxyEndpointFactory>();
```
WeihanLi, line 37: why not use `services.TryAddSingleton<ProxyEndpointFactory>();` directly

davidfowl:
I think we're going to need an explanation of the overall design that comes out of this change since it looks like it's evolved a bunch from the original purpose of the PR.

Kahbazi:
The purpose of this PR is to enable users to add their own Metadata to Endpoints that are created by YARP and it can be done by `ReverseProxyConventionBuilder` .
```diff
- void MapReverseProxy(...)
+ ReverseProxyConventionBuilder MapReverseProxy(...)
```
The challenge was that when calling `MapReverseProxy()` it will eventually call `IRuntimeRouteBuilder.Build()` and it both creates the `RouteConfig` **and** `Endpoint`. So there's no chance for the user convention to be applied to the `Endpoint`.
So I introcude `ProxyEndpointFactory` and move the responsibility of creating `Endpoint` from `IRuntimeRouteBuilder` to it, so the `Endpoint` could be created at a later point in time.
For the same reason I change the `ProxyConfigManager.Endpoints` getter from simply returning the `Endpoint` to Initializing the `Endpoint` first and then return them.

This is the API that user can use to add `Metadata`.
```csharp
public class ReverseProxyConventionBuilder
{
    // Configure all routes
    public void Add(Action<EndpointBuilder> convention); 
    // Configure all routes based on their Cluster
    public ReverseProxyConventionBuilder ConfigureClusters(Action<IEndpointConventionBuilder, Cluster> convention);
    // Configure all routes based on their Route
    public ReverseProxyConventionBuilder ConfigureRoutes(Action<IEndpointConventionBuilder, ProxyRoute> convention);
    // Configure all routes for a specific Cluster
    public ReverseProxyConventionBuilder ConfigureCluster(string clusterId, Action<IEndpointConventionBuilder> convention);
    // Configure a specific Route
    public ReverseProxyConventionBuilder ConfigureRoute(string routeId, Action<IEndpointConventionBuilder> convention);
}
```

davidfowl:
I like the design. Do those configure callbacks get run every time the configuration changes?

Kahbazi:
> I like the design. Do those configure callbacks get run every time the configuration changes?

If the configuration changes causes a change on routes.
https://github.com/microsoft/reverse-proxy/blob/dc2334b172896dc8f41777d3f3d364fe78d416ae/src/ReverseProxy/Service/Management/ProxyConfigManager.cs#L167-L171

https://github.com/microsoft/reverse-proxy/blob/dc2334b172896dc8f41777d3f3d364fe78d416ae/src/ReverseProxy/Service/Management/ProxyConfigManager.cs#L103-L113

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+
+        public ReverseProxyConventionBuilder ConfigureClusters(Action<IEndpointConventionBuilder, Cluster> convention)
+        {
+            void Action(EndpointBuilder endpointBuilder)
+            {
+                var routeConfig = endpointBuilder.Metadata.OfType<RouteConfig>().Single();
+
+                var config = _proxyConfigProvider.GetConfig();
```
Tratcher, line 50: We shouldn't be reloading config here, GetConfig can return different information every time. If we need something then it should be preserved on the RouteConfig or EndpointBuilder.Metadata.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,130 @@
...
+                    convention(conventionBuilder);
+                }
+            }
+
+            Add(Action);
+
+            return this;
+        }
+
+        public ReverseProxyConventionBuilder ConfigureRoute(string routeId, Action<IEndpointConventionBuilder> convention)
```
Tratcher, line 97: Should there also be overloads like `ConfigureRoute(string routeId, Action<IEndpointConventionBuilder, ProxyRoute> convention)`?

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -62,15 +64,53 @@ internal class ProxyConfigManager : EndpointDataSource, IProxyConfigManager, IDi
...
+
         // EndpointDataSource
 
         /// <inheritdoc/>
-        public override IReadOnlyList<Endpoint> Endpoints => Volatile.Read(ref _endpoints);
+        public override IReadOnlyList<Endpoint> Endpoints
+        {
+            get
+            {
+                Initialize();
```
Tratcher, line 84: This needs to be lazy the first time because MapReverseProxy forces the config to load before we have the _conventions, right? The proxy data model can be built, but the endpoints can't be built yet? And the only signal we have that someone is done providing conventions is when the route table starts getting built?

Include a comment for that so we don't remove this and break conventions.

src/ReverseProxy/Service/DynamicEndpoint/ProxyEndpointFactory.cs
```diff
@@ -0,0 +1,121 @@
...
+{
+    internal class ProxyEndpointFactory
+    {
+        private static readonly IAuthorizeData _defaultAuthorization = new AuthorizeAttribute();
+        private static readonly IEnableCorsAttribute _defaultCors = new EnableCorsAttribute();
+        private static readonly IDisableCorsAttribute _disableCors = new DisableCorsAttribute();
+
+        private RequestDelegate _pipeline;
+
+        public void AddEndpoint(List<Endpoint> endpoints,
```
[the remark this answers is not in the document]
Kahbazi: I tried to bring the same APIs as MVC have. `ActionEndpointFactory` uses the same API.
https://github.com/dotnet/aspnetcore/blob/15f341f8ee556fa0c2825cdddfe59a88b35a87e2/src/Mvc/Mvc.Core/src/Routing/ActionEndpointFactory.cs#L36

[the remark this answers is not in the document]
Tratcher: ActionEndpointFactory seems to use that pattern because it's capable of generating multiple endpoints. That's not the case here so let's simplify the API.

src/ReverseProxy/Configuration/ReverseProxyIEndpointRouteBuilderExtensions.cs
```diff
@@ -50,16 +53,27 @@ public static void MapReverseProxy(this IEndpointRouteBuilder endpoints, Action<
...
+            var proxyEndpointFactory = endpoints.ServiceProvider.GetRequiredService<ProxyEndpointFactory>();
+            proxyEndpointFactory.SetProxyPipeline(app);
 
-            var configManager = endpoints.ServiceProvider.GetRequiredService<IProxyConfigManager>();
+            return GetOrCreateDataSource(endpoints).DefaultBuilder;
+        }
+
+        private static ProxyConfigManager GetOrCreateDataSource(IEndpointRouteBuilder endpoints)
+        {
+            var dataSource = (ProxyConfigManager)endpoints.DataSources.OfType<IProxyConfigManager>().FirstOrDefault();
```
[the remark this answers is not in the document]
Kahbazi: I removed the `IProxyConfigManager`.
I added the check like in `MapControllers` and `MapRazorPages`, but I understand your concern. 
How about moving the the pipeline from `MapReverseProxy(IEndpointRouteBuilder)` to `AddReverseProxy(IServiceCollection)`?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
Kahbazi: I'm passing `RouteConfig.ProxyRoute` and `RouteConfig.ClusterInfo.Config.Cluster` but I'm using `Signal` for it. Is it ok?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
Kahbazi: Sounds good. I've added those too.

src/ReverseProxy/Service/Config/RuntimeRouteBuilder.cs
```diff
@@ -21,23 +21,13 @@ namespace Microsoft.ReverseProxy.Service
     /// </summary>
     internal class RuntimeRouteBuilder : IRuntimeRouteBuilder
```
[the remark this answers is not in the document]
Kahbazi: I like the idea of having a class with the responsibility of just creating the internal type (`RouteConfig`, `Transforms` ) from public types that user can provide.

src/ReverseProxy/Configuration/ReverseProxyIEndpointRouteBuilderExtensions.cs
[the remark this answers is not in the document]
Tratcher: No, the pipeline setup should stay in Configure/MapReverseProxy. The check is fine, we can decide later if it makes sense to call MapReverseProxy more than once.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            // The lock is shared with the data source. We want to lock here
+            // to avoid mutating this list while its read in the data source.
+            lock (_lock)
+            {
+                _conventions.Add(convention);
+            }
+        }
+
+        public ReverseProxyConventionBuilder ConfigureClusters(Action<IEndpointConventionBuilder, Cluster> convention)
```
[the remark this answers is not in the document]
Tratcher: These will need doc comments.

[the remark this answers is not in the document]
Tratcher: I realize that the naming here may be confusing. What this does is let you configure each endpoint based on its associated cluster data, you're not actually configuring the cluster.

I've been thinking about these wrong. Every convention has access to all of the data, so ConfigureRoutes and ConfigureClusters don't need to be separate, they could be progressive overloads:
```C#
// Configures the endpoint for all routes
Configure(Action<IEndpointConventionBuilder> convention);
Configure(Action<IEndpointConventionBuilder, ProxyRoute> convention);
Configure(Action<IEndpointConventionBuilder, ProxyRoute, Cluster> convention);
// Configures the endpoint for a specific route
ConfigureRoute(string routeId, Action<IEndpointConventionBuilder> convention);
ConfigureRoute(string routeId, Action<IEndpointConventionBuilder, ProxyRoute> convention);
ConfigureRoute(string routeId, Action<IEndpointConventionBuilder, ProxyRoute, Cluster> convention);
// Configures the endpoint for all routes that map to a specific cluster
ConfigureCluster(string clusterId, Action<IEndpointConventionBuilder> convention);
ConfigureCluster(string clusterId, Action<IEndpointConventionBuilder, ProxyRoute> convention);
ConfigureCluster(string clusterId, Action<IEndpointConventionBuilder, ProxyRoute, Cluster> convention);
```

Given that structure, we probably only need the first set of Configure overloads, not ConfigureRoute or ConfigureCluster since those are something you can already do with the base Configure overload and a simple if check on the name.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,164 @@
...
+        public void Add(Action<EndpointBuilder> convention)
+        {
+            if (convention == null)
+            {
+                throw new ArgumentNullException(nameof(convention));
+            }
+
+            // The lock is shared with the data source. We want to lock here
+            // to avoid mutating this list while its read in the data source.
+            lock (_lock)
```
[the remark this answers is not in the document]
Kahbazi: I tried to follow `ControllerActionEndpointConventionBuilder` and add a lock.
https://github.com/dotnet/aspnetcore/blob/6a99743d337d205bdf78333cfaac33db993c1034/src/Mvc/Mvc.Core/src/Builder/ControllerActionEndpointConventionBuilder.cs#L40-L43
Should I remove it?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,295 @@
...
+            return this;
+        }
+
+        /// <summary>
+        /// Configures the endpoint for a specific route
+        /// </summary>
+        /// <param name="routeId">The route id to aplly convention to its endpoint.</param>
+        /// <param name="convention">The convention to add to the builder.</param>
+        /// <returns></returns>
+        public ReverseProxyConventionBuilder ConfigureRoute(string routeId, Action<IEndpointConventionBuilder> convention)
```
Tratcher, line 118: The experience question we need to answer is how common you'll write code to configure a route or cluster by name vs inspecting all the endpoints and reacting to other fields like metadata. i.e. do we need all 9 of these overloads, or only the first three Configure ones? The cluster and route specific ones can be built on top of the first three Configure conventions if needed.

I'd suggest limiting it to just the Configure overloads for now. We can revisit as we see demand for common variations.

test/ReverseProxy.Tests/Service/Management/ProxyConfigManagerTests.cs
```diff
@@ -232,36 +232,36 @@ public async Task GetChangeToken_SignalsChange()
             Assert.NotNull(readEndpoints2);
         }
 
-        [Fact]
-        public async Task LoadAsync_RequestVersionValidationError_Throws()
-        {
-            const string TestAddress = "https://localhost:123/";
+        //[Fact]
+        //public async Task LoadAsync_RequestVersionValidationError_Throws()
```
Tratcher, line 236: ?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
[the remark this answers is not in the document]
JamesNK: YARP routes can be defined in config. Does YARP listen to config changes and repopulate routes when config changes?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,295 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.ReverseProxy.Abstractions;
+using Microsoft.ReverseProxy.RuntimeModel;
+using Microsoft.ReverseProxy.Service;
+
+namespace Microsoft.ReverseProxy
+{
+    public class ReverseProxyConventionBuilder : IEndpointConventionBuilder
```
JamesNK, line 11: Am I right that all YARP routes are defined in config? (i.e. a JSON file) And then the various `Configure` methods on this builder are to apply metadata to a subset of those routes?

It does feel a little like the line is being blurred between JSON configuration and code configuration. (I don't know whether that is desirable or bad yet). Are there scenarios where people have needed this?

As an alternative, perhaps there could be a strongly typed API for adding and working with individual routes in code:

```cs
endpoints.MapReverseProxyRoute(route =>
{
   route.Foo = "x";
   // set various routing things
}).RequireAuthorization();
```

I'm not sure if adding `ProxyRoute` instances when mapping routing is a good thing to do.

Instead, an extension to map an individual route from config (would combine with other routes if MapReverseProxy() is also called):

```cs
endpoints.MapReverseProxyRoute("RouteId").RequireAuthorization();
endpoints.MapReverseProxy();
```

Tratcher: > Am I right that all YARP routes are defined in config? (i.e. a JSON file) And then the various `Configure` methods on this builder are to apply metadata to a subset of those routes?

YARP can load routes from IConfiguration or an arbitrary data source like Service Fabric via IProxyConfigProvider. Both produce the POCOs like ProxyRoute. The point of Configure here is to dynamically add metadata to Endpoints like AuthZ. We're not going to be able to define all possible endpoint metadata in the config sources. You get access to the ProxyRoute and Cluster so you can check if that route/cluster needs additional metadata on the endpoint.

> It does feel a little like the line is being blurred between JSON configuration and code configuration. (I don't know whether that is desirable or bad yet). Are there scenarios where people have needed this?

This came up few times before we'd added built in support for CORS and AuthZ. We expect there are other kinds of metadata people will want to add in the future.

> As an alternative, perhaps there could be a strongly typed API for adding and working with individual routes in code:
> 
> ```cs
> endpoints.MapReverseProxyRoute(route =>
> {
>    route.Foo = "x";
>    // set various routing things
> }).RequireAuthorization();
> ```
> 
> I'm not sure if adding `ProxyRoute` instances when mapping routing is a good thing to do.

I could see that for some very simple proxy scenarios, but we don't have any requirements yet for adding proxy routes here, they all come from outside sources.

> Instead, an extension to map an individual route from config (would combine with other routes if MapReverseProxy() is also called):
> 
> ```cs
> endpoints.MapReverseProxyRoute("RouteId").RequireAuthorization();
> endpoints.MapReverseProxy();
> ```

How would those two cases combine? Would MapReverseProxy load all routes and then MapReverseProxyRoute would tweak a specific route? That looks nice and clean but I worry it wouldn't scale well to many routes. I also anticipate the metadata would be determined by fields in the route other than the name (there's a property bag for people to add their own data).

That's how we got to the shape proposed here. The routes and clusters are loaded from an external source, you can inspect all of them with a single action, and add custom endpoint metadata to each. There are overloads that references routes and clusters by name but I don't think we need them.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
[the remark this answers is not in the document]
Tratcher: > YARP routes can be defined in config. Does YARP listen to config changes and repopulate routes + endpoints when config changes?

Yes, but that doesn't start until after Startup.Configure finishes and Endpoints are built for the first time.

> I think this lock is here because there is nothing stopping someone storing a convention builder post-startup and adding to it. That won't do anything unless the endpoint datasource also updates. Because user code can call `conventionBuilder.Add` at anytime, and endpoints can regenerate at anytime, this lock is to provide safety.

That's an odd departure from our usual guidelines about objects not being thread safe unless we have a known scenario that requires it, especially when as you say it's ineffective without the datasource also updating. If we were going to support adding conventions after startup then we'd want to be more explicit about it and make sure the end-to-end worked.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
davidfowl: Can we update the proposed shape in the PR title? I'm having a hard time tracking where we are here.

JamesNK: > How would those two cases combine? Would MapReverseProxy load all routes and then MapReverseProxyRoute would tweak a specific route?

Map method calls don't need to eagerly create endpoints when called. Think of each one providing configuration that is then used when the endpoint data source is resolved.

You can see that sort of thing in action in MVC:

https://github.com/dotnet/aspnetcore/blob/15f341f8ee556fa0c2825cdddfe59a88b35a87e2/src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
https://github.com/dotnet/aspnetcore/blob/940dcb0aba247da054ee435380d533e04944093c/src/Mvc/Mvc.Core/src/Routing/ControllerActionEndpointDataSource.cs

e.g. you can combine MapController, MapControllerRoute, MapDefaultControllerRoute, etc

test/ReverseProxy.Tests/Service/Management/ProxyConfigManagerTests.cs
Kahbazi: Sorry, I forgot to uncomment this test.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,295 @@
...
+        {
+            _ = convention ?? throw new ArgumentNullException(nameof(convention));
+
+            void Action(EndpointBuilder endpointBuilder)
+            {
+                var routeConfig = endpointBuilder.Metadata.OfType<RouteConfig>().Single();
+
+                var cluster = routeConfig.Cluster?.Config.Cluster;
+                var proxyRoute = routeConfig.ProxyRoute;
+                if (proxyRoute != null && cluster != null)
```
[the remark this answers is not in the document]
Kahbazi: I removed the null check for proxy route

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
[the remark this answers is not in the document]
Kahbazi: I removed the lock around adding conventions.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,295 @@
...
+                _conventions.Add(convention);
+            }
+        }
+
+        /// <summary>
+        /// Configures the endpoint for all routes 
+        /// </summary>
+        /// <param name="convention">The convention to add to the builder.</param>
+        /// <returns></returns>
+        public ReverseProxyConventionBuilder Configure(Action<IEndpointConventionBuilder> convention)
```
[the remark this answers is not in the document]
Kahbazi: I've added some tests.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -124,7 +165,11 @@ private async Task ReloadConfigAsync()
 
             try
             {
-                await ApplyConfigAsync(newConfig);
+                var hasChanged = await ApplyConfigAsync(newConfig);
+                if (hasChanged)
+                {
+                    CreateEndpoints();
+                }
```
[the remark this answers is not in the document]
Kahbazi: Where do you suggest we should stash the endpoints now that RouteConfig is too early to create the endpoints? 

Also do we need to re-create the endpoints of a cluster if its destinations changed?

Tratcher (commented):
Design update:
MapReverseProxy now returns a `ReverseProxyConventionBuilder : IEndpointConventionBuilder`. This has a new `Configure` API with three overloads. The purpose of the API is to let you provide one or more conventions that examine all of the routes loaded from config and append any custom endpoint metadata those routes need.

E.g. A convention could check for all routes that use cluster1 and set AllowAnonymous auth for those endpoints, where auth is normally configured per route. A developer could also insert their own endpoint metadata based on route or cluster extensible metadata fields.

New API:
```
Configure(Action<IEndpointConventionBuilder> convention)
Configure(Action<IEndpointConventionBuilder, ProxyRoute> convention)
Configure(Action<IEndpointConventionBuilder, ProxyRoute, Cluster> convention)
```

Clarifications from the prior discussion:
- The ProxyRoute and Cluster types are given here for information purposes, they're not modifiable at this point the the lifecycle. https://github.com/microsoft/reverse-proxy/issues/179 tracks clarifying that.
- We don't want to define new routes in Startup, we want routes to be sourced from an external store so they can always be updated without restarting.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -21,61 +20,111 @@
...
+        private void CreateEndpoints()
+        {
+            var endpoints = new List<Endpoint>();
+            foreach (var existingRoute in _routeManager.GetItems())
+            {
+                var runtimeConfig = existingRoute.Config;
+                // Only rebuild the endpoint for modified routes or clusters.
+                var endpoint = existingRoute.CachedEndpoint;
+                if (endpoint == null)
+                {
```
Kahbazi, line 119:
```suggestion
            var routes = _routeManager.GetItems();
            var endpoints = new List<Endpoint>(routes.Count);
            foreach (var existingRoute in routes)
            {
                // Only rebuild the endpoint for modified routes or clusters.
                var endpoint = existingRoute.CachedEndpoint;
                if (endpoint == null)
                {
                    var runtimeConfig = existingRoute.Config;                
```

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -156,8 +214,9 @@ private async Task ApplyConfigAsync(IProxyConfig config)
             }
 
             // Update clusters first because routes need to reference them.
-            UpdateRuntimeClusters(configuredClusters);
-            UpdateRuntimeRoutes(configuredRoutes);
+            var clustersChanged = UpdateRuntimeClusters(configuredClusters);
+            var routesChanged = UpdateRuntimeRoutes(configuredRoutes);
+            return routesChanged || clustersChanged;
```
Kahbazi, line 219:
```suggestion
            return routesChanged;
```
Now that cluster changes are propagated on route changes with revision, I think it's possible to only rely on `routesChanged` to recreate endpoints. 

There's no need to recreate the endpoints if an unused cluster is added, edited or removed, right?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,110 @@
...
+                var conventionBuilder = new EndpointBuilderConventionBuilder(endpointBuilder);
+                convention(conventionBuilder, proxyRoute.DeepClone(), cluster?.DeepClone());
+            }
+
+            Add(Action);
+
+            return this;
+        }
+
+        internal class EndpointBuilderConventionBuilder : IEndpointConventionBuilder
```
JamesNK, line 95:
```suggestion
        private class EndpointBuilderConventionBuilder : IEndpointConventionBuilder
```

Type used anywhere else?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// </summary>
+        /// <param name="convention">The convention to add to the builder.</param>
+        /// <returns></returns>
+        public ReverseProxyConventionBuilder Configure(Action<IEndpointConventionBuilder, ProxyRoute> convention)
+        {
+            _ = convention ?? throw new ArgumentNullException(nameof(convention));
+
+            void Action(EndpointBuilder endpointBuilder)
+            {
+                var routeConfig = endpointBuilder.Metadata.OfType<RouteConfig>().Single();
```
JamesNK, line 61: To avoid this I think you could derive EndpointBuilder (or RouteEndpointBuilder) and have a RouteConfig property. Then you can cast to that type here and access the property.

JamesNK (approved):
Minor comments

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: I think you're right. The strangest change I can think of is if an in-use cluster were removed, but rebuilding routes in that case doesn't seem helpful.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
Tratcher: It looks like RouteEndpointBuilder is sealed so I can't derive from it, I'd probably have to duplicate it. I'm going to leave this for now.

Tratcher:
@Kahbazi thanks for the hard work on this one, I'm glad we were able to work through it eventually.

Issue #222, opened 2020-06-02
Add more benchmark scenarios


halter73:
This is a follow up to https://github.com/microsoft/reverse-proxy/issues/40.

There are countless proxy scenarios and variations that can be benchmarked, but currently our proxy benchmarks only capture the following scenario:

- Using unencrypted connections and making HTTP/1.1 GET requests to both the proxy and backend

This scenario varies by response body size: 10B, 100B, 1KB, 10KB, 100KB plaintext responses are currently tested. Limiting response body sizes to just 32KB and 255KB should reduce the total number of scenarios allowing us to add more interesting scenarios without creating information overload.

The next proxy benchmark scenarios we plan to add are:

- Using TLS connections and making HTTP/2 GET requests to both the proxy and backend
- Using TLS connections and making HTTP/1.1 GET requests to both the proxy and backend
- Using unencrypted connections and making HTTP/1.1 POST requests to both the proxy and backend

After those scenarios are completed. Here are even more scenarios we can consider.

- Logging completely disabled
- Requests with bodies
- Variations on headers
  - Large request headers
  - Large response headers
  - Request/Response header transformation
- gRPC through the rpoxy
  - gRPC streaming

davidni:
Sugg adding larger payloads as well e.g. **10MB**.

For context, see e.g. [an issue I had with HttpSysServer on ASP .NET Core 2](https://github.com/aspnet/HttpSysServer/pull/418). A trivial HttpSys flag passed incorrectly resulted in massive impact (more than 20x slower throughput in that case). I expect most customers will encounter large payloads at one point or another, and making sure we're not adding perf cliffs along the way would be nice.

karelz:
Triage:

We added most of all these benchmarks: https://aka.ms/aspnet/benchmarks
Couple of them are in progress of being added.

Closing, current status is enough for 1.0 for now.

PR #330, opened 2020-07-16
Use pipe to copy stream

Kahbazi:


Kahbazi:
> ` The type or namespace name 'Pipelines' does not exist in the namespace 'System.IO' (are you missing an assembly reference?`

Do I need to do anything to use Pipelines in dotnet 5?

Tratcher:
Is the premise that this will improve performance? @halter73 can you help with the benchmarks and the Pipelines issue?

halter73:
I was hoping you could help with the Pipelines issue. This PR compiles (after fixing an error in StreamCopier.cs:75) for the net5.0 target but fails for the netcoreapp3.1 target.

System.IO.Pipelines is a bit weird in that it isn't included the base .NET Core runtime but is included in the ASP.NET Core runtime, but this was already the case well before 3.1.

Adding the PackageReference to Microsoft.ReverseProxy.csproj "fixes" the issue, but that shouldn't be necessary given the Microsoft.AspNetCore.App FrameworkReference! Indeed it works without the PackageReference when targeting net5.0.

Updating the SDK version in global.json from 5.0.100-preview.6.20310.4 to 3.1.302 *also* fixes the issue, but then we cannot build the net5.0 target ofc. Updating the version to 5.0.100-preview.7.20330.3 instead does not fix the issue.

I wonder if we're experiencing the same ref pack issues we saw in the 3.1.102 back in February still in our 5.0.100 preview SDKs. Is this a possibility @wtgodbe

P.S. @Tratcher Why do we use a Microsoft.AspNetCore.App FrameworkReference instead of using Microsoft.NET.Sdk.Web?

Tratcher:
> P.S. @Tratcher Why do we use a Microsoft.AspNetCore.App FrameworkReference instead of using Microsoft.NET.Sdk.Web?

No good reason, change it if it fixes the issue.

wtgodbe:
The Pipelines issue is because the `FrameworkReference` to `Microsoft.AspNetCore.App` is bringing in the 3.1.2 targeting pack, which has that bug. My guess is that we're getting that from the version of 3.1 that's bundled with VS, but I'm not 100% sure.

You could work around this by adding something like the following to Directory.Build.Targets, on the condition that `%(TargetFramework)` is `netcoreapp3.1`:

https://github.com/dotnet/aspnetcore/blob/master/Directory.Build.targets#L138-L145

Except updating `Microsoft.AspNetCore.App`, and setting `LatestRuntimeFrameworkVersion` to 3.1.3.

Tratcher:
I worked around the pipelines issue in another repo by adding `<PackageReference Include="System.IO.Pipelines" Version="4.7.2" />`

halter73:
I mentioned the PackageReference workaround, but I think @wtgodbe's workaround is better. I don't want to merge the PackageReference workaround if it could mess up NuGet package authoring.

Tratcher:
I think this is now unblocked now that we have moved to the preview8 SDK. Can you rebase?

Tratcher:
Maybe not...

Kahbazi:
Any solution for fixing the Pipelines?

brporter:
Do we have any benchmarks on this PR that demonstrates it's performance impact (good or bad?)

Tratcher:
> Do we have any benchmarks on this PR that demonstrates it's performance impact (good or bad?)

Not yet, but I do consider that a requirement before merging this PR. @sebastienros for help with that.

sebastienros:
Command to run the current state is here (change to use the yarp scenario):
https://github.com/aspnet/Benchmarks/tree/master/scenarios#sample-4

And to use a locally build version of the changes (with the files that are impacted by this PR):
https://github.com/aspnet/Benchmarks/tree/master/scenarios#how-to-upload-custom-files

But this PR should be rebased on master first

halter73:
To get coverage for the BodyReader half of this, we're going to want to add at least one scenario with a request body (#222). I'm addding HTTPS+HTTP/2 coverage with https://github.com/aspnet/Benchmarks/pull/1587 but haven't looked at request bodies yet.

Tratcher:
Side note: though we just shipped preview6 today, we don't plan to immediately update to the 5.0.0 RTM SDK. We're still deciding if we want to ship preview7 before or after 5.0.0 ships.

We can still test locally to see if that will fix the pipelines dependency issue though.

ManickaP:
**5.0 RTM fixes the build issue.**

It also brings back functioning `dotnet test` from the repo root, so :+1: 

Unfortunately, this change breaks this test: [`ProxyAsync_RequestBodyCanceledAfterResponse_Reported`](https://github.com/microsoft/reverse-proxy/blob/3d4ac5a84eee532b4c6b5a4a9a21fb262abc63dc/test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs#L1080-L1080)

My changes can be found here: https://github.com/ManickaP/reverse-proxy/commit/9a149ff352fe96e9983423e2d51349c21484c7db They're based on this PR and they also include additional fixes in compiler directives to make it all build.

Tratcher:
We decided to proceed with the RTM SDK update since it fixes this pipelines issue. https://github.com/microsoft/reverse-proxy/pull/488. You should be able to rebase on that once it's merged. The compiler directive fixups were fixed in a prior PR.

Kahbazi:
I couldn`t figure out the way to fix this test https://github.com/microsoft/reverse-proxy/blob/3d4ac5a84eee532b4c6b5a4a9a21fb262abc63dc/test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs#L1080-L1080

I could use some help.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -42,25 +38,37 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(Stream input,
...
-                    var read = await input.ReadAsync(buffer.AsMemory(), cancellation);
+                    var result = await input.ReadAsync(cancellation);
+                    var buffer = result.Buffer;
 
-                    // End of the source stream.
-                    if (read == 0)
+                    foreach (var item in buffer)
                     {
-                        return (StreamCopyResult.Success, null);
+                        if (cancellation.IsCancellationRequested)
```
ManickaP, line 48: This is causing the failing test.

If you get rid of it, the test will pass. However, the cancellation is then checked only between subsequent `ReadAsync` not per each item in the buffer (but it's passed to `WriteAsync` so it's not ignored for the whole buffer).
I'm not sure if this is the correct solution in this scenario though.

Also you can swap these two lines in the test:
https://github.com/microsoft/reverse-proxy/blob/3b2668a67cebd47af574fa358c63bd9553c8bed5/test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs#L1206-L1207
That helps as well. But once again, I didn't write it originally so I'm not 100% sure it's the right way to fix it.

ManickaP (commented):
I looked into to the failing test and found two possibilities for the fix.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -42,25 +38,37 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(Stream input,
...
+                        {
+                            return (StreamCopyResult.Canceled, new OperationCanceledException(cancellation));
+                        }
+                        reading = false;
+                        await output.WriteAsync(item, cancellation);
+                        totalBytes += buffer.Length;
```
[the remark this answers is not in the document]
Kahbazi: Before `StreamCopier` was static, `totalBytes` and `iops` were used in logs. I'll remove them.

MihaZupan:
I may be missing something here, but what are we actually achieving with the current implementation?
It looks like we're just wrapping the streams into a PipeReader/Writer, doing interleaved Read/Writes. Afaik that will do nothing but add IO.Pipelines overhead over Stream Read/Write calls.

Shouldn't we be using a Pipe here and running two (Read+Write) tasks in parallel?

If we want to log more details about the time spent reading/writing, we will need a way to hook into reads to get those timings. If StreamCopier accepts a PipeReader/Writer, we don't have a way to do so. If it keeps accepting Streams instead, we can wrap those to gather timings.

Tratcher:
> Shouldn't we be using a Pipe here and running two (Read+Write) tasks in parallel?

Kestrel uses pipes natively, so if we use the pipe reader/writer on the AspNet side of the copies it should remove a layer and improve performance. That said, you're right that this PR seems to go well beyond that. E.g. I'm not sure why pipes are being used in the upgrade code path when both sides are streams. Nor do I see why a pipe reader was created around HttpClient's response body stream.

@Kahbazi let's stick with using Kestrel's BodyReader and BodyWriter for now, no wrapping of other streams for now. I realize that will require multiple overloads of StreamCopier.CopyAsync, but that's OK.

MihaZupan:
If we're using the PipeReader/Writers from Kestrel, should we give up on tracking time spent in Read/Write for telemetry?
The only way to achieve it in that case (that I can think of) is to introduce another set of overloads in StreamCopier and pass in Streams directly at call sites if telemetry is enabled.

ManickaP:
@Kahbazi do you plan to continue working on this PR? Or should we take over? We're interested in the work done here and we would like to continue working on it.

Kahbazi:
@ManickaP Sorry for the long delay. I will work on this PR. I will rebase and add the telemetry for the other overloads.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -152,11 +153,232 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
...
+
+                    var memory = output.GetMemory();
+
+                    reading = true;
+                    var bytesRead = await input.ReadAsync(memory, cancellation);
+
+                    reading = false;
+                    output.Advance(bytesRead);
+
+                    var result = await output.FlushAsync();
```
Kahbazi, line 349: Is this the correct implementation of copying from `Stream` to `PIpeWriter`? Do I need to calculate telemetry for read and write at the same time?

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -152,11 +153,232 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
...
+                                        lastTime = clock.GetStopwatchTimestamp();
+                                        nextTransferringEvent = lastTime + stopwatchTicksBetweenTransferringEvents;
+                                    }
+                                }
+                            }
+                        }
+                    }
+
+                    // Tell the PipeReader how much of the buffer has been consumed.
+                    input.AdvanceTo(result.Buffer.End);
```
Tratcher, line 267: AdvanceTo needs to be called in a finally block.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
Tratcher: FlushAsync is the 'write' operation and it should have the same telemetry as the other writes.

Tratcher:
```suggestion
                    var memory = output.GetMemory();

                    reading = true;
                    var bytesRead = await input.ReadAsync(memory, cancellation);

                    reading = false;
                    output.Advance(bytesRead);

                    var result = await output.FlushAsync(cancellation);
```

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -152,11 +153,232 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
...
+                    var result = await output.FlushAsync();
+
+                    if (bytesRead == 0)
+                    {
+                        break;
+                    }
+                }
+
+                reading = false;
+                await output.CompleteAsync();
```
Tratcher, line 358: Don't complete the pipe, it would prevent us from sending trailers.
```suggestion
```

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -152,11 +153,232 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
...
+
+                    // Stop reading if there's no more data coming.
+                    if (result.IsCompleted)
+                    {
+                        break;
+                    }
+                }
+
+                // Mark the PipeReader as complete.
+                await input.CompleteAsync();
```
Tratcher, line 278: The server will clean this up.
```suggestion
```

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -624,8 +624,8 @@ void ReportResult(HttpContext context, bool reqeuest, StreamCopyResult result, E
             }
         }
 
-        private async Task<(StreamCopyResult, Exception)> CopyResponseBodyAsync(HttpContent destinationResponseContent, Stream clientResponseStream,
-            CancellationToken cancellation)
+        private async Task<(StreamCopyResult, Exception)> CopyResponseBodyAsync(HttpContent destinationResponseContent, PipeWriter clientResponseStream,
+             CancellationToken cancellation)
```
MihaZupan, line 628: nit: looks like unintentional indent

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -152,11 +153,284 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
...
+                            if (firstReadTime == -1)
+                            {
+                                firstReadTime = currentReadTime;
+                            }
+                        }
+                    }
+
+                    try
+                    {
+                        if (!result.Buffer.IsEmpty)
```
MihaZupan, line 229: Is this check needed?

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -152,11 +153,284 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
...
+                        {
+                            foreach (var item in result.Buffer)
+                            {
+                                if (cancellation.IsCancellationRequested)
+                                {
+                                    return (StreamCopyResult.Canceled, new OperationCanceledException(cancellation));
+                                }
+                                reading = false;
```
MihaZupan, line 237: This can be moved outside of the loop

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -152,11 +153,284 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
...
+                    if (cancellation.IsCancellationRequested)
+                    {
+                        return (StreamCopyResult.Canceled, new OperationCanceledException(cancellation));
+                    }
+
+                    reading = false;
+
+                    try
+                    {
+                        var result = await output.FlushAsync(cancellation);
```
MihaZupan, line 381: @Tratcher does Kestrel call `CancelPendingFlush`? If so I assume we'd want to look at the `result.IsCanceled`

MihaZupan (commented):
Telemetry looks good to me.

Just a heads up, #257 will likely be merged first so there will be a bit more churn around telemetry because of `IClock` changes.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
Tratcher: @halter73 Kestrel wouldn't call CancelPendingFlush on the PipeWriter, right?

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -152,11 +153,284 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest
...
+                }
+
+                while (true)
+                {
+                    if (cancellation.IsCancellationRequested)
+                    {
+                        return (StreamCopyResult.Canceled, new OperationCanceledException(cancellation));
+                    }
+
+                    var memory = output.GetMemory();
```
Tratcher, line 344: The size of the buffers used here will have a huge impact on throughput. We'll need to test different sizes when benchmarking.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
halter73: Kestrel doesn't call CancelPendingFlush() on the BodyWriter. In general, when Kestrel does call CancelPendingRead() or CancelPendingFlush(), it will ensure that this isn't observable to the application.

samsp-msft:
@karelz & @MihaZupan - who is working on getting this PR merged?

karelz:
I believe this is the one @ManickaP is working on -- it is also assigned to her ...

src/ReverseProxy/Service/Proxy/StreamCopier.cs
ManickaP: Moved outside to bypass the loop and return early.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
ManickaP: Done.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
ManickaP: Removed.

ManickaP:
I did some perf measurements. I played with the buffer sizes:
1. For stream to stream copy: https://github.com/microsoft/reverse-proxy/blob/6c0d85ab6b6f8ef8693c80e80563bd594db1024b/src/ReverseProxy/Service/Proxy/StreamCopier.cs#L21
2. For stream to pipe writer copy: https://github.com/microsoft/reverse-proxy/blob/6c0d85ab6b6f8ef8693c80e80563bd594db1024b/src/ReverseProxy/Service/Proxy/StreamCopier.cs#L337
And I compared the results to the master where I played with same buffer size as in 1.

Conclusion is that the pipelines do not bring any perf improvements that cannot be achieved by tweaking 1. For stream to stream copy either 64kB or 32kB seem to be the most efficient. For 2. the default (4kB) is the best, but it doesn't bring any additional perf improvements. I was only able to achieve the same values as for streams with optimized buffer.

However, a small change brought some additional gains. It was actually getting rid of the copier and replace it with single line:
```C#
input.CopyToAsync(output, DefaultBufferSize, cancellation)
```
So we might want to consider special casing for when telemetry is not on and bypass the whole method.

### Some numbers (for 10kB payload):
10kB payload showed the biggest differences, I tried 1kB as well as 100kB. And all is Linux.

#### The Best Current Master
| load                   |         |
| ---------------------- | ------- |
| CPU Usage (%)          | 34      |
| Cores usage (%)        | 406     |
| Working Set (MB)       | 41      |
| Start Time (ms)        | 110     |
| First Request (ms)     | 246     |
| Requests               | 934 110 |
| Bad responses          | 0       |
| Mean latency (us)      | 4 104   |
| Max latency (us)       | 49 437  |
| Requests/sec           | 62 369  |
| Requests/sec (max)     | 67 348  |
| Read throughput (MB/s) | 615,24  |

#### The Best Pipes (1. = 64kB; 2. = 4kB)
| load                   |         |
| ---------------------- | ------- |
| CPU Usage (%)          | 33      |
| Cores usage (%)        | 399     |
| Working Set (MB)       | 42      |
| Start Time (ms)        | 108     |
| First Request (ms)     | 251     |
| Requests               | 971 612 |
| Bad responses          | 0       |
| Mean latency (us)      | 3 945   |
| Max latency (us)       | 53 757  |
| Requests/sec           | 64 885  |
| Requests/sec (max)     | 69 686  |
| Read throughput (MB/s) | 640,62  |

#### The Best Streams (1. = 64kB)
| load                   |         |
| ---------------------- | ------- |
| CPU Usage (%)          | 33      |
| Cores usage (%)        | 402     |
| Working Set (MB)       | 42      |
| Start Time (ms)        | 110     |
| First Request (ms)     | 244     |
| Requests               | 974 876 |
| Bad responses          | 0       |
| Mean latency (us)      | 3 932   |
| Max latency (us)       | 53 776  |
| Requests/sec           | 65 109  |
| Requests/sec (max)     | 69 622  |
| Read throughput (MB/s) | 642,72  |

#### The Best Overall (`CopyToAsync` with 32kB buffer)
| load                   |         |
| ---------------------- | ------- |
| CPU Usage (%)          | 35      |
| Cores usage (%)        | 418     |
| Working Set (MB)       | 42      |
| Start Time (ms)        | 110     |
| First Request (ms)     | 241     |
| Requests               | 987 214 |
| Bad responses          | 0       |
| Mean latency (us)      | 3 883   |
| Max latency (us)       | 57 535  |
| Requests/sec           | 65 928  |
| Requests/sec (max)     | 71 307  |
| Read throughput (MB/s) | 650,64  |

ManickaP:
The comparison against the merged #661 from yesterday:
| load                   | base10k | pipes10k |          |
| ---------------------- | ------- | -------- | -------- |
| CPU Usage (%)          |      31 |       31 |    0,00% |
| Cores usage (%)        |     366 |      371 |   +1,37% |
| Working Set (MB)       |      41 |       41 |    0,00% |
| Start Time (ms)        |     110 |      108 |   -1,82% |
| First Request (ms)     |     264 |      261 |   -1,14% |
| Requests               | 799 807 |  790 978 |   -1,10% |
| Bad responses          |       0 |        0 |          |
| Mean latency (us)      |   4 793 |    4 847 |   +1,13% |
| Max latency (us)       |  54 746 |  130 423 | +138,23% |
| Requests/sec           |  53 419 |   52 798 |   -1,16% |
| Requests/sec (max)     |  62 379 |   63 936 |   +2,49% |
| Read throughput (MB/s) |  526,76 |   521,48 |   -1,00% |

Unfortunately at the moment, this changes is performance-wise on par with the stream-to-stream original. The root cause is that underneath it's stream-to-pipe or pipe-to-stream and not pipe-to-pipe (all the way through, not just stream disguised as pipe). And since it triples the amount of code, it's currently not worth the added maintenance. So I'll close this PR for now.

@Kahbazi This doesn't mean that this PR was all for naught. It allowed us to investigate and learn more about pipe/stream performance which will be helpful in the future. We really **appreciate** the work you've done here and how much it helped us. Thank you!

Issue #137, opened 2020-05-04
Configuring the HttpClientHandler (& Client Certs)


Tratcher:
### What should we add or change to make your life better?

The application needs to be able to configure attributes on the outbound HttpClient / SocketHttpHandler / HttpClientHandler per backend and/or backend endpoint, especially client certificates.

Client certificates for backend endpoints should be configurable in the config file as well as in code.

### Why is this important to you?

Service-to-service mutual authentication.

Tratcher:
https://github.com/microsoft/reverse-proxy/issues/113

Tratcher:
Note we've now had two partners need to customize server cert validation because of self-signed certs. The workaround at the moment is to replace the entire ProxyHttpClientFactory in DI.

Tratcher:
Consolidated from https://github.com/microsoft/reverse-proxy/issues/253
#### Requirements:
* Ability to specify `SocketsHttpHandler.SslClientAuthenticationOptions.RemoteCertificateValidationCallback` per backend
* Ability for that to be set up to depend on Backend config. Internally, we have an implementation that works efficiently using `Signals`
   * When a backend's config changes, a derived signal produces a new cert validation callback lambda
    * In the hot-path, cert validation just calls the latest validation lambda produced by the signal, which always tracks the backend config

### Why is this important to you?
Compliance requirement. YARP isn't an option for us without this.

Tratcher:
Also: The ability to add a DelegatingHandler for advanced customization. https://github.com/microsoft/reverse-proxy/issues/283

samsp-msft:
Need to be able to tweak the handler on a per cluster basis.

Need to consider the association between httpClient and clusters/destinations. Are httpclient instances created on a per cluster basis. What happens if destinations are used in multiple clusters.

Tratcher:
Island gateway explicitly does not want a destination in multiple clusters to share connections or HttpClient instances. They want to limit the impact different clients/routes can have on each other.

This is the current behavior.

Tratcher:
Implementation notes:
IProxyHttpClientFactory is split into two APIs,  CreateNormalClient() and CreateUpgradableClient(). The only difference is this additional line in the upgrade client:
https://github.com/microsoft/reverse-proxy/blob/24c0c5a969bb025dd7a355f0b62ee4cbac96c99f/src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs#L63
[...]

PR #332, opened 2020-07-16
Remove CreateUpgradableClient from the client factory 

Tratcher:
Contributes to #137. We still need to finish the design around the client factory but this part was well understood. The separate handler with its own config is not needed for WebSocket/Upgrade scenarios. The pooling setting was an optimization copied from ClientWebSocket. 

I need to do some refactoring in HttpProxy and it will be easier if I get this out of the way first.

### DIFF
--- a/src/ReverseProxy/Service/Proxy/Infrastructure/IProxyHttpClientFactory.cs
+++ b/src/ReverseProxy/Service/Proxy/Infrastructure/IProxyHttpClientFactory.cs
@@ -23,30 +23,13 @@ namespace Microsoft.ReverseProxy.Service.Proxy.Infrastructure
         /// </summary>
         /// <remarks>
         /// <para>
-        /// Each call to <see cref="CreateNormalClient()"/> is guaranteed
-        /// to return a new <see cref="HttpMessageInvoker"/> instance.
-        /// It is generally not necessary to dispose of the <see cref="HttpMessageInvoker"/>
-        /// as the <see cref="IProxyHttpClientFactory"/> tracks and disposes resources
-        /// used by the <see cref="HttpClient"/>.
-        /// </para>
-        /// </remarks>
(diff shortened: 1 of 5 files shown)


src/ReverseProxy/Service/Proxy/Infrastructure/IProxyHttpClientFactory.cs
```diff
@@ -23,30 +23,13 @@ internal interface IProxyHttpClientFactory : IDisposable
...
-        /// Each call to <see cref="CreateUpgradableClient()"/> is guaranteed
+        /// Each call to <see cref="CreateClient()"/> is guaranteed
         /// to return a new <see cref="HttpMessageInvoker"/> instance.
         /// It is generally not necessary to dispose of the <see cref="HttpMessageInvoker"/>
         /// as the <see cref="IProxyHttpClientFactory"/> tracks and disposes resources
         /// used by the <see cref="HttpMessageInvoker"/>.
         /// </para>
         /// </remarks>
-        HttpMessageInvoker CreateUpgradableClient();
+        HttpMessageInvoker CreateClient();
```
halter73, line 33: Would you ever need to know a HttpClient creation type whether the request is upgradable? Maybe the request be passed into CreateClient().

Tratcher: > Would you ever need to know a HttpClient creation type whether the request is upgradable? Maybe the request be passed into CreateClient().

Not that I know of. Upgraded requests should work fine with the normal handler. If you really needed to do something special per request you could do that inside the returned client.

halter73: I just now reread my original comment. I'm glad you at least understood my word salad 😆

Issue #255, opened 2020-06-19
POST not working in IIS/Express due to IsUpgradableRequest


maeneak:
-- Admin Edit: Unrelated issue content hidden --

<details>

I'm currently unable to get Request Transforms to work on NETCore 3.1 in VS19 16.6.5.
```
    {
        "RouteId": "oh/rest",
        "BackendId": "oh",
        "Match": {
          "Host": "localhost",
          "Path": "/rest/{*remainder}"
        },
        "Transforms": [
          { "X-Forwarded": "" },
          { "RequestHeaderOriginalHost": "true" }
        ]
      }
```
From wireshark:
```
POST /rest/items/Flush2Relay_Switch1 HTTP/1.1\r\n
    Host: otherhost:8080\r\n
    Accept: application/json, text/plain, */*\r\n
    Accept-Encoding: gzip, deflate, br\r\n
    Accept-Language: en-US, en; q=0.9\r\n
    Cache-Control: no-cache\r\n
    Connection: close\r\n
     [truncated]Cookie: HIDDDEN
    Pragma: no-cache\r\n
    Referer: https://localhost:44354/paperui/index.html\r\n
    User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.106 Safari/537.36 Edg/83.0.478.54\r\n
    Origin: https://localhost:44354\r\n
    sec-fetch-site: same-origin\r\n
    sec-fetch-mode: cors\r\n
    sec-fetch-dest: empty\r\n
    X-Forwarded-Proto: https\r\n
    X-Forwarded-Host: localhost:44354\r\n
    X-Forwarded-For: ::1\r\n
    Content-Length: 0\r\n
    \r\n
    [Full request URI: http://otherhost:8080/rest/items/Flush2Relay_Switch1]
    [HTTP request 1/1]
    [Response in frame: 52458]
```

The proxy is proxying correctly but the incoming request is not being modified. Is this feature still in development?

</details>

Tratcher:
Transforms were introduced in preview2 which has not been released yet. You can either build the repo locally or wait for preview2 to release (soon).

maeneak:
Awesome thanks mate. Also POST data is not getting proxied, is this in the pipeline as well?

Tratcher:
POST data should already work. The request above has Content-length 0?

maeneak:
[...]

PR #335, opened 2020-07-17
Mitigate POST issue in IIS 

Tratcher:
This is a temporary mitigation for #255. IIS marks all requests as upgradable, and we don't proxy the request body for upgradable requests, so that breaks all POST requests in IIS.

Mitigation: Constrain upgrades to WebSocket requests. In theory Upgrade can be used for other things but we haven't seen that in practice.

We've had at least three separate reports of this now, so I want to get this mitigation into preview3.

### DIFF
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -16,6 +16,7 @@ using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Features;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
 using Microsoft.ReverseProxy.Service.Metrics;
 using Microsoft.ReverseProxy.Service.Proxy.Infrastructure;
 using Microsoft.ReverseProxy.Service.RuntimeModel.Transforms;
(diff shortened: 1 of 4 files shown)


src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -87,8 +88,10 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
             // :::::::::::::::::::::::::::::::::::::::::::::
             // :: Step 1: Create outgoing HttpRequestMessage
             var upgradeFeature = context.Features.Get<IHttpUpgradeFeature>();
-            var isUpgrade = upgradeFeature?.IsUpgradableRequest ?? false;
-            // Default to HTTP/1.1 for proxying upgradable requests. This is already the default as of .NET Core 3.1
+            var isUpgrade = (upgradeFeature?.IsUpgradableRequest ?? false)
+                // Mitigate https://github.com/microsoft/reverse-proxy/issues/255, IIS considers all requests upgradeable.
+                && string.Equals("WebSocket", context.Request.Headers[HeaderNames.Upgrade], StringComparison.OrdinalIgnoreCase);
```
halter73, line 93: Why not use `StringComparison.Ordinal`?

Tratcher: We treat most request values as ignore case by default. If a client specified "websocket" I'd still want to match it.

halter73: But the perf! 😆

Issue #323, opened 2020-07-13
502 - Response when routing to Azure App Website - GET requests proxied with Transfer-Encoding


jeremielumandong:
### Describe the bug
Can you please show me the correct settings I need to add to allow Azure App Service application to be loaded in the Proxy. I'm trying to replace our existing Node Reverse Proxy to this one and I can't make it work. The node version is loading properly the website.

### Steps:

Added a simple Angular App as Front End as Azure App Service and configure that as the destination address.
Calling it from the YARP Proxy Server returns 502. I tried removing the ARR affinity in the Azure Website and still not working.

### Error:

{StatusCode: 502, ReasonPhrase: 'Bad Gateway', Version: 1.1, Content: System.Net.Http.HttpConnectionResponseContent, Headers:
{
  Server: Microsoft-IIS/10.0
  Date: Mon, 13 Jul 2020 19:40:54 GMT
  Content-Type: text/html
  Content-Length: 1477
}}

Tratcher:
Can you please share the config that you already tried?

jeremielumandong:
My sample config below. 

"ReverseProxy": {
    "Clusters": {
      "sample_ui": {
        "Destinations": {
          "backend_ui": {
            "Address": "https://proxyui.azurewebsites.net/"
          }
        }
      }
    },
    "Routes": [
      {
        "RouteId": "route1",
        "ClusterId": "sample_ui",
        "Match": {
          "Methods": [
            "GET",
            "POST"
          ],
          "Hosts": [
            "localhost"
          ],
          "Path": "/ui/{**catchall}"
        },
        "Transforms": [
          {
            "PathRemovePrefix": "/ui"
          },
          {
            "X-Forwarded": ""
          }
        ]
      }
    ]
  }

Tratcher:
Hmmm, this indicates an error in the application.
"502 - Web server received an invalid response while acting as a gateway or proxy server.
There is a problem with the page you are looking for, and it cannot be displayed. When the Web server (while acting as a gateway or proxy) contacted the upstream content server, it received an invalid response from the content server."

However, it only happens when going through YARP. I was able to repro this against your site and one random other azurewebsites.net site. I'll dig a little deeper.

Tratcher:
Notes:
- It still doesn't work if I use Fiddler as the client to send an empty request, so it's not related to any of the request fields provided by the browser.
[...]

PR #336, opened 2020-07-18
Be specific about what requests will have bodies 

Tratcher:
#323 HttpProxy today always generates an HttpContent request body. That makes HttpClient assume there's a body and it adds the Transfer-Encoding: chunked header to HTTP/1.1 requests. Servers like Azure Web Sites will reject GET requests that have a Transfer-Encoding: chunked header.

In AspNetCore apps we usually tell people to just read the body to see if there is one, but we don't want to do that too soon in the proxy or else it can trigger side-effects like 100-continue.

I read through a the specs for each HTTP version and it's a bit of a mess. See the inline comments for details.

TODO: Tests

### DIFF
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -156,7 +156,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
             // :::::::::::::::::::::::::::::::::::::::::::::
             // :: Step 2: Setup copy of request body (background) Downstream --► Proxy --► Upstream
             // Note that we must do this before step (3) because step (3) may also add headers to the HttpContent that we set up here.
-            var bodyToUpstreamContent = SetupCopyBodyUpstream(context.Request.Body, upstreamRequest, in proxyTelemetryContext, isStreamingRequest, longCancellation);
+            var bodyToUpstreamContent = SetupCopyBodyUpstream(context.Request, upstreamRequest, in proxyTelemetryContext, isStreamingRequest, longCancellation);
 
             // :::::::::::::::::::::::::::::::::::::::::::::
             // :: Step 3: Copy request headers Downstream --► Proxy --► Upstream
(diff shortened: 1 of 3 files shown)


src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -372,11 +372,69 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
...
         }
 
-        private StreamCopyHttpContent SetupCopyBodyUpstream(Stream source, HttpRequestMessage upstreamRequest, in ProxyTelemetryContext proxyTelemetryContext, bool isStreamingRequest, CancellationToken cancellation)
+        private StreamCopyHttpContent SetupCopyBodyUpstream(HttpRequest request, HttpRequestMessage upstreamRequest, in ProxyTelemetryContext proxyTelemetryContext, bool isStreamingRequest, CancellationToken cancellation)
         {
+            // If we generate an HttpContent without a Content-Length then for HTTP/1.1 HttpClient will add a Transfer-Encoding: chunked header
+            // even if it's a GET request. Some servers reject requests containing a Transfer-Encoding header if they're not expecting a body.
+            // Try to be as specific as possible about the client's intent to send a body. The one thing we don't want to do is to start
+            // reading the body early because that has side-effects like 100-continue.
+            bool? hasBody;
```
halter73, line 381: Can we make this a normal bool?

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -372,11 +372,69 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
...
+                && string.Equals("chunked", transferEncoding.ToString(), StringComparison.OrdinalIgnoreCase))
+            {
+                hasBody = true;
+            }
+            else if (contentLength.HasValue)
+            {
+                hasBody = contentLength > 0;
+            }
+            // Kestrel HTTP/2: There are no required headers that indicate if there is a request body so we need to sniff other fields.
+            else if (!ProtocolHelper.IsHttp2(request.Protocol))
```
halter73, line 408: It's probably slightly better to check for HTTP/1.x in case someone enable experimental HTTP/3 support.

Tratcher: I'd have to check 1.0 as well then. What I really want is a < HTTP2 but we don't have those APIs.

I honestly don't know if this logic even works for HTTP/3, I haven't read through the spec closely enough.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -372,11 +372,69 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
...
+            {
+                hasBody = true;
+            }
+            // https://tools.ietf.org/html/rfc7231#section-4.3.1
+            // A payload within a GET/HEAD/DELETE/CONNECT request message has no defined semantics; sending a payload body on a
+            // GET/HEAD/DELETE/CONNECT request might cause some existing implementations to reject the request.
+            else if (HttpMethods.IsGet(method)
+                || HttpMethods.IsHead(method)
+                || HttpMethods.IsDelete(method)
+                || HttpMethods.IsConnect(method))
```
halter73, line 426: So if one of these HTTP/2 GET/HEAD/DELETE/CONNECT do wind up having a body, we effectively just strip it from the proxied request, right? Now that we verified there's no "Expect: 100-continue", maybe we should try the TryRead trick you were suggesting before considering the request bodyless. It still wouldn't be perfect, but it seems like a reasonable improvement.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
halter73: Neither do I, but I think the HTTP/2 logic is more likely to work with HTTP/3 than the HTTP/1.x logic. I don't think it's a huge deal since hopefully this logic will be replaced soon.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
Tratcher: Reading the body is a can of worms I really don't want to open. If we get a stronger hint from Kestrel in 5.0 I'd prefer to #if this code out.

halter73: If we don't have to worry about 100-continue, why would TryRead be a can of worms?

Tratcher: Would I also have to call Advance? Can it throw IO exceptions?

Tratcher: Filed https://github.com/dotnet/aspnetcore/issues/24175 for followup

PR #338, opened 2020-07-21
Adding catch-all Path

sebastienros:
Without this I get 

```
fai: Microsoft.ReverseProxy.Service.RouteValidator[51]
      Route `route` requires Hosts or Path specified. Set the Path to `/{**catchall}` to match all requests.
```

### DIFF
--- a/samples/BenchmarkApp/Startup.cs
+++ b/samples/BenchmarkApp/Startup.cs
@@ -37,7 +37,7 @@ namespace BenchmarkApp
             {
                 { "Routes:0:RouteId", "route" },
                 { "Routes:0:ClusterId", "cluster" },
-                { "Routes:0:Match:Host", new Uri(urls.Split(';', 1)[0]).Host },
+                { "Routes:0:Match:Hosts:0", new Uri(urls.Split(';', 1)[0]).Host },
                 { "Routes:0:Match:Path", "/{**catchall}" }
             };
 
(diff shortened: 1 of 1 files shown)


samples/BenchmarkApp/Startup.cs
```diff
@@ -38,6 +38,7 @@ public void ConfigureServices(IServiceCollection services)
                 { "Routes:0:RouteId", "route" },
                 { "Routes:0:ClusterId", "cluster" },
                 { "Routes:0:Match:Host", new Uri(urls.Split(';', 1)[0]).Host },
```
sebastienros, line 40: Is this still supported? Some articles mention "Hosts" instead

Kahbazi: No, `string Host` has been changed to `IReadOnlyList<string> Hosts` in https://github.com/microsoft/reverse-proxy/commit/3e84b854e73bce20f469b688000a526e08644073

Kahbazi, line 40: This should work.
```suggestion
                { "Routes:0:Match:Hosts:0", new Uri(urls.Split(';', 1)[0]).Host },
```

samples/BenchmarkApp/Startup.cs
```diff
@@ -37,7 +37,8 @@ public void ConfigureServices(IServiceCollection services)
             {
                 { "Routes:0:RouteId", "route" },
                 { "Routes:0:ClusterId", "cluster" },
-                { "Routes:0:Match:Host", new Uri(urls.Split(';', 1)[0]).Host },
+                { "Routes:0:Match:Hosts:0", new Uri(urls.Split(';', 1)[0]).Host },
+                { "Routes:0:Match:Path", "/{**catchall}" }
```
Tratcher, line 41: You don't need the catch-all now that you've fixed the Host-Hosts.
```suggestion
```

Tratcher:
Stephen's fix is probably better: https://github.com/microsoft/reverse-proxy/pull/339

PR #339, opened 2020-07-21
Use catchall route in BenchmarkApp

halter73:
@sebastienros This should fix the issues you were seeing. Now, instead of the requiring the Host header from the load test client match the BenchmarkApp's --urls parameter, the BenchmarkApp should proxy any request.


Tratcher:
Conflicts with https://github.com/microsoft/reverse-proxy/pull/338.

Tratcher:
This version is probably better.

halter73:
Oops! Didn't see the other one until now. I do think it's better to just remove the host matching, so I'll just keep this open for now.

Issue #345, opened 2020-07-22
Bug in remainder/catchall or unlisted breaking change in preview 3


amweiss:
### Describe the bug
I think this is a bug and not intentional, but if it's intentional it was not listed as a breaking change for preview3.

It seems like the remainder/catchall style routing items are no longer working and the replacement value name is added to the route instead of the value.

git bisect shows this as the source of the issue:
```
e25fb189613907753972700b87b7f142ed23e34e is the first bad commit
commit e25fb189613907753972700b87b7f142ed23e34e
Date:   Tue Jul 21 22:12:42 2020 +0430

    Add Helper methods for Transforms (#324)
```

### To Reproduce
* Run the ReverseProxy in samples.
* GET `http://localhost:5000/api/testing/stuff/v1/name?query=something&trace=other`
* Expected to have that route to `http://localhost:10010/foo/testing/bar/v1/name?query=something&trace=other`
* Instead it gets routed to: `http://localhost:10010/foo/%7Bplugin%7D/bar/%7Bremainder%7D?query=something&trace=other`

```
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://localhost:5000/api/testing/stuff/v1/name?query=something&trace=other - -
dbug: Microsoft.AspNetCore.HostFiltering.HostFilteringMiddleware[0]
      Wildcard detected, all requests with hosts will be allowed.
dbug: Microsoft.AspNetCore.Routing.Matching.DfaMatcher[1001]
      2 candidate(s) found for the request path '/api/testing/stuff/v1/name'
dbug: Microsoft.AspNetCore.Routing.Matching.DfaMatcher[1005]
      Endpoint 'route1' with route pattern '/api/{plugin}/stuff/{*remainder}' is valid for the request path '/api/testing/stuff/v1/name'
dbug: Microsoft.AspNetCore.Routing.Matching.DfaMatcher[1005]
      Endpoint 'route2' with route pattern '/{**catchall}' is valid for the request path '/api/testing/stuff/v1/name'
dbug: Microsoft.AspNetCore.Routing.EndpointRoutingMiddleware[1]
      Request matched endpoint 'route1'
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
      Executing endpoint 'route1'
info: Microsoft.ReverseProxy.Middleware.ProxyInvokerMiddleware[31]
      Operation started: ReverseProxy.Proxy
[...]

PR #346, opened 2020-07-22
Pass PathString.Value to prevent encoding

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/345


Tratcher:
Thanks

PR #357, opened 2020-08-03
Rework the route loading pattern

Tratcher:
#8 #277

High level changes:
- Added the new public IProxyConfigProvider and IProxyConfig types for loading routes from code. These are using the existing ProxyRoute and Cluster objects. If we want to customize the config file binding structure we will likely introduce custom types for that scenario and then map them internally to the common type. The shape of those route and cluster types is a whole different design discussion.
- Removed the redundant ParsedRoute type
- IRouteValidator changed to IConfigValidator, added support for validating clusters, and made public.
- ICluster/RoutesRepo's removed
- Combine IProxyDynamicEndpointDataSource and IReverseProxyConfigManager as IProxyConfigManager. It builds config and exposes the results to routing as an EndpointDataSource.
- Changed the error handling to throw on startup. There's one thorny issue here because validation is async and startup isn't. See ReverseProxyIEndpointRouteBuilderExtensions
- Split the sample to demonstrate code vs config

TODO:
- [x] Make the route validator public
- [x] Self-review
- [x] Tests
- [x] PR description
- [x] Docs - I'll work on these while you review
- [ ] Proof of concept (Service fabric? K8?) - Follow-up


src/ReverseProxy/Abstractions/Config/IConfigValidator.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Provides a method to validate routes and clusters.
+    /// </summary>
+    public interface IConfigValidator
+    {
+        /// <summary>
+        /// Validates a route and returns all errors
+        /// </summary>
+        Task<IList<Exception>> ValidateRouteAsync(ProxyRoute route);
```
davidfowl, line 19: ValueTask, these are synchronous 90% of the time.

src/ReverseProxy/Service/Config/DynamicConfigBuilder.cs
```diff
@@ -3,45 +3,39 @@
...
         {
             _filters = filters ?? throw new ArgumentNullException(nameof(filters));
-            _parsedRouteValidator = parsedRouteValidator ?? throw new ArgumentNullException(nameof(parsedRouteValidator));
-            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
-            _sessionAffinityProviders = sessionAffinityProviders?.ToProviderDictionary() ?? throw new ArgumentNullException(nameof(sessionAffinityProviders));
-            _affinityFailurePolicies = affinityFailurePolicies?.ToPolicyDictionary() ?? throw new ArgumentNullException(nameof(affinityFailurePolicies));
+            _configValidator = configValidator ?? throw new ArgumentNullException(nameof(configValidator));
         }
 
         public async Task<DynamicConfigRoot> BuildConfigAsync(IReadOnlyList<ProxyRoute> routes, IReadOnlyList<Cluster> clusters, CancellationToken cancellation)
```
davidfowl, line 27: This DynamicConfigRoot should be removed.

Tratcher: In favor of `Task<(IList<ProxyRoute>, IDictionary<string, Cluster>)>`?

davidfowl: Don't return anything just mutate the inputs.

Tratcher: We don't own the inputs, IProxyConfigProvider gave them to us. That's why the clone them before we apply any changes.

src/ReverseProxy/Configuration/ReverseProxyIEndpointRouteBuilderExtensions.cs
```diff
@@ -52,7 +53,8 @@ public static void MapReverseProxy(this IEndpointRouteBuilder endpoints, Action<
             var routeBuilder = endpoints.ServiceProvider.GetRequiredService<IRuntimeRouteBuilder>();
             routeBuilder.SetProxyPipeline(app);
 
-            var dataSource = (EndpointDataSource)endpoints.ServiceProvider.GetRequiredService<IProxyDynamicEndpointDataSource>();
+            var configManager = endpoints.ServiceProvider.GetRequiredService<IProxyConfigManager>();
+            var dataSource = configManager.LoadAsync().GetAwaiter().GetResult();
```
davidfowl, line 57: big comment here

src/ReverseProxy/Service/Config/ConfigValidator.cs
```diff
@@ -0,0 +1,263 @@
...
+            ValidateMethods(errors, route.Match.Methods, route.RouteId);
+            errors.AddRange(_transformBuilder.Validate(route.Transforms));
+            await ValidateAuthorizationPolicyAsync(errors, route.AuthorizationPolicy, route.RouteId);
+            await ValidateCorsPolicyAsync(errors, route.CorsPolicy, route.RouteId);
+
+            return errors;
+        }
+
+        // Note this performs all validation steps without short circuiting in order to report all possible errors.
+        public Task<IList<Exception>> ValidateClusterAsync(Cluster cluster)
```
davidfowl, line 100: ValueTask

src/ReverseProxy/Service/Config/ConfigValidator.cs
```diff
@@ -0,0 +1,263 @@
...
+        {
+            _transformBuilder = transformBuilder ?? throw new ArgumentNullException(nameof(transformBuilder));
+            _authorizationPolicyProvider = authorizationPolicyProvider ?? throw new ArgumentNullException(nameof(authorizationPolicyProvider));
+            _corsPolicyProvider = corsPolicyProvider ?? throw new ArgumentNullException(nameof(corsPolicyProvider));
+            _sessionAffinityProviders = sessionAffinityProviders?.ToProviderDictionary() ?? throw new ArgumentNullException(nameof(sessionAffinityProviders));
+            _affinityFailurePolicies = affinityFailurePolicies?.ToPolicyDictionary() ?? throw new ArgumentNullException(nameof(affinityFailurePolicies));
+        }
+
+        // Note this performs all validation steps without short circuiting in order to report all possible errors.
+        public async Task<IList<Exception>> ValidateRouteAsync(ProxyRoute route)
```
davidfowl, line 74: ValueTask

src/ReverseProxy/Service/Config/DynamicConfigBuilder.cs
davidfowl: Does this type need to exist then? It seems like layer that runs validation and returns cloned validated inputs. I think `DynamicConfigRoot` and `IDynamicConfigBuilder` as a concept needs to be removed

davidfowl:
Can we rename Config -> Configuration? That's been bugging me

src/ReverseProxy/Abstractions/Config/IDynamicConfigBuilder.cs
```diff
@@ -16,6 +17,6 @@ internal interface IDynamicConfigBuilder
         /// <summary>
         /// Creates a <see cref="DynamicConfigRoot"/> object representing the current desired proxy dynamic configurations.
         /// </summary>
-        Task<DynamicConfigRoot> BuildConfigAsync(CancellationToken cancellation);
+        Task<DynamicConfigRoot> BuildConfigAsync(IReadOnlyList<ProxyRoute> routes, IReadOnlyList<Cluster> clusters, CancellationToken cancellation);
```
davidfowl, line 20: This doesn't seem like a replaceable service

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
```diff
@@ -71,5 +71,71 @@ ProxyRoute IDeepCloneable<ProxyRoute>.DeepClone()
                 Transforms = Transforms?.Select(d => new Dictionary<string, string>(d, StringComparer.OrdinalIgnoreCase)).ToList<IDictionary<string, string>>(),
             };
         }
+
+        // Used to diff for config changes
+        internal int GetConfigHash()
+        {
+            var hash = 0;
```
davidfowl, line 78: Not for this PR but we should be using HashCode to combine the hashcode here.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -6,51 +6,122 @@
...
-        public ReverseProxyConfigManager(
-            ILogger<ReverseProxyConfigManager> logger,
+        private List<Endpoint> _endpoints = new List<Endpoint>(0);
+        private CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
+        private IChangeToken _changeToken;
+
+        public ProxyConfigManager(
+            ILogger<ProxyConfigManager> logger,
+            IProxyConfigProvider provider,
             IDynamicConfigBuilder configBuilder,
```
davidfowl, line 46: nuke this and do it all in here.

src/ReverseProxy/Service/Config/ConfigValidator.cs
```diff
@@ -0,0 +1,263 @@
...
+
+            if (string.IsNullOrEmpty(cluster.SessionAffinity.Mode))
+            {
+                cluster.SessionAffinity.Mode = SessionAffinityConstants.Modes.Cookie;
+            }
+
+            var affinityMode = cluster.SessionAffinity.Mode;
+            if (!_sessionAffinityProviders.ContainsKey(affinityMode))
+            {
+                errors.Add(new ArgumentException($"No matching ISessionAffinityProvider found for the session affinity mode `{affinityMode}` set on the cluster `{cluster.Id}`."));
```
davidni, line 248: nitL use `nameof(ISessionAffinityProvider)` to improve resiliency against future refactorings.

src/ReverseProxy/Service/Config/ConfigValidator.cs
```diff
@@ -0,0 +1,263 @@
...
+            }
+
+            if (string.IsNullOrEmpty(cluster.SessionAffinity.FailurePolicy))
+            {
+                cluster.SessionAffinity.FailurePolicy = SessionAffinityConstants.AffinityFailurePolicies.Redistribute;
```
davidni, line 253: ditto, shouldn't mutate objects being validated.

src/ReverseProxy/Service/Config/RuntimeRouteBuilder.cs
```diff
@@ -64,19 +63,19 @@ public RouteConfig Build(ParsedRoute source, ClusterInfo cluster, RouteInfo runt
 
             // TODO: Handle arbitrary AST's properly
```
davidni, line 64: You can remove this comment about handling arbitrary AST's

src/ReverseProxy/Service/Config/RuntimeRouteBuilder.cs
```diff
@@ -64,19 +63,19 @@ public RouteConfig Build(ParsedRoute source, ClusterInfo cluster, RouteInfo runt
...
             // Catch-all pattern when no path was specified
-            var pathPattern = string.IsNullOrEmpty(source.Path) ? "/{**catchall}" : source.Path;
+            var pathPattern = string.IsNullOrEmpty(source.Match.Path) ? "/{**catchall}" : source.Match.Path;
 
             // TODO: Propagate route priority
             var endpointBuilder = new AspNetCore.Routing.RouteEndpointBuilder(
-                requestDelegate: _pipeline ?? Invoke,
+                requestDelegate: _pipeline,
                 routePattern: AspNetCore.Routing.Patterns.RoutePatternFactory.Parse(pathPattern),
                 order: 0);
```
davidni, line 72: Sugg renaming the route field `Priority` to `Order`, and setting it here, thus resolving the TODO in line 68 as well.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -6,51 +6,122 @@
...
+
+            IProxyConfig newConfig;
+            try
+            {
+                newConfig = manager._provider.GetConfig();
+            }
+            catch (Exception ex)
+            {
+                Log.ErrorReloadingConfig(manager._logger, ex);
+                // If we can't load the config then we can't listen for changes anymore.
```
davidni, line 107: This is catastrophic, shouldn't we let it all explode and crash in this case?

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -6,51 +6,122 @@
...
+                await manager.ApplyConfigAsync(newConfig);
+            }
+            catch (Exception ex)
+            {
+                Log.ErrorApplyingConfig(manager._logger, ex);
+            }
+
+            if (newConfig.ChangeToken.ActiveChangeCallbacks)
+            {
+                manager._changeSubscription?.Dispose();
```
davidni, line 122: Shouldn't this be done sooner, before any of the possible early bail outs? Or alternatively in a finally block?

test/ReverseProxy.Tests/Common/InMemoryConfigProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+    }
+}
+
+namespace Microsoft.ReverseProxy.Configuration
+{
+    public class InMemoryConfigProvider : IProxyConfigProvider
+    {
+        private InMemoryConfig _config;
+
+        public InMemoryConfigProvider(IReadOnlyList<ProxyRoute> routes, IReadOnlyList<Cluster> clusters)
```
davidni, line 29: Null checks?

test/ReverseProxy.Tests/Utilities/TaskExtensions.cs
```diff
@@ -16,6 +17,80 @@ namespace Microsoft.ReverseProxy.Utilities
...
+
+            var cts = new CancellationTokenSource();
+            if (task == await Task.WhenAny(task, Task.Delay(timeout, cts.Token)))
+            {
+                cts.Cancel();
+                return await task;
+            }
+            else
+            {
+                throw new TimeoutException(CreateMessage(timeout, filePath, lineNumber));
```
davidni, line 61: `task` will go unobserved, sugg. setting up a dummy continuation to avoid spurious errors downstream.

test/ReverseProxy.Tests/Service/Config/ConfigValidatorTests.cs
```diff
@@ -0,0 +1,443 @@
...
+            var result = await validator.ValidateRouteAsync(route);
+
+            Assert.NotEmpty(result);
+            Assert.Contains(result, err => err.Message.Equals("Missing Route Id."));
+        }
+
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        [InlineData("example.com,")]
```
davidni, line 88: Shouldn't this be valid?

src/ReverseProxy/Abstractions/Config/IProxyConfigManager.cs
```diff
@@ -1,20 +1,19 @@
...
     /// </summary>
-    public interface IReverseProxyConfigManager
+    internal interface IProxyConfigManager
     {
         /// <summary>
-        /// Applies latest configurations obtained from <see cref="IDynamicConfigBuilder"/>.
+        /// Load the first data at startup. May throw.
         /// </summary>
-        Task ApplyConfigurationsAsync(CancellationToken cancellation);
+        Task<EndpointDataSource> LoadAsync();
```
davidni, line 17: Sugg `InitialLoadAsync` to make intent clearer.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -6,51 +6,122 @@
...
+            // We intend this to crash the app so we don't try listening for further changes.
+            var config = _provider.GetConfig();
+            await ApplyConfigAsync(config);
+
+            if (config.ChangeToken.ActiveChangeCallbacks)
+            {
+                _changeSubscription = config.ChangeToken.RegisterChangeCallback(ReloadConfigAsync, this);
+            }
+
+            return this;
```
davidni, line 83: Sugg wrapping in a try + catch + rethrow more specific exception, plus logging a Critical error

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -6,51 +6,122 @@
...
-            var config = await _configBuilder.BuildConfigAsync(cancellation);
+            var dynamicConfig = await _configBuilder.BuildConfigAsync(config.Routes, config.Clusters, default);
 
-            UpdateRuntimeClusters(config);
-            UpdateRuntimeRoutes(config);
+            UpdateRuntimeClusters(dynamicConfig);
+            UpdateRuntimeRoutes(dynamicConfig);
+        }
+
+        private static async void ReloadConfigAsync(object state)
```
davidni, line 95: I understand `LoadAsync` and `ReloadConfigAsync` have different exception handling semantics, but would be nice if they could share code. Consider refactoring so that the core functionality is shared: getting configs, applying them, and subscribing to changes.

src/ReverseProxy/Service/Config/DynamicConfigBuilder.cs
Tratcher: Yeah, I was just looking at consolidating it since it's lost 2/3 of its code in this PR already.

Tratcher:
> Can we rename Config -> Configuration? That's been bugging me

I'd like to find a different name entirely that doesn't easily confuse with IConfiguration. Ideas?

samples/ReverseProxy.Config.Sample/Properties/launchSettings.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "profiles": {
+    "ReverseProxy.Sample": {
+      "commandName": "Project",
+      "launchBrowser": false,
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      }
+    },
+    "IIS Express": {
```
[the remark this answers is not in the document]
Tratcher: Yes, YARP works fine with IIS Express.

test/ReverseProxy.FunctionalTests/Common/InMemoryConfigProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+    }
+}
+
+namespace Microsoft.ReverseProxy.Configuration
+{
+    public class InMemoryConfigProvider : IProxyConfigProvider
+    {
+        private InMemoryConfig _config;
+
+        public InMemoryConfigProvider(IReadOnlyList<ProxyRoute> routes, IReadOnlyList<Cluster> clusters)
```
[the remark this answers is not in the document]
Tratcher: Nulls are fine. I've added some tests that make sure we handle null inputs from the provider.

src/ReverseProxy/Configuration/ConfigurationOptions.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.ReverseProxy.Abstractions;
+
+namespace Microsoft.ReverseProxy.Configuration
+{
+    internal class ConfigurationOptions
```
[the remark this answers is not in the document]
Tratcher: I'd like to avoid prefixing every internal class with "Proxy", "Proxy" is already in the namespace.

We also need some new naming to distinguish between IConfiguration related scenarios and the general proxy configuration.

src/ReverseProxy/Service/Config/ConfigValidator.cs
```diff
@@ -0,0 +1,263 @@
...
+            var errors = new List<Exception>();
+
+            if (string.IsNullOrEmpty(route.RouteId))
+            {
+                errors.Add(new ArgumentException("Missing Route Id."));
+            }
+
+            if ((route.Match.Hosts == null || route.Match.Hosts.Count == 0 || route.Match.Hosts.Any(host => string.IsNullOrEmpty(host))) && string.IsNullOrEmpty(route.Match.Path))
+            {
+                errors.Add(new ArgumentException($"Route `{route.RouteId}` requires Hosts or Path specified. Set the Path to `/{{**catchall}}` to match all requests."));
```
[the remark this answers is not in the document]
Tratcher: Hmm, it's in a lot of the error and log messages. I'll fix it.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -3,66 +3,254 @@
...
+
+            if (config.ChangeToken.ActiveChangeCallbacks)
+            {
+                _changeSubscription = config.ChangeToken.RegisterChangeCallback(ReloadConfigAsync, this);
+            }
+
+            return this;
+        }
+
+        private static async void ReloadConfigAsync(object state)
```
davidfowl, line 90: I just noticed this is async void. This isn't the right way to fix this. We should be using a bounded channel with a size of 1. I have that code locally.

src/ReverseProxy/Service/Config/RuntimeRouteBuilder.cs
Tratcher: Makes sense, but I'll separate that item. https://github.com/microsoft/reverse-proxy/issues/358

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: That doesn't seem like the kind of choice a library should be making for you. @samsp-msft ?

test/ReverseProxy.Tests/Utilities/TaskExtensions.cs
Tratcher: We've been using these test helpers elsewhere for quite a while without issue. I'm going to leave it unless we see an issue.

test/ReverseProxy.Tests/Service/Config/ConfigValidatorTests.cs
Tratcher: It's invalid in this case because it gets split into "example.com" and "" where the empty host is invalid.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: Our framework guidelines are to log or throw, not both. Otherwise the same exception will probably be logged multiple times.

I can add an exception wrapper so you get consistent root error message.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: The two are small enough and right next to each other that I'm not too worried about them getting out of sync. And as you said, the behavior is actually quite different.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: To fix the overlapped config reload issue you mean? That's not an issue because with IChangeTokens we have to re-subscribe after every config load. It's impossible to have an overlapped reload if we don't check for the next one until we're done with the current one.

The async void is because we have an async method triggered from a sync notification.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
davidni: I'd expect the library to either (a) give me chance to recover (i.e. make it possible to react to future updates); or (b) fail fast, not pretending everything is fine. Not doing either of these seems error prone and makes this exceedingly hard to detect in a production environment.

src/ReverseProxy/Service/RuntimeModel/RouteConfig.cs
```diff
@@ -51,7 +50,7 @@ internal sealed class RouteConfig
 
         public Transforms Transforms { get; }
 
-        public bool HasConfigChanged(ParsedRoute newConfig, ClusterInfo cluster)
+        public bool HasConfigChanged(ProxyRoute newConfig, ClusterInfo cluster)
         {
             return Cluster != cluster
                 || !ConfigHash.Equals(newConfig.GetConfigHash());
```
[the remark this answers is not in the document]
Tratcher: I think your only alternative is to hold onto the original ProxyRoute used to create this RouteConfig and do a strong Equals check against it. Anything else would be trying to reverse engineer the ProxyRoute properties from RouteConfig properties.

Filed https://github.com/microsoft/reverse-proxy/issues/359.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
davidfowl: It can overlap because its async void.

Tratcher: The registration for the next callback happens within this method, at the end after we've processed the new config. It doesn't matter that it's async void, the execution is strictly linear.
- Load the first config, register for callback
- Reload config, register for callback
- Reload config, register for callback
...

davidfowl: I just noticed this code isn't using OnChange anymore.

davidfowl: Still, I'm going to remove this async void if you don't in this PR 😄

Tratcher: I'm curious to see what you have in mind. Share a gist?

davidfowl:
```C#
_changeSubscription = config.ChangeToken.RegisterChangeCallback(() => _ = ReloadConfigAsync(), this);
```

Tratcher: Oh it's not actually different, you just really wanted the async void gone. I can respect that 😁 .

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: A few points:
A) Yes it's bad if the reloads ever stop happening.
B) It should be pretty easy for IProxyConfigProvider implementations to avoid this situation. They're not supposed to signal the prior notification until everything is ready, all GetConfig has to do is return pre-prepped data.
C) We don't have to make a decision about this right now. I'll file an follow-up issue and we can work out how to best handle catastrophic failures.

test/ReverseProxy.FunctionalTests/Common/InMemoryConfigProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            _config = new InMemoryConfig(routes, clusters);
+        }
+
+        public IProxyConfig GetConfig() => _config;
+
+        public void Update(IReadOnlyList<ProxyRoute> routes, IReadOnlyList<Cluster> clusters)
+        {
+            var oldConfig = _config;
+            _config = new InMemoryConfig(routes, clusters);
```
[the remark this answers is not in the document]
Tratcher: I'll make the field volatile so it's always current.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: https://github.com/microsoft/reverse-proxy/issues/368

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -3,66 +3,266 @@
...
+        public override IChangeToken GetChangeToken() => Volatile.Read(ref _changeToken);
+
+        // IProxyConfigManager
+
         /// <inheritdoc/>
-        public async Task ApplyConfigurationsAsync(CancellationToken cancellation)
+        public async Task<EndpointDataSource> InitialLoadAsync()
+        {
+            // Trigger the first load immediately and throw if it fails.
+            // We intend this to crash the app so we don't try listening for further changes.
```
alnikola, line 78: "listening for" -> "listening to"?
```suggestion
            // We intend this to crash the app so we don't try listening to further changes.
```

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -3,66 +3,266 @@
...
+
+            IProxyConfig newConfig;
+            try
+            {
+                newConfig = _provider.GetConfig();
+            }
+            catch (Exception ex)
+            {
+                Log.ErrorReloadingConfig(_logger, ex);
+                // If we can't load the config then we can't listen for changes anymore.
```
alnikola, line 115:
```suggestion
                // If we can't load the config then we can't listen to changes anymore.
```

alnikola (approved):
LGTM besides wording related comments.

docs/docfx/articles/configproviders.md
```diff
@@ -0,0 +1,141 @@
...
+If the `IChangeToken` supports `ActiveChangeCallbacks`, once the proxy has processed the initial set of configuration it will register a callback with this token. Note the proxy does not support polling for changes.
+
+When the provider wants to provide new configuration to the proxy it should first load that configuration in the background, optionally validate it using the [IConfigValidator](xref:Microsoft.ReverseProxy.Service.IConfigValidator), and only then signal the `IChangeToken` from the prior `IProxyConfig` instance that new data is available. The proxy will call `GetConfig()` again to retrieve the new data.
+
+There are important differences when reloading configuration vs the first configuration load.
+- The new configuration will be diffed against the current one and only modified routes or clusters will be updated. The update will be applied atomically and will only affect new requests, not request currently in progress.
+- Any errors in the reload process will be logged and the application to continue using the last known good configuration.
+- If `GetConfig()` throws the proxy will be unable to listen for future changes because `IChangeToken`s are single use.
```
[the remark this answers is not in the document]
Tratcher: listen for is correct here.

docs/docfx/articles/configproviders.md
```diff
@@ -0,0 +1,141 @@
...
+        {
+            var routes = new[]
+            {
+                new ProxyRoute()
+                {
+                    RouteId = "route1",
+                    ClusterId = "cluster1",
+                    Match =
+                    {
+                        Path = "{**catch-all}"
```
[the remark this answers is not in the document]
Tratcher: catch-all is the term used in the routing docs.
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.1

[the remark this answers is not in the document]
alnikola: Should then all occurrences of "catchall" be changed to "catch-all" them? There is a number of them. [Example](https://github.com/microsoft/reverse-proxy/blob/47c6a6bdcc19e843820bdc4330f940501bacf78f/src/ReverseProxy/Service/Config/ConfigValidator.cs#L86).

PR #361, opened 2020-08-07
Fix policy configuration example.

ericwj:
```
error CS1061: 'IList<IAuthorizationRequirement>' does not contain a definition for 'RequireAuthenticatedUser' and no accessible extension method 'RequireAuthenticatedUser' accepting a first argument of type 'IList<IAuthorizationRequirement>' could be found (are you missing a using directive or an assembly reference?)
```


Tratcher:
Thanks

ericwj:
Thank you instead! Love this project.

Issue #359, opened 2020-08-07
Checking for route/cluster changes


Tratcher:
The current pattern for checking for route changes saves a hash of the old ProxyRoute and compares it to a hash of the new ProxyRoute instance. That's not a strong equality check, nor very efficient for negative cases.
https://github.com/microsoft/reverse-proxy/blob/444e7aa5f377363f1842223515a710e3cff14630/src/ReverseProxy/Service/RuntimeModel/RouteConfig.cs#L54-L58

Proposal:
Save the original ProxyRoute in a private field and implement Equals on ProxyRoute.

Do something similar for Cluster.

PR #366, opened 2020-08-11
Implement Equals for ProxyRoute & Cluster

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/359
Fixes https://github.com/microsoft/reverse-proxy/issues/373

Since `ParsedRoute` type is removed by https://github.com/microsoft/reverse-proxy/pull/357, I create a branch of the PR in order to prevent future merge conflicts. So this could be merged after https://github.com/microsoft/reverse-proxy/pull/357.

### DIFF
--- a/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs
+++ b/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs
@@ -102,6 +102,7 @@ namespace Microsoft.ReverseProxy.Abstractions
                 && LoadBalancingOptions.Equals(cluster1.LoadBalancing, cluster2.LoadBalancing)
                 && SessionAffinityOptions.Equals(cluster1.SessionAffinity, cluster2.SessionAffinity)
                 && HealthCheckOptions.Equals(cluster1.HealthCheck, cluster2.HealthCheck)
+                && ProxyHttpClientOptions.Equals(cluster1.HttpClient, cluster2.HttpClient)
                 && CaseInsensitiveEqualHelper.Equals(cluster1.Destinations, cluster2.Destinations, Destination.Equals)
                 && CaseInsensitiveEqualHelper.Equals(cluster1.Metadata, cluster2.Metadata);
         }
--- a/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs
+++ b/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.ReverseProxy.Abstractions.ClusterDiscovery.Contract;
+using Microsoft.ReverseProxy.Utilities;
 
 namespace Microsoft.ReverseProxy.Abstractions
 {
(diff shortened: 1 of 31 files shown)


Tratcher:
#357 has been merged, please rebase.

src/ReverseProxy/Utilities/EqualHelper.cs
```diff
@@ -0,0 +1,124 @@
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.ReverseProxy.Utilities
+{
+    internal class EqualHelper
```
halter73, line 6: Maybe name this CaseInsensitiveEqualHelper to ensure this never gets used when capitalization matters.

src/ReverseProxy/Utilities/CaseInsensitiveEqualHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        internal static bool Equals(IList<IDictionary<string, string>> list1, IList<IDictionary<string, string>> list2)
+        {
+            if (list1 == null && list2 == null)
```
halter73, line 43: Nit:

```suggestion
            if ((list1?.Count ?? 0) == 0 && (list2?.Count ?? 0) == 0)
```

src/ReverseProxy/Service/RuntimeModel/ClusterConfig.cs
```diff
@@ -110,7 +114,12 @@ public ClusterSessionAffinityOptions(bool enabled, string mode, string failurePo
 
             public string FailurePolicy { get; }
 
-            public IReadOnlyDictionary<string, string> Settings { get;  }
+            public IReadOnlyDictionary<string, string> Settings { get; }
+        }
+
+        internal bool HasConfigChanged(Cluster configCluster)
```
halter73, line 120: Where is this being called from?

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs
```diff
@@ -76,5 +77,23 @@ Cluster IDeepCloneable<Cluster>.DeepClone()
                 Metadata = Metadata?.DeepClone(StringComparer.OrdinalIgnoreCase),
             };
         }
+
+        internal bool Equals(Cluster cluster)
+        {
+            return string.Equals(Id, cluster.Id, StringComparison.OrdinalIgnoreCase)
+                && CircuitBreakerOptions.Equals(cluster.CircuitBreakerOptions)
```
Tratcher, line 84: If CircuitBreakerOptions on this object is null then this will throw a NullReferenceException right?

Are there tests for these new methods?

Tratcher: Also, make sure the other `cluster` isn't null.

src/ReverseProxy/Service/RuntimeModel/RouteConfig.cs
```diff
@@ -52,8 +51,8 @@ internal sealed class RouteConfig
 
         public bool HasConfigChanged(ProxyRoute newConfig, ClusterInfo cluster)
         {
-            return Cluster != cluster
-                || !ConfigHash.Equals(newConfig.GetConfigHash());
+            return !ClusterInfo.Equals(Cluster, cluster)
```
Tratcher, line 54: We don't need a deep equality check on the cluster here, a reference equality check is enough to know that we're pointing at the same cluster. When config changes the ClusterInfo is preserved and only the ClusterConfig is updated. See
https://github.com/microsoft/reverse-proxy/blob/8b9fa3d97200138096887e9014633c8f3f6700dc/src/ReverseProxy/Service/Management/ProxyConfigManager.cs#L323

We don't need to rebuild all the routes in that case, they're still pointing at a valid ClusterInfo.

Tratcher (approved):

Tratcher:
Thanks @Kahbazi

PR #376, opened 2020-08-25
Change the doc publishing to track a release branch

### LINKED ISSUE #344: Create versioned docs
Have different versions of the docs for different releases.

We're making breaking changes between releases. As part of that we try to update the docs in the same PR, but that means the docs are out of sync with the released version until the next release.

Proposal: Have two branches for the docs, one based on the current release and one based on master. https://microsoft.github.io/reverse-proxy/ should point at the current release.
Or does docfx have built in versioning?

Tratcher:
Fixes #344 Our docs live in the repo and get updated live with commits to master. However, when we make breaking changes the published docs get out of sync with the last release (e.g. preview4) causing confusion for some users.

I've changed the doc publishing process to explicitly track a release/docs branch. See the updated readme.

### DIFF
--- a/docs/docfx/readme.md
+++ b/docs/docfx/readme.md
@@ -12,4 +12,8 @@ The build will produce a series of HTML files in the `_site` directory. Many of 
 
 ## Publishing the docs
 
-The docs are automatically built and published by a [GitHub Action](https://github.com/microsoft/reverse-proxy/blob/master/.github/workflows/docfx_build.yml) on every push to `master`. The built `_site` directory is pushed to the `gh-pages` branch and served by [https://microsoft.github.io/reverse-proxy/](https://microsoft.github.io/reverse-proxy/).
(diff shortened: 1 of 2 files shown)


docs/docfx/readme.md
```diff
@@ -12,4 +12,8 @@ The build will produce a series of HTML files in the `_site` directory. Many of
 
 ## Publishing the docs
 
-The docs are automatically built and published by a [GitHub Action](https://github.com/microsoft/reverse-proxy/blob/master/.github/workflows/docfx_build.yml) on every push to `master`. The built `_site` directory is pushed to the `gh-pages` branch and served by [https://microsoft.github.io/reverse-proxy/](https://microsoft.github.io/reverse-proxy/).
+The docs are automatically built and published by a [GitHub Action](https://github.com/microsoft/reverse-proxy/blob/master/.github/workflows/docfx_build.yml) on every push to `release/docs`. The built `_site` directory is pushed to the `gh-pages` branch and served by [https://microsoft.github.io/reverse-proxy/](https://microsoft.github.io/reverse-proxy/). Maintaining a seperate branch for the released docs allows us to choose when to publish them and with what content, and without modifying the build scripts each release.
```
halter73, line 15: What's the motivation behind move docs to their own branch? Keeping the docs in the same branch as code seems really nice for keeping the docs and code in sync. Are we planning on documenting multiple versions simultaneously in this branch?

Tratcher: The docs are staying in the master branch with the code. This is only about what shows on the website. I want the website to show the docs for the last release preview rather than for the nightly state.

Issue #138, opened 2020-05-04
Proxy error handling


Tratcher:
### Describe the bug
The proxy should catch exceptions from the HttpProxy and convert them to 503s or other statuses as appropriate. This also ties into our [retry](https://github.com/microsoft/reverse-proxy/issues/56) middleware thoughts. Regardless, a common exception like this shouldn't reach Kestrel (which would turn it into a 500).

### To Reproduce
Proxy to a server that isn't currently running.
```
info: Microsoft.ReverseProxy.Core.Middleware.ProxyInvokerMiddleware[9]
      Proxying to https://localhost:10000/api/dump
info: Microsoft.ReverseProxy.Core.Middleware.ProxyInvokerMiddleware[31]
      Operation started: ReverseProxy.Proxy
info: Microsoft.ReverseProxy.Core.Middleware.ProxyInvokerMiddleware[33]
      Operation ended: ReverseProxy.Proxy, 4113.4367ms, error: No connection could be made because the target machine actively refused it.
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[1]
      Executed endpoint 'backend1/route1'
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HLVGE3ID4IFL", Request id "0HLVGE3ID4IFL:00000001": An unhandled exception was thrown by the application.
System.Net.Http.HttpRequestException: No connection could be made because the target machine actively refused it.
 ---> System.Net.Sockets.SocketException (10061): No connection could be made because the target machine actively refused it.
   at System.Net.Http.ConnectHelper.ConnectEventArgs.OnCompleted(SocketAsyncEventArgs _)
   at System.Net.Sockets.MultipleConnectAsync.AsyncFail(Exception e)
   at System.Net.Sockets.MultipleConnectAsync.InternalConnectCallback(Object sender, SocketAsyncEventArgs args)
   at System.Net.Sockets.SocketAsyncEventArgs.OnCompleted(SocketAsyncEventArgs e)
   at System.Net.Sockets.SocketAsyncEventArgs.ExecutionCallback(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
[...]

PR #406, opened 2020-09-08
Handling proxy errors

Tratcher:
https://github.com/microsoft/reverse-proxy/issues/138

- Restructure HttpProxy to unify the flow between normal and upgrade requests. Reduces code duplication.
- Catch exceptions from HttpMessageInvoker and response body copying. Set 502s.
- Use a feature to preserve the exception
- Catch exceptions from the request body. Attribute blame to either the client or destination.
- Log exceptions
- More granular exception handling by type. E.g. HttpRequestException vs OperationCancelledException.
- Having the ProxyInvokerMiddledware set the error feature with its own error conditions. Note there's no exception in this case.

New API:
```C#
    /// <summary>
    /// Stores errors that occurred when proxying the request to the destination.
    /// </summary>
    public interface IProxyErrorFeature
    {
        /// <summary>
        /// The specified ProxyError.
        /// </summary>
        public ProxyError Error { get; }

        /// <summary>
        /// An exception that occurred when proxying the request to the destination.
        /// </summary>
        public Exception Exception { get; }
    }
```

Also ProxyError with values like RequestBodyCanceled, RequestBodyClient, ResponseBodyDestination, UpgradeRequestClient, etc..


Tratcher:
Reviewers, please take a first look to make sure this method of error handling makes sense, and that I've covered the important classes of errors.

src/ReverseProxy/Service/Proxy/ProxyException.cs
```diff
@@ -0,0 +1,49 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.ReverseProxy.Service.Proxy
+{
+    /// <summary>
+    /// Used to report errors that occur while proxying a request to the destination.
+    /// </summary>
+    public class ProxyException : Exception
```
Tratcher, line 11: Note to self: replace this with the proxy error feature, it never gets thrown anyways.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -46,18 +46,49 @@ public async Task CopyAsync(Stream source, Stream destination, CancellationToken
             {
                 while (true)
                 {
-                    cancellation.ThrowIfCancellationRequested();
+                    if (cancellation.IsCancellationRequested)
+                    {
+                        return (StreamCopyResult.Canceled, new OperationCanceledException(cancellation));
```
Kahbazi, line 51: I think the `StackTrace` of this exception would be null. Is this an issue?

Tratcher: Yes it's null, but I don't expect that to cause any issues. That's possible on any exception so the consumer would have to guard against it.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -224,122 +186,42 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
                 await context.Response.CompleteAsync();
```
halter73, line 186: What if there was an error prior to writing to the response body? Would this prevent user middleware from changing the status code from a 502 (or whatever) to something custom?

src/ReverseProxy/Service/Proxy/ProxyErrorCodes.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        /// <summary>
+        /// No error.
+        /// </summary>
+        None,
+
+        /// <summary>
+        /// Failed to connect or send the request.
+        /// </summary>
+        Request,
```
halter73, line 19: > Failed to connect or send the request.

Isn't this also the ProxyError used if the request is successfully sent, but the server closes/resets the connection without sending a response status line and/or headers? I get that's it's hard to disambiguate, but we should probably at least mention that in the doc comment.

halter73 (commented):
Looking good. Mostly nits.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -84,134 +84,96 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
...
+            HttpResponseMessage destinationResponse;
+            try
+            {
+                destinationResponse = await httpClient.SendAsync(destinationRequest, shortCancellation);
+            }
+            catch (OperationCanceledException oex)
+            {
+                // We're not sure if this was canceled by a timeout or a client disconnect.
+                ReportProxyError(context, ProxyErrorCode.RequestCanceled, oex);
+                context.Response.StatusCode = StatusCodes.Status502BadGateway;
```
halter73, line 118: How would a client disconnect lead to an OperationCanceledException?

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -677,6 +788,34 @@ public static void Proxying(ILogger logger, string targetUrl)
...
+            {
+                _proxyError(logger, errorCode, GetMessage(errorCode), ex);
+            }
+
+            private static string GetMessage(ProxyErrorCode errorCode)
+            {
+                return errorCode switch
+                {
+                    ProxyErrorCode.None => throw new NotSupportedException("A more specific error code must be used"),
+                    ProxyErrorCode.Request => "An error was encountered when sending the request to the destination.",
```
halter73, line 802:
```suggestion
                    ProxyErrorCode.Request => "An error was encountered before receiving a response.",
```

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -677,6 +788,34 @@ public static void Proxying(ILogger logger, string targetUrl)
...
+                    ProxyErrorCode.RequestCanceled => "The request was canceled while sending it to the destination.",
```
halter73, line 803:
```suggestion
                    ProxyErrorCode.RequestCanceled => "The request was canceled before receiving a response.",
```

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -677,6 +788,34 @@ public static void Proxying(ILogger logger, string targetUrl)
...
+                    ProxyErrorCode.ResponseBodyCanceled => "Copying the response body was canceled.",
+                    ProxyErrorCode.ResponseBodyClient => "The client reported an error when copying the response body.",
+                    ProxyErrorCode.ResponseBodyDestination => "The destination reported an error when copying the response body.",
+                    ProxyErrorCode.UpgradeRequestCanceled => "Copying the upgraded request body was canceled.",
+                    ProxyErrorCode.UpgradeRequestClient => "The client reported an error when copying the upgraded request body.",
+                    ProxyErrorCode.UpgradeRequestDestination => "The destination reported an error when copying the upgraded request body.",
+                    ProxyErrorCode.UpgradeResponseCanceled => "Copying the upgraded response body was canceled.",
+                    ProxyErrorCode.UpgradeResponseClient => "The client reported an error when copying the upgraded response body.",
+                    ProxyErrorCode.UpgradeResponseDestination => "The destination reported an error when copying the upgraded response body.",
```
halter73, line 815: Is it time to add a Resources.resx?

src/ReverseProxy/Service/Proxy/ProxyErrorCodes.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <summary>
+        /// Canceled when trying to connection or send the request.
+        /// </summary>
```
halter73, line 23: Ditto about updating the comment to indicate it's any cancellation before receiving a response.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -84,134 +84,96 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
...
+            // :: Step 5: Copy response status line Client ◄-- Proxy ◄-- Destination
+            // :: Step 6: Copy response headers Client ◄-- Proxy ◄-- Destination
+            CopyResponseStatusAndHeaders(destinationResponse, context, transforms.ResponseHeaderTransforms);
 
-            // :::::::::::::::::::::::::::::::::::::::::::::
-            // :: Step 6: Copy response headers Downstream ◄-- Proxy ◄-- Upstream
-            CopyHeadersToDownstream(upstreamResponse, context, transforms.ResponseHeaderTransforms);
+            // Note as of 5.0 HttpResponse.Content never returns null.
+            // https://github.com/dotnet/runtime/blame/8fc68f626a11d646109a758cb0fc70a0aa7826f1/src/libraries/System.Net.Http/src/System/Net/Http/HttpResponseMessage.cs#L46
+            if (destinationResponse.StatusCode == HttpStatusCode.SwitchingProtocols && destinationResponse.Content != null)
```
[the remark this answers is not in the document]
Tratcher: I'll make it throw.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -534,8 +420,70 @@ static void AddHeader(HttpRequestMessage request, string headerName, StringValue
...
         }
 
-        private void CopyHeadersToDownstream(HttpResponseMessage source, HttpContext context, IReadOnlyDictionary<string, ResponseHeaderTransform> transforms)
+        private void HandleRequestBodyFailure(HttpContext context, StreamCopyResult requestBodyCopyResult, Exception requestBodyError, Exception additionalError)
         {
+            ProxyErrorCode requestBodyErrorCode;
+            int statusCode;
+            switch (requestBodyCopyResult)
+            {
+                // Failed while trying to copy the request body from the client. It's ambiguous if the request or response failed first.
```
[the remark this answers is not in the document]
Tratcher: If we're concurrently reading request data from the client and writing response data to the client and the connection aborts, then both the read and write operations can fail at about the same time. From our perspective it's ambiguous which failed first, and it doesn't really matter. We're in a similar situation on the destination side.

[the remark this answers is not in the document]
halter73: Ah. I understood this to mean that we cannot distinguish between the client aborting the request or the destination aborting the response. As long as we're correctly assigning the blame to the client or destination (and it looks like we are), that's fine. You're right that when a socket closes on a potentially full-duplex stream it's hard to distinguish which direction failed first.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -224,122 +186,42 @@ public HttpProxy(ILogger<HttpProxy> logger, ProxyMetrics metrics)
...
             if (destinationAddress == null || destinationAddress.Length < 8)
             {
                 throw new ArgumentException(nameof(destinationAddress));
             }
 
+            // Default to HTTP/1.1 for proxying upgradeable requests. This is already the default as of .NET Core 3.1
+            // Otherwise request HTTP/2 and let HttpClient fallback to HTTP/1.1 if it cannot establish HTTP/2 with the target.
+            // This is done without extra round-trips thanks to ALPN. We can detect a downgrade after calling HttpClient.SendAsync
+            // (see Step 3 below). TBD how this will change when HTTP/3 is supported.
+            var httpVersion = isUpgradeRequest ? ProtocolHelper.Http11Version : ProtocolHelper.Http2Version;
```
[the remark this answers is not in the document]
Tratcher: The HTTP version is the main check HttpClient does since it's related to the connection establishment.

HttpClient only has one upgrade check, and it's on the response when checking the 101:
https://github.com/dotnet/runtime/blob/7600720a60b9b1146807c4a71b104f98ed10bef1/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnection.cs#L651-L654

That code doesn't exist in the HTTP/2 code path.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -534,8 +420,70 @@ static void AddHeader(HttpRequestMessage request, string headerName, StringValue
...
+                // Failed while trying to copy the request body to the destination. It's ambiguous if the request or response failed first.
+                case StreamCopyResult.OutputError:
+                    requestBodyErrorCode = ProxyErrorCode.RequestBodyDestination;
+                    statusCode = StatusCodes.Status502BadGateway;
+                    break;
+                // Canceled while trying to copy the request body, either due to a client disconnect or a timeout. This probably caused the response to fail as a secondary error.
+                case StreamCopyResult.Canceled:
+                    requestBodyErrorCode = ProxyErrorCode.RequestBodyCanceled;
+                    // We don't use 504 timed out here because we can't tell why it was canceled.
+                    statusCode = StatusCodes.Status502BadGateway;
```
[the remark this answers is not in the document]
Tratcher: Cancellation could have been triggered by the RequestAborted token. There's no timeout on the long token, only the short one.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -534,8 +420,70 @@ static void AddHeader(HttpRequestMessage request, string headerName, StringValue
...
+                    throw new NotImplementedException(requestBodyCopyResult.ToString());
+            }
+
+            ReportProxyError(context, requestBodyErrorCode, new AggregateException(requestBodyError, additionalError));
+
+            // We don't know if the client is still around to see this error, but set it for diagnostics to see.
+            if (!context.Response.HasStarted)
+            {
+                // Nothing has been sent to the client yet, we can still send a good error response.
+                context.Response.Clear();
```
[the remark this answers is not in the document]
Tratcher: It's pretty basic and only works in limited scenarios.

src/ReverseProxy/Service/Proxy/ProxyErrorCodes.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <summary>
+        /// Failed when writing the upgraded response body to the client.
+        /// </summary>
+        UpgradeResponseClient,
+
+        /// <summary>
+        /// Failed when reading the upgraded response body from the destination.
+        /// </summary>
+        UpgradeResponseDestination,
```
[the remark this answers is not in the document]
Tratcher: Yeah, request/response is a little odd for upgraded connections. That said, I choose it because it's consistent with the request non-upgraded entries and it's unambiguous about the direction.  Input/Output is directionally ambiguous, both the client and destination have one of each. Other ideas?

src/ReverseProxy/Service/Proxy/HttpProxy.cs
Tratcher: The tokens are all linked to RequestAborted which fires for client disconnects. HttpProxy can't tell if the short token fires due to disconnect or timeout.

Tratcher: Hmm, maybe if it fires and we backsniff to see if RequestAborted has triggered or not? It seems like a layering issue, but plausible

src/ReverseProxy/Service/Proxy/HttpProxy.cs
Tratcher: Not if I can avoid it 😁. If we're not going to LOC then it's just overhead, especially when the project is churning.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
Tratcher: Ok, let me try reshuffling how we pass parameters into HttpProxy so it owns the timeout.

src/ReverseProxy/Service/Proxy/StreamCopyHttpContent.cs
```diff
@@ -119,29 +124,45 @@ protected override async Task SerializeToStreamAsync(Stream stream, TransportCon
...
+                // Some protocols (e.g. gRPC) may rely on specific bytes being sent, and HttpClient's buffering would prevent it.
+                // AutoFlushingStream delegates to the provided stream, adding calls to FlushAsync on every WriteAsync.
+                // Note that HttpClient does NOT call Flush on the underlying socket, so the perf impact of this is expected to be small.
+                // This statement is based on current knowledge as of .NET Core 3.1.201.
+                stream = new AutoFlushingStream(stream);
+            }
 
-                await _streamCopier.CopyAsync(_source, stream, _cancellation);
-                _tcs.TrySetResult(true);
+            // Immediately flush request stream to send headers
```
ManickaP, line 136: Is it still needed if https://github.com/dotnet/runtime/issues/860 is fixed?

src/ReverseProxy/Service/Proxy/StreamCopyHttpContent.cs
```diff
@@ -119,29 +124,45 @@ protected override async Task SerializeToStreamAsync(Stream stream, TransportCon
...
+            // https://github.com/dotnet/corefx/issues/39586#issuecomment-516210081
+            try
+            {
+                await stream.FlushAsync(_cancellation);
+            }
+            catch (OperationCanceledException oex)
+            {
+                _tcs.TrySetResult((StreamCopyResult.Canceled, oex));
```
ManickaP, line 144: Is not ending the method here intentional? Previously, an exception from `FlushAsync` would kill the method.

src/ReverseProxy/Service/Proxy/StreamCopyHttpContent.cs
```diff
@@ -119,29 +124,45 @@ protected override async Task SerializeToStreamAsync(Stream stream, TransportCon
...
+
+            var (result, error) = await _streamCopier.CopyAsync(_source, stream, _cancellation);
+            _tcs.TrySetResult((result, error));
+            // Check for errors that weren't the result of the destination failing.
+            // We have to throw something here so the transport knows the body is incomplete.
+            // We can't re-throw the original exception since that would cause concurrency issues.
+            // We need to wrap it.
+            if (result == StreamCopyResult.InputError)
+            {
+                throw new IOException(string.Empty, error);
```
ManickaP, line 159:
```suggestion
                throw new IOException(error.Message, error);
```

ManickaP (commented):
I checked only `StreamCopyHttpContent.cs`

src/ReverseProxy/Service/Proxy/StreamCopyHttpContent.cs
Tratcher: I think it's still needed for https://github.com/dotnet/runtime/issues/30295#issuecomment-516210081. In a bidirectional streaming scenario the request body might not be immediately available. Is it still the case that SocketHttpHandler doesn't automatically flush request headers until the request body is written? For HTTP/1.x or HTTP/2?

src/ReverseProxy/Service/Proxy/StreamCopyHttpContent.cs
Tratcher: That would duplicate the message. I'll put in a real one.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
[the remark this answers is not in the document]
halter73: Could we make this a 504 when `!RequestAborted.IsCancellationRequested`?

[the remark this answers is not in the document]
Tratcher: I've updated the comment.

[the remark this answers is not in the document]
Tratcher: Why? It shouldn't be possible for it to be cancelled in due to anything except the RequestAborted token. Note if SendAsync throws OperationCancelledException that's where I'm handling 504s.

src/ReverseProxy/Service/Proxy/ProxyErrorCodes.cs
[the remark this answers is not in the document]
halter73: In my mind, from the perspective to the proxy, client-to-destination data is input and destination-to-client data is output. Even from the perspective of the client or destination, it would be weird to call data going  to the client "input".

That said, this was just a nit. At least the meaning is still clear the way it currently is.

halter73 (approved):

src/ReverseProxy/Service/Proxy/HttpProxy.cs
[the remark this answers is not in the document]
halter73: Does that mean there's currently no way to timeout a request/response after response headers are received?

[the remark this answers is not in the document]
Tratcher: Only through the RequestAborted token. 

In the common case the request body will be fully uploaded before the response headers are received, so the request timeout will apply to that direction. For the response body you don't want a timeout as it doesn't scale to response sizes, but rather a data rate. (You may also want a data rate for the request body). Kestrel will enforce a minimum inbound and outbound data rate between the proxy and the client. We may need to add some between the proxy and the destination. I'll file something.

In the gRCP streaming and WebSocket sceanrios you don't want any timeouts or fixed rates.

src/ReverseProxy/Service/Proxy/StreamCopyHttpContent.cs
ManickaP: You're right, flush is not automatic between headers and content.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -46,18 +46,49 @@ public async Task CopyAsync(Stream source, Stream destination, CancellationToken
...
+                        read = await input.ReadAsync(buffer.AsMemory(), cancellation);
+                    }
+                    catch (OperationCanceledException oex)
+                    {
+                        return (StreamCopyResult.Canceled, oex);
+                    }
+                    catch (Exception ex)
+                    {
+                        return (StreamCopyResult.InputError, ex);
+                    }
```
davidni, line 67: You can avoid try/catch blocks within the loop entirely with a nifty trick, just replace the `iops` counter with two variables `readOps` and `writeOps`, incremented respectively before each readAsync and before each WriteAsync. Then in the uber catch block, `readOps == writeOps` indicates WriteAsync failed; otherwise, it was ReadAsync.

This would reduce code duplication, and may improve perf. it's what we did in our fork in Island Gateway.

davidni (approved):
Looks good and takes care of many code smells we used to have, thanks @Tratcher!

src/ReverseProxy/Service/Proxy/StreamCopier.cs
Tratcher: Do you have metrics for that? As I understand it try/catch blocks are cheep when there aren't any exceptions, it's the process of throwing an exception that's expensive.

src/ReverseProxy/Service/Proxy/ProxyError.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        /// <summary>
+        /// Failed writing the request body to the destination.
+        /// </summary>
+        RequestBodyDestination,
+
+        /// <summary>
+        /// Canceled while copying the response body.
+        /// </summary>
+        ResponseBodyCanceled,
```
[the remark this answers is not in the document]
Tratcher: They happen at different stages in the request lifecycle. E.g. RequestCanceled refers specifically to HttpMessageInvoker.SendAsync throwing and only happens prior to receiving the response headers. If you get ResponseBodyCanceled then you know you at least received the response headers.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -534,8 +449,71 @@ static void AddHeader(HttpRequestMessage request, string headerName, StringValue
...
+
+            ResetOrAbort(context, isCancelled: requestBodyCopyResult == StreamCopyResult.Canceled);
+        }
+
+        private async Task HandleRequestFailureAsync(HttpContext context, StreamCopyHttpContent requestContent, Exception requestException)
+        {
+            // Check for request body errors, these may have triggered the response error.
+            if (requestContent?.ConsumptionTask.IsCompleted == true)
+            {
+                var (requestBodyCopyResult, requestBodyException) = await requestContent.ConsumptionTask;
```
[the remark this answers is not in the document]
Tratcher: We're pretty explicit about avoiding Task.Result wherever possible. Let's hold off on task pattern optimizations until we run a profiler. As this is only an failure path I doubt it will register.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -547,24 +525,169 @@ private void CopyHeadersToDownstream(HttpResponseMessage source, HttpContext con
...
+            {
+                var error = result switch
+                {
+                    StreamCopyResult.InputError => ProxyError.UpgradeResponseDestination,
+                    StreamCopyResult.OutputError => ProxyError.UpgradeResponseClient,
+                    StreamCopyResult.Canceled => ProxyError.UpgradeResponseCanceled,
+                    _ => throw new NotImplementedException(result.ToString()),
+                };
+                ReportProxyError(context, error, exception);
+            }
```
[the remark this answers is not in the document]
Tratcher: Yeah, it's a bit complicated, but I think it's necessary if you want to fully attribute the failure to the right party. E.g. who faulted first and why.

The abort token is only triggered if one direction failed. It's possible for one direction to end gracefully and the connection to be in a half-closed state.

I tried re-working it, let me know what you think.

src/ReverseProxy/Middleware/ProxyInvokerMiddleware.cs
```diff
@@ -71,35 +72,30 @@ public async Task Invoke(HttpContext context)
                 throw new InvalidOperationException($"Chosen destination has no configs set: '{destination.DestinationId}'");
             }
 
-            using (var shortCts = CancellationTokenSource.CreateLinkedTokenSource(context.RequestAborted))
+            // TODO: Make this configurable on a route rather than create it per request?
```
davidfowl, line 75: Yes, why is this per request in the first place..

Tratcher: To limit scope creep on this PR. These options were not configurable before, and we have an issue to make them configurable like timeouts, but that's a non-trivial amount of additional work unrelated to everything else here.

The main reason I changed this now is to pass the timeout into the proxy rather than a cancellation token. Letting HttpProxy own the cancellation tokens helped it figure out when things timed out vs when they were canceled for other reasons.

PR #411, opened 2020-09-11
Removed '*Options' suffixes from 'Abstractions.Cluster' properties

### LINKED ISSUE #396: Remove '*Options' suffixes from 'Abstractions.Cluster' properties
There are inconsistently used '*Options' suffixes on [Cluster properties](https://github.com/microsoft/reverse-proxy/blob/master/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs#L26). It's suggested to remove them.

Marusyk:
Removed '*Options' suffixes 

Fixes #396

Please review
Thank you in advance

### DIFF
--- a/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs
+++ b/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs
@@ -23,17 +23,17 @@ namespace Microsoft.ReverseProxy.Abstractions
         /// <summary>
         /// Circuit breaker options.
         /// </summary>
-        public CircuitBreakerOptions CircuitBreakerOptions { get; set; }
+        public CircuitBreakerOptions CircuitBreaker { get; set; }
 
         /// <summary>
         /// Quota options.
         /// </summary>
-        public QuotaOptions QuotaOptions { get; set; }
+        public QuotaOptions Quota { get; set; }
 
         /// <summary>
         /// Partitioning options.
         /// </summary>
-        public ClusterPartitioningOptions PartitioningOptions { get; set; }
+        public ClusterPartitioningOptions Partitioning { get; set; }
 
         /// <summary>
         /// Load balancing options.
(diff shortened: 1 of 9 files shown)


src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs
```diff
@@ -23,17 +23,17 @@ public sealed class Cluster : IDeepCloneable<Cluster>
         /// <summary>
         /// Circuit breaker options.
         /// </summary>
-        public CircuitBreakerOptions CircuitBreakerOptions { get; set; }
+        public CircuitBreakerOptions CircuitBreaker { get; set; }
```
Tratcher, line 26: Do some of these sound odd in the singular?
```suggestion
        public CircuitBreakerOptions CircuitBreakers { get; set; }
```
Also HealthChecks, Quotas. The others seem ok.

Kahbazi:
I think the docs also needs to be changed.

https://github.com/microsoft/reverse-proxy/blob/87e463ca4e64aa1c1b48c13881a4e41bb334d954/docs/docfx/articles/proxyhttpclientconfig.md#L14

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/Cluster.cs
Marusyk: So, should I change it to CircuitBreakers, HealthChecks, and Quotas?

Tratcher: Yes, unless you can give a reason not to?

Marusyk: I would prefer to leave it as is.

CircuitBreaker - [Circuit breaker design pattern](https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern), but not Circuit breakers. 
HealthCheck - we set up options of one health check (timeout, interval etc.). There is no possibility to set up many checks, so, HealthCheck seems to be ok.
Quotas - don't know, can be changed

Tratcher: Fair enough. We'll do an end-to-end API review later in the release.

Tratcher (approved):

Tratcher:
Thanks

PR #414, opened 2020-09-14
Disable broken validation steps

Tratcher:
These two steps were broken when we moved to the preview8 SDK because the machines that run the post-build validation don't have the new version of VS yet. I'll file an issue to re-enable.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -169,6 +169,9 @@ stages:
       enableSymbolValidation: false
       # It's a private repo in github so this won't pass until we create an internal mirror
       enableSourceLinkValidation: false
+      # Broken by the move to the 5.0.0-preview8 SDK
+      enableSigningValidation: false
+      enableNugetValidation: false
       # This is to enable SDL runs part of Post-Build Validation Stage
       SDLValidationParameters:
         enable: true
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -169,6 +169,9 @@ stages:
       enableSymbolValidation: false
       # It's a private repo in github so this won't pass until we create an internal mirror
       enableSourceLinkValidation: false
+      # Broken by the move to the 5.0.0-preview8 SDK
```
dougbu, line 172: Suggest leaving this comment out because these validations won't be needed at all once https://dnceng.visualstudio.com/internal/_git/dotnet-release/pullrequest/10338 is in.

dougbu (approved):

azure-pipelines.yml
Tratcher: As in we'll be able to leave enableSigningValidation set to false permanently? Or that we'll be able to revert this PR?

dougbu: > As in we'll be able to leave enableSigningValidation set to false permanently?

Yes

Tratcher, line 172:
```suggestion
```

PR #417, opened 2020-09-15
Add CredScan suppressions

MihaZupan:


.config/CredScanSuppressions.json
```diff
@@ -0,0 +1,37 @@
+{
+    "tool": "Credential Scanner",
+    "suppressions": [
+        {
+            "_justification": "Legitimate key/cert used for testing",
+            "file": [
+                "\\test\\TestCertificates\\testCert.pfx",
```
ManickaP, line 7: Why backslashes? And why the starting backslash? BTW, would this work on Linux? And shouldn't we suppress the whole directory? Is it even possible `test/TestCertificates/*`?

.config/CredScanSuppressions.json
```diff
@@ -0,0 +1,37 @@
...
+            ]
+        },
+        {
+            "_justification": "Generic example password in documentation",
+            "file": "\\docs\\docfx\\articles\\proxyhttpclientconfig.md",
+            "placeholder": "1234abc"
+        },
+        {
+            "_justification": "Fake password used in test code",
+            "file": "\\test\\ReverseProxy.Tests\\Configuration\\CertificateConfigLoaderTests.cs",
```
ManickaP, line 25: This is fragile, once someone changes the text it won't work. Could we suppress this inline, like: https://github.com/dotnet/runtime/blob/master/src/libraries/System.Diagnostics.Process/tests/ProcessStartInfoTests.cs#L357

.config/CredScanSuppressions.json
```diff
@@ -0,0 +1,37 @@
...
+                "\\test\\TestCertificates\\https-ecdsa-protected.key",
+                "\\test\\TestCertificates\\https-ecdsa.key",
+                "\\test\\TestCertificates\\https-rsa-protected.key",
+                "\\test\\TestCertificates\\https-rsa.key"
+            ]
+        },
+        {
+            "_justification": "Generic example password in documentation",
+            "file": "\\docs\\docfx\\articles\\proxyhttpclientconfig.md",
+            "placeholder": "1234abc"
```
ManickaP, line 21: Doesn't this line: https://github.com/microsoft/reverse-proxy/blob/master/docs/docfx/articles/proxyhttpclientconfig.md#L21 trigger the cred scan?
If not, than cool. If yes, we might exclude the whole file.

.config/CredScanSuppressions.json
MihaZupan: > Why backslashes? 

I started with https://github.com/dotnet/aspnetcore/blob/master/.config/CredScanSuppressions.json as the template.
I see we use forward slashes in runtime, I will update to match runtime.

> BTW, would this work on Linux?

I assume so if aspnetcore isn't seeing problems

> Is it even possible test/TestCertificates/*?

I tried using * or just the directory but that doesn't seem to work. Docs also just say `An absolute path or file name`.

.config/CredScanSuppressions.json
MihaZupan: Didn't trigger for me

.config/CredScanSuppressions.json
MihaZupan: Changed, though inline suppressions seem just as fragile to me.

ManickaP (approved):
LGTM, thanks!

PR #423, opened 2020-09-17
Update CredScan suppressions

MihaZupan:
Use config-based suppressions for all passwords in the flagged `.cs` files.

### DIFF
--- a/.config/CredScanSuppressions.json
+++ b/.config/CredScanSuppressions.json
@@ -16,6 +16,21 @@
             ]
         },
         {
+            "_justification": "Unit test dummy credentials",
+            "file": [
+                "test/ReverseProxy.Tests/Configuration/CertificateConfigLoaderTests.cs",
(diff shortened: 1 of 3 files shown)


Tratcher:
Any idea why the inline suppressions didn't work?

ManickaP:
> Any idea why the inline suppressions didn't work?

Quoting CredScan: `NOTE: It is not enough to remove the secrets and push another change. You must reset your branch according to the instructions.`

It looks like it checks each of the pulled commits individually. We'd need to bypass it with a special commit. Or once it's there, we might try go back to inline suppression (the pulled change won't contain any secrets in any commits).

MihaZupan:
My understanding is that we would have to remove them from history if we actually wanted to remove them. Here we've we're just telling CredScan that it's wrong in these cases.

Tratcher (approved):

MihaZupan:
From what I see code-mirror is happy with this now

PR #438, opened 2020-09-23
programatically -> programmatically

halter73:
Probably should have double checked the spelling before merging #436 instead of immediately after 😆

### DIFF
--- a/docs/docfx/articles/configproviders.md
+++ b/docs/docfx/articles/configproviders.md
@@ -3,7 +3,7 @@
 Introduced: preview4
 
 ## Introduction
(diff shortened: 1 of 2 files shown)


Tornhoof:
I used the spelling from: https://github.com/microsoft/reverse-proxy/blob/master/docs/docfx/articles/configproviders.md (that's where I pasted the value from). You might want to change that too.

Tratcher (approved):

Tratcher:
Note the way our docs are branched right now this won't show up publicly until we branch for preview6.

PR #448, opened 2020-09-29
Implement header based routing 

Tratcher:
Fixes #448

Design points:
- The ability to specify one or more header and value(s) that must be present on a request in order to route.
- Exposed from both code and config.
  - Should we allow you to specify a route using only a header? Today routes must have a path or host.
  - In config I've included both a Value and Values properties for convivence in the common single value scenario. Values takes priority if set. We may also want to do this with Host and Hosts, Header and Headers.
- Headers are lower priority matches than paths, methods, and hosts. One unanticipated side effect of this I saw was if one route specifies methods and another specifies headers, the methods route will always win as more specific. The headers check only comes into play if the higher priority checks are equal.
- Comparisons are case-insensitive by default to match the rest of routing, but you can opt-into case-sensitive checks.
  - Headers do not have consistent semantics like other parts of the request. Casing shouldn't matter in the average scenario, but there are some data types such as base64 encoded values where casing significantly changes the interpretation.
  - The case sensitivity option is implemented separately from match modes. When I included some of the match modes we may add in the future, and tried merging them with the case in/sensitive option it got very verbose and redundant. E.g. ExactHeaderCaseSensitive, ExactHeaderCaseInsensitive, ExactValueCaseSensitive, ExactValueCaseInsensitive, etc.. Even regexes have an optional, separate case sensitivity flag.
- Only a limited set of header matching rules have been implemented, scoped to the customer requirements. See the docs for details.
  - Multi-header and multi-value headers are not supported.
  - Complex matching such as regex patterns are not supported.
  - The matching rules are specified separately from the values because there's no well-defined pattern syntax for header formats short of a full regex. The customer requirements so far only call for basic value matching.
- Routing has some advanced performance features like INodeBuilderPolicy that we are not implementing in this first pass. They add significant complexity and it's not clear yet if they're needed for our scenarios.

Other oddities:
- Bumped the 3.1 dependencies to 3.1.8 to pick up a fix in routing.

### DIFF
--- a/docs/docfx/articles/header-routing.md
+++ b/docs/docfx/articles/header-routing.md
@@ -1,0 +1,283 @@
+---
+uid: header-routing
+title: Header Routing
+---
+
+# Header Based Routing
+
+Proxy routes specified in [config](configfiles.md) or via [code](configproviders.md) must include at least a path or host to match against. In addition to these, a route can also specify one or more headers that must be present on the request.
+
+### Precedence
+
+The default route match precedence order is 1) path, 2) method, 3) host, 4) headers. That means a route which specifies methods and no headers will match before a route which specifies headers and no methods. This can be overridden by setting the `Order` property on a route.
+
+## Configuration
+
+Headers are specified in the `Match` section of a proxy route.
+
+If multiple headers rules are specified on a route then all must match for a route to be taken. OR logic must be implemented either within a header rule or as separate routes.
+
+Configuration:
+```JSON
+    "Routes": [
+      {
+        "RouteId": "route1",
+        "ClusterId": "cluster1",
+        "Match": {
+          "Path": "{**catch-all}",
+          "Headers": [
+            {
+              "Name": "header1",
+              "Values": [ "value1" ],
+              "Mode": "ExactHeader"
+            }
+          ]
+        }
+      },
+      {
+        "RouteId": "route2",
+        "ClusterId": "cluster1",
+        "Match": {
+          "Path": "{**catch-all}",
+          "Headers": [
+            {
+              "Name": "header2",
+              "Values": [ "1prefix", "2prefix" ],
+              "Mode": "HeaderPrefix"
+            }
+          ]
+        }
+      },
+      {
+        "RouteId": "route3",
+        "ClusterId": "cluster1",
+        "Match": {
+          "Path": "{**catch-all}",
+          "Headers": [
+            {
+              "Name": "header3",
+              "Mode": "Exists"
+            }
+          ]
+        }
+      },
+      {
+        "RouteId": "route4",
+        "ClusterId": "cluster1",
+        "Match": {
+          "Path": "{**catch-all}",
+          "Headers": [
+            {
+              "Name": "header4",
+              "Values": [ "value1", "value2" ],
+              "Mode": "ExactHeader"
+            },
+            {
+              "Name": "header5",
+              "Mode": "Exists"
+            }
+          ]
+        }
+      }
+    ]
+```
+
+Code:
+```C#
+    var routes = new[]
+    {
+        new ProxyRoute()
+        {
+            RouteId = "route1",
+            ClusterId = "cluster1",
+            Match =
+            {
+                Path = "{**catch-all}",
+                Headers = new[]
+                {
+                    new RouteHeader()
+                    {
+                        Name = "Header1",
+                        Values = new[] { "value1" },
+                        Mode = HeaderMatchMode.ExactHeader
+                    }
+                }
+            }
+        },
+        new ProxyRoute()
+        {
+            RouteId = "route2",
+            ClusterId = "cluster1",
+            Match =
+            {
+                Path = "{**catch-all}",
+                Headers = new[]
+                {
+                    new RouteHeader()
+                    {
+                        Name = "Header2",
+                        Values = new[] { "1prefix", "2prefix" },
+                        Mode = HeaderMatchMode.HeaderPrefix
+                    }
+                }
+            }
+        },
+        new ProxyRoute()
+        {
+            RouteId = "route3",
+            ClusterId = "cluster1",
+            Match =
+            {
+                Path = "{**catch-all}",
+                Headers = new[]
+                {
+                    new RouteHeader()
+                    {
+                        Name = "Header3",
+                        Mode = HeaderMatchMode.Exists
+                    }
+                }
+            }
+        },
+        new ProxyRoute()
+        {
+            RouteId = "route4",
+            ClusterId = "cluster1",
+            Match =
+            {
+                Path = "{**catch-all}",
+                Headers = new[]
+                {
+                    new RouteHeader()
+                    {
+                        Name = "Header4",
+                        Values = new[] { "value1", "value2" },
+                        Mode = HeaderMatchMode.ExactHeader
+                    },
+                    new RouteHeader()
+                    {
+                        Name = "Header5",
+                        Mode = HeaderMatchMode.Exists
+                    }
+                }
+            }
+        }
+    };
+```
+
+## Contract
+
+[RouteHeaderData](xref:Microsoft.ReverseProxy.Configuration.Contract.RouteHeaderData) defines the configuration contract, where [RouteHeader](xref:Microsoft.ReverseProxy.Abstractions.RouteHeader) defines the code contract. The only difference in these contracts is that RouteHeaderData defines an extra `Value` property for convenience, `Values` is preferred if specified.
+
+### Name
+
+The header name to check for on the request. A non-empty value is required. This field is case-insensitive per the HTTP RFCs.
+
+### Values
+
+A list of possible values to search for. The header must match at least one of these values according to the specified `Mode`. At least one value is required unless `Mode` is set to `Exists`.
+
+### Mode
+
+[HeaderMatchMode](xref:Microsoft.ReverseProxy.Abstractions.HeaderMatchMode) specifies how to match the value(s) against the request header. The default is `ExactHeader`.
+- ExactHeader - The header must match in its entirety, subject to the value of `IsCaseSensitive`. Only single headers are supported. If there are multiple headers with the same name then the match fails.
+- HeaderPrefix - The header must match by prefix, subject to the value of `IsCaseSensitive`. Only single headers are supported. If there are multiple headers with the same name then the match fails.
+- Exists - The header must exist and contain any non-empty value.
+
+### IsCaseSensitive
+
+Indicates if the value match should be performed as case sensitive or insensitive. The default is `false`, insensitive.
+
+## Examples
+
+These examples use the configuration specified above.
+
+### Scenario 1 - Exact Header Match
+
+A request with the following header will match against route1.
+```
+Header1: Value1
+```
+
+A header with multiple values is not currently supported and will _not_ match.
+```
+Header1: Value1, Value2
+```
+
+Multiple headers with the same name are not currently supported and will _not_ match.
+```
+Header1: Value1
+Header1: Value2
+```
+
+### Scenario 2 - Multiple Values
+
+Route2 defined multiple values to search for in a header ("1prefix", "2prefix"), any of the values are acceptable. It also specified the `Mode` as `HeaderPrefix`, so any header that starts with those values is acceptable.
+
+Any of the following headers will match route2.
+```
+Header2: 1prefix
+```
+```
+Header2: 2prefix
+```
+```
+Header2: 1prefix-extra
+```
+```
+Header2: 2prefix-extra
+```
+
+A header with multiple values is not currently supported and will _not_ match.
+```
+Header2: 1prefix, 2prefix
+```
+
+Multiple headers with the same name are not currently supported and will _not_ match.
+```
+Header2: 1prefix
+Header2: 2prefix
+```
+
+### Scenario 3 - Exists
+
+Route3 only requires that the header "Header3" exists with any non-empty value
+
+The following is an example that will match route3.
+```
+Header3: value
+```
+
+An empty header will _not_ match.
+```
+Header3:
+```
+
+This mode does support headers with multiple values and multiple headers with the same name since it does not look at the header contents. The following will match.
+```
+Header3: value1, value2
+```
+```
+Header3: value1
+Header3: value2
+```
+
+### Scenario 4 - Multiple Headers
+
+Route4 requires both `header4` and `header5`, each matching according to their specified `Mode`. The following headers will match route4:
+```
+Header4: value1
+Header5: AnyValue
+```
+```
+Header4: value2
+Header5: AnyValue
+```
+
+These will _not_ match route4 because they are missing one of the required headers:
+```
+Header4: value2
+```
+```
+Header5: AnyValue
+```
--- a/src/ReverseProxy/Abstractions/RouteDiscovery/Contract/HeaderMatchMode.cs
+++ b/src/ReverseProxy/Abstractions/RouteDiscovery/Contract/HeaderMatchMode.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using Microsoft.ReverseProxy.Service.Routing;
+
 namespace Microsoft.ReverseProxy.Abstractions
 {
     /// <summary>
@@ -9,7 +11,7 @@ namespace Microsoft.ReverseProxy.Abstractions
     public enum HeaderMatchMode
     {
         /// <summary>
-        /// The header must match in its entirety, subject to the value of <see cref="IHeaderMetadata.CaseSensitive"/>.
+        /// The header must match in its entirety, subject to the value of <see cref="IHeaderMetadata.IsCaseSensitive"/>.
         /// Only single headers are supported. If there are multiple headers with the same name then the match fails.
         /// </summary>
         ExactHeader,
@@ -20,7 +22,7 @@ namespace Microsoft.ReverseProxy.Abstractions
         // ValuePrefix,
 
         /// <summary>
-        /// The header must match by prefix, subject to the value of <see cref="IHeaderMetadata.CaseSensitive"/>.
+        /// The header must match by prefix, subject to the value of <see cref="IHeaderMetadata.IsCaseSensitive"/>.
         /// Only single headers are supported. If there are multiple headers with the same name then the match fails.
         /// </summary>
         HeaderPrefix,
--- a/src/ReverseProxy/Abstractions/RouteDiscovery/Contract/HeaderMatchMode.cs
+++ b/src/ReverseProxy/Abstractions/RouteDiscovery/Contract/HeaderMatchMode.cs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Microsoft.ReverseProxy.Service.Routing
+namespace Microsoft.ReverseProxy.Abstractions
 {
     /// <summary>
     /// How to match header values.
--- a/src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyMatch.cs
+++ b/src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyMatch.cs
@@ -76,12 +76,7 @@ namespace Microsoft.ReverseProxy.Abstractions
                 return true;
             }
 
-            if ((headers1?.Count ?? 0) == 0 && (headers2?.Count ?? 0) == 0)
-            {
-                return true;
-            }
-
-            if (headers1 != null && headers2 == null || headers1 == null && headers2 != null)
+            if (headers1 == null || headers2 == null)
             {
                 return false;
             }
--- a/src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyMatch.cs
+++ b/src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyMatch.cs
@@ -34,11 +34,10 @@ namespace Microsoft.ReverseProxy.Abstractions
         /// </summary>
         // public ICollection<KeyValuePair<string, string>> QueryParameters { get; set; }
 
-        // TODO:
         /// <summary>
-        /// Only match requests that contain all of these request headers.
+        /// Only match requests that contain all of these headers.
         /// </summary>
-        // public ICollection<KeyValuePair<string, string>> Headers { get; set; }
+        public IReadOnlyList<RouteHeader> Headers { get; set; }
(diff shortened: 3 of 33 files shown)


src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyMatch.cs
```diff
@@ -65,7 +64,42 @@ internal static bool Equals(ProxyMatch proxyMatch1, ProxyMatch proxyMatch2)
...
+            {
+                return true;
+            }
+
+            if ((headers1?.Count ?? 0) == 0 && (headers2?.Count ?? 0) == 0)
+            {
+                return true;
+            }
+
+            if (headers1 != null && headers2 == null || headers1 == null && headers2 != null)
```
alnikola, line 84: It can be simplified because we checked at line 74 that they are not equal, thus if one is null the other is guaranteed to be **not null**.
```suggestion
            if (headers1 == null || headers2 == null)
```

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyMatch.cs
```diff
@@ -65,7 +64,42 @@ internal static bool Equals(ProxyMatch proxyMatch1, ProxyMatch proxyMatch2)
...
+        {
+            if (ReferenceEquals(headers1, headers2))
+            {
+                return true;
+            }
+
+            if ((headers1?.Count ?? 0) == 0 && (headers2?.Count ?? 0) == 0)
+            {
+                return true;
+            }
```
alnikola, line 82: Nullability checks can be removed if this is moved below the next `if` block.

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -182,9 +182,37 @@ private void Convert(ProxyMatch proxyMatch, ProxyMatchData data)
...
 
-        private CircuitBreakerOptions Convert(CircuitBreakerData data)
+        private static IReadOnlyList<RouteHeader> Convert(IReadOnlyList<RouteHeaderData> headers)
+        {
+            if (headers == null)
+            {
+                return null;
+            }
+
+            return headers.Select(data => Convert(data)).ToArray();
```
alnikola, line 195: Can we reduce it to the single line?
```suggestion
            return headers?.Select(data => Convert(data)).ToArray();
```

src/ReverseProxy/Utilities/CaseSensitiveEqualHelper.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            if ((list1?.Count ?? 0) == 0 && (list2?.Count ?? 0) == 0)
+            {
+                return true;
+            }
+
+            if (list1 != null && list2 == null || list1 == null && list2 != null)
+            {
+                return false;
+            }
```
alnikola, line 26: Can it be simplified if the checks are rearranged as suggested in a comment on ProxyMatch.Equals?

test/ReverseProxy.Tests/IntegrationTests/RoutingTests.cs
```diff
@@ -0,0 +1,246 @@
...
+
+            // Positive
+            var request = new HttpRequestMessage();
+            request.Headers.Add("header1", "value1");
+            var response = await client.SendAsync(request);
+            response.EnsureSuccessStatusCode();
+            Assert.Equal("route1", response.Headers.GetValues("route").SingleOrDefault());
+
+            // Negative
+            response = await client.GetAsync("/");
```
alnikola, line 109: Let's add 2 more negative cases:
1. `request.Headers.Add("header2", "value1")`
2. `request.Headers.Add("header1", null)`

test/ReverseProxy.Tests/IntegrationTests/RoutingTests.cs
```diff
@@ -0,0 +1,246 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.ReverseProxy.Abstractions;
+using Xunit;
+
+namespace Microsoft.ReverseProxy.IntegrationTests
+{
+    public class RoutingTests
```
alnikola, line 20: I believe we should add tests verifying the routing match precedence order:
1. Path
2. Method
3. Host
4. Headers

test/ReverseProxy.Tests/IntegrationTests/RoutingTests.cs
```diff
@@ -0,0 +1,246 @@
...
+            request.Headers.Add("header2", "value2");
+            response = await client.SendAsync(request);
+            response.EnsureSuccessStatusCode();
+            Assert.Equal("route3", response.Headers.GetValues("route").SingleOrDefault());
+
+            // Negative
+            response = await client.GetAsync("/");
+            Assert.Equal(HttpStatusCode.NotFound, response.StatusCode);
+        }
+
```
alnikola, line 203: Add negative tests for multi-value headers.

alnikola (changes requested):
Looks good overall besides some comments on test coverage below.

JamesNK:
> Routing has some advanced performance features like INodeBuilderPolicy that we are not implementing in this first pass. They add significant complexity and it's not clear yet if they're needed for our scenarios.

The main benefit is performance. It moves logic from being executed per-request to at startup when the DFA is built.

alnikola (approved):
LGTM

docs/docfx/articles/header-routing.md
```diff
@@ -0,0 +1,289 @@
...
+                    {
+                        Name = "Header4",
+                        Values = new[] { "value1", "value2" },
+                        Mode = HeaderMatchMode.ExactHeader
+                    },
+                    new RouteHeader()
+                    {
+                        Name = "Header5",
+                        Mode = HeaderMatchMode.Exists
+                    }
```
javiercn, line 161: nit: I'm not going to go a lot into the design, but you could consider
```
                    new RouteHeader()
                    {
                        Name = "Header5"
                    }
```

As a way to indicate that you only care about the header being present.

docs/docfx/articles/header-routing.md
```diff
@@ -0,0 +1,289 @@
...
+This mode does support headers with multiple values and multiple headers with the same name since it does not look at the header contents. The following will match.
+```
+Header3: value1, value2
+```
+```
+Header3: value1
+Header3: value2
+```
+
+### Sceanrio 4 - Multipe Headers
```
javiercn, line 271: nit
```suggestion
### Scenario 4 - Multipe Headers
```

docs/docfx/articles/header-routing.md
```diff
@@ -0,0 +1,289 @@
+---
+uid: header-routing
+title: Header Routing
+---
+
+# Header Based Routing
```
javiercn, line 6: Question:

Is there interest in capturing the header value as part of the route values?

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/HeaderMatchMode.cs
```diff
@@ -0,0 +1,35 @@
...
+        // TODO: Matches individual values from multi-value headers (split by coma, or semicolon for cookies).
+        // Also supports multiple headers of the same name.
+        // ExactValue,
+        // ValuePrefix,
+
+        /// <summary>
+        /// The header must match by prefix, subject to the value of <see cref="IHeaderMetadata.CaseSensitive"/>.
+        /// Only single headers are supported. If there are multiple headers with the same name then the match fails.
+        /// </summary>
+        HeaderPrefix,
```
javiercn, line 28: Is this matching the header name?

Does the case sensitivity make sense here? (I thought the docs said that the header match was case insensitive [here](https://github.com/microsoft/reverse-proxy/pull/448/files#diff-ad8d0e4b392186631caa1ee7a2f1b874R174))

src/ReverseProxy/Service/Config/RuntimeRouteBuilder.cs
```diff
@@ -76,6 +76,15 @@ public RouteConfig Build(ProxyRoute source, ClusterInfo cluster, RouteInfo runti
                 endpointBuilder.Metadata.Add(new AspNetCore.Routing.HostAttribute(source.Match.Hosts.ToArray()));
             }
 
+            if (source.Match.Headers != null)
+            {
+                foreach (var header in source.Match.Headers)
+                {
+                    endpointBuilder.Metadata.Add(
+                        new HeaderMetadata(header.Name, header.Values, header.Mode, header.CaseSensitive));
```
javiercn, line 84: nit: The `CaseSensitive` name tripped me off, maybe you could name it differently, like "Casing" with an enum instead of a bool, `IsCaseSensitive` or `CaseSensitivity`

docs/docfx/articles/header-routing.md
Tratcher: Not that I know of, at least not for the simple matching patterns supported so far. I expect that's something you'd want when using a pattern matching syntax like paths do.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/HeaderMatchMode.cs
Tratcher: No, its matching a prefix of the header value, optional case sensitivity still makes sense there.

src/ReverseProxy/Service/Routing/HeaderMatcherPolicy.cs
```diff
@@ -0,0 +1,231 @@
...
+                        else if (metadata.Mode == HeaderMatchMode.Exists)
+                        {
+                            // We were asked to match as long as the header exists, and it *does* exist
+                            matched = true;
+                        }
+                        // Multi-value headers are not supported.
+                        // Note a single entry may also contain multiple values, we don't distinguish, we only match on the whole header.
+                        else if (requestHeaderValues.Count == 1)
+                        {
+                            var requestHeaderValue = requestHeaderValues.ToString();
```
javiercn, line 96: Is this going to allocate a new string?

Tratcher: No, it just forces the type conversion from a StringValue of array length 1 to a string.

src/ReverseProxy/Service/Config/RuntimeRouteBuilder.cs
Tratcher: `bool IsCaseSensitive` would work.

src/ReverseProxy/Service/Routing/HeaderMatcherPolicy.cs
```diff
@@ -0,0 +1,231 @@
...
+            return matchMode switch
+            {
+                HeaderMatchMode.ExactHeader => MemoryExtensions.Equals(requestHeaderValue, metadataHeaderValue, comparison),
+                HeaderMatchMode.HeaderPrefix => requestHeaderValue != null && metadataHeaderValue != null
+                    && MemoryExtensions.StartsWith(requestHeaderValue, metadataHeaderValue, comparison),
+                _ => throw new NotImplementedException(matchMode.ToString()),
+            };
+        }
+
+        private class HeaderMetadataEndpointComparer : IComparer<Endpoint>
```
javiercn, line 132: Is there really a need to disambiguate endpoints based on the casing or one endpoint being "more" specific than the other?

It feels brittle to route to A instead of B based on them having matching the same value with the only difference that one matches it exactly and the other one does not. IMO it's better to not support that type of thing and throw an error indicating there is an ambiguity, the current behavior can lead to very subtle hard to detect errors.

docs/docfx/articles/header-routing.md
Tratcher: I considered it but it seemed error prone, too easy to accidentally leave out the value. By using the mode the intent is a lot clearer and we can provide better argument validation and errors.

src/ReverseProxy/Service/Routing/HeaderMatcherPolicy.cs
Tratcher: Fair, we can scale this back to only consider the number of header rules.

docs/docfx/articles/header-routing.md
javiercn: The scenario that I'm thinking of is transforming an outgoing request based on the contents of a header. For example a request to /api with the version on the header to /v1/api in the downstream request, but not sure if that's a scenario.

Tratcher: Since we primarily support an exact match against the header, a given route already knows what the value is and can hard code that into the transform for that route.

This does get more interesting for the Exists or Prefix modes where the value is more fluid, but I'd wait to pursue that until we have some specific requirements for it.

javiercn: Fair enough

javiercn (approved):
Looks great!

docs/docfx/articles/header-routing.md
javiercn: Fair enough, I think it'll be more common to require a value than just the header being present.

src/ReverseProxy/Service/Routing/HeaderMatcherPolicy.cs
```diff
@@ -0,0 +1,174 @@
...
+            _ = candidates ?? throw new ArgumentNullException(nameof(candidates));
+
+            for (var i = 0; i < candidates.Count; i++)
+            {
+                if (!candidates.IsValidCandidate(i))
+                {
+                    continue;
+                }
+
+                var metadataList = candidates[i].Endpoint.Metadata.GetOrderedMetadata<IHeaderMetadata>();
```
JamesNK, line 61: I’m not sure about this. I believe all our other matcher policies have metadata that contains all required info.

Using GetOrderedMetadata means that an action couldn’t override the behavior of a controller.

I’m reviewing on my phone so it is hard to check :smile:

Tratcher: That's true, they would be cumulative ANDs.

We looked at trying to define multiple header matches in one rule/metadata but it was very complicated/nested. Other metadata like Method, Host, etc. are simple lists of strings, not complex rules with multiple options each. Note the Authorize attribute is the other one that uses GetOrderedMetadata and it's cumulative, sortof.
https://github.com/dotnet/aspnetcore/blob/1c27ba1bbdc3aac6b483965eb7150011a97ed5b7/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs#L47-L48

javiercn: The metadata that you put in place could be a list of rules? I'm not sure how clean that would be. HttpMethodMetadata does that I think.

javiercn: We will have to think about the semantics here when there are more than one of these things applied to a controller and an action.

Tratcher: > The metadata that you put in place could be a list of rules? I'm not sure how clean that would be. HttpMethodMetadata does that I think.

Yeah, HttpMethodMetadata is a simple `string[]`. We'd need a `RouteHeader[]` and I don't see how you'd be able to build that in a single attribute.

javiercn: We separate the metadata from the attribute, and we combine them at the MVC level.

Endpoint metadata != Controller/Action metadata

Tratcher: Headers are also something you can have more than one of, so it makes more sense for them to be cumulative unless it's for the same header name?
```
[Header("h1", "v1")]
Controller
{
  [Header("h2", "v2")]
  Action
}
```

javiercn: Just to be clear, I'm not complaining about anything specific. I don't know what the correct behavior will be and we will cross that bridge when we get there.

javiercn: > Headers are also something you can have more than one of, so it makes more sense for them to be cumulative unless it's for the same header name?
> 
> ```
> [Header("h1", "v1")]
> Controller
> {
>   [Header("h2", "v2")]
>   Action
> }
> ```

I'm happy if you come up with a proposal and we take it from there. I don't think we should block on this at the moment, just make a note for the future.

samsp-msft:
> * Headers are lower priority matches than paths, methods, and hosts. One unanticipated side effect of this I saw was if one route specifies methods and another specifies headers, the methods route will always win as more specific. The headers check only comes into play if the higher priority checks are equal.

Is there a way to add a manual priority/weight to routes to bypass the default ordering. For example if I want all requests with a debug header to go to a specific destination, could I flag that rule in some way to push it to the top of the stack?

javiercn:
> Is there a way to add a manual priority/weight to routes to bypass the default ordering. For example if I want all requests with a debug header to go to a specific destination, could I flag that rule in some way to push it to the top of the stack?

If you are able to specify the route order explicitly that takes precedence over anything else. Routing supports that, I'm not sure if YARP exposes it, but it would be something good to expose.

Tratcher:
Yes, YARP exposes the Order option on routes.

docs/docfx/articles/header-routing.md
```diff
@@ -0,0 +1,289 @@
...
+This mode does support headers with multiple values and multiple headers with the same name since it does not look at the header contents. The following will match.
+```
+Header3: value1, value2
+```
+```
+Header3: value1
+Header3: value2
+```
+
+### Scenario 4 - Multipe Headers
```
Kahbazi, line 271:
```suggestion
### Scenario 4 - Multiple Headers
```

src/ReverseProxy/Service/Routing/HeaderMatcherPolicy.cs
Tratcher: > Endpoint metadata != Controller/Action metadata

That's a good point. I've re-worked it to use a single metadata type with a list. We'll worry about how to generate that list for MVC later.

src/ReverseProxy/Utilities/CaseInsensitiveEqualHelper.cs
```diff
@@ -15,12 +15,7 @@ internal static bool Equals(IReadOnlyList<string> list1, IReadOnlyList<string> l
                 return true;
             }
 
-            if ((list1?.Count ?? 0) == 0 && (list2?.Count ?? 0) == 0)
-            {
-                return true;
-            }
-
-            if (list1 != null && list2 == null || list1 == null && list2 != null)
+            if (list1 == null || list2 == null)
```
halter73, line 18: Much nicer!

halter73 (changes requested):

docs/docfx/articles/header-routing.md
```diff
@@ -0,0 +1,289 @@
...
+### Mode
+
+[HeaderMatchMode](xref:Microsoft.ReverseProxy.Abstractions.HeaderMatchMode) specifies how to match the value(s) against the request header. The default is `ExactHeader`.
+- ExactHeader - The header must match in its entirety, subject to the value of `IsCaseSensitive`. Only single headers are supported. If there are multiple headers with the same name then the match fails.
+- HeaderPrefix - The header must match by prefix, subject to the value of `IsCaseSensitive`. Only single headers are supported. If there are multiple headers with the same name then the match fails.
+- Exists - The header must exist and contain any non-empty value.
+
+### IsCaseSensitive
+
+Indicates if the value match should be performed as case sensitive or insensitive. The default is `false`, insensitive.
```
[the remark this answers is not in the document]
Tratcher: I don't know. It's more of an escape hatch for people, I'm not sure how much it will get used.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/HeaderMatchMode.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// The header must match in its entirety, subject to the value of <see cref="IHeaderMetadata.IsCaseSensitive"/>.
+        /// Only single headers are supported. If there are multiple headers with the same name then the match fails.
+        /// </summary>
+        ExactHeader,
+
+        // TODO: Matches individual values from multi-value headers (split by coma, or semicolon for cookies).
+        // Also supports multiple headers of the same name.
+        // ExactValue,
+        // ValuePrefix,
```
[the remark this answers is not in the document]
Tratcher: We're not doing these right now, we'll wait for feedback first. I left them here as examples of possible future values.

I don't know about Contains, it has mixed connotations from string.Contains and ICollection.Contains, we'd want the latter.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/RouteHeader.cs
```diff
@@ -0,0 +1,72 @@
...
+            return new RouteHeader()
+            {
+                Name = Name,
+                Values = Values?.ToArray(),
+                Mode = Mode,
+                IsCaseSensitive = IsCaseSensitive,
+            };
+        }
+
+        internal static bool Equals(RouteHeader header1, RouteHeader header2)
```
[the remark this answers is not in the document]
Tratcher: To simplify the calling code when either value might be null.

docs/docfx/articles/header-routing.md
```diff
@@ -0,0 +1,289 @@
...
+
+[RouteHeaderData](xref:Microsoft.ReverseProxy.Configuration.Contract.RouteHeaderData) defines the configuration contract, where [RouteHeader](xref:Microsoft.ReverseProxy.Abstractions.RouteHeader) defines the code contract. The only difference in these contracts is that RouteHeaderData defines an extra `Value` property for convenience, `Values` is preferred if specified.
+
+### Name
+
+The header name to check for on the request. A non-empty value is required. This field is case-insensitive per the HTTP RFCs.
+
+### Value
+
+A convenience overload of `Values` for the common single value scenario. `Values` is preferred if specified.
```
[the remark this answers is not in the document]
Tratcher: It is in the sample configuration for route1. 
```
            {
              "Name": "header1",
              "Value": "value1"
            }
```

I expect this to be the most common usage so it seemed to warrant special casing. This is the kind of thing we had in mind when we split the config model from the code mode. I can cut it for now and revisit when we look at the overall usability of the config model.

halter73 (approved):

PR #450, opened 2020-10-01
Add helper methods for query transforms

Kahbazi:


src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
```diff
@@ -170,5 +170,35 @@ public static void AddTransformForwarded(this ProxyRoute proxyRoute, bool useFor
...
+            var type = append ? "Append" : "Set";
+
+            proxyRoute.Transforms.Add(new Dictionary<string, string>
+            {
+                ["QueryRouteParameter"] = queryName,
+                [type] = routeValueKey
+            });
+        }
+
+        public static void AddTransformQueryValueParameter(this ProxyRoute proxyRoute, string queryName, string value, bool append = true)
```
Tratcher, line 185:
```suggestion
        public static void AddTransformQueryValueParameter(this ProxyRoute proxyRoute, string queryKey, string value, bool append = true)
```

src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
```diff
@@ -170,5 +170,35 @@ public static void AddTransformForwarded(this ProxyRoute proxyRoute, bool useFor
...
+            var type = append ? "Append" : "Set";
+
+            proxyRoute.Transforms.Add(new Dictionary<string, string>
+            {
+                ["QueryValueParameter"] = queryName,
+                [type] = value
+            });
+        }
+
+        public static void AddTransformRemoveQueryParameter(this ProxyRoute proxyRoute, string queryName)
```
Tratcher, line 196:
```suggestion
        public static void AddTransformRemoveQueryParameter(this ProxyRoute proxyRoute, string queryKey)
```

Tratcher (approved):
One small naming change and then this is ready.

Tratcher:
Thanks

PR #455, opened 2020-10-06
Make IHttpProxy public 

### LINKED ISSUE #408: Simple proxying
Many users have very basic needs from a proxy. They don't need load balancing, affinity, or even routing, or they address those concerns in their own way. What these users really need is the code from [HttpProxy](https://github.com/microsoft/reverse-proxy/blob/c1d65e944d2205c2d442d45785e7875bd87c0fde/src/ReverseProxy/Service/Proxy/HttpProxy.cs) that manages the transition between HttpContext and HttpRequestMessage. They want to call this from their own controller or middleware to forward specific requests to specific destinations.

Consider directly exposing HttpProxy (or a more targeted API) for these use cases.

Tratcher:
Fixes #408 Multiple partners have asked to directly access the IHttpProxy to proxy requests since they handle most of the other aspects themselves (discovery, routing, health checks, etc.). Making it public is easy, but there are some things we want to clean up on the signature first. The main change is removing ProxyTelemetryContext from IHttpProxy.ProxyAsync which has a cascading effect down to StreamCopier. 

@MihaZupan The metrics removed from StreamCopier are a good example of something we may want to replace with new diagnostics.

An additional change I hadn't anticipated was needing to make many of the transform classes public, only the base classes were public before.


Tratcher:
(Note, not ready for review, but marking as ready to debug a CI issue).

Tratcher:
(Actually ready for review this time 😁 )

docs/docfx/articles/direct-proxying.md
```diff
@@ -0,0 +1,105 @@
...
+
+Re-using a client for requests to the same destination is recommended for performance reasons as it allows you to re-use pooled connections. A client may also be re-used for requests to different destinations if the configuration is the same.
+
+### Transforms
+
+The request and response can be modified by providing [transforms](transforms.md) on the [RequestProxyOptions](xref:Microsoft.ReverseProxy.Service.Proxy.RequestProxyOptions).
+
+### Error handling
+
+IHttpProxy catches exceptions and timeouts from the HTTP client, logs them, and converts them to 5xx status codes or aborts the response. The error details, if any, can be accessed from the [IProxyErrorFeature](xref:Microsoft.ReverseProxy.Service.Proxy.IProxyErrorFeature) as shown above.
```
halter73, line 105: What if I want to manually retry a failed request?

docs/docfx/articles/direct-proxying.md
```diff
@@ -0,0 +1,105 @@
...
+        });
+    });
+}
+```
+
+### The HTTP Client
+
+The http client may be customized, but the above example is recommended for common proxy scenarios.
+
+Always use HttpMessageInvoker rather than HttpClient, HttpClient buffers responses by default. Buffering breaks streaming scenarios and increases memory usage and latency.
```
halter73, line 95: We should also point out that response buffering can make handling partial responses easier since the status code and headers of the proxied response can still be modified at that time. This is necessary if you want the proxy to retry a request in this case.

docs/docfx/articles/direct-proxying.md
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+    app.UseRouting();
+    app.UseAuthorization();
+    app.UseEndpoints(endpoints =>
+    {
+        endpoints.Map("/{**catch-all}", async httpContext =>
+        {
+            await httpProxy.ProxyAsync(
+                httpContext, "https://localhost:10000/", httpClient, proxyOptions);
```
halter73, line 78: Adding a prefix to the path works, right? It might be cool to show it off.

```suggestion
                httpContext, "https://localhost:10000/prefix/", httpClient, proxyOptions);
```

docs/docfx/articles/direct-proxying.md
Tratcher: We decided in triage to block the use of HttpClient for now, so no need to include any endorsements for it. If we want to support response buffering that can also be done as an explicit feature of HttpProxy or as a DelegatingHandler in a way that does not breaking streaming scenarios.

docs/docfx/articles/direct-proxying.md
Tratcher: If the response hasn't already started then that can be done by calling the ProxyAsync again. The normal error handling code sets the status code but does not otherwise start the response. You may also be able to use Polly?

docs/docfx/articles/direct-proxying.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Example
+
+### Create a new project
+
+Follow the [Getting Started](xref:getting_started) guide to create a project and add the Microsoft.ReverseProxy nuget dependency.
+
+### Update Startup
+
+In this example the IHttpProxy is registered in DI, injected into the `Startup.Configure` method, and used to proxy requests from a specific route to `https://localhost:10000/`.
```
Tratcher, line 36:
```suggestion
In this example the IHttpProxy is registered in DI, injected into the `Startup.Configure` method, and used to proxy requests from a specific route to `https://localhost:10000/prefix/`.
```

halter73 (approved):

docs/docfx/articles/direct-proxying.md
Kahbazi: If I want to retry a failed request, I do need to buffer the request body and rewind it for retry, right?

Tratcher: Yes, though requests with bodies are not generally considered retriable as you can't tell if the server processed it. E.g. you wouldn't want to double submit a form.

Issue #292, opened 2020-07-02
Customized SocketsHttpHandler should not break distributed tracing


samsp-msft:
https://github.com/dotnet/runtime/issues/31261

cijothomas:
Is this issue created just to track the issue from runtime? Or any other workaround planned here? I wonder if it is even possible to have any workaround at all.

Tratcher:
It's only for tracking so far. No workaround has been proposed.

cijothomas:
> It's only for tracking so far. No workaround has been proposed.

Thanks. We were following up internally and was wondering if anyone found workarounds.

karelz:
Discussion with @samsp-msft: This is not solvable in 5.0 timeframe of .NET Core - hence removing the label ".NET runtime schedule".
YARP 1.0.0 should try to implement the workaround we recommend on main issue (keeping it in 1.0.0 milestone).

Tratcher:
@karelz would this be a good one for @MihaZupan?

jmezach:
We just ran into this issue which is a bit of a bummer because we were kinda expecting this to work out of the box. We have New Relic set up to monitor our environment but it is unable to correlate the requests going into YARP and onwards to the downstream service.

Is there something we can help out with?

Tratcher:
If you'd like to experiment with the workarounds for YARP by all means. E.g. you could try copying the DiagnosticsHandler into YARP and wiring it up.

MihaZupan:
Fixed by #456 

We exposed a new class `ActivityPropagationHandler` for injecting tracing headers based on the current activity.

It is used by the HttpInvoker factory by default, unless disabled by setting `PropagateActivityContext` to false explicitly.

Issue #149, opened 2020-05-08
Use HttpClientHandler for Proxy


Kahbazi:
### What should we add or change to make your life better?
Use `HttpClientHandler` for proxy instead of `SocketsHttpHandler`
https://github.com/microsoft/reverse-proxy/blob/bc9298322206f35e7d658183464cdf806280143f/src/ReverseProxy.Core/Service/Proxy/Infra/ProxyHttpClientFactory.cs#L33-L42

### Why is this important to you?
`HttpClientHandler` creates a [`DiagnosticsHandler`](https://github.com/dotnet/runtime/blob/4b03513e20e0caf9ee34817eb74356903f1bb33e/src/libraries/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs) which adds `Request-Id` header and writes diagnostic events.

karelz:
Is that a good idea for the proxy to automatically add additional headers without explicit opt-in from the proxy dev?

Also note that `SocketsHttpHandler` provides bunch of useful customizations which `HttpClientHandler` does not and we are even adding more (Connection abstractions, connection pool API) which won't be exposed in `HttpClientHandler`.

I would rather view this as duplicate of https://github.com/dotnet/runtime/issues/31261 -- `SocketsHttpHandler` should have ability to opt-in into additional DiagnosticsHandler logging (or some similar solution to the end-to-end problem).

karelz:
Duplicate of https://github.com/dotnet/runtime/issues/31261

PR #456, opened 2020-10-07
Wire up DiagnosticsHandler

jmezach:
As discussed in #292 I've wired up the `DiagnosticsHandler` in the `ProxyHttpClientFactory` as a workaround to make sure that the trace context is propagated to the downstream services. I also had to bring in the `DiagnosticsHandlerLoggingStrings` class to ensure that it compiles.

I've already seen it working on my local development machine in that I receive a `traceparent` header in the downstream service. I'm a little bit hesitant to put this into production to see if it solves our problem with distributed tracing (using New Relic) so I figured I might as well create a draft PR here first to get some input on things that could be improved.


src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,235 @@
...
+    /// DiagnosticHandler notifies DiagnosticSource subscribers about outgoing Http requests
+    /// </summary>
+    internal sealed class DiagnosticsHandler : DelegatingHandler
+    {
+        /// <summary>
+        /// DiagnosticHandler constructor
+        /// </summary>
+        /// <param name="innerHandler">Inner handler: Windows or Unix implementation of HttpMessageHandler.
+        /// Note that DiagnosticHandler is the latest in the pipeline </param>
+        public DiagnosticsHandler(HttpMessageHandler innerHandler) : base(innerHandler)
```
Kahbazi, line 26: How about splitting `DiagnosticsHandler` into two parts? 

1. Writing Diagnostic events which is always on.
2. Adding request id header, which is disable by default and there could be an switch in `ProxyHttpClientContext` to enable it for each cluster.

jmezach: @kahbazi I can understand why you'd might want to configure this on a per-cluster basis. Not entirely sure if you'd want this to be an opt-in or an opt-out though. It was my understanding from #292 that if YARP hadn't customized the `SocketsHttpHandler` it would just work out-of-the-box so it makes sense to me that it would be on by default, but you'd have an option to disable it if you want?

That being said, I wouldn't mind splitting these. Do you have a suggestion for how we would name both of them? And would it just be the `InjectHeaders()` parts that would be split out?

Tratcher:
@MihaZupan can you supervise this PR?

Tratcher:
@jmezach are there any tests you could bring over or add for this?

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
Kahbazi: >@Kahbazi I can understand why you'd might want to configure this on a per-cluster basis. Not entirely sure if you'd want this to be an opt-in or an opt-out though.

Based on this comment https://github.com/microsoft/reverse-proxy/issues/149#issuecomment-626050960, I think maybe the team wants to go with opt-in options.

>That being said, I wouldn't mind splitting these. Do you have a suggestion for how we would name both of them? And would it just be the InjectHeaders() parts that would be split out?

Yes I think it's only the `InjectHeaders()` method. We could also keep the `DiagnosticsHandler` class as it is and only add an option to its constructor and use it in `InjectHeaders()` method.

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,235 @@
...
+                loggingRequestId = Guid.NewGuid();
+                s_diagnosticListener.Write(DiagnosticsHandlerLoggingStrings.RequestWriteNameDeprecated,
+                    new
+                    {
+                        Request = request,
+                        LoggingRequestId = loggingRequestId,
+                        Timestamp = timestamp
+                    }
+                );
+            }
```
MihaZupan, line 101: It should be safe to remove all these deprecated events

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,235 @@
...
+        internal static bool IsEnabled()
+        {
+            // check if there is a parent Activity (and propagation is not suppressed)
+            // or if someone listens to HttpHandlerDiagnosticListener
+            return s_enableActivityPropagation && (Activity.Current != null || s_diagnosticListener.IsEnabled());
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
+            CancellationToken cancellationToken)
+        {
```
MihaZupan, line 39: We should have an `IsEnabled()` check here and fall through to the inner handler.

Runtime does that before calling into the `DiagnosticsHandler`

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
jmezach: Done

jmezach:
@Tratcher I did find this [test class](https://github.com/dotnet/runtime/blob/ba7c518a494ef7aa87159816e07576eefe14290f/src/libraries/System.Net.Http/tests/FunctionalTests/DiagnosticsTests.cs) in dotnet/runtime but that class is abstract. It looks like there's the [SocketsHttpHandlerTest](https://github.com/dotnet/runtime/blob/fdc6472bd1de22b3fe67e232ca86d0593d0b7ce7/src/libraries/System.Net.Http/tests/FunctionalTests/SocketsHttpHandlerTest.cs) class which inherits from that base class but surely we don't want to include all those tests in here?

@MihaZupan I also found out that the DiagnosticsHandler I based this on was a slightly older version from dotnet/corefx rather than dotnet/runtime. The latter seems to be slightly different in its implementation. Should I base this implementation on that?

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
jmezach: @MihaZupan Isn't that already handled by [this piece of code](https://github.com/rr-wfm/reverse-proxy/blob/53f57694104d2b4260a3d0908fbb4ee0d2a8dd98/src/ReverseProxy/Telemetry/DiagnosticsHandler.cs#L55-L69)?

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
jmezach: I've implemented this and added a `PropagateActivityContext` flag to the `ProxyHttpClientOptions` although I'm open for alternative names for this property.

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
MihaZupan: No, that check is looking at whether the listener is enabled.

Without checking IsEnabled first, we're still creating an activity and injecting headers even if activity propagation is disabled or the current activity isn't set.

jmezach: Ah, you're right. I've added an IsEnabled() check after the argument validation now.

Kahbazi:
Now that `DiagnosticHandler` is adding `traceparent` and `tracestate` header, what should happened to a request if it already has these headers? 
Is it ok that the proxy request has two `traceparent` and `tracestate` headers?
Should the user add a transform to prevent the original `traceparent` to be proxied or it can be removed in `DiagnosticHandler`?

jmezach:
@Kahbazi That's an interesting question. I did some testing and it seems that if we send a `traceparent` header along with the request to YARP it will just send that exact same header to the downstream service and it adds a `Request-Id` header (which includes the value of `traceparent` among other things).

I'm not too sure about the specifics of the W3C Trace Context specification, but this behaviour does make sense to me. Essentially the reverse proxy is just another service in the chain if you're looking at distributed tracing. Were you expecting different behaviour?

Kahbazi:
Oh my bad! I looked at the code and it's only sending the `traceparent` if it doesn't exists, so we won't get multiple headers.
https://github.com/microsoft/reverse-proxy/blob/cbc128da9d266908b2baecb56e3e73a5967bd8b2/src/ReverseProxy/Telemetry/DiagnosticsHandler.cs#L183
But I'm wondering how's the distributed tracing when the request have a `traceparent` and YARP is sending that instead of getting it from current Activity.

If I have ZipKin distributed tracing I expect to see the charts like this. Here _Backend_ parent is _Proxy_ and _Proxy_ parent is _Client_
image
I could be wrong but now I think we will see this which _Client_ is parent for both _Backend_ and _Proxy_.
image

jmezach:
@Kahbazi Yeah, you might be on to something there since I'm seeing the traceparent being transmitted as is to the downstream service.

Perhaps @Tratcher and @MihaZupan could chime in here. Is this what we're expecting to see here? I'm getting a feeling that we might have to do somethings differently in the case of YARP.

jmezach:
@Kahbazi I've run a couple of experiments to validate your theory. I currently have an API gateway setup using YARP and I've recompiled it using the changes in this PR. I then setup OpenTelemetry for both the API gateway and the downstream service which is an ASP.NET Core API that simply echoes the incoming headers in the response body. I've setup the Jaeger exporter with OpenTelemetry since I couldn't get the Zipkin exporter to work, or at least not on my machine.

If I don't send a `traceparent` header along with my request to the API gateway (YARP) I'm seeing the following in Jaeger:

Screenshot 2020-10-09 at 20 51 45

When I sent along a `traceparent` header with a dummy value along with my request to the API gateway (YARP) I'm seeing this:

Screenshot 2020-10-09 at 20 51 58

So yes, it seems that when we sent along a `traceparent` header with the incoming request to YARP it messes up the distributed tracing to the downsteam service and the chain is broken.

jmezach:
@Kahbazi @MihaZupan @Tratcher I've added an additional check for the `currentActivity.ParentId`. When that is set we remove the existing `traceparent` header before setting the new one. Not entirely sure if that is correct, but at least now I'm seeing all spans show up in Jaeger even if I send along a `traceparent` with my initial request to YARP.

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,232 @@
...
+            if (!_injectHeaders)
+            {
+                return;
+            }
+
+            if (currentActivity.IdFormat == ActivityIdFormat.W3C)
+            {
+                if (!string.IsNullOrEmpty(currentActivity.ParentId))
+                {
+                    request.Headers.Remove(DiagnosticsHandlerLoggingStrings.TraceParentHeaderName);
```
Kahbazi, line 185: What if a user wants to add its own `traceparent` header in a Transform?

jmezach: Hmm, yeah, that could be an issue. Not sure how we could tackle that though. I believe transforms are performed before we get to this handler so how would we know whether the header was added by a transform or if it was in the original request? If the original request didn't have such a header then `ParentId` on the `currentActivity` would be `null` which is why I added the `if`. But if the original request has such a header and a transform has modified it in some way then I don't think we'll be able to detect that here, unless we add some metadata to the header somehow.

jmezach:
@MihaZupan Is there something more that I can do here? I think there's a couple of design issues here that we need to tackle to take this further, but would love some input on that.

jmezach:
@Tratcher Is there anything I can do to get this moving?

MihaZupan:
Sorry for the late reply.

Correct me if I'm wrong, but the main thing we want from this is to propagate the trace context via headers, to enable distributed tracing.

Do you actually need the diagnostic events written here?

Looking at all the DiagnosticHandler-related issues I gather there are a few pain points:
1. It's internal, so can't be used when working with SocketsHttpHandler directly for example dotnet/runtime#31261 (this is fine, this PR is fixing that)
2. It's not configurable in a reliable and reasonable manner dotnet/runtime#31862
3. Because of 2), there are problems when users don't want the hidden internal Activity creation dotnet/runtime#41072

What I would suggest is to drop the writing of events and creation of activity, and only bring over the `InjectHeaders` portion of DiagnosticHandler. If anyone requires those events/activities, they can always layer them on top.

The behavior I would expect would be:
- If the propagation is disabled (which is the default here) OR `Activity.Current` isn't set, nothing happens
- If trace headers already exist, they are removed - this is fine as it's an opt-in feature
- We add the trace headers with info from the Current activity

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,232 @@
...
+            // we expect baggage to be empty or contain a few items
+            using (var e = currentActivity.Baggage.GetEnumerator())
+            {
+                if (e.MoveNext())
+                {
+                    var baggage = new List<string>();
+                    do
+                    {
+                        var item = e.Current;
+                        baggage.Add(new NameValueHeaderValue(WebUtility.UrlEncode(item.Key), WebUtility.UrlEncode(item.Value)).ToString());
```
MihaZupan, line 219: I realize this is taken from runtime, but these should really be using `Url.EscapeDataString`

jmezach:
@MihaZupan I can totally live with that, makes everything a lot simpler imho. I just did a rebase of my branch on the latest changes and I'll get to work on this. Perhaps we should also rename the handler to ActivityPropagationHandler?

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
jmezach: Fixed

src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs
```diff
@@ -65,7 +66,7 @@ public HttpMessageInvoker CreateClient(ProxyHttpClientContext context)
             }
 
             Log.ProxyClientCreated(_logger, context.ClusterId);
-            return new HttpMessageInvoker(handler, disposeHandler: true);
+            return new HttpMessageInvoker(new DiagnosticsHandler(handler, newClientOptions.PropagateActivityContext), disposeHandler: true);
```
Tratcher, line 69: If PropagateActivityContext is false, do you even need to add the DiagnosticsHandler? It would be more efficient to exclude it.

jmezach: Yes, I agree, that's not necessary. Changed that to either passing the `SocketsHttpHandler` or the `DiagnosticsHandler` to the `HttpMessageInvoker` based on the value of `PropagateActivityContext`.

jmezach:
With my latest changes I'm now seeing the following in Jaeger:

Screenshot 2020-11-04 at 21 32 47

One thing to note is that we now have only 2 spans, one for the YARP-based proxy and one for the downstream service. Previously we would see 3 spans, one for the YARP-based proxy, one for the HTTP call from the proxy to the downstream service and one for the downstream service. I'm unsure as to what would be considered "correct" here. Personally I can live with what I'm seeing here, but I don't know if this is what people expect to happen.

src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs
```diff
@@ -65,6 +66,12 @@ public HttpMessageInvoker CreateClient(ProxyHttpClientContext context)
             }
 
             Log.ProxyClientCreated(_logger, context.ClusterId);
+
+            if (newClientOptions.PropagateActivityContext)
+            {
+                return new HttpMessageInvoker(new DiagnosticsHandler(handler), disposeHandler: true);
```
MihaZupan, line 72:
```suggestion
                handler = new DiagnosticsHandler(handler);
```

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,109 @@
...
+                var responseTask = base.SendAsync(request, cancellationToken);
+
+                return await responseTask.ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+                // we'll report task status in HttpRequestOut.Stop
+                throw;
+            }
+        }
```
MihaZupan, line 61: Try-catch can be removed entirely

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// <summary>
+        /// DiagnosticHandler constructor
+        /// </summary>
+        /// <param name="innerHandler">Inner handler: Windows or Unix implementation of HttpMessageHandler.
+        /// Note that DiagnosticHandler is the latest in the pipeline </param>
+        public DiagnosticsHandler(HttpMessageHandler innerHandler) : base(innerHandler)
+        {
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
```
MihaZupan, line 29:
```suggestion
        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
```

This doesn't have to be async, we can just inject the header and return `base.SendAsync`

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        #region private
+
+        private void InjectHeaders(Activity currentActivity, HttpRequestMessage request)
+        {
+            if (currentActivity.IdFormat == ActivityIdFormat.W3C)
+            {
+                request.Headers.Remove(DiagnosticsHandlerLoggingStrings.TraceParentHeaderName);
+
+                if (!request.Headers.Contains(DiagnosticsHandlerLoggingStrings.TraceParentHeaderName))
```
MihaZupan, line 71: We're removing the header, so no need to check contains again

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,109 @@
...
+                    {
+                        request.Headers.TryAddWithoutValidation(DiagnosticsHandlerLoggingStrings.TraceStateHeaderName, currentActivity.TraceStateString);
+                    }
+                }
+            }
+            else
+            {
+                request.Headers.Remove(DiagnosticsHandlerLoggingStrings.RequestIdHeaderName);
+
+                if (!request.Headers.Contains(DiagnosticsHandlerLoggingStrings.RequestIdHeaderName))
```
MihaZupan, line 84: Same here, this can be just

```c#
request.Headers.Remove(DiagnosticsHandlerLoggingStrings.RequestIdHeaderName);
request.Headers.TryAddWithoutValidation(DiagnosticsHandlerLoggingStrings.RequestIdHeaderName, currentActivity.Id);
```

src/ReverseProxy/Telemetry/DiagnosticsHandlerLoggingStrings.cs
```diff
@@ -0,0 +1,19 @@
...
+        public const string ActivityName = "System.Net.Http.HttpRequestOut";
+        public const string ActivityStartName = "System.Net.Http.HttpRequestOut.Start";
+
+        public const string RequestIdHeaderName = "Request-Id";
+        public const string CorrelationContextHeaderName = "Correlation-Context";
+
+        public const string TraceParentHeaderName = "traceparent";
+        public const string TraceStateHeaderName = "tracestate";
+    }
+}
```
MihaZupan, line 19: We're only using `traceparent` and `tracestate` from this now, we can just move these constants into the handler

src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs
jmezach: I tried this, but it doesn't work since `handler` is typed as `SocketsHttpHandler` currently. I can change that to `HttpMessageHandler`, but then I would need to change some more things in this method since it accesses properties that are only defined on `SocketsHttpHandler`. Frankly I thought that wasn't worth it in this case, but I'd be happy to be proven wrong ;).

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,98 @@
...
+            // This code won't be called unless consumer unsubscribes from DiagnosticListener right after the check.
+            // So some requests happening right after subscription starts might not be instrumented. Similarly,
+            // when consumer unsubscribes, extra requests might be instrumented
+
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request)); //, SR.net_http_handler_norequest); TODO: Is this really necessary?
+            }
+
+            // If we are on at all, we propagate current activity information
```
MihaZupan, line 49: These comments are now outdated

src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs
MihaZupan: In that case it's fine as it is, I don't think we'll be adding more layered handlers in the near future.

MihaZupan:
> Perhaps we should also rename the handler to ActivityPropagationHandler?

Sounds good, would avoid some confusion between the types as this one has very different behavior now.

> One thing to note is that we now have only 2 spans, one for the YARP-based proxy and one for the downstream service. Previously we would see 3 spans, one for the YARP-based proxy, one for the HTTP call from the proxy to the downstream service and one for the downstream service. I'm unsure as to what would be considered "correct" here. Personally I can live with what I'm seeing here, but I don't know if this is what people expect to happen.

I think this is okay as the default. A user can always add more activities wherever they see fit, but if we added them ourselves, we would have to add another switch to allow disabling it. This PR still saves everyone the work of dealing with headers, which is very valuable. @Tratcher would you agree with that?

As we're now not creating any extra activities/events, we should consider making it an opt-out instead of opt-in (there is a bit of an opt-in just by having the Activity set in the first place).

src/ReverseProxy/Telemetry/DiagnosticsHandler.cs
jmezach: Fixed

MihaZupan:
> As we're now not creating any extra activities/events, we should consider making it an opt-out instead of opt-in (there is a bit of an opt-in just by having the Activity set in the first place).

The team view on this is to make it **opt-out by default**, until proven otherwise.
The good thing is we actually have a switch for it, unlike with HttpHandler.

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ProxyHttpClientOptions.cs
```diff
@@ -17,6 +17,8 @@ public sealed class ProxyHttpClientOptions
 
         public int? MaxConnectionsPerServer { get; set; }
 
+        public bool PropagateActivityContext { get; set; }
```
Tratcher, line 20: If we're going to default to true, then it might be easiest to make these properties nullable all the way up like MaxConnectionsPerServer. That way we can apply the default in the factory only if a value isn't set, the default isn't baked into the options types.

src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
+            CancellationToken cancellationToken)
+        {
+            // This handler is conditionally inserted by the ProxyHttpClientFactory based on the configuration
+            // If inserted it will insert the necessary headers to propagate the current activity context to
+            // the downstream service, if there is a current activity
+
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request)); //, SR.net_http_handler_norequest); TODO: Is this really necessary?
```
Tratcher, line 44:
```suggestion
                throw new ArgumentNullException(nameof(request));
```

test/ReverseProxy.Tests/Service/Proxy/Infrastructure/ProxyHttpClientFactoryTests.cs
```diff
@@ -99,14 +100,28 @@ public void CreateClient_ApplyMaxConnectionsPerServer_Success()
...
+        public void CreateClient_ApplyPropagateActivityContext_Success()
+        {
+            var factory = new ProxyHttpClientFactory(Mock<ILogger<ProxyHttpClientFactory>>().Object);
+            var options = new ClusterProxyHttpClientOptions(default, default, default, default, true);
+            var client = factory.CreateClient(new ProxyHttpClientContext { NewOptions = options });
+
+            var handler = GetHandler(client);
+
+            Assert.NotNull(handler);
+            // TODO Assert.Equal(true, handler.prop);
```
Tratcher, line 113: ?

Tratcher:
> I think this is okay as the default. A user can always add more activities wherever they see fit, but if we added them ourselves, we would have to add another switch to allow disabling it. This PR still saves everyone the work of dealing with headers, which is very valuable. @Tratcher would you agree with that?

Simpler seems better until we have a requirement for more.

test/ReverseProxy.Tests/Service/Proxy/Infrastructure/ProxyHttpClientFactoryTests.cs
jmezach: Yeah, the problem here is that the `handler` variable here is the `SocketsHttpHandler` so at this point I don't have anything externally visible that I can do the assertion on. I could have the `GetHandler` method return a `HttpMessageHandler` but then we'll need to change all the other tests. Or we could add an out parameter to the `GetHandler` method that can contain the `ActivityPropagationHandler` and then do the assertion on that. Either option isn't ideal imho, so maybe you have some other suggestions for this test?

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ProxyHttpClientOptions.cs
jmezach: Fixed

test/ReverseProxy.Tests/Service/Proxy/Infrastructure/ProxyHttpClientFactoryTests.cs
```diff
@@ -99,14 +100,28 @@ public void CreateClient_ApplyMaxConnectionsPerServer_Success()
...
         }
 
+        [Fact]
+        public void CreateClient_ApplyPropagateActivityContext_Success()
+        {
+            var factory = new ProxyHttpClientFactory(Mock<ILogger<ProxyHttpClientFactory>>().Object);
+            var options = new ClusterProxyHttpClientOptions(default, default, default, default, true);
+            var client = factory.CreateClient(new ProxyHttpClientContext { NewOptions = options });
+
+            var handler = GetHandler(client);
```
Tratcher, line 110: If you want to assert that the handler was applied then try adding an optional bool `expectActivityHandler` to GetHandler and put the assert in there. That way you don't have to modify the other tests.

jmezach: Yup, that works. Fixed :).

src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+        private const string CorrelationContextHeaderName = "Correlation-Context";
+
+        private const string TraceParentHeaderName = "traceparent";
+        private const string TraceStateHeaderName = "tracestate";
+
+        /// <summary>
+        /// ActivityPropagationHandler constructor
+        /// </summary>
+        /// <param name="innerHandler">Inner handler: Windows or Unix implementation of HttpMessageHandler.
+        /// Note that ActivityPropagationHandler is the latest in the pipeline </param>
```
MihaZupan, line 30: I think this comment can be removed as it adds no value

src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+            var currentActivity = Activity.Current;
+            if (currentActivity != null)
+            {
+                InjectHeaders(currentActivity, request);
+            }
+
+            return base.SendAsync(request, cancellationToken);
+        }
+
+        #region private
```
MihaZupan, line 57: No need for a region

src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        private void InjectHeaders(Activity currentActivity, HttpRequestMessage request)
+        {
+            if (currentActivity.IdFormat == ActivityIdFormat.W3C)
+            {
+                request.Headers.Remove(TraceParentHeaderName);
```
MihaZupan, line 63:
```suggestion
                request.Headers.Remove(TraceParentHeaderName);
                request.Headers.Remove(TraceStateHeaderName);
```

We should remove the TraceState header as well (even if we don't write our own)

MihaZupan (commented):
Some comments, otherwise LGTM

src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
jmezach: Yes, that makes sense. Fixed!

MihaZupan (approved):
We could add a test to assert the header modifications, but the behavior LGTM.

jmezach:
I've been trying to write some tests for the `ActivityPropagationHandler`, however I'm running into an issue with setting the current activity. I've tried doing `using var activity = new Activity()`, but that fails to compile stating that `Activity` must be implicitly convertible to `System.IDisposable`. As far I know the activity type is disposable, but perhaps that is only on .NET 5 and not in .NET Core 3.1.

I've also tried setting `Activity.Current` explicitly, but it just returns `null` after setting the property. I must be doing something wrong ;).

MihaZupan:
> As far I know the activity type is disposable, but perhaps that is only on .NET 5 and not in .NET Core 3.1.

I believe the `IDisposable` interface was added in 5.0.

Would you be okay with me pushing test changes to this PR branch directly?

jmezach:
@MihaZupan Sure, go ahead :).

MihaZupan:
Looks like I don't have the perms to push onto your branch directly, could you pick up the commit/changes from my branch: https://github.com/MihaZupan/reverse-proxy/commit/ff6e65fba2f57ae74be7b27323420038c76878c2

jmezach:
@MihaZupan Done. I had to make one additional change in order for things to build on my dev machine. There was a using statement missing.

MihaZupan:
There's a conflict with the master branch now (small conflict in the test file)

jmezach:
@MihaZupan I fixed the merge conflict.

src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.ReverseProxy.Telemetry
+{
+    /// <summary>
+    /// ActivityPropagationHandler propagates the current Activity to the downstream service
+    /// </summary>
+    internal sealed class ActivityPropagationHandler : DelegatingHandler
```
Tratcher, line 17: This handler should be public. We expect people to replace the http client factory and they'll need this handler if they still want activities in their custom client.

Tratcher (approved):

src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
jmezach: Sure thing!

jmezach:
@MihaZupan One thing I noticed is that you've put the tests for the handler into Service/Proxy/Infrastructure while the actual handler is in Telemetry. I would expect the tests and corresponding class in the same folder structure, but do you have a preference as to whether the handler should move or the tests?

Tratcher:
> I would expect the tests and corresponding class in the same folder structure

Yes, please move it.

jmezach:
@Tratcher Sure thing. Done :).

PR #476, opened 2020-10-19
Expose immutable Metadata from RouteConfig

ittennull:
Since extensibility of the yarp is work in progress yet I use Metadata to add some additional information that I can later use to run my own middelware. Without Metadata being accessible through Endpoint->Metadata->RouteConfig->Metadata I need to build my own store for metadata which is unfortunate because I already have everything.

### DIFF
--- a/src/ReverseProxy/Service/RuntimeModel/RouteConfig.cs
+++ b/src/ReverseProxy/Service/RuntimeModel/RouteConfig.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using Microsoft.ReverseProxy.Abstractions;
 using Microsoft.ReverseProxy.Service.RuntimeModel.Transforms;
 
@@ -36,6 +37,7 @@ namespace Microsoft.ReverseProxy.RuntimeModel
             Order = proxyRoute.Order;
             Cluster = cluster;
             Transforms = transforms;
+            Metadata = proxyRoute.Metadata.ToImmutableDictionary();
         }
 
         public RouteInfo Route { get; }
@@ -48,6 +50,8 @@ namespace Microsoft.ReverseProxy.RuntimeModel
         public IReadOnlyList<AspNetCore.Http.Endpoint> Endpoints { get; }
 
         public Transforms Transforms { get; }
+
+        public IReadOnlyDictionary<string, string> Metadata { get; }
 
         public bool HasConfigChanged(ProxyRoute newConfig, ClusterInfo cluster)
         {
--- a/src/ReverseProxy/Service/RuntimeModel/RouteConfig.cs
+++ b/src/ReverseProxy/Service/RuntimeModel/RouteConfig.cs
@@ -37,7 +37,7 @@ namespace Microsoft.ReverseProxy.RuntimeModel
             Order = proxyRoute.Order;
             Cluster = cluster;
             Transforms = transforms;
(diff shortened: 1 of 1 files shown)


src/ReverseProxy/Service/RuntimeModel/RouteConfig.cs
```diff
@@ -49,6 +51,8 @@ internal sealed class RouteConfig
 
         public Transforms Transforms { get; }
 
+        public IReadOnlyDictionary<string, string> Metadata { get; }
```
Tratcher, line 54: How does this help you when RouteConfig itself is still internal? You'd probably need to flow the data somewhere accessible like IReverseProxyFeature:
https://github.com/microsoft/reverse-proxy/blob/a5184af7a838083d89e10dde466f6774af4f14d6/src/ReverseProxy/Middleware/DestinationInitializerMiddleware.cs#L62

Tratcher:
For reference, can you get an example of the kind of information you are adding to the route metadata?

src/ReverseProxy/Service/RuntimeModel/RouteConfig.cs
ittennull: Ah, right. I saw RouteConfig pop up in the debugger and when I opened it it didn't expose metadata so I decided to add it but I didn't notice that the whole class was internal. Ok, thanks for clarifying it, I'll try to find another way. Although in general it would be very helpful to have runtime access to the original configuration that was used to create a route

ittennull:
> For reference, can you get an example of the kind of information you are adding to the route metadata?

Sure. I'm still in a process of estimating yarp and I'm working on a POC that has the functionality that is currently used in production. I'm thinking on separating configuration from the RP itself so that I don't need to update RP every time I add new route. It should look like this
```
RP ---watch_and_load---> combined JSON config <---update--- ConfigUpdaterApp
                                                                ↑       ↑
                                                  route_collection1    route_collection2
```
I want combined json config to be a standard config for RP, no additional stuff. That means currently that I need to rely on metadata if I want something custom (transform extensibility isn't there yet). Something custom is for example a transformation that is not available in RP now or another middleware that I want to activate on the route. For example I want to short circuit certain paths and don't let them go through. I add metadata item `{ "ShortCircuit", ""}`. Now in my RP I add a ShortCircuitMiddleware that needs to somehow understand which routes should be short circuited. Since all the meatadata is available on the endpoint (only hidden because of internal classes) I could use 
```
if (context.GetEndpoint()?.Metadata.GetMetadata<RouteConfig>()?.Metadata?.TryGetValue("ShortCircuit", out var _))
    // short circuit the request
```

So I want to reuse existing ProxyRoute structure to add functionality and I need an access to this additional data

Kahbazi:
After https://github.com/microsoft/reverse-proxy/pull/328 got merged, you could add your custom metadata to YARP endpoints.

Tratcher:
@Kahbazi I think they're asking about different metadata in this case. Endpoint.Metadata is a routing concept, but RouteConfig.Metadata is just extra config input.

Kahbazi:
@Tratcher I was thinking about adding ProxyRoute.Metadata directly to Endpoint.Metadata or creates an object from ProxyRoute.Metadata and add it to Endpoint.Metadata in `ReverseProxyConventionBuilder.ConfigureRoutes(Action<IEndpointConventionBuilder, ProxyRoute> convention)`.

Issue #282, opened 2020-07-01
Outbound http connections to destinations are configurable 


samsp-msft:
https://github.com/dotnet/runtime/issues/987

samsp-msft:
Runtime feature: Enables control of the version to prevent connections from being downgraded.

Tratcher:
A partner requires this for use with IHttpProxy because they prefer HTTP/1.1 for their scenarios. We currently default to HTTP/2.

Proposal:
RequestProxyOptions is used with IHttpProxy and could also be exposed as per-cluster configuration.
https://github.com/microsoft/reverse-proxy/blob/4bd161b6b357e2f146159dd8aef843b6e67fc4a7/src/ReverseProxy/Service/Proxy/RequestProxyOptions.cs#L9-L22

This would also address https://github.com/microsoft/reverse-proxy/issues/319 (configurable timeouts).

Make the http version and version policy (5.0 only) available on the options and then use them when constructing requests in HttpProxy.

Should the version property be nullable to allow you to preserve the incoming request version?

Tratcher:
The version policy should also be added to the RequestParametersTransformContext so it can be customized by transforms.
https://github.com/microsoft/reverse-proxy/blob/6311df4abe1ce10c6274c5212600011886cac2c6/src/ReverseProxy/Service/RuntimeModel/Transforms/RequestParametersTransformContext.cs#L22

PR #479, opened 2020-10-19
HTTP Version Policy

ManickaP:
Contributes to #282

### DIFF
--- a/src/ReverseProxy/Service/Proxy/RequestProxyOptions.cs
+++ b/src/ReverseProxy/Service/Proxy/RequestProxyOptions.cs
@@ -28,18 +28,14 @@ namespace Microsoft.ReverseProxy.Service.Proxy
         /// Preferred version of the outgoing request.
         /// The default is HTTP/2.0.
         /// </summary>
-        public Version Version { get; set; } = System.Net.HttpVersion.Version20;
+        public Version Version { get; set; } = HttpVersion.Version20;
 
-#if NET5_0
+#if NET
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/ReverseProxy/Service/Proxy/RequestProxyOptions.cs
```diff
@@ -22,9 +24,23 @@ public class RequestProxyOptions
...
 
-        // Future:
-        // ResponseBodyTimeout - The time allowed to receive the full response body. Default to infinite. Not applied to Upgraded requests or gRPC streams.
-        // HttpVersion - Default to HTTP/2?
-        // HttpVersionPolicy - Default to OrLower?
+        /// <summary>
+        /// Preferred version of the outgoing request.
+        /// The default is HTTP/2.0.
+        /// </summary>
+        public Version Version { get; set; } = System.Net.HttpVersion.Version20;
```
Tratcher, line 31:
```suggestion
        public Version Version { get; set; } = HttpVersion.Version20;
```

Tratcher (approved):
Looks good. Try `#if NET` in the product code as well and we'll see how it works.

Note there's a second half to #282 that we can discuss after this PR which is making RequestProxyOptions configurable per cluster.

src/ReverseProxy/Service/Proxy/RequestProxyOptions.cs
ManickaP: Fixed

Issue #428, opened 2020-09-18
Github action doc build no longer works


Tratcher:
https://github.com/microsoft/reverse-proxy/runs/1135982884?check_suite_focus=true

We updated to the Preview8 SDK which requires VS 16.8.0-preview2, but the github agents only have 16.7.
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#visual-studio-enterprise-2019

For preview 5 we should probably run the steps manually, after we get all the preview5 edits made.
https://github.com/microsoft/reverse-proxy/actions/runs/261817348/workflow

@MihaZupan

Tratcher:
A workaround might be to revert the SDK only in that release/docs branch, but that might complicate our branching even more.

Tratcher:
This was worked around by building the docs locally and checking them in. The Github Action was modified to sync the checked in docs rather than build them.

Revert after 16.8 RTMs in November and gets deployed to the github agents (when?).

karelz:
Triage: Wait for November - it will go away, we can manually workaround until then.

Tratcher:
ETA: Nov 16-20
https://github.com/actions/virtual-environments/issues/1891

PR #491, opened 2020-10-24
Update docs

davidfowl:
This is what happened when I cloned and built the repo


Tratcher (approved):
https://github.com/microsoft/reverse-proxy/issues/428 This is a temporary annoyance until VS 16.8 RTM gets deployed on the CI agents so we can go back to the CI building the docs.

If it's too annoying then we can delete these files in master in the interim and put the .gitignore entry back.

davidfowl:
It's definitely too annoying 😄 . It embeds the branch name it turns out...

PR #493, opened 2020-10-25
Some cleanup

davidfowl:
- Use a cached Task to represent true and false
- Remove nested classes around session affinity
- Use nameof for SessionAffinityConstants
- Updated calling code and tests

PS: this PR is mostly cosmetic but the nested classes have been bugging me for a long time. There's more but I was looking at the session affinity logic.

### DIFF
--- a/src/ReverseProxy/Utilities/TaskUtilities.cs
+++ b/src/ReverseProxy/Utilities/TaskUtilities.cs
@@ -12,6 +12,6 @@ namespace Microsoft.ReverseProxy.Utilities
     internal static class TaskUtilities
     {
         internal static readonly Task<bool> TrueTask = Task.FromResult(true);
-        internal static readonly Task<bool> FalseTask = Task.FromResult(true);
+        internal static readonly Task<bool> FalseTask = Task.FromResult(false);
     }
 }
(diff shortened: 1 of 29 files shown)


src/ReverseProxy/Utilities/TaskUtilities.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.ReverseProxy.Utilities
+{
+    internal static class TaskUtilities
+    {
+        internal static readonly Task<bool> TrueTask = Task.FromResult(true);
```
davidfowl, line 14: @stephentoub I wish we had this in the box 😄

stephentoub: https://github.com/dotnet/runtime/issues/22627

davidfowl: FromResult returning a cached task would fix lots of code automagically.

stephentoub: Right, that's my preference as well, at least for the bools.

ManickaP (approved):

Tratcher (approved):

PR #496, opened 2020-10-26
Update of releasing steps 

ManickaP:
Shuffled the sections a bit so that it makes more sense IMO.
Filled in little details encountered while doing the release.


Tratcher (approved):
Good follow through 😁.

PR #459, opened 2020-10-07
Active and passive health checks

alnikola:
### Abstract
From time to time, destinations can get unhealthy and start failing requests processing due to various reasons, thus to prevent request failures and maintain a good quality of service YARP must monitor destinations health status and stop sending traffic to the ones became unhealthy until they have recovered.

Fixes #228 

### Health checks
This PR implements active and passive health check mechanisms where the former periodically probes destination with dedicated HTTP requests and the latter watches for proxied client requests outcomes.

There are additionally 2 basic policies implemented:
- Transport failure rate limit - it's a passive policy marking a destination as unhealthy when proxied requests failure rate reached the specified limit
- Consecutive failure threshold - it's an active policy marking a destination as unhealthy when the specified number of consecutive probes failed

### Configuration
All but one of active and passive health check settings are set on the cluster level. The only exception is an optional "Destination/Health" element specifying a separate active health check endpoint. The actual active health check probing URI is composed as "Destination/Address" + (optional) "Destination/Health" + "Cluster/HealthChecks/Path".

**Active health check settings**
- Enabled - indicates whether active checks are enabled. Default `false`.
- Interval - period health check requests are sent. Default "00:01:00".
- Timeout - health check request timeout. Default "00:00:10".
- Policy - name of policy evaluating destination health status based on probing results.
- Path - health check path on all cluster's destinations.

**Passive health check settings**
- Enabled - indicates whether passive checks are enabled. Default `false`.
- Policy - name of policy evaluating destination health based on proxied requests successes and failures.
- ReactivationPeriod - period to reset an `Unhealthy` destination's passive health status back to `Healthy`. Default `00:05:00`.

### API
Configuration
```JSON
{
    "Clusters": {
        "cluster1": {
            "LoadBalancing": {
                "Mode": "Random"
            },
            "HealthChecks": {
                "Active": {
                    "Enabled": "true",
                    "Interval": "00:05:00",
                    "Timeout": "00:00:30",
                    "Policy": "ConsecutiveFailures",
                    "Path": "/Heath"
                },
                "Passive": {
                    "Enabled": "true",
                    "Policy": "TransportFailureRate",
                    "ReactivationPeriod": "00:05:00"
                }
            },
            "Destinations": {
                "cluster1/destination1": {
                    "Address": "https://localhost:10000/",
                    "Health": "https://localhost:4040/"
                },
                "cluster1/destination2": {
                    "Address": "http://localhost:10010/"
                }
           },
           "Metadata": {
[...]

Issue #83, opened 2020-04-23
Panic threshold


Tratcher:
https://github.com/microsoft/reverse-proxy/pull/78#discussion_r412653804
@davidni: Something to keep in mind is what to do when a large portion of endpoints in a backend are unhealthy. Envoy for example has a `panic mode` where it will prefer to try to route to endpoints it knows to be unhealthy, because it may be better than the alternative of not routing at all. See e.g. [Panic threshold](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/panic_threshold)

Tratcher:
Design notes: Such a panic check might happen here when first pulling the list of healthy destinations. If it's empty, or below some threshold, consider using the entire list of destinations instead.
https://github.com/microsoft/reverse-proxy/blob/a6c755c9db27e34dafdc030f501f0e215471d55c/src/ReverseProxy/Middleware/DestinationInitializerMiddleware.cs#L51-L59

Alternate proposal: The health check service could have that kind of logic built in when building the list of healthy destinations.

karelz:
Triage: In 1.0 we need panic response/policy

Tratcher:
Design notes:
 - If you tried to implement this as part of health checks today: 
   -  For active health checks you could do it by replacing the IDestinationHealthUpdater, having it check the results and overriding them if it determined that too many were marked as unhealthy.
   - Replacing the health state results in a loss of information and should be avoided.
   - Even IDestinationHealthUpdater would have trouble doing this for passive health checks. They only check and update one destination at a time.
   - I think this rules out trying to build a panic mode into the health check system.
 - Doing the panic check in middleware is pretty simple:
   - This has the advantage of not changing any persistent state, only request state.
   - This is still some per-request overhead, especially if you wanted to do any other kind of filtering.
```
public Task PanicMiddleware(HttpContext context, Func<Task> next)
{
    var proxyFeature = context.Features.Get<IReverseProxyFeature>();

    if (proxyFeature.AvailableDestinations.Count == 0) // or a percentage of AllDestinations
    {
        // Panic!
        proxyFeature.AvailableDestinations = proxyFeature.AllDestinations;
    }

    return next();
}
```
- What if we made the AvailableDestinations generation code extensible?
[...]

PR #499, opened 2020-10-27
Passive health policy controlling request transport failures rate in proxying them to destinations

### LINKED ISSUE #498: Transport failure rate passive health check policy
Implement a passive health check policy marking a destination as unhealthy when a percentage of failed requests in the specified time period exceeded the given limit. Initially, it implemented as `TransportFailureRateHealthPolicy` in #459, but then was moved out of the scope.

alnikola:
`TransportFailureRateHealthPolicy` calculates the proxied requests failure rate for each destination and marks it as unhealthy if the specified limit is exceeded.

Fixes #498


alnikola:
@davidfowl @Tratcher This is a follow-up PR for a passive `TransportFailureRateHealthPolicy` descoped from #459.

alnikola:
Oops. I accidentally pushed a rebuilt documentation.

src/ReverseProxy/Service/HealthChecks/TransportFailureRateHealthPolicy.cs
```diff
@@ -0,0 +1,152 @@
...
+            UpdateDestinationHealth(cluster.Config, destination, newHealth);
+        }
+
+        private DestinationHealth EvaluateProxiedRequest(ClusterInfo cluster, DestinationInfo destination, bool failed)
+        {
+            var history = _requestHistories.GetOrCreateValue(destination);
+            var metadataName = TransportFailureRateHealthPolicyOptions.FailureRateLimitMetadataName;
+            var rateLimitEntry = _clusterFailureRateLimits.GetValue(cluster, c => new ParsedMetadataEntry<double>(TryParse, c, TransportFailureRateHealthPolicyOptions.FailureRateLimitMetadataName));
+            var rateLimit = rateLimitEntry.GetParsedOrDefault(_policyOptions.DefaultFailureRateLimit);
+            lock (history)
```
davidfowl, line 58: This will be a bottleneck.

src/ReverseProxy/Service/HealthChecks/TransportFailureRateHealthPolicy.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            var error = context.Features.Get<IProxyErrorFeature>();
+            var newHealth = EvaluateProxiedRequest(cluster, destination, error != null);
+            UpdateDestinationHealth(cluster.Config, destination, newHealth);
+        }
+
+        private DestinationHealth EvaluateProxiedRequest(ClusterInfo cluster, DestinationInfo destination, bool failed)
+        {
+            var history = _requestHistories.GetOrCreateValue(destination);
+            var metadataName = TransportFailureRateHealthPolicyOptions.FailureRateLimitMetadataName;
```
davidfowl, line 55: This is unused.

src/ReverseProxy/Service/HealthChecks/TransportFailureRateHealthPolicy.cs
```diff
@@ -0,0 +1,152 @@
...
+                return failureRate < rateLimit ? DestinationHealth.Healthy : DestinationHealth.Unhealthy;
+            }
+        }
+
+        private void UpdateDestinationHealth(ClusterConfig cluster, DestinationInfo destination, DestinationHealth newPassiveHealth)
+        {
+            var state = destination.DynamicState;
+            if (newPassiveHealth != state.Health.Passive)
+            {
+                destination.DynamicState = new DestinationDynamicState(state.Health.ChangePassive(newPassiveHealth));
```
Tratcher, line 74: This is where we need to be concerned about triggering signal updates synchronously from a request thread. This can cause a sync, locked, rebuild of the healthy destinations list, right?

src/ReverseProxy/Service/HealthChecks/TransportFailureRateHealthPolicy.cs
alnikola: I believe I can implement a lock-free algorithm here, but first let's agree on whether or not we are satisfied with how Signal is used because depending on that discussion outcome this whole policy can require substantial changes.

src/ReverseProxy/Service/HealthChecks/TransportFailureRateHealthPolicy.cs
davidfowl: Right, this is exactly what I'm worried about.

alnikola: I do agree that the concerns are valid, but would also like to dig into the details the code flow.

1. Normally, assigning `Unhealthy` to `destination.DynamicState.Health.Passive` will happen only once in a reactivation period (scale of minutes) for each destination because right after setting `Unhealthy` the destination gets excluded from consideration by next requests.
2. It's correct that `ClusterInfo.DynamicState.HealthyDestinations` can get rebuilt synchronously here, but under normal conditions it shouldn't happen frequently for the same cluster. In extreme cases, where most of cluster's destinations getting unhealthy, we should probably anyways enable "panic" mode where `HealthyDestinations` rebuilding will be suspended.

P.S. I will file an issue for "panic" cluster's health mode.

alnikola: All in all, I would firstly address locking-related concerns (e.g. possibility of a deadlock), and only then performance-related ones.

Tratcher: It might be better to kick this out to a service similar to the reactivation scheduler, one that could A) move the rebuild off of the request thread, and B) debounce some of the changes, e.g. if many requests to a destination fail in close proximity, you don't want to update the state many times. We might be able to build both of those capabilities into Signals directly.

alnikola: >B) debounce some of the changes, e.g. if many requests to a destination fail in close proximity

It's kind of already done by this if clause `if (newPassiveHealth != state.Health.Passive)` that prevents consecutive updates of the same destination. Meaning, once it got unhealthy, its health won't be updated till a reactivation.

alnikola: Though, it costs one Signal read per request.

alnikola: I found an existing issue for panic mode #83.

alnikola: I implemented IDestinationHealthUpdater applying changes to active and passive destination health properties asynchronously as well as handling bulk updates in a thread-safe way. Later, it can also implement the panic mode feature. Please, check it out.

src/ReverseProxy/Service/HealthChecks/ConsecutiveFailuresHealthPolicy.cs
```diff
@@ -19,51 +19,36 @@ internal class ConsecutiveFailuresHealthPolicy : IActiveHealthCheckPolicy
...
             {
-                cluster.ResumeHealthyDestinationUpdates();
+                var destination = probingResults[i].Destination;
+
+                var count = _failureCounters.GetOrCreateValue(destination);
+                var newHealth = EvaluateHealthState(threshold, probingResults[i].Response, count);
+                newHealthPairs[i] = (destination, newHealth);
             }
+
+            _ = _healthUpdater.SetActiveAsync(cluster, newHealthPairs);
```
Tratcher, line 51: @davidfowl was your original concern here that health probes could be running in parallel, or that the updates were happening in parallel?

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
```diff
@@ -0,0 +1,115 @@
...
+            var updateSyncToken = new object();
+            if (!_activeSynchTokens.TryAdd(cluster, updateSyncToken))
+            {
+                // Don't allow concurrent updates.
+                return Task.CompletedTask;
+            }
+
+            return Task.Run(() =>
+            {
+                if (!_activeSynchTokens.TryGetValue(cluster, out var currentToken))
```
Tratcher, line 36: I don't see how you could get into this state? The only remove happens at the end of this method, there is no cancel.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        public DestinationHealthUpdater(IReactivationScheduler reactivationScheduler, ILogger<DestinationHealthUpdater> logger)
+        {
+            _reactivationScheduler = reactivationScheduler ?? throw new ArgumentNullException(nameof(reactivationScheduler));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public Task SetActiveAsync(ClusterInfo cluster, IEnumerable<(DestinationInfo Destination, DestinationHealth NewHealth)> newHealthPairs)
+        {
+            var updateSyncToken = new object();
```
Tratcher, line 27: The value isn't used for anything, it could be static, or null, or the cluster itself?

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    cluster.ResumeHealthyDestinationUpdates();
+                    _activeSynchTokens.TryRemove(cluster, out _);
+                }
+            });
+        }
+
+        public Task SetPassiveAsync(ClusterInfo cluster, DestinationInfo destination, DestinationHealth newHealth, TimeSpan reactivationPeriod)
+        {
+            return Task.Run(() =>
```
Tratcher, line 77: Can we do this without a closure? It's going to be pretty common.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
```diff
@@ -0,0 +1,115 @@
...
+
+namespace Microsoft.ReverseProxy.Service.HealthChecks
+{
+    internal class DestinationHealthUpdater : IDestinationHealthUpdater
+    {
+        private readonly IReactivationScheduler _reactivationScheduler;
+        private readonly ILogger<DestinationHealthUpdater> _logger;
+        private readonly ConcurrentDictionary<ClusterInfo, object> _activeSynchTokens = new ConcurrentDictionary<ClusterInfo, object>();
+
+        public DestinationHealthUpdater(IReactivationScheduler reactivationScheduler, ILogger<DestinationHealthUpdater> logger)
```
Tratcher, line 19: The IReactivationScheduler could be absorbed by DestinationHealthUpdater, it's the only caller right now right?

src/ReverseProxy/Service/HealthChecks/ConsecutiveFailuresHealthPolicy.cs
davidfowl: Updates triggered from the request.

src/ReverseProxy/Service/HealthChecks/IDestinationHealthUpdater.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Threading.Tasks;
+using Microsoft.ReverseProxy.RuntimeModel;
+
+namespace Microsoft.ReverseProxy.Service.HealthChecks
+{
+    /// <summary>
+    /// Updates destinations' health states when it's requested by a health check policy
+    /// while taking into account not only the new evaluated value but also the overall current cluster's health state.
+    /// </summary>
+    public interface IDestinationHealthUpdater
```
davidfowl, line 15: We don't need this. We need to support signals dispatching off the caller's thread.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
alnikola: It was a leftover from another lock-based approach (token was locked on). Replaced with `null`.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
alnikola: In my opinion, they have different responsibilities. `IDestinationHealthUpdater` decides whether or not apply health state changes requested by a policy and actually applies them. The former part (decision making) is not implemented in this PR, but can be added as part of #83.

src/ReverseProxy/Service/HealthChecks/IDestinationHealthUpdater.cs
alnikola: As far as I see it, the only reliable way to avoid deadlocks or slowing down a request is to do "heavy" Signal updates in background. Do you suggest to fire up a background task right from SignalContext? Or do you have a completely different approach to addressing this issue in mind?

This abstraction also synchronizes concurrent bulk update operations involving Pause/Resume healthy destination collection rebuilding.

davidfowl: > Do you suggest to fire up a background task right from SignalContext?

Yes. I think we should do that. Similar to how TaskCompletionSource has RunContinuationsAsynchronously, this is similar. We don't need a new abstraction for this.

alnikola: I see. Though, currently signal's value is updated through a simple property setter which cannot be parameterized and I'm not sure whether it's a good idea to always run updates in background. Thus, we might need to introduce a `Signal.SetValueAsync()` method or something similar.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
alnikola: I was thinking about adding a cancellation logic, but now it seems to soon since we have not yet decided whether this type should exist at all. It's been deleted.

alnikola:
Converting it to a draft until signals removal is completed in #505.

src/ReverseProxy/Service/HealthChecks/ConsecutiveFailuresHealthPolicy.cs
alnikola: `SetActive` updates the active health state **synchronously** now.

src/ReverseProxy/Service/HealthChecks/IDestinationHealthUpdater.cs
alnikola: Signals have been removed.

alnikola:
Remaining work:
0. Address a few currently open comments.
1. Unit tests for DestinationHealthUpdater
2. Decision on debouncing passive health state changes
3. Documentation update taking into account all of the above and the introduced policy itself

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
alnikola: Actually, you can be right. `ReactivationScheduler` contains only few lines of production code (besides logging) which can be seen as dealing with health state updates. So, it could make sense to merge the `IReactivationScheduler` into this new `DestinationHealthUpdater`.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+            }
+
+            if (changed)
+            {
+                cluster.UpdateDynamicState();
+            }
+        }
+
+        public Task SetPassiveAsync(ClusterInfo cluster, DestinationInfo destination, DestinationHealth newHealth, TimeSpan reactivationPeriod)
```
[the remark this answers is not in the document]
alnikola: Previously, @davidfowl advises to return a task and dismiss it in such case.

src/ReverseProxy/Service/HealthChecks/IDestinationHealthUpdater.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// will be reset to <see cref="DestinationHealth.Unknown"/>. Otherwise, it's not used.</param>
+        /// <returns><see cref="Task"/> representing a passive health update operation.</returns>
+        Task SetPassiveAsync(ClusterInfo cluster, DestinationInfo destination, DestinationHealth newHealth, TimeSpan reactivationPeriod);
+
+        /// <summary>
+        /// Sets the active health values on the given destinations.
+        /// </summary>
+        /// <param name="cluster">Cluster.</param>
+        /// <param name="newHealths">New active health values.</param>
+        void SetActive(ClusterInfo cluster, IEnumerable<(DestinationInfo Destination, DestinationHealth NewHealth)> newHealths);
```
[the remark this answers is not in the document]
alnikola: Got it. I was not sure what is our guidance on this.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
alnikola: Changed to a closure-less call to `TaskFactory.StartNew`.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
alnikola: `ReactivationScheduler` is merged into `DestinationHealthUpdater`.

src/ReverseProxy/Service/HealthChecks/IDestinationHealthUpdater.cs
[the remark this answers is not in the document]
alnikola: Fixed.

src/ReverseProxy/Service/HealthChecks/TransportFailureRateHealthPolicy.cs
alnikola: Hmm, it seems implementing a lock-free algorithm here requires a untrivial amount of effort. Do you really think it can noticeably hurt perf? It does a little work under the lock which is mainly about updating counters. Actual health state change is done outside.

If it does look like a bottleneck, I'd suggest we address it later in a separate PR. What do you think?

Tratcher: This lock is per destination and destinations are expected to hit thousands of RPS. It's definitely going to cause some contention, but we'll have to measure it to see how much.

Doing the measurements and optimizations in a follow-up item should be OK.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            return Task.Factory.StartNew(s =>
+            {
+                (var c, var d, var h, var p) = ((ClusterInfo, DestinationInfo, DestinationHealth, TimeSpan)) s;
+                var healthState = d.Health;
+                if (h != healthState.Passive)
+                {
+                    healthState.Passive = h;
+                    ScheduleReactivation(c, d, h, p);
+                    c.UpdateDynamicState();
```
Tratcher, line 65: UpdateDynamicState is the only part that needs to be in a background task. It would be better to do the health update and reactivation scheduling synchronously, those are quick and time sensitive.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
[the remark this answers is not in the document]
davidfowl: Let me clarify, if this interface is meant to be a fire and forget interface then it's on to return void here *but* the implementation should use a discard _.

If we want to give the caller the option of awaiting then the caller should use the discard.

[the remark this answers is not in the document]
Tratcher: I checked with David and he said it would be OK for this method to be void and to do the discard internally: `_ = Task.Factory.StartNew`. What he wants to avoid is any usage of `async void`.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
alnikola: Fixed.

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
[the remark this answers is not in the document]
alnikola: Discarding the task internally will jeopardize unit testing where currently I can simply call `await`, but without a task, would have to use delays.

alnikola:
@Tratcher @davidfowl I changed `ClusterInfo.UpdateDynamicState` in such a way that only one of multiple concurrent calls takes effect. `ProxyConfigManager` now calls an internal blocking `ForceUpdateDynamicState` that waits till a lock gets free to apply changes.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -59,13 +60,38 @@ public ClusterConfig Config
...
+                if (force)
+                {
+                    Monitor.Enter(_stateLock, ref lockTaken);
+                }
+                else
+                {
+                    Monitor.TryEnter(_stateLock, ref lockTaken);
+                }
+
+                if (!lockTaken)
```
Tratcher, line 90: This isn't quite enough, it can cause you to miss updates that happened after the prior rebuild started. We can't rely on ForceUpdateDynamicState being called again in the future, we don't know when that will be. Any call to UpdateDynamicStateInternal needs to guarantee that a fresh rebuild happens, even if one is already in progress. The trick is to only de-duplicate rebuilds that haven't started yet.

alnikola: Yes, that's correct, but I'm not sure if that's a real problem because of the following. If there are several health state updates/rebuilt requests which are so close that they concurrently call this method, I think there is no reliable way to order them by causality chain. How can we be sure that an update made 5ms after another one is actually newer?

Tratcher: I don't think this method cares who triggered which update, it's safer to rebuild when asked. If one update is in progress and four more are requested before it finishes, then at least one more rebuild needs to happen (but not all 4).

alnikola: But why do you think that if in example 1 rebuild gets request at T0 and then at T1 (5ms later) 3 more rebuilts are requested again, we should do 1 + 1 rebuild instead just 1 going on? Could you may be give some specific case that concerns you the most?

Tratcher: The rebuild that's in progress is scanning the list of destinations d1->dN. If the state of d1 changes after it's been scanned then another rebuild is required (as signaled by another call to UpdateDynamicStateInternal). If 3 more rebuilds were requested, we only have to do one full rebuild to ensure we pick up new changes anywhere in the list.

alnikola: Ok, I see the point. It means if we got at least one concurrent rebuild request while the current one is going we have to start it over again. I agree it kind of make sense. Though, I'm a bit afraid that we can see "long recalculation tails" with that approach. But, it's probably unavoidable.

Tratcher: Yeah, we'll never end up doing more builds than requested, but hopefully we can de-duplicate.

The next question is how to track the +1 rebuild. I'd considered looping inside the lock, but that could end up blocking one caller for a long time.

docs/docfx/articles/dests-health-checks.md
```diff
@@ -242,21 +245,80 @@ endpoints.MapReverseProxy(proxyPipeline =>
...
+There are two sets of policy parameters defined globally and on per cluster level.
+
+Global parameters are set via the options mechanism using `TransportFailureRateHealthPolicyOptions` type with the following properties:
+
+- `DetectionWindowSize` - period of time while detected failures are kept and taken into account in the rate calculation. Default is `00:01:00`.
+- `MinimalTotalCountThreshold` - minimal total number of requests which must be proxied to a destination within the detection window before this policy starts evaluating the destination's health and enforcing the failure rate limit. Default is `10`.
+- `DefaultFailureRateLimit` - default failure rate limit for a destination to be marked as unhealhty that is applied if it's not set on a cluster's metadata. The value is in range `(0,1)`. Default is `0.3 (30%)`.
```
Tratcher, line 256:
```suggestion
- `DefaultFailureRateLimit` - default failure rate limit for a destination to be marked as unhealthy that is applied if it's not set on a cluster's metadata. The value is in range `(0,1)`. Default is `0.3` (30%).
```

docs/docfx/articles/dests-health-checks.md
```diff
@@ -242,21 +245,80 @@ endpoints.MapReverseProxy(proxyPipeline =>
...
...analyzing responses returned by destinations. For each response from a destination belonging to a cluster with enabled passive health checks, `PassiveHealthCheckMiddleware` invokes an [IPassiveHealthCheckPolicy](xref:Microsoft.ReverseProxy.Service.HealthChecks.IPassiveHealthCheckPolicy) specified for the cluster. The policy analyzes the given response, evaluates a new destination's passive health state and calls [IDestinationHealthUpdater](xref:Microsoft.ReverseProxy.Service.HealthChecks.IDestinationHealthUpdater) to actually update [DestinationHealthState.Passive](xref:Microsoft.ReverseProxy.RuntimeModel.DestinationHealthState.Passive) value. Update happens asynchronously in the background and doesn't block the main pipeline. When a destination gets marked as unhealthy, it stops receiving new requests until it gets reactivated after a configured period. Reactivation means the destination's `DestinationHealthState.Passive` state is reset from `Unhealthy` to `Unknown` and the cluster's list of healthy destinations is rebuilt to include it. A reactivation is scheduled by `IDestinationHealthUpdater` right after setting the destination's `DestinationHealthState.Passive` to `Unhealthy`.
```
Tratcher, line 273:
```suggestion
The main component is [PassiveHealthCheckMiddleware](xref:Microsoft.ReverseProxy.Middleware.PassiveHealthCheckMiddleware) sitting in the request pipeline and analyzing responses returned by destinations. For each response from a destination belonging to a cluster with enabled passive health checks, `PassiveHealthCheckMiddleware` invokes an [IPassiveHealthCheckPolicy](xref:Microsoft.ReverseProxy.Service.HealthChecks.IPassiveHealthCheckPolicy) specified for the cluster. The policy analyzes the given response, evaluates a new destination's passive health state and calls [IDestinationHealthUpdater](xref:Microsoft.ReverseProxy.Service.HealthChecks.IDestinationHealthUpdater) to actually update [DestinationHealthState.Passive](xref:Microsoft.ReverseProxy.RuntimeModel.DestinationHealthState.Passive) value. The update happens asynchronously in the background and doesn't block the request pipeline. When a destination gets marked as unhealthy, it stops receiving new requests until it gets reactivated after a configured period. Reactivation means the destination's `DestinationHealthState.Passive` state is reset from `Unhealthy` to `Unknown` and the cluster's list of healthy destinations is rebuilt to include it. A reactivation is scheduled by `IDestinationHealthUpdater` right after setting the destination's `DestinationHealthState.Passive` to `Unhealthy`.
```

docs/docfx/articles/dests-health-checks.md
```diff
@@ -242,21 +245,80 @@ endpoints.MapReverseProxy(proxyPipeline =>
...
-      (schedule a reactivation)
+      IDestinationHealthUpdater --(Asynchronously update passive state)--> DestinationInfo.Health.Passive
                   |
                   V
-        IReactivationScheduler --(set to Unknown)--> DestinationInfo.Health.Passive
+      (Schedule a reactivation) --(Set to Unknown)--> DestinationInfo.Health.Passive
+```
+
+### Extensibility
+There is one main extensibility point in the passive health check subsystem.
```
Tratcher, line 292:
```suggestion
There is one main extensibility point in the passive health check subsystem, the `IPassiveHealthCheckPolicy`.
```

src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        private Task Reactivate(ClusterInfo cluster, DestinationInfo destination)
+        {
+            var healthState = destination.Health;
+            if (healthState.Passive == DestinationHealth.Unhealthy)
+            {
+                healthState.Passive = DestinationHealth.Unknown;
+                cluster.UpdateDynamicState();
+                Log.PassiveDestinationHealthResetToUnkownState(_logger, destination.DestinationId);
```
Tratcher, line 88: Log before continuing with the update, it's non-trivial.
```suggestion
                healthState.Passive = DestinationHealth.Unknown;
                Log.PassiveDestinationHealthResetToUnkownState(_logger, destination.DestinationId);
                cluster.UpdateDynamicState();
```

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
alnikola: Deduplication improved to allow both of the first and the last rebuilds proceed. Check it out.

src/ReverseProxy/Service/HealthChecks/IDestinationHealthUpdater.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// Asynchronously sets the passive health on the given <paramref name="destination"/>.
+        /// </summary>
+        /// <param name="cluster">Cluster.</param>
+        /// <param name="destination">Destination.</param>
+        /// <param name="newHealth">New passive health value.</param>
+        /// <param name="reactivationPeriod">If <paramref name="newHealth"/> is <see cref="DestinationHealth.Unhealthy"/>,
+        /// this parameter specifies a reactivation period after which the destination's <see cref="DestinationHealthState.Passive"/> value
+        /// will be reset to <see cref="DestinationHealth.Unknown"/>. Otherwise, it's not used.</param>
+        /// <returns><see cref="Task"/> representing a passive health update operation.</returns>
+        Task SetPassiveAsync(ClusterInfo cluster, DestinationInfo destination, DestinationHealth newHealth, TimeSpan reactivationPeriod);
```
Tratcher, line 27: I'm still not sure how we communicate to developers that a passive health check policy should not wait for this task, it's intended only for tests. We'll need to revisit how we test this so it doesn't leak into the public API.

src/ReverseProxy/Service/HealthChecks/TransportFailureRateHealthPolicy.cs
Tratcher: Have you filed this yet?

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -59,35 +61,80 @@ public ClusterConfig Config
...
-                        return passive != DestinationHealth.Unhealthy && active != DestinationHealth.Unhealthy;
-                    }).ToList().AsReadOnly();
+                        lockTaken = false;
+                        _updateRequests.Release();
+                    }
+                }
+            }
+            finally
+            {
+                // This releases the semaphore in case of an exception thrown by 'lock' call.
```
Tratcher, line 133: `lock` can throw?

Can't you accomplish the same thing by inverting the `lock` and the inner `try`? I don't see why you need a try/finally both inside and outside the lock.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -59,35 +61,80 @@ public ClusterConfig Config
...
+        {
+            UpdateDynamicStateInternal(force: true);
+        }
+
+        private void UpdateDynamicStateInternal(bool force)
         {
             // Prevent overlapping updates. If there are multiple signals that state needs to be updated,
             // we want to ensure that updates don't conflict with each other. E.g. if state changes
             // while an update is already in progress, the next update should wait until the current one finishes
             // to ensure they don't race to set _dynamicState and end up with the stale one overwriting the fresh one.
```
Tratcher, line 78: Update the comment to explain the reason for and the mechanics of the double lock.

Tratcher (approved):

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -59,35 +61,80 @@ public ClusterConfig Config
...
-                var allDestinations = DestinationManager.Items;
-                var healthyDestinations = allDestinations;
+                if (force)
+                {
+                    lockTaken = true;
+                    _updateRequests.Wait();
+                }
+                else
+                {
+                    lockTaken = _updateRequests.Wait(0);
```
halter73, line 89: Why use a SemaphoreSlim at all rather than Monitor.TryEnter?

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -59,35 +61,80 @@ public ClusterConfig Config
         /// Recreates the DynamicState data.
         /// </summary>
         public void UpdateDynamicState()
+        {
+            UpdateDynamicStateInternal(force: false);
```
halter73, line 65: So it's okay if this doesn't actually update _dynamicState? Why is that? Is there guaranteed to be calls to this method on a regular interval or every request or something like that? This should probably be renamed to something like TryUpdateDynamicState or MaybeUpdateDynamicState.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
Tratcher: SemaphoreSlim lets two threads in, not just one. That's how we ensure the current + 1 behavior we want.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
Tratcher: It's OK for this to exit early so long as another update is already queued ahead of it on the lock. So long as that update hasn't started yet, any additional requested update would be redundant.

davidfowl:
This health check isn't on by default right? You have to turn on the policy?

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
alnikola: Hmm, it happens to be `lock` can throw only in case of [obj being null](https://docs.microsoft.com/en-us/dotnet/api/system.threading.monitor.enter?view=net-5.0#System_Threading_Monitor_Enter_System_Object_System_Boolean__) which is not possible here. So, the outer try/finally is not necessary. I will remove it.

src/ReverseProxy/Service/HealthChecks/TransportFailureRateHealthPolicy.cs
alnikola: Done #538

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
alnikola: Outer try/finally removed, but the inner one it kept as is because the semaphore must be released while still holding the lock. I added a comment there explaining why it's important.

src/ReverseProxy/Service/HealthChecks/IDestinationHealthUpdater.cs
alnikola: Task has been removed from the public API because I do agree it looked extremely confusing. For testing purposes, there is an internal `Task SetPassiveAsync` method has been added directly on the concrete `DestinationHealthUpdater` class.

alnikola:
> This health check isn't on by default right? You have to turn on the policy?

@davidfowl That's correct. Health checks are disabled by default, they can be enabled in the configuration on per cluster basis as it's explained in the docs.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
alnikola: I do agree we should somehow reflect on the name an eventual character of this method. But, we can do it as a small follow-up PR after a discussion.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
halter73: I didn't notice that this was releasing the semaphore before releasing the lock. Clever.

PR #500, opened 2020-10-29
Log HttpStatusCode as int instead of enum

Kahbazi:
Log status code as integer in order to be consistent with https://github.com/aspnet/Announcements/issues/396


Tratcher (approved):

Tratcher:
You managed to make a PR about status codes with PR number being 500 😁.

Issue #158, opened 2020-05-12
E2E logging coverage


karelz:
Should cover:
* YARP code
* ASP.NET Core components
* dotnet/runtime components

karelz:
We need design of the list.
@davidni can you please post the required list?

davidni:
Here's the timing diagram of one duplex proxy request and the things we keep track of:

![image](https://user-images.githubusercontent.com/5666247/93514330-ee6f0180-f8db-11ea-9cb9-ff905f91d2e9.png)

### Raw timing markers we collect for each request

Measure|Description
-|-
`t1_aspNetCoreRouting`|Marks just after ASP .NET Endpoint Routing has executed the routing logic to pick a matching route for an incoming request.
`t2_pickDestination`|Marks just after Island Gateway has selected one endpoint of the target service to route to.
`t3_buildTargetUrl`|Marks just after Island Gateway has computed the final target uri to proxy a request to.
`t4_reqHeadersStart`|Marks just before Island Gateway issues the outbound HTTPs call to the target endpoint. In code, this occurs just before we call `HttpMessageInvoker.SendAsync`.
`t5_reqBodyStart`|Marks the start of the stream-copy operation copying bytes from the client's request body to the backend.
`t6a1_reqBodyFirstRead`|Marks when the first call to ReadAsync completes on the incoming request stream. This measures when Island Gateway received the first byte in the request body from the client.
`t6a2_reqBodyEnd`|Marks when the stream-copy operation on the request path (i.e. client to Island Gateway to backend) completed.
`t6b1_respHeadersReceived`|Marks when Island Gateway received response headers from the backend service
`t6b2_respBodyStart`|Marks the start of the stream-copy operation copying bytes from the backend's response body to the client.
`t6b3_respBodyFirstRead`|Marks when the first call to ReadAsync completes on the backend's response stream. This measures when Island Gateway received the first byte in the response body from the backend.
`t6b4_respBodyEnd`|Marks when the stream-copy operation on the response path (i.e. backend to Island Gateway to client) completed.
`t6b5_respEnd`|Marks when Island Gateway has completed its response to the client. In HTTP2 requests, this includes sending END_STREAM to the client.

### Aggregates we collect for each request

Measure|Description
-|-
`reqBytes`|Total number of bytes proxied on the request path (i.e. from client to Island Gateway to backend).
`reqOps`|Total IO `ReadAsync` operations performed on the request path (i.e. from client to Island Gateway to backend).
[...]

PR #501, opened 2020-10-29
Add YARP Telemetry instrumentation

MihaZupan:
Contributes to #158

- Removes `IOperationLogger` and associated glue
- Adds a `Microsoft.ReverseProxy` `EventSource` and the related `HttpProxy`/`StreamCopier` instrumentation
- Adds public `ProxyStage` enum

Marked some open questions with `TODO` or `PR REVIEW` comments.


MihaZupan:
The failing tests are the timing tests. They're flaky, I will figure out how to make those more stable.

src/ReverseProxy/Telemetry/ProxyStage.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Telemetry
+{
+    public enum ProxyStage : int
```
Tratcher, line 6: Internal?

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
```diff
@@ -0,0 +1,206 @@
...
+        [NonEvent]
+        public void ProxyFailed()
+        {
+            Interlocked.Increment(ref _failedRequests);
+        }
+
+        [Event(3, Level = EventLevel.Informational)]
+        public void ProxyStage(ProxyStage stage)
+        {
+            if (IsEnabled(EventLevel.Informational, EventKeywords.All))
```
Tratcher, line 77: Should ProxyStage and ContentTransferring use EventLevel.Verbose?

src/ReverseProxy/Telemetry/ProxyStage.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        InvokerSendAsyncStart = 1,
+        InvokerSendAsyncStop,
```
Tratcher, line 9: There are a few different components named `Invoker`.
```suggestion
        SendAsyncStart = 1,
        SendAsyncStop,
```

test/ReverseProxy.Tests/Service/Proxy/StreamCopierTests.cs
```diff
@@ -101,5 +266,30 @@ public override void Write(byte[] buffer, int offset, int count)
...
+
+            public SlowStream(Stream innerStream, TimeSpan waitTime)
+                : base(innerStream)
+            {
+                _waitTime = waitTime;
+            }
+
+            public override async ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default)
+            {
+                await Task.Delay(_waitTime);
```
Tratcher, line 284: The tests must not use Task.Delay or similar pauses, it makes them flaky and slow. Abstract the timing mechanism to avoid this. Alexander recently added the IUptimeClock for a similar scenario.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -93,6 +94,42 @@ public HttpProxy(ILogger<HttpProxy> logger)
...
             _ = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
             _ = proxyOptions ?? throw new ArgumentNullException(nameof(proxyOptions));
+
+            return ProxyTelemetry.Log.IsEnabled()
+                ? ProxyWithTelemetryAsync()
+                : ProxyAsyncCore(context, destinationPrefix, httpClient, proxyOptions);
+
+            async Task ProxyWithTelemetryAsync()
+            {
+                ProxyTelemetry.Log.ProxyStart(context);
```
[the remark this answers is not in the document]
MihaZupan: These should have minimal overhead on their own, I separated them out so it's clearer from the diff what is happening.
I will inline the logic into the main method.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -22,18 +24,83 @@ internal static class StreamCopier
...
         /// See: <see href="https://github.com/dotnet/aspnetcore/blob/080660967b6043f731d4b7163af9e9e6047ef0c4/src/Http/Shared/StreamCopyOperationInternal.cs"/>.
         /// </remarks>
-        public static async Task<(StreamCopyResult, Exception)> CopyAsync(Stream input, Stream output, CancellationToken cancellation)
+        public static Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest, Stream input, Stream output, CancellationToken cancellation)
         {
             _ = input ?? throw new ArgumentNullException(nameof(input));
             _ = output ?? throw new ArgumentNullException(nameof(output));
 
             // TODO: Consider System.IO.Pipelines for better perf (e.g. reads during writes)
+            return ProxyTelemetry.Log.IsEnabled()
```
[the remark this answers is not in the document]
MihaZupan: The main overhead here comes from the the GetTimestamp calls, at ~50 ns on my machine or ~70ns on an Azure F4s_v2 VM per Read/Write pair. The check can be made inline cheaply to avoid duplication.

src/ReverseProxy/Telemetry/ProxyStage.cs
MihaZupan: I assume we'd want to have this exposed as public to make consumption easier, otherwise anyone listening to events would have to look at the source and maintain their own copy of the enum.

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
MihaZupan: We could change the EventLevel for these, or we could also think about introducing Keyword flags for allowing more granular control over which events we fire.

I don't believe we will reach a significant number of ProxyStage values either, so even that could be made into flags instead, allowing users to have fine-grained control over exactly which events they actually care about.

test/ReverseProxy.Tests/Service/Proxy/StreamCopierTests.cs
MihaZupan: I'm looking at that interface, but I don't understand its purpose. It's registered in DI but looks like it is not used anywhere in product code. The concrete type is only used in a few tests, where it's not adding any value over just looking at `Environment.TickCount64` directly.

Was the plan to expose it publicly and use that to query time instead of TickCount/Stopwatch?

src/ReverseProxy/Service/Proxy/StreamCopier.cs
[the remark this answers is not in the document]
MihaZupan: Moved the checks inline

src/ReverseProxy/Telemetry/ProxyStage.cs
MihaZupan: Renamed

test/ReverseProxy.Tests/Service/Proxy/StreamCopierTests.cs
Tratcher: It was added for the passive health checks, but they got moved to another PR.
https://github.com/microsoft/reverse-proxy/pull/499

Talk to Alexander about it.

alnikola: I added #508 to introduce `ITimer` and stabilize all time-dependent tests.

src/ReverseProxy/Telemetry/ProxyStage.cs
Tratcher: The event data is all serialize as primitives right? In the proposed event consumption library aren't we going to have to provide our own types for most things regardless? I'd suggest starting with this as internal until you work out the consumption model.

MihaZupan: Sounds good, will change it to internal

src/ReverseProxy/Microsoft.ReverseProxy.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFrameworks>net5.0;netcoreapp3.1</TargetFrameworks>
     <OutputType>Library</OutputType>
     <RootNamespace>Microsoft.ReverseProxy</RootNamespace>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
davidfowl, line 8: EventSource 😢

test/ReverseProxy.Tests/Service/Proxy/StreamCopierTests.cs
MihaZupan: Discussed this with Alexander, I will adapt the `IUptimeClock` for high-precision (Stopwatch) timing support.

src/ReverseProxy/Telemetry/ProxyStage.cs
MihaZupan: Changed.

test/ReverseProxy.Tests/Service/Proxy/StreamCopierTests.cs
MihaZupan: Now using `IUptimeClock`.

Should the interface be renamed to just `IClock` as 'uptime' is just misleading at this point?

Also Alexander FYI I added a `ManualClock` class for tests (akin to `UptimeClockStub` in your PR).

src/ReverseProxy/Utilities/MonotonicTimer.cs
```diff
@@ -22,7 +22,7 @@ public sealed class MonotonicTimer : IMonotonicTimer
         private static readonly TimeSpan _minimalInterval = TimeSpan.FromMilliseconds(0.1);
 
         /// <summary>
-        /// Use a System.Diagnostics.StopWatch to measure time. Even though it has a poorer resolution, it serves this purpose very well.
+        /// Use a System.Diagnostics.StopWatch to measure time.
```
MihaZupan, line 25: Am I missing something or is this comment just wrong?
Afaik, `Stopwatch` is *the* high-resolution time provider in .Net (at least in its current implementation).

src/ReverseProxy/Telemetry/ProxyStage.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    internal enum ProxyStage : int
+    {
+        SendAsyncStart = 1,
+        SendAsyncStop,
+        RequestContentTransferStart,
+        RequestContentTransferStop,
+        ResponseContentTransferStart,
+        ResponseContentTransferStop,
+        ResponseUpgradeStart,
```
Tratcher, line 14: Unpaired `Start`?
```suggestion
        ResponseUpgrade,
```

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
```diff
@@ -0,0 +1,206 @@
...
+                var routeId = routeConfig?.Route.RouteId;
+
+                var destinations = context.Features.Get<IReverseProxyFeature>()?.AvailableDestinations;
+                var destinationId = destinations != null && destinations.Count != 0
+                    ? destinations[0].DestinationId
+                    : null;
+
+                ProxyStart(cluserId, routeId, destinationId);
+            }
+        }
```
Tratcher, line 52:
```suggestion
```

src/ReverseProxy/Utilities/IUptimeClock.cs
```diff
@@ -4,10 +4,12 @@
...
 {
     /// <summary>
-    /// Measures the time passed since the application start.
+    /// Abstraction over time providers (Environment.TickCount64, Stopwatch.GetTimestamp)
     /// </summary>
     internal interface IUptimeClock
     {
         long TickCount { get; }
+
+        long GetStopwatchTimestamp();
```
Tratcher, line 13:
```suggestion
        // See Stopwatch.GetTimestamp
        long Timestamp { get; }
```

src/ReverseProxy/Utilities/MonotonicTimer.cs
Tratcher: I don't think this type is even used anywhere. Feel free to remove it.

Tratcher (commented):
Much better

src/ReverseProxy/Utilities/IUptimeClock.cs
MihaZupan: I wanted to be very explicit about what it is to avoid more confusion.

For example TickCount is in ms while Timestamp is in whatever frequency the underlying Stopwatch is

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -596,6 +619,7 @@ void ReportResult(HttpContext context, bool reqeuest, StreamCopyResult result, E
                     _ => throw new NotImplementedException(result.ToString()),
                 };
                 ReportProxyError(context, error, exception);
+                ProxyTelemetry.Log.ProxyFailed();
```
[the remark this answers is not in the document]
MihaZupan: Right now ProxyFailed isn't an event, it's only an interlocked increment.

We can [make it an event instead](https://github.com/microsoft/reverse-proxy/pull/501/files#diff-9c97aefe4504c8caed6234adc15a7d8544edd083e9958105ac71126acb4d386fR65-R67), in which case I agree, it should include the ProxyError or the exception message.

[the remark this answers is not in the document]
Tratcher: Yeah, getting access to failure details seems important.

src/ReverseProxy/Utilities/IUptimeClock.cs
Tratcher: Does IUptimeClock also need to expose the frequency?

src/ReverseProxy/Middleware/ProxyInvokerMiddleware.cs
```diff
@@ -85,9 +79,7 @@ public async Task Invoke(HttpContext context)
                 cluster.ConcurrencyCounter.Increment();
                 destination.ConcurrencyCounter.Increment();
 
-                await _operationLogger.ExecuteAsync(
```
davidfowl, line 88: Happy to see these closures finally going away.

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
```diff
@@ -0,0 +1,206 @@
...
+using System.Diagnostics;
+using System.Diagnostics.Tracing;
+using System.Threading;
+using Microsoft.AspNetCore.Http;
+using Microsoft.ReverseProxy.Middleware;
+using Microsoft.ReverseProxy.RuntimeModel;
+
+namespace Microsoft.ReverseProxy.Telemetry
+{
+    [EventSource(Name = "Microsoft.ReverseProxy")]
```
davidfowl, line 14: Use the constructor to specify the name to avoid reflection.

https://github.com/dotnet/aspnetcore/blob/b1dcacabec1aeacef72c9aa2909f1cb49993fa73/src/Hosting/Hosting/src/Internal/HostingEventSource.cs#L25

MihaZupan: Changed.

src/ReverseProxy/Telemetry/ProxyStage.cs
MihaZupan: Changed.

src/ReverseProxy/Utilities/IUptimeClock.cs
MihaZupan: I don't see the need for it, this is just an internal abstraction over fetching time so that we can have stable tests.

If we wanted a cleaner API I think we should just expose a TimeSpan - we don't right now to save a few instructions on repeated conversions.

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        private long _startedRequests;
+        private long _stoppedRequests;
+        private long _failedRequests;
+
+        [Event(1, Level = EventLevel.Informational)]
+        private void ProxyStart(string clusterId, string routeId, string destinationId)
+        {
+            WriteEvent(eventId: 1, clusterId, routeId, destinationId);
+        }
```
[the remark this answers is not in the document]
MihaZupan: I agree with this, the only question I have is what the ProxyInvoker event should be called. `ProxyInvokerInvoke`? 😕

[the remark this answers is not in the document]
Tratcher: ProxyInvokerStart?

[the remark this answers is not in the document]
MihaZupan: Are we expecting custom IHttpProxy implementations to be common?

Naming it Start would imply we're also adding a Stop, which would be redundant with the built-in HttpProxy.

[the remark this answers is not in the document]
Tratcher: > Are we expecting custom IHttpProxy implementations to be common?

I don't think so.

> Naming it Start would imply we're also adding a Stop, which would be redundant with the built-in HttpProxy.

How about ProxyInvoke or ProxyInvoked?

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -22,32 +25,80 @@ internal static class StreamCopier
...
+                long contentLength = 0;
+                long iops = 0;
+                long firstReadTime = -1;
+                long readTime = 0;
+                long writeTime = 0;
+                long lastTime = 0;
+                long nextTransferringEvent = 0;
+
+                // TODO: Should this be configurable
+                long stopwatchTicksBetweenTransferringEvents = 0;
```
Tratcher, line 50: This is a maximum value, right? If reads or writes block then there is no event until they complete? So worst case you'd get one event per stream per read/write per second? And none at all for requests that complete in less than a second? That seems OK for now.

Tratcher:
And yes, renaming the clock abstraction to IClock makes sense.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
MihaZupan: > This is a maximum value, right?

Yes, we're firing it at most once a second.

> If reads or writes block then there is no event until they complete? So worst case you'd get one event per stream per read/write per second?

Correct, if a Read+Write took >= 1s, you'd get an event after every Read+Write pair.

> And none at all for requests that complete in less than a second?

Yes, if StreamCopier takes less than 1s, you won't see any transferring events (you will still get a final Transfered event).

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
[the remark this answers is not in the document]
MihaZupan: Added `ProxyInvoke`

src/ReverseProxy/Service/Proxy/HttpProxy.cs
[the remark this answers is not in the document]
MihaZupan: Changed `ProxyFailed` to be an event taking the ProxyError enum

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -22,32 +25,80 @@ internal static class StreamCopier
...
+                        }
+                    }
+
                     // End of the source stream.
                     if (read == 0)
                     {
+                        if (telemetryEnabled)
+                        {
+                            // PR REVIEW:
+                            // Should this event be logged on failure as well (instead of XContentTransferStop)?
```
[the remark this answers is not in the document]
MihaZupan: Right now we log `XContentTransferStart` and `XContentTransferStop` ProxyStage events in try/finally.

If we were to log ContentTransferred on failure as well, there'd be no need to have the `Stop` ProxyStage event.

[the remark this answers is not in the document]
Tratcher: Ok, try it.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -74,6 +143,17 @@ public static async Task<(StreamCopyResult, Exception)> CopyAsync(Stream input,
...
+
+                if (telemetryEnabled)
+                {
+                    ProxyTelemetry.Log.ProxyStage(isRequest ? ProxyStage.RequestContentTransferStop : ProxyStage.ResponseContentTransferStop);
+                }
+            }
+
+            static long StopwatchTicksToDateTimeTicks(long stopwatchTicks)
+            {
+                var dateTimeTicksPerStopwatchTick = (double)TimeSpan.TicksPerSecond / Stopwatch.Frequency;
```
alnikola, line 155: The same question as above about a direct access to `Stopwatch.Frequency`.

src/ReverseProxy/Telemetry/ProxyStage.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Telemetry
+{
+    internal enum ProxyStage : int
```
alnikola, line 6: Do we need an explicit base type here?

test/ReverseProxy.Tests/Service/Proxy/StreamCopyHttpContentTests.cs
```diff
@@ -25,7 +26,7 @@ public async Task CopyToAsync_InvokesStreamCopier()
             var source = new MemoryStream(sourceBytes);
             var destination = new MemoryStream();
 
-            var sut = new StreamCopyHttpContent(source, autoFlushHttpClientOutgoingStream: false, CancellationToken.None);
+            var sut = new StreamCopyHttpContent(source, autoFlushHttpClientOutgoingStream: false, new Clock(), CancellationToken.None);
```
alnikola, line 29: Is there a plan to add telemetry verification here as well?

alnikola (commented):
A couple of comments, but overall it definitely looks good.

src/ReverseProxy/Service/Proxy/ProxyError.cs
```diff
@@ -6,7 +6,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
     /// <summary>
     /// Errors reported when proxying a request to the destination.
     /// </summary>
-    public enum ProxyError
+    public enum ProxyError : int
```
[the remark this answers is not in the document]
MihaZupan: Just being really explicit about the backing type, as we can't change it in the future.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -22,32 +25,80 @@ internal static class StreamCopier
         /// Based on <c>Microsoft.AspNetCore.Http.StreamCopyOperationInternal.CopyToAsync</c>.
         /// See: <see href="https://github.com/dotnet/aspnetcore/blob/080660967b6043f731d4b7163af9e9e6047ef0c4/src/Http/Shared/StreamCopyOperationInternal.cs"/>.
         /// </remarks>
-        public static async Task<(StreamCopyResult, Exception)> CopyAsync(Stream input, Stream output, CancellationToken cancellation)
+        public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest, Stream input, Stream output, IClock clock, CancellationToken cancellation)
         {
             _ = input ?? throw new ArgumentNullException(nameof(input));
             _ = output ?? throw new ArgumentNullException(nameof(output));
 
+            var telemetryEnabled = ProxyTelemetry.Log.IsEnabled();
```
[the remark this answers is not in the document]
MihaZupan: If it recognizes `telemetryEnabled` can't change, maybe?

We're not doing `IsEnabled` checks inline in any product code here (YARP), which is the same pattern AspNetCore uses. Personally I prefer the readability of that approach over doing IsEnabled checks everywhere - it moves that responsibility onto the EventSource.

In general, I don't think we care too much about how trimmed down this package is?

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -22,32 +25,80 @@ internal static class StreamCopier
...
+                long nextTransferringEvent = 0;
+
+                // TODO: Should this be configurable
+                long stopwatchTicksBetweenTransferringEvents = 0;
+
+                if (telemetryEnabled)
+                {
+                    ProxyTelemetry.Log.ProxyStage(isRequest ? ProxyStage.RequestContentTransferStart : ProxyStage.ResponseContentTransferStart);
+
+                    stopwatchTicksBetweenTransferringEvents = Stopwatch.Frequency; // 1 second
```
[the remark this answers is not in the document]
MihaZupan: See https://github.com/microsoft/reverse-proxy/pull/501#discussion_r516765874, I don't think it's needed

src/ReverseProxy/Service/Proxy/StreamCopier.cs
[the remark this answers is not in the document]
Tratcher: IsEnabled is runtime state, right? The linker is a compile time operation and shouldn't apply in this scenario. The Jitter is the one that would be optimizing for the true/false code paths at runtime, and it can do that regardless of how you structure it.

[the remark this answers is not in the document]
MihaZupan: It is a runtime check, but the linker has an option to remove EventSource code entirely by stubbing out `EventSource.IsEnabled()` to be const false and then removing unreachable code.

test/ReverseProxy.Tests/Service/Proxy/StreamCopyHttpContentTests.cs
MihaZupan: We're just delegating to `StreamCopier` inside, which is already tested, so duplicating that here seems redundant.

Is there another scenario you see that should be covered?

src/ReverseProxy/Service/Proxy/StreamCopier.cs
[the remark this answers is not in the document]
MihaZupan: Changed.

Tratcher (approved):

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
```diff
@@ -0,0 +1,202 @@
...
+    internal sealed class ProxyTelemetry : EventSource
+    {
+        public static readonly ProxyTelemetry Log = new ProxyTelemetry();
+
+        private IncrementingPollingCounter _startedRequestsPerSecondCounter;
+        private PollingCounter _startedRequestsCounter;
+        private PollingCounter _currentRequestsCounter;
+        private PollingCounter _failedRequestsCounter;
+
+        private long _startedRequests;
```
ManickaP, line 21: Where is this being updated? I don't see any increments in this file.

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
```diff
@@ -0,0 +1,202 @@
...
+        protected override void OnEventCommand(EventCommandEventArgs command)
+        {
+            if (command.Command == EventCommand.Enable)
+            {
+                _startedRequestsCounter ??= new PollingCounter("requests-started", this, () => Volatile.Read(ref _startedRequests))
+                {
+                    DisplayName = "Requests Started",
+                };
+
+                _startedRequestsPerSecondCounter ??= new IncrementingPollingCounter("requests-started-rate", this, () => Volatile.Read(ref _startedRequests))
```
ManickaP, line 108: How does this work? It says the counter contains requests per seconds but it's reading the same value as `"requests-started"` counter. Is there any computational magic behind `IncrementingPollingCounter`?

src/ReverseProxy/Service/Proxy/StreamCopier.cs
```diff
@@ -22,28 +25,68 @@ internal static class StreamCopier
         /// Based on <c>Microsoft.AspNetCore.Http.StreamCopyOperationInternal.CopyToAsync</c>.
         /// See: <see href="https://github.com/dotnet/aspnetcore/blob/080660967b6043f731d4b7163af9e9e6047ef0c4/src/Http/Shared/StreamCopyOperationInternal.cs"/>.
         /// </remarks>
-        public static async Task<(StreamCopyResult, Exception)> CopyAsync(Stream input, Stream output, CancellationToken cancellation)
+        public static async Task<(StreamCopyResult, Exception)> CopyAsync(bool isRequest, Stream input, Stream output, IClock clock, CancellationToken cancellation)
```
ManickaP, line 28: Should we rename `StreamCopier` to something that would tie it together with `HttpProxy`?
It shouldn't be used for anything else than copying proxy contents now, if I understand this change properly.

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
MihaZupan: Yes, `IncrementingPollingCounter` has an internal field for last-value that it substracts, so it only reports the change in the last interval.

src/ReverseProxy/Telemetry/ProxyTelemetry.cs
MihaZupan: Removed by mistake in 9ed69c0cd0da621b277d6c3ae52e5161b85ccd86, added it back now.

src/ReverseProxy/Service/Proxy/StreamCopier.cs
MihaZupan: > It shouldn't be used for anything else than copying proxy contents now, if I understand this change properly.

Correct.

> Should we rename StreamCopier

It's also going to be using Pipes in the future, so maybe just `ContentCopier`?
I can make the change in a future PR if we want to change the name

ManickaP: I like the `ContentCopier` name! But I don't insist on changing it here in this PR, we can do it later.

davidfowl: BodyCopier

ManickaP (approved):
LGTM

PR #505, opened 2020-10-30
Remove signals

### LINKED ISSUE #504: Remove Signals
We've reviewed our usage of signals and decided that we're not making sufficient use of them to justify their complexity. The only place we make derived signal updates is for health checks and that can be managed by health checks directly.

Tratcher:
Fixes #504

- Signals have been removed to reduce overall project complexity
  - We weren't using them right in a lot of cases anyways
- When a component needs to change state, it does so, but then it needs to manually signal to interested parties to re-compute their own state.
- I've changed how Destination state is tracked. It used to be updated by replacing the whole state object, but that process was racy and could cause unrelated changes to be dropped. E.g. updating active and passive health at the same time might cause one of the two to be overwritten.
  - Now Destination health state is directly mutable volatile properties (I didn't see a need for snapshots at this layer). The consumer (ClusterInfo) is the one responsible for composing the active and passive states when making decisions.
  - This also helped fix another issue where a destination would never recover if the active health checks were disabled while it was marked unhealthy. Now, health state for disabled checks is ignored when computing the composite state.

There's one outstanding update to the ReactivationScheduler where it needs access to the ClusterInfo in order to signal the health change. @alnikola do you want to take care of that in your PR when you complete the passive health checks work?


src/ReverseProxy/Service/Management/ItemManagerBase.cs
```diff
@@ -98,9 +96,9 @@ protected virtual void OnItemChanged(T item, bool added)
         protected virtual void OnItemRemoved(T item)
         {}
 
-        private void UpdateSignal()
+        private void UpdateSnapshot()
         {
-            _signal.Value = _items.Select(kvp => kvp.Value).ToList().AsReadOnly();
+            _snapshot = _items.Select(kvp => kvp.Value).ToList().AsReadOnly();
```
davidfowl, line 101: Why are we still doing this?

src/ReverseProxy/Service/Management/IItemManager.cs
```diff
@@ -15,10 +14,9 @@ internal interface IItemManager<T>
         where T : class
```
davidfowl, line 14: Do we even need this type anymore? It's basically an observable collection using signals.

src/ReverseProxy/Service/Management/ItemManagerBase.cs
Tratcher: Because _items isn't thread safe so we create a snapshot under the lock. What did you have in mind? Switch to ConcurrentDictionary?

src/ReverseProxy/Service/Management/IItemManager.cs
Tratcher: Maybe not? The most interesting question about changing it would be how to wire up the change notifications.

Changing it seems out of scope for this PR.

src/ReverseProxy/Service/HealthChecks/ReactivationScheduler.cs
```diff
@@ -37,8 +37,9 @@ private Task Reactivate(DestinationInfo destination)
             var state = destination.DynamicState;
             if (state.Health.Passive == DestinationHealth.Unhealthy)
             {
-                destination.DynamicState = new DestinationDynamicState(state.Health.ChangePassive(DestinationHealth.Unknown));
+                destination.DynamicState.Health.Passive = DestinationHealth.Unknown;
                 Log.PassiveDestinationHealthResetToUnkownState(_logger, destination.DestinationId);
+                // TODO: Update cluster by calling cluster.UpdateDynamicState()
```
Tratcher, line 42: @alnikola do you want to take care of this as part of your passive health checks PR? The ReactivationScheduler needs to pass in the ClusterInfo along with the DestinationInfo so it can trigger updates.

src/ReverseProxy/Service/Management/ItemManagerBase.cs
halter73: I think making _items a ConcurrentDictionary makes some sense. Even if we do that, we'd want to take snapshots assuming that we read the values more than we update them. Either way, couldn't this be rewritten to the following?

```suggestion
            _snapshot = GetItems().AsReadOnly();
```

This makes me wonder why GetItems() exists.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -23,82 +20,67 @@ namespace Microsoft.ReverseProxy.RuntimeModel
...
+
+                healthyDestinations = allDestinations.Where(destination =>
+                {
+                    // Only consider the current state if those checks are enabled.
+                    var state = destination.DynamicState;
+                    var active = activeEnabled ? state.Health.Active : DestinationHealth.Unknown;
+                    var passive = passiveEnabled ? state.Health.Passive : DestinationHealth.Unknown;
+
+                    // Filter out unhealthy ones. Unknown state is OK, all destinations start that way.
+                    return passive != DestinationHealth.Unhealthy && active != DestinationHealth.Unhealthy;
```
halter73, line 79: With this unsynchronized access of Active and Passive health, an earlier call to UpdateDynamicState() with stale health info could get pre-empted (let's say the first call to UpdateDynamicState() gets preempted between assigning healthyDestinations and _dynamicState for instance for the sake of example) and clobber the _dynamicState set by a later a later call to UpdateDynamicState() with fresh health data. _dynamicState and the DestinationHealth fields being volatile doesn't save you from this kind of race.

Since this data is cached, if there are no further calls to UpdateDynamicState() (which AFAIK is possible if there are no further health state changes), the stale _dynamicState could be used indefinitely.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -23,82 +20,67 @@ namespace Microsoft.ReverseProxy.RuntimeModel
...
+                }).ToList().AsReadOnly();
+            }
+
+            _dynamicState = new ClusterDynamicState(allDestinations, healthyDestinations);
```
halter73, line 83: With this unsynchronized, cached calculation of healthy destinations, an earlier call to UpdateDynamicState() with stale health info could get pre-empted (let's say the first call to UpdateDynamicState() gets preempted between assigning healthyDestinations and _dynamicState for instance for the sake of example) and clobber the _dynamicState set by a later a later call to UpdateDynamicState() with fresh health data. The _dynamicState and DestinationHealth fields being volatile doesn't save you from this kind of race.

Since this data is cached, if there are no further calls to UpdateDynamicState() (which AFAIK is possible if there are no further health state changes), the stale _dynamicState could be used indefinitely.

src/ReverseProxy/Service/RuntimeModel/DestinationDynamicState.cs
```diff
@@ -5,11 +5,6 @@ namespace Microsoft.ReverseProxy.RuntimeModel
 {
     public sealed class DestinationDynamicState
     {
-        public DestinationDynamicState(CompositeDestinationHealth health)
-        {
-            Health = health;
-        }
-
-        public CompositeDestinationHealth Health { get; }
+        public DestinationHealthState Health { get; } = new DestinationHealthState();
```
halter73, line 8: Why does DestinationDynamicState still exist if DestinationHealthState is mutable? Can't we just replace usages of DestinationDynamicState with direct usages of DestinationHealthState now?

src/ReverseProxy/Service/Management/ItemManagerBase.cs
davidfowl: We don't want to allocate a new list when items change, so we need something concurrent or we need to lock and mutate.

src/ReverseProxy/Service/Management/IItemManager.cs
davidfowl: I'll file an issue. We don't use change notifications anywhere... But it is out of scope.

src/ReverseProxy/Service/Management/ItemManagerBase.cs
Tratcher: I actually expect more changes than reads here. E.g. Reads are only triggered when we're pretty sure something has already changed, and we may change more than one thing before the read happens.

Note we end up making a copy regardless when the health checks are enabled because we filter the list. I'll try changing to a ConcurrentDictionary since the contention should be relatively low (e.g. not updated per-request, and only on background threads).

src/ReverseProxy/Service/Management/IItemManager.cs
Tratcher: We use change notifications in the active health checks.
https://github.com/microsoft/reverse-proxy/blob/fa71a5fcdf2fc33bc08038600aace3b8e828b058/src/ReverseProxy/Service/HealthChecks/ActiveHealthCheckMonitor.cs#L69

src/ReverseProxy/Service/RuntimeModel/DestinationDynamicState.cs
Tratcher: In theory there can be more dynamic state here besides health. I can flatten it for now though.

alnikola: I just realized that if we keep changing the public health check API here, I won't be able to correctly implement samples in [the documentation PR](https://github.com/microsoft/reverse-proxy/pull/509) until after this PR has been merged and the API is stabilized. Could we minimize such changes so I can work on the docs PR in parallel?

src/ReverseProxy/Service/Management/ItemManagerBase.cs
davidfowl: That's bad, but we can do these changes in stages. We don't want GC pressure when clusters frequently update (we don't have any performance scenarios that test those sorts of scenarios).

src/ReverseProxy/Service/HealthChecks/ReactivationScheduler.cs
alnikola: Yes, I can do it in #499, but it would mean the code needs to be merged in master in a non-working state to complete this PR.

src/ReverseProxy/Service/Management/IItemManager.cs
davidfowl: The `IClusterChangeListener` can be implemented without the IItemManager\<....\> types.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
alnikola: As long as the atomicity is ensured, it doesn't really matter which of the concurrent operations wins - the one started 10 ms sooner or 10 ms later. If we ended up with a stale value there are 2 cases. The first is enabled health checks. In this case, health state will be eventually overwritten by an up-to-date value. The second is disabled health checks. In this case, we will simply ignore the stale health state.

alnikola (commented):
Looks good overall. Besides one open issue with excessive allocations in ItemManagerBase.

src/ReverseProxy/Service/Management/ItemManagerBase.cs
davidfowl: I *just* re-read the code and remembered how these updates happen. It seems like we'd be better off with an immutable dictionary instead of the whole IItemManager thing. We can change the internal implementation for now then look to change calling code once we're done with that.

src/ReverseProxy/Service/Management/IItemManager.cs
alnikola: Are you suggesting to store `object`s? Is the performance of virtual generic methods the main concern here?

davidfowl: No I don't think any of this needs to change for this PR but the IItemManager originally existed because it exposed signals. Now it's just a collection with snapshots that I think doesn't need to exist. We can fire these change events from the ProxyConfigurationManager directly.

alnikola: But we do still need sometimes to read the whole Items collection. [ProxyConfigManager.InitialLoadAsync](https://github.com/microsoft/reverse-proxy/blob/fa71a5fcdf2fc33bc08038600aace3b8e828b058/src/ReverseProxy/Service/Management/ProxyConfigManager.cs#L103) is an example.

davidfowl: It's not *an* example, it's literally the only example.

src/ReverseProxy/Service/RuntimeModel/DestinationDynamicState.cs
davidfowl: The API won't be stable until we're done with 1.0. We haven't even done an API review as yet. More things will be renamed.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -23,82 +20,67 @@ namespace Microsoft.ReverseProxy.RuntimeModel
     /// </remarks>
     public sealed class ClusterInfo
     {
-        private readonly DelayableSignal<Unit> _destinationsStateSignal;
+        private volatile ClusterDynamicState _dynamicState = new ClusterDynamicState(Array.Empty<DestinationInfo>(), Array.Empty<DestinationInfo>());
```
davidfowl, line 23: Can we rename `ClusterDynamicState` to destinations or destination state?

src/ReverseProxy/Service/Management/IItemManager.cs
alnikola: There is [another one](https://github.com/microsoft/reverse-proxy/blob/fa71a5fcdf2fc33bc08038600aace3b8e828b058/src/ReverseProxy/Service/Management/ProxyConfigManager.cs#L334).

davidfowl: That's in the `ProxyConfigManager` itself, the thing doing the rebuild of the route data source. It's literally looping over things it just added. I think the meta point is that these collections don't need to exist outside of the ProxyConfigManager, at least currently. If we design something that needs to expose the collection we can but to date what we have is something that was written to allow subscribing to changes via signals and instead we have the `IClusterChangeListener` so it really doesn't need to exist.

src/ReverseProxy/Service/RuntimeModel/DestinationDynamicState.cs
alnikola: I mean stable for a period which is enough to complete, review and merge that PR. :)

alnikola: So we can then consistently rename everything in master.

src/ReverseProxy/Service/Management/IItemManager.cs
alnikola: Yes, I do agree. My point only was that **sometimes** it can be convenient to explicitly enumerate all the items.

davidfowl: Ah yes, that's fine to expose if we need it.

src/ReverseProxy/Service/HealthChecks/ReactivationScheduler.cs
Tratcher: ReactivationScheduler is only used by the passive checks you're still working on, so that seems OK.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
Tratcher: Hmm, health state can change even as we scan the list to filter out the unhealthy destinations. I see your point though, even if the state changed on a destination after we scanned it, that's fine so long as there's another complete scan scheduled. Maybe I should put a lock around this to make sure the scans are at least sequential.

alnikola: I'd suggest to treat each destination's health state independently. We need to only ensure that the health state is read exactly once per scan per destination, but I don't see any real issues with destination's health being updated while we are scanning.

Tratcher: Scenario:
- Active checks update health and trigger a list update
- Scan 1 starts.
- Passive checks update health and trigger a list update
- Scan 2 starts.
- Scan 2 finishes first for threading reasons.
- Scan 1 finishes.

It doesn't matter if scan 1 picks up the changes from the passive checks or not. The concern with the above is that scan 1 might finish after scan 2 for threading reasons and the passive check updates picked up by scan 2 might be overwritten by the results from scan 1 that missed them. The lock avoids this by making sure scan 2 doesn't start until scan 1 finishes.

src/ReverseProxy/Service/RuntimeModel/DestinationDynamicState.cs
Tratcher: The plan was to get this PR merged before you finished passive health checks, right? You're already going to need to update the docs and samples as part of the passive checks, or hold them until the passive work is finished.

I'm more concerned about getting things into the shape we want them than about some short term conflicts. If the docs PR gets merged first then I'll plan to react to it here.

src/ReverseProxy/Service/Management/ItemManagerBase.cs
Tratcher: #513 to track cleaning up the IItemManager types.

src/ReverseProxy/Service/Management/IItemManager.cs
Tratcher: #513 for IItemManager cleanup

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
Tratcher: Hmm, it's cluster state that contains destinations. What if we trimmed it to `ClusterState`?

davidfowl: Maybe we'll come up with better names when we do the folder shuffle.

Tratcher:
I think this is ready for a final review.

src/ReverseProxy/Service/Management/ItemManagerBase.cs
```diff
@@ -14,10 +12,10 @@ internal abstract class ItemManagerBase<T> : IItemManager<T>
     {
         private readonly object _lockObject = new object();
         private readonly Dictionary<string, T> _items = new Dictionary<string, T>(StringComparer.OrdinalIgnoreCase);
-        private readonly Signal<IReadOnlyList<T>> _signal = SignalFactory.Default.CreateSignal<IReadOnlyList<T>>(new List<T>().AsReadOnly());
+        private volatile IReadOnlyList<T> _snapshot = Array.Empty<T>();
 
         /// <inheritdoc/>
-        public IReadableSignal<IReadOnlyList<T>> Items => _signal;
+        public IReadOnlyList<T> Items => _snapshot;
```
halter73, line 18: Is there any reason to use GetItems() over this getter? Can we delete GetItems()?

src/ReverseProxy/Service/Management/ItemManagerBase.cs
halter73, line 101: Nit:
```suggestion
            _snapshot = _items.Values.ToList().AsReadOnly();
```

halter73 (approved):

src/ReverseProxy/Service/Management/ItemManagerBase.cs
Tratcher: The current proposal is to replace this whole class (#513) so I'm not going to mess with it.

davidfowl:
It's worth noting @Tratcher that we still have a chokepoint at the `_stateLock` on each cluster when we update the dynamic state. This will result

Tratcher:
Yes, if there's a lot of churn per cluster for passive health checks. The first concern would be to make sure the updates from passive checks are queued to a background thread. From there, I'd be curious to see how many destinations and how much health churn you need before the contention becomes a problem.

davidfowl:
I think we still need to do debouncing for passive health checks

Tratcher:
At what level? Debouncing changes per destination, or debouncing when re-building the list?

I'd expect destination level debouncing to be done within the policy itself, though there is understandably a lot of parallel activity there.

Debouncing the list rebuild could be done either within UpdateDynamicState or in a shared caller

halter73:
Debouncing at both levels might make sense.

PR #511, opened 2020-11-03
Fix CaseSensitiveEqualHelper null check

MihaZupan:
Randomly saw this while browsing through code, not related to any PR


Tratcher (approved):
Good catch.

PR #518, opened 2020-11-05
ITimer abstracts out a real Timer

### LINKED ISSUE #508: Add ITimer abstraction to stabilize unit tests
`EntityActionScheduler` tests rely on an actual `Timer` implementation and therefore are time-dependent. This causes a noticeable flakiness of CI pipeline, so it's suggested to introduce a `ITimer` abstraction and migrate the tests to it in order to stabilize them.

alnikola:
This PR adds `ITimer` abstraction and makes all timer-dependent tests use it to improve their stability and performance.

Fixes #508


davidfowl (approved):
LGTM

Tratcher (commented):
Much better, thanks.

PR #521, opened 2020-11-08
Implement manual configuration binding for proxy config

### LINKED ISSUE #481: Use explicit config binding
@davidfowl feedback:

There are currently three data models in the code base for the same structures (route, cluster, destination, etc.).
A) The runtime model used when processing requests. This is read-only, but does allow dynamic updates.
B) The code based configuration model
C) The `*Data` types like `ClusterData` used when binding from IConfiguration. After binding these are manually converted to the (B) model.

Problems:
- It's hard to maintain this many models
- Dynamic IConfiguration binding is not linker friendly, any property on the data types referenced only by reflection would have its setter trimmed.
- The dynamic data binding might not be as flexible as we want it to be. E.g. We might want to allow single values or arrays for properties like hosts and methods.

Proposal:
- Remove the `*Data` types.
- Manually read the IConfiguration entries and write directly to the code based configuration model. That's almost what we do already when converting between models:
https://github.com/microsoft/reverse-proxy/blob/7a6f66ccf886dfb331d607c8aa498a0c80e58856/src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs#L136-L157

davidfowl:
- Removed the intermediate *Data types. This makes it clear that there are really 2 models, the input configuration and the runtime model the proxy uses.
- Removed use of IOptionMonitor\<T\> and bind directly to configuration due to inability to handle parsing or validation errors.

Fixes #481

PS: This change lead me to file this issue https://github.com/dotnet/runtime/issues/44381


test/ReverseProxy.Tests/Configuration/ConfigurationConfigProviderTests.cs
```diff
@@ -536,71 +541,88 @@ public void CachedCertificateIsDisposed_RemoveItFromCache()
             Assert.Same(certificateConfig[4], cachedCertificates[2]);
         }
 
+        private void TriggerOnChange(IConfigurationRoot configurationRoot)
+        {
+            // This method is protected so we use reflection to trigger it. The alternative is to wrap or implement
+            // a custom configuration provider and expose OnReload as a public method
+            var reload = typeof(ConfigurationProvider).GetMethod("OnReload", BindingFlags.NonPublic | BindingFlags.Instance);
```
davidfowl, line 548: This is slightly hacky but it let me skip writing a custom configuration provider that replicates the in memory provider 😄. I can if this feels too hacky though...

alnikola: Since it's in tests, I think it's fine as long as it saves redundant code lines.

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -120,225 +127,247 @@ private void UpdateSnapshot(ConfigurationData data)
...
             {
                 return null;
             }
 
             return new CircuitBreakerOptions
             {
-                MaxConcurrentRequests = data.MaxConcurrentRequests,
-                MaxConcurrentRetries = data.MaxConcurrentRetries,
+                MaxConcurrentRequests = int.TryParse(section[nameof(CircuitBreakerOptions.MaxConcurrentRequests)], out var maxConcurrentRequests) ? maxConcurrentRequests : default,
+                MaxConcurrentRetries = int.TryParse(section[nameof(CircuitBreakerOptions.MaxConcurrentRetries)], out var maxConcurrentRetries) ? maxConcurrentRetries : default,
```
Tratcher, line 238: InvariantCulture, NumberStyles.None

Make a helper for this?

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -120,225 +127,247 @@ private void UpdateSnapshot(ConfigurationData data)
...
-            if (data == null)
+            if (section?.Exists() is false)
             {
                 return null;
             }
 
             return new LoadBalancingOptions
             {
-                Mode = Enum.Parse<LoadBalancingMode>(data.Mode),
+                Mode = Enum.TryParse<LoadBalancingMode>(section[nameof(LoadBalancingOptions.Mode)], out var mode) ? mode : default,
```
Tratcher, line 280: Should we use a more aggressive failure policy than TryParse? E.g. Parse? We already throw exceptions if we can't load certificates, why not fail if we can't parse the int's, enums, etc? Silent failures here seem impossible to troubleshoot.

Tratcher (commented):
Don't all of the int, double, and TimeSpan TryParse calls need culture and style specified?

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
davidfowl: I don't have an issue with that. I think I initially had that and ended up with regressions but I can do it again. Only parse if not-null.

davidfowl:
Updated

src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        internal static bool ReadBool(this IConfiguration configuration, string name)
+        {
+            return configuration[name] is string value ? bool.Parse(value) : default;
+        }
+
+        internal static Version ReadVersion(this IConfiguration configuration, string name)
+        {
+            return configuration[name] is string value ? Version.Parse(value + (value.Contains('.') ? "" : ".0")) : default;
```
Tratcher, line 37: The `is` here handles the null check right? This code is going to fail if config returns an empty string instead, it will try to parse ".0".

Tratcher: Version is a class, return null, not default.

src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        internal static TEnum ReadEnum<TEnum>(this IConfiguration configuration, string name) where TEnum : struct
+        {
+            return configuration[name] is string value ? Enum.Parse<TEnum>(value, ignoreCase: true) : default;
+        }
+
+        internal static bool ReadBool(this IConfiguration configuration, string name)
+        {
+            return configuration[name] is string value ? bool.Parse(value) : default;
+        }
```
Tratcher, line 33: Many of the cases where you're parsing primitives should be returning nullables rater than returning the default (0, false, etc.). In fact, I think we should clarify our policy so that all options properties should be nullable and not set by default. That lets the implementation decide what default to apply if none is specified, especially since many of the implementations will replaceable.

E.g. this should be nullable and not set. Let the factory choose the default.
https://github.com/microsoft/reverse-proxy/blob/6d04918bfe18827d7f64d8b12f437faa209db648/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ProxyHttpClientOptions.cs#L14

Tratcher (commented):
Delete the new ProxyHttpRequestData now that you've rebased.

src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
davidfowl: > The is here handles the null check right? This code is going to fail if config returns an empty string instead, it will try to parse ".0".

Why shouldn't it throw. default is null.

src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
davidfowl: I can do this in a different PR.

Tratcher (approved):

src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
Tratcher: https://github.com/microsoft/reverse-proxy/issues/531

src/ReverseProxy/Configuration/CertificateConfigLoader.cs
```diff
@@ -23,13 +23,25 @@ public CertificateConfigLoader(IWebHostEnvironment hostEnvironment)
         }
 
         /// <inheritdoc/>
-        public X509Certificate2 LoadCertificate(CertificateConfigData certificateConfig)
+        public X509Certificate2 LoadCertificate(IConfigurationSection certificateSection)
         {
-            if (certificateConfig is null)
+            if (certificateSection?.Exists() is false)
```
halter73, line 28: This PR uses `is false` in a bunch of places when it really wants to check if the `bool?` `!= true`.

```suggestion
            if (certificateSection?.Exists() != true)
```

davidfowl: I'm just going to remove all of these nullable checks. The section is never null anyways.

PR #522, opened 2020-11-09
fix for issue 517, ProxyRouteTransformExtension methods failing when the ProxyRoute.Transforms property is null

### LINKED ISSUE #517: Adding transform on route in code gives Null Reference Exception when not defining empty transform first
### Describe the bug
Adding transform on route in code gives Null Reference Exception when not defining an empty transform first. The code below (other route properties left out for clarity) gives a null reference exception since the "Transforms" property is not initialized. A workaround is to specify the an empty List (uncomment the line in the code below).

### To Reproduce

```
var route = new ProxyRoute()
{
    RouteId = "route1",
    // Transforms = new List<IDictionary<string, string>>(),
};
route.AddTransformPathPrefix("/");
```

### Further technical details

- Latest preview
- .Net 5

jrunyen:
Implementing fix discussed in [issue 517](https://github.com/microsoft/reverse-proxy/issues/517). These changes will initialize the ProxyRoute.Transforms property in each of the extension methods if it is null. fixes #517

### DIFF
--- a/src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
+++ b/src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
@@ -11,6 +11,8 @@ namespace Microsoft.ReverseProxy.Service
         /// </summary>
         public static void AddTransformPathSet(this ProxyRoute proxyRoute, PathString path)
         {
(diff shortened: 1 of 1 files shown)


Tratcher (approved):

Tratcher:
Thanks

PR #523, opened 2020-11-09
Performing HttpClient vs HttpMessageInvoker check in HttpProxy

### LINKED ISSUE #458: HttpClient vs HttpMessageInvoker buffering issue
We decided to use HttpMessageInvoker with IHttpProxy and IProxyHttpClientFactory to cut out some unnecessary layers for a modest perf benefit. However, because HttpClient derives from HttpMessageInvoker customers can still return it from the factory and/or pass it to IHttpProxy. They might do this if they were for example using the IHttpClientFactory API to produce their client, or just out of habit.

Problem:
When HttpClient overrides HttpMessageInvoker.SendAsync it changes the response behavior to fully buffered (HttpCompletionOption.ResponseContentRead) and there's no way to override that from the HttpMessageInvoker API. Fully buffering responses would have significant impacts on the proxy performance and functionality including:
- High memory usage
- Increased latency
- Timeouts would apply to the full response body, not just the headers
- Broken streaming scenarios like gRPC and WebSockets

Possible Mitigations:
A) Doc guidance. We already document several other things a developer could mess up when creating their own client like these:
```c#
            new SocketsHttpHandler()
            {
                UseProxy = false,
                AllowAutoRedirect = false,
                UseCookies = false
            }
```
Fully buffering is worse than any of those, but calling it out in the docs might be a good start.
B) Downcast check for HttpClient and hard fail.
C) Downcast check for HttpClient and call the non-buffering overload of SendAsync. That makes the scenario work, but incurs some (small?) performance overhead even if you aren't using HttpClient.

Recommendation: A (docs guidance) should be done regardless. We could do B initially and wait to see if we get feedback that people require C instead.

jrunyen:
As discussed in issue 458, the HttpClient SendAsync overload changes the response buffering behavior which results in a performance impact. This change performs a downcast check on the httpClient parameter and hard fails if it is of type HttpClient. 

Fixes #458

### DIFF
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -95,13 +95,14 @@ namespace Microsoft.ReverseProxy.Service.Proxy
         {
             _ = context ?? throw new ArgumentNullException(nameof(context));
             _ = destinationPrefix ?? throw new ArgumentNullException(nameof(destinationPrefix));
+            _ = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
             _ = proxyOptions ?? throw new ArgumentNullException(nameof(proxyOptions));
 
             // HttpClient overload for SendAsync changes response behavior to fully buffered which impacts performance
             // See discussion in https://github.com/microsoft/reverse-proxy/issues/458
-            if (httpClient is HttpClient || httpClient == null)
+            if (httpClient is HttpClient)
             {
-                throw new ArgumentNullException(nameof(httpClient));
+                throw new InvalidOperationException($"{nameof(httpClient)} should be of type HttpMessageInvoker, not HttpClient");
             }
 
             ProxyTelemetry.Log.ProxyStart(destinationPrefix);
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -102,7 +102,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
             // See discussion in https://github.com/microsoft/reverse-proxy/issues/458
             if (httpClient is HttpClient)
             {
-                throw new ArgumentException($"{nameof(httpClient)} should be of type HttpMessageInvoker, not HttpClient");
+                throw new ArgumentException($"The http client must be of type HttpMessageInvoker, not HttpClient", nameof(httpClient));
             }
 
             ProxyTelemetry.Log.ProxyStart(destinationPrefix);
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -98,6 +98,13 @@ public HttpProxy(ILogger<HttpProxy> logger, IClock clock)
             _ = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
             _ = proxyOptions ?? throw new ArgumentNullException(nameof(proxyOptions));
 
+            // HttpClient overload for SendAsync changes response behavior to fully buffered which impacts performance
+            // See discussion in https://github.com/microsoft/reverse-proxy/issues/458
+            if (httpClient is HttpClient)
+            {
+                throw new InvalidOperationException($"{nameof(httpClient)} should be of type HttpMessageInvoker, not HttpClient");
```
Tratcher, line 105: ArgumentException?

Tratcher (commented):
Please add a regression test as well.

jrunyen:
Probably should have had the test on there from the get go, my bad... Agreed on the exception type, fixed that as well. Let me know if it all checks out.

test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs
```diff
@@ -1551,6 +1551,19 @@ public async Task ProxyAsync_UpgradableRequest_ResponseBodyCopyError_CancelsRequ
             events.AssertContainProxyStages(upgrade: true);
         }
 
+        [Fact]
+        public async Task ProxyAsync_WithHttpClient_Fails()
```
alnikola, line 1555: Why is it called **WinHttpClient**_Fails whereas it creates a default HttpClient relying internally on a managed `SocketsHttpHandler`?

Tratcher: @alnikola that's not "Win", it's "With". This is fine.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -98,6 +98,13 @@ public HttpProxy(ILogger<HttpProxy> logger, IClock clock)
             _ = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
             _ = proxyOptions ?? throw new ArgumentNullException(nameof(proxyOptions));
 
+            // HttpClient overload for SendAsync changes response behavior to fully buffered which impacts performance
+            // See discussion in https://github.com/microsoft/reverse-proxy/issues/458
+            if (httpClient is HttpClient)
+            {
+                throw new ArgumentException($"{nameof(httpClient)} should be of type HttpMessageInvoker, not HttpClient");
```
Tratcher, line 105:
```suggestion
                throw new ArgumentException($"The http client must be of type HttpMessageInvoker, not HttpClient", nameof(httpClient));
```

Tratcher (approved):

Tratcher:
Thanks

PR #539, opened 2020-11-12
Rebuild docs for preview7

### LINKED ISSUE #458: HttpClient vs HttpMessageInvoker buffering issue

### LINKED ISSUE #515: Documentation: not all "Improve this Doc" links are working
### Describe the bug
Minor issue, but not all "Improve this Doc" links on the documentation (https://microsoft.github.io/reverse-proxy/articles) are working, those links resolve to a 404.

### To Reproduce

- Go to urls below and click the "Improve this Doc":

1. https://microsoft.github.io/reverse-proxy/articles/configfiles.html
2. https://microsoft.github.io/reverse-proxy/articles/configproviders.html
3. https://microsoft.github.io/reverse-proxy/articles/header-routing.html
4. https://microsoft.github.io/reverse-proxy/articles/transforms.html

Tratcher:
- Fixed up links to types that have changed like ClusterData.
- Added the warning about HttpClient, fixes #458
- General rebuild to pick up preview7 changes.
- Make sure the edit links point to `release/docs`, fixes #515

The interesting changes can be reviewed by looking at the separate commits.


Tratcher:
Merging to complete the release.

PR #542, opened 2020-11-13
Merge RequestOptions types alternative.

### LINKED ISSUE #525: Merge `RequestProxyOptions` with `ClusterProxyHttpRequestOptions`
### Motivation
Reduce the number of types representing the very same thing

Discussed offline: we'll want to see how we can unify this with RequestProxyOptions so we don't have to allocate that per request. The big differences are:
- RequestProxyOptions has Transforms on it which come from routeConfig.
- RequestProxyOptions has the Timeout which should be easy to move/mirror.
- struct vs class, readonly vs mutable
- Naming

Some possibilities:
- We come up with a type that works in both scenarios and use that on ClusterConfig.
- RequestProxyOptions becomes a readonly struct so at least we don't have to allocate it. We'd still end up copying the paramters across.
- Transforms get passed into IHttpProxy.ProxyAsync as a separate parameter, making the options types more compatible.

Follow up to #512 PR: https://github.com/microsoft/reverse-proxy/pull/512/files#r517037962

ManickaP:
This is an alternative to #534.

Instead of using `ClusterProxyHttpRequestOptions` in `RequestProxyOption`, this PR splits transforms from `RequestProxyOption`. And then uses `RequestProxyOption` in cluster configuration.

Also renames RequestTimeout --> Timeout. As it was pointed out that the name is redundant.

Fixes #525


Tratcher (approved):
Much better.

Also update the docs at https://github.com/microsoft/reverse-proxy/blame/dd6aa43513c5832583df15000942fadf37ceba89/docs/docfx/articles/direct-proxying.md#L55

Tratcher:
Remember to set the milestone on the PR and the issue.

PR #546, opened 2020-11-13
Remove pre-built docs

Tratcher:
#428 hasn't been completely resolved, but we expect it to be before we plan to do another release. In the meantime these local doc artifacts are annoying. Worst case we can bring them back in the release/docs branch if we need them.


Tratcher:
*ping*

halter73 (approved):

Issue #565, opened 2020-11-20
YARP config doesn't support a negative order


NGloreous:
### Describe the bug
If you try to provide a negative value for a proxy route order, config load fails with a InvalidOperationException due to a FormatException when parsing the order.

The issue is in ConfigurationReadingExtensions.cs, ReadInt32 calls int.Parse passing in NumberStyles.None which doesn't allow negative values. 

```c#
        internal static int? ReadInt32(this IConfiguration configuration, string name)
        {
            return configuration[name] is string value ? int.Parse(value, NumberStyles.None, CultureInfo.InvariantCulture) : null;
        }
```

Since the default order is 0, it would be nice to be able to specify negative values so some routes take precedence over the routes with out order specified. 

### To Reproduce

Try using the following config:

```JSON
{
  "ReverseProxy": {
    "Routes": [
      {
        "RouteId": "route1",
        "ClusterId": "cluster1",
        "Order": -1,
        "Match": {
          "Path": "{**catch-all}"
        },
      }
    ],
    "Clusters": {
      "cluster1": {
        "Destinations": {
          "cluster1/destination1": {
            "Address": "https://example.com/"
          }
        }
      }
    }
  }
}
```

```
System.InvalidOperationException: Unable to load or apply the proxy configuration.
       ---> System.FormatException: Input string was not in a correct format.
         at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
         at System.Number.ParseInt32(ReadOnlySpan`1 value, NumberStyles styles, NumberFormatInfo info)
         at System.Int32.Parse(String s, NumberStyles style, IFormatProvider provider)
         at Microsoft.Extensions.Configuration.ConfigurationReadingExtensions.ReadInt32(IConfiguration configuration, String name)
         at Microsoft.ReverseProxy.Configuration.ConfigurationConfigProvider.CreateRoute(IConfigurationSection section)
         at Microsoft.ReverseProxy.Configuration.ConfigurationConfigProvider.UpdateSnapshot()
         at Microsoft.ReverseProxy.Configuration.ConfigurationConfigProvider.GetConfig()
         at Microsoft.ReverseProxy.Service.Management.ProxyConfigManager.InitialLoadAsync()
```

Tratcher:
This was a regression when we refactored the config bindings. Should be an easy fix.

PR #566, opened 2020-11-20
Support negative number in config

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/565

### DIFF
--- a/src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
+++ b/src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Extensions.Configuration
     {
         internal static int? ReadInt32(this IConfiguration configuration, string name)
         {
-            return configuration[name] is string value ? int.Parse(value, NumberStyles.None, CultureInfo.InvariantCulture) : null;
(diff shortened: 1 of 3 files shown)


src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
```diff
@@ -9,7 +9,7 @@ internal static class ConfigurationReadingExtensions
     {
         internal static int? ReadInt32(this IConfiguration configuration, string name)
         {
-            return configuration[name] is string value ? int.Parse(value, NumberStyles.None, CultureInfo.InvariantCulture) : null;
+            return configuration[name] is string value ? int.Parse(value, NumberStyles.AllowLeadingSign, CultureInfo.InvariantCulture) : null;
```
Kahbazi, line 12: Is `NumberStyles.AllowLeadingSign` good enough, or Should it be `NumberStyles.Integer` which has `AllowLeadingWhite`, `AllowTrailingWhite`, and `AllowLeadingSign`?

Tratcher (commented):
Nice. Can you also update the test for Order to use a negative value?
https://github.com/microsoft/reverse-proxy/blob/b34e66a4394e6c313f6f88331d049edc6fadadd3/test/ReverseProxy.Tests/Configuration/ConfigurationConfigProviderTests.cs#L307

Tratcher (approved):

Tratcher:
Thanks

PR #567, opened 2020-11-23
Fix table layout in transforms.md doc

jboelter:


Tratcher (approved):

Tratcher:
Thanks for the fix. This won't be visible immediately due to branching and #428, but I'll take care of those.

PR #571, opened 2020-11-24
Add Telemetry Consumption package

MihaZupan:
Opening for feedback on the design and implementation.

Includes APIs for 6 EventSources:
- Proxy (YARP)
- Kestrel
- Http
- Sockets
- NetSecurity
- NameResolution

For each of those, providing 4 components:
- `I{Name}TelemetryConsumer` - interface with methods mirroring EventSource's Events that are relevant within the scope of an HTTP request
- `{Name}Metrics` - Immutable POCO containing properties mirroring EventSource's EventCounters
- `I{Name}MetricsConsumer` - interface with a single method `void On{Name}Metrics(Metrics oldMetrics, Metrics newMetrics)`
- Internal `{Name}EventListenerService` - internal infrastructure for dispatching events and metrics, exposed as an extension method on `IReverseProxyBuilder`: `Add{Name}TelemetryListener`

The added tests are not exhaustive for every event and parameter.


src/Microsoft.ReverseProxy.TelemetryConsumption/IReverseProxyBuilderTelemetryExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            builder.Services.AddHttpContextAccessor();
+            builder.Services.TryAddHostedService<NetSecurityEventListenerService>();
+            return builder;
+        }
+
+        public static IReverseProxyBuilder AddSocketsTelemetryListener(this IReverseProxyBuilder builder)
+        {
+            builder.Services.AddHttpContextAccessor();
+            builder.Services.TryAddHostedService<SocketsEventListenerService>();
```
davidfowl, line 60: AddHostedService does this already.

MihaZupan: Removed the redundant helper

src/ReverseProxy.TelemetryConsumption/Proxy/ProxyEventListenerService.cs
```diff
@@ -0,0 +1,230 @@
...
+                var previous = _previousMetrics;
+                _previousMetrics = metrics;
+                _currentMetrics = new ProxyMetrics();
+
+                if (previous is null)
+                {
+                    return;
+                }
+
+                foreach (var consumer in _serviceProvider.GetServices<IProxyMetricsConsumer>())
```
WeihanLi, line 223: Should we consider for consuming in parallel?

MihaZupan: These callbacks execute inline within the request, so the expected scenario is to have very few (just 1) consumer and for it to be very cheap and execute quickly.
I don't believe we would gain anything from parallelizing consumers (and the cost of it would likely outweigh any benefit).

src/ReverseProxy.TelemetryConsumption/Kestrel/KestrelEventListenerService.cs
```diff
@@ -0,0 +1,190 @@
...
+                    metrics.CurrentUpgradedRequests = value;
+                    break;
+
+                default:
+                    return;
+            }
+
+            const int TotalEventCounters = 10;
+
+            if (++_eventCountersCount == TotalEventCounters)
```
alnikola, line 168: Could you please move the increment out of `if` expression? It seems a bit unusual.

src/ReverseProxy.TelemetryConsumption/Kestrel/KestrelEventListenerService.cs
```diff
@@ -0,0 +1,190 @@
...
+                var previous = _previousMetrics;
+                _previousMetrics = metrics;
+                _currentMetrics = new KestrelMetrics();
+
+                if (previous is null)
+                {
+                    return;
+                }
+
+                foreach (var consumer in _serviceProvider.GetServices<IKestrelMetricsConsumer>())
```
alnikola, line 183: If I understood it correctly, `IKestrelMetricsConsumer` will be resolved for each `EventWrittenEventArgs` which seems a lot. Would it be better to resolve it once and store on a field?

src/ReverseProxy.TelemetryConsumption/NameResolution/NameResolutionEventListenerService.cs
```diff
@@ -0,0 +1,160 @@
...
+            _httpContextAccessor = httpContextAccessor ?? throw new ArgumentNullException(nameof(httpContextAccessor));
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+        public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+        protected override void OnEventSourceCreated(EventSource eventSource)
+        {
+            if (eventSource.Name == "System.Net.NameResolution")
```
alnikola, line 37: There are plenty of string comparisons done without explicitly specifying a comparer to use. Is it considered safe enough?

src/ReverseProxy.TelemetryConsumption/NameResolution/NameResolutionEventListenerService.cs
```diff
@@ -0,0 +1,160 @@
...
+                    metrics.AverageLookupDuration = TimeSpan.FromMilliseconds(value);
+                    break;
+
+                default:
+                    return;
+            }
+
+            const int TotalEventCounters = 2;
+
+            if (++_eventCountersCount == TotalEventCounters)
```
alnikola, line 138: Could you please move the increment out of `if`?

src/ReverseProxy.TelemetryConsumption/NameResolution/NameResolutionEventListenerService.cs
```diff
@@ -0,0 +1,160 @@
...
+                var previous = _previousMetrics;
+                _previousMetrics = metrics;
+                _currentMetrics = new NameResolutionMetrics();
+
+                if (previous is null)
+                {
+                    return;
+                }
+
+                foreach (var consumer in _serviceProvider.GetServices<INameResolutionMetricsConsumer>())
```
alnikola, line 153: The same concern as above about resolving a service on each call.

src/ReverseProxy.TelemetryConsumption/NameResolution/NameResolutionMetrics.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.ReverseProxy.Telemetry.Consumption
+{
+    public sealed class NameResolutionMetrics
```
alnikola, line 8: Please, add comments to public API.

src/ReverseProxy.TelemetryConsumption/NetSecurity/INetSecurityMetricsConsumer.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Telemetry.Consumption
+{
+    public interface INetSecurityMetricsConsumer
```
alnikola, line 6: Please, add comments to public API.

src/ReverseProxy.TelemetryConsumption/NetSecurity/INetSecurityTelemetryConsumer.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Security.Authentication;
+
+namespace Microsoft.ReverseProxy.Telemetry.Consumption
+{
+    public interface INetSecurityTelemetryConsumer
```
alnikola, line 9: Please, add comments to public API.

alnikola (commented):
Some comments.

src/ReverseProxy.TelemetryConsumption/NameResolution/NameResolutionEventListenerService.cs
MihaZupan: We're going after Ordinal comparison so I would say yes

src/ReverseProxy.TelemetryConsumption/Kestrel/KestrelEventListenerService.cs
MihaZupan: For the metrics, we're resolving the consumers on each EventCounters update - so once a second, which doesn't seem problematic.

For events, we are resolving it for each event.
Since we're relying on DI for lifetime management, we have to do it each time. If we documented that we disallow transient consumer services (since they don't make much sense to use for events anyway), we could avoid the cost on each event and instead use an `AsyncLocal` to cache the resolved services.

MihaZupan:
@davidni Could you please look at this as well to make sure the public APIs are matching your expectations/needs

src/ReverseProxy.TelemetryConsumption/Microsoft.ReverseProxy.Telemetry.Consumption.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
Tratcher, line 4: What's the story for 3.1? Is is possible to collect anything there?

src/ReverseProxy.TelemetryConsumption/Microsoft.ReverseProxy.Telemetry.Consumption.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <OutputType>Library</OutputType>
+    <LangVersion>9.0</LangVersion>
+    <Version>0.0.1-dev</Version>
+    <AssemblyVersion>0.0.1.0</AssemblyVersion>
+    <FileVersion>0.0.1.0</FileVersion>
```
Tratcher, line 9:
```suggestion
```

test/ReverseProxy.FunctionalTests/TelemetryConsumptionTests.cs
```diff
@@ -0,0 +1,257 @@
...
+                    if (stopwatch.Elapsed > TimeSpan.FromSeconds(10))
+                    {
+                        throw new TimeoutException($"Timed out waiting for {name}");
+                    }
+                    await Task.Delay(10);
+                }
+            }
+        }
+
+        private sealed class MetricsConsumer :
```
Tratcher, line 233: How closely do these Telemetry and Metrics tests reflect the actual expected usage? If this isn't how we expect it to be used then please add a sample that shows the expected usage.

test/ReverseProxy.FunctionalTests/Common/TestEnvironment.cs
```diff
@@ -0,0 +1,132 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Server.Kestrel.Core;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.ReverseProxy.Abstractions;
+using Microsoft.ReverseProxy.Utilities.Tests;
+
+namespace Microsoft.ReverseProxy.Common
+{
+    public class TestEnvironment
```
ManickaP, line 19: Thanks for extracting this.

ManickaP (commented):
I commented only on the first set of consumer classes but the comments apply to all of them.

test/ReverseProxy.FunctionalTests/TelemetryConsumptionTests.cs
```diff
@@ -0,0 +1,257 @@
...
+            {
+                var httpClient = new HttpClient();
+                await httpClient.GetStringAsync($"{uri}/{guid}");
+            });
+
+            var stages = Assert.Single(consumers, c => c.PathAndQuery != null && c.PathAndQuery.Contains(guid)).Stages;
+
+            var expected = new[]
+            {
+                "OnRequestStart-Kestrel",
```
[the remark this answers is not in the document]
MihaZupan: Most of these could be `nameof`, but some are explicitly different since some names repeat.

Specifically both Runtime Http and Kestrel include `RequestStart` and `RequestStop` events.

src/ReverseProxy.TelemetryConsumption/Http/HttpEventListenerService.cs
```diff
@@ -0,0 +1,293 @@
...
+        public Task StartAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+        public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+        protected override void OnEventSourceCreated(EventSource eventSource)
+        {
+            if (eventSource.Name == "System.Net.Http")
+            {
+                var arguments = new Dictionary<string, string> { { "EventCounterIntervalSec", MetricsOptions.IntervalSeconds.ToString() } };
+                EnableEvents(eventSource, EventLevel.LogAlways, EventKeywords.None, arguments);
```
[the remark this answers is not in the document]
MihaZupan: We are pairing the events based on the shared `ExecutionContext` (specifically the `HttpContextAccessor`'s AsyncLocal).
So we're using the same mechanism `ActivityId`s are using to flow, we are just using it directily instead of going through the pain of processing Guids manually.

src/ReverseProxy.TelemetryConsumption/Microsoft.ReverseProxy.Telemetry.Consumption.csproj
MihaZupan: It's possible to collect the `Proxy` and `Kestrel` events (runtime only added them in 5.0).

For Kestrel, we would have to modify the interface as more fields were added to the event in 5.0.

I will add the 3.1 target as well, it shouldn't be too messy - just a few ifdefs.

src/ReverseProxy.TelemetryConsumption/Http/HttpEventListenerService.cs
```diff
@@ -0,0 +1,293 @@
...
+            }
+        }
+
+        protected override void OnEventWritten(EventWrittenEventArgs eventData)
```
shirhatti, line 45: `OnEventWritten` can run before the derived constructor on an EventListener runs.- https://github.com/dotnet/runtime/issues/38804#issuecomment-655186978

You may need to guard against it

shirhatti: See https://github.com/shirhatti/Metrics/blob/master/src/Microsoft.Extensions.Metrics/Listener.cs#L67 for an example

MihaZupan: Added more checks to avoid this in https://github.com/microsoft/reverse-proxy/pull/571/commits/d11d635aadb0bedc5f60861bdccd9672fa7fe4c5.

From what I can tell, the linked example has a race condition:
After setting `_initialized` but before looping over the queue in the ctor, new `EvnetSource`s may still be added to the queue and enabled in `OnEventSourceCreated`, and will thus be enabled twice (don't know if `EventSource` internals have any checks guarding against this).

src/ReverseProxy.TelemetryConsumption/Http/HttpEventListenerService.cs
```diff
@@ -0,0 +1,293 @@
...
+        {
+            if (eventData.EventId < 1)
+            {
+                if (eventData.EventId == -1)
+                {
+                    OnEventCounters(eventData);
+                }
+
+                return;
```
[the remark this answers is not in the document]
MihaZupan: Added `MinEventId`, `MaxEventId` constants and checks

src/ReverseProxy.TelemetryConsumption/MetricsOptions.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Telemetry.Consumption
+{
+    internal static class MetricsOptions
+    {
+        // TODO: Should this be publicly configurable? It's currently only visible to tests to reduce execution time
+        public static float IntervalSeconds { get; set; } = 1;
```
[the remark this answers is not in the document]
MihaZupan: Changed

src/ReverseProxy.TelemetryConsumption/Microsoft.ReverseProxy.Telemetry.Consumption.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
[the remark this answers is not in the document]
MihaZupan: Added a description.

It will be published as a separate package with a dependency on the core proxy package.

src/ReverseProxy.TelemetryConsumption/Microsoft.ReverseProxy.Telemetry.Consumption.csproj
MihaZupan: Added a target for 3.1 (along with a bunch of ifdefs).

3.1 surface includes:
- Proxy events and metrics identical to 5.0.
- Kestrel events (with modified signatures).

It does not include any runtime events/metrics and Kestrel metrics.

src/ReverseProxy/Telemetry/ProxyStage.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.ReverseProxy.Telemetry
 {
-    internal enum ProxyStage : int
+    public enum ProxyStage : int
```
Tratcher, line 6: I thought everything was supposed to be decoupled to the point where you didn't even need a dependency on the core package? Is this type the only dependency? Would duplicating it in the telemetry package work?

MihaZupan: Yes, technically this is the only type we depend on.
As long as the values are kept in sync, we can duplicate it in the Telemetry package and remove a dependency on the core package entirely.

The only other thing the telemetry package depends on is that consumers of the library target a preview >= 7, otherwise they will not observe proxy events (wouldn't break anything tho).

MihaZupan:
We're currently relying on the package reference in 3 places:
- `ProxyStage` enum (currently internal in core package)
- `ProxyError` enum (public)
- `IReverseProxyBuilder` interface for the `AddXTelemetryListener` extension methods

The extension methods could be swapped to use `IServiceCollection` instead and the `ProxyStage` enum could be duplicated and only exposed in the Consumption package.

But we would have to duplicate the already public `ProxyError` enum.

I think we should:
- keep the package reference to the core package
- use the `IReverseProxyBuilder` interface
- duplicate and expose the `ProxyStage` enum in the consumption package only
- use the core package's `ProxyError`

@Tratcher thoughts?

src/ReverseProxy.TelemetryConsumption/IReverseProxyBuilderTelemetryExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public static IReverseProxyBuilder AddProxyTelemetryListener(this IReverseProxyBuilder builder)
+        {
+            builder.Services.AddHttpContextAccessor();
+            builder.Services.AddHostedService<ProxyEventListenerService>();
+            return builder;
+        }
+
+        public static IReverseProxyBuilder AddKestrelTelemetryListener(this IReverseProxyBuilder builder)
```
Tratcher, line 30: Kestrel, HttpClient, DNS, etc. specific extensions don't belong on IReverseProxyBuilder, move these to IServiceCollection. At that point I'm not sure if it makes sense to still have the others on IReverseProxyBuilder.

Tratcher (commented):
That plan makes sense, but see my comment below about using IReverseProxyBuilder.

src/ReverseProxy.TelemetryConsumption/Http/HttpEventListenerService.cs
shirhatti: `EnableEvents` is safe to call twice.

src/ReverseProxy.TelemetryConsumption/IReverseProxyBuilderTelemetryExtensions.cs
MihaZupan: Changed all to use IServiceCollection instead.

MihaZupan:
@Tratcher is there anything else that should be done here before merging the PR?

Tratcher (approved):
My only outstanding question is https://github.com/microsoft/reverse-proxy/pull/571/files#r535719286

MihaZupan:
I added a sample telemetry & metrics consumer to Code.Sample. More examples can be added as part of #676.

I also wrapped the metrics consumers in a try/catch, passing any exceptions to ILogger instead of letting them propagate to the [CounterGroup thread](https://github.com/dotnet/runtime/blob/004eb7556063cd3f5c508287089266e2571fb437/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/CounterGroup.cs#L242) and crashing the process.

PR #573, opened 2020-11-24
Dispose FabricClient

Kahbazi:
Does each one of these wrappers need their own instance of `FabricClient` or is it possible to create only one instance and share it?

### DIFF
--- a/src/ReverseProxy.ServiceFabric/FabricWrapper/HealthClientWrapper.cs
+++ b/src/ReverseProxy.ServiceFabric/FabricWrapper/HealthClientWrapper.cs
@@ -1,19 +1,28 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using System.Fabric;
 using System.Fabric.Health;
 
 namespace Microsoft.ReverseProxy.ServiceFabric
 {
     /// <inheritdoc/>
-    internal class HealthClientWrapper : IHealthClientWrapper
+    internal class HealthClientWrapper : IHealthClientWrapper , IDisposable
     {
+        private readonly FabricClient _fabricClient;
         private readonly FabricClient.HealthClient _healthClient;
 
         public HealthClientWrapper()
         {
-            _healthClient = new FabricClient().HealthManager;
+            _fabricClient = new FabricClient();
+            _healthClient = _fabricClient.HealthManager;
+        }
+
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            _fabricClient.Dispose();
         }
 
         /// <inheritdoc/>
--- a/src/ReverseProxy.ServiceFabric/FabricWrapper/HealthClientWrapper.cs
+++ b/src/ReverseProxy.ServiceFabric/FabricWrapper/HealthClientWrapper.cs
@@ -1,28 +1,19 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System;
 using System.Fabric;
 using System.Fabric.Health;
 
 namespace Microsoft.ReverseProxy.ServiceFabric
 {
     /// <inheritdoc/>
-    internal class HealthClientWrapper : IHealthClientWrapper , IDisposable
+    internal class HealthClientWrapper : IHealthClientWrapper
     {
-        private readonly FabricClient _fabricClient;
         private readonly FabricClient.HealthClient _healthClient;
 
-        public HealthClientWrapper()
+        public HealthClientWrapper(IFabricClientWrapper fabricClientWrapper)
         {
-            _fabricClient = new FabricClient();
-            _healthClient = _fabricClient.HealthManager;
-        }
-
-        /// <inheritdoc/>
-        public void Dispose()
-        {
-            _fabricClient.Dispose();
+            _healthClient = fabricClientWrapper.FabricClient.HealthManager;
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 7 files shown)


Tratcher:
> Does each one of these wrappers need their own instance of `FabricClient` or is it possible to create only one instance and share it?

What's the lifetime of each of these services? And what are the concurrency limitations of the FabricClient?

Kahbazi:
> What's the lifetime of each of these services? And what are the concurrency limitations of the FabricClient?

All the wrappers are registered as singleton.

The [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.fabric.fabricclient?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(System.Fabric.FabricClient)%3Bk(SolutionItemsProject)%3Bk(DevLang-csharp)%26rd%3Dtrue&view=azure-dotnet#remarks), recommend to share, so it seems there's no limitation.
> It is highly recommended that you share FabricClients as much as possible. This is because the FabricClient has multiple optimizations such as caching and batching that you would not be able to fully utilize otherwise.

Tratcher:
Interesting. Would FabricClient be shared as its own DI service? With or without at test abstraction?

Kahbazi:
There could be an `IFabricClientWrapper` interface in case the user wants to create `FabricClient` without the parameterless constructor. Also I haven't seen any tests for wrappers.

Tratcher:
Ok, if `IHealthClientWrapper` for example is already the test abstraction then another abstraction might not be needed.

src/ReverseProxy.ServiceFabric/FabricWrapper/HealthClientWrapper.cs
```diff
@@ -11,9 +11,9 @@ internal class HealthClientWrapper : IHealthClientWrapper
     {
         private readonly FabricClient.HealthClient _healthClient;
 
-        public HealthClientWrapper()
+        public HealthClientWrapper(IFabricClientWrapper fabricClientWrapper)
         {
-            _healthClient = new FabricClient().HealthManager;
+            _healthClient = fabricClientWrapper.FabricClient.HealthManager;
```
alnikola, line 16: I cannot understand what benefit the new interface gives to us, if we still directly accessing `FabricClient`. Could you please clarify it?

alnikola (commented):
Not sure about the purpose of this PR.

src/ReverseProxy.ServiceFabric/FabricWrapper/HealthClientWrapper.cs
Kahbazi: The other options would be to inject `FabricClient` directly and I don't know if it's ok that YARP add a service to IoC container that it doesn't own.

alnikola: If the goal is to make `FabricClient` unit test friendly, we would need a complete wrapper over it. Meaning a new interface mirroring all the FabricClient API.

Kahbazi: That wasn't my intention, since there's no test for the client wrappers. Do we need to have tests for them?

alnikola: No right now at least. Sorry, I might have missed the point. Could you please explain the purpose of this PR? Is it to dispose `FabricClient`? If I remember correctly it's supposed to live as long as YARP application lives.

@davidni Could you please comment on that?

Kahbazi: This PR serves two purpose:

1. Share the instance of `FabricClient` between client wrappers as recommended by [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.fabric.fabricclient?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(System.Fabric.FabricClient)%3Bk(SolutionItemsProject)%3Bk(DevLang-csharp)%26rd%3Dtrue&view=azure-dotnet#remarks).
2. Dispose `FabricClient` as recommended by https://github.com/microsoft/reverse-proxy/pull/257#discussion_r445704082

> If I remember correctly it's supposed to live as long as YARP application lives.

Normally yes, but I could have a process which start a new `WebHost` (with YARP in it) every hour and every hour a new instance of `FabricClient` would be created which never dies until process is dead.

alnikola: Ok, I got it now. But it still feels wrong to just expose FabricClient belonging to Service Fabric API directly on our new YARP interface. A wrapper is supposed to 'wrap' something. Is it possible to mirror FabricClient's methods we use on the IFabricClientWrapper interface? I expect there are not so much of them.

Kahbazi: `IHealthClientWrapper` also uses ServiceFabric types like `HealthReport` and `HealthReportSendOptions`. Is this a problem too?

Here are the methods that other wrapper are using
`FabricClient.ServiceManagementClient.GetServiceManifestAsync`
`FabricClient.HealthClient.ReportHealth`
`FabricClient.PropertyManagementClient.GetPropertyAsync`
`FabricClient.PropertyManagementClient.EnumeratePropertiesAsync`
`FabricClient.QueryClient.GetApplicationListAsync`
`FabricClient.QueryClient.GetServiceListAsync`
`FabricClient.QueryClient.GetServiceTypeListAsync`
`FabricClient.QueryClient.GetPartitionListAsync`
`FabricClient.QueryClient.GetReplicaListAsync`

Also all the wrapper interfaces are internal tight now.

I think most of these methods are already mirrored and we have wrapper for application, service and replica, but there are no wrapper for HealthReport.

alnikola: It seems there are too many of them, so let's keep the current version with exposing `FabricClient` directly from wrapper.

alnikola (approved):
LGTM

alnikola:
Thank you!

PR #574, opened 2020-11-25
Consume IHttpRequestBodyDetectionFeature

Tratcher:
Problem: HttpClient always adds `Transfer-Encoding: chunked` to outgoing HTTP/1.1 requests if the HttpRequestMessage has an HttpContent, even if it's empty. Some servers won't accept requests like GETs that have a `Transfer-Encoding: chunked` header, so YARP has to avoid creating an HttpContent when it doesn't need one.

There are some ambiguous cases where we can't tell if the request has a body by just looking at headers, such as HTTP/2 where Transfer-Encoding is obsolete. Today we mitigate this with a series of heuristics about methods, versions, headers, etc., but it's not foolproof, only the underlying server knows for sure.

This PR consumes the new .NET 5 IHttpRequestBodyDetectionFeature that can be used in pace of any heuristics to check if the server is expecting a request body. This should be both more reliable and more efficient. I've also simplified some of the heuristics and removed some corner cases.

### DIFF
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -337,12 +337,16 @@ namespace Microsoft.ReverseProxy.Service.Proxy
             var hasBody = true;
             var contentLength = request.Headers.ContentLength;
             var method = request.Method;
-            // https://tools.ietf.org/html/rfc7231#section-4.3.8
-            // A client MUST NOT send a message body in a TRACE request.
-            if (HttpMethods.IsTrace(method))
+
+#if NET
+            var canHaveBodyFeature = request.HttpContext.Features.Get<IHttpRequestBodyDetectionFeature>();
+            if (canHaveBodyFeature != null)
             {
-                hasBody = false;
+                // 5.0 servers provide a definitive answer for us.
+                hasBody = canHaveBodyFeature.CanHaveBody;
             }
+            else
+#endif
             // https://tools.ietf.org/html/rfc7230#section-3.3.3
             // All HTTP/1.1 requests should have Transfer-Encoding or Content-Length.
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -365,21 +369,16 @@ public HttpProxy(ILogger<HttpProxy> logger, IClock clock)
...
-            // A client MUST NOT generate a 100-continue expectation in a request that does not include a message body.
-            else if (request.Headers.TryGetValue(HeaderNames.Expect, out var expect)
-                && expect.Count == 1
-                && string.Equals("100-continue", expect.ToString(), StringComparison.OrdinalIgnoreCase))
-            {
-                hasBody = true;
-            }
             // https://tools.ietf.org/html/rfc7231#section-4.3.1
             // A payload within a GET/HEAD/DELETE/CONNECT request message has no defined semantics; sending a payload body on a
             // GET/HEAD/DELETE/CONNECT request might cause some existing implementations to reject the request.
```
ManickaP, line 374: I don't see the whole picture here probably, but isn't this a bit bold assumption? I mean especially GET and DELETE. I've recently seen question about how to send DELETE with content in runtime repo, that's why I'm asking.

alnikola (approved):
LGTM

src/ReverseProxy/Service/Proxy/HttpProxy.cs
Tratcher: Yes, that's exactly the problem with the heuristics approach and why we added this new feature in 5.0. The RFC says the behavior of DELETE with content is undefined, but people still try to do it on rare occasions. Hopefully when they do it will be caught by one of the other heuristics like content-length or transfer-encoding. This methods check is a last resort for specific HTTP/2 scenarios.

PR #578, opened 2020-11-25
Handle SPDY upgrades

### LINKED ISSUE #467: SPDY Connection upgrades with kubernetes API server not working.
### Describe the bug
I am trying to use YARP as reverse proxy behind a kubernetes cluster. Normal HTTP requests work fine but API's which require connection upgrade are not working(For example: kubectl exec). I tried using IHttpUpgradeFeature as well to make YARP understand if this request is a upgrade request or not. But with that change as well it's not working.

### To Reproduce

- You would need to have a kubernetes cluster.

- Get your K8s cluster's CA data and use it as client certificate and point the destination address to k8s api server.

- Deploy nginx on to your k8s cluster by following this doc:  https://kubernetes.io/docs/tasks/run-application/run-stateless-application-deployment/#creating-and-exploring-an-nginx-deployment 
-  Change the api server endpoint in your kubeconfig to yarp proxy server.
- Run kubectl exec command with the pod information form your previously created deployment.

This is the response we get from k8s API server when we use YARP as reverse proxy in between:
error: unable to upgrade connection: unable to upgrade: unable to negotiate protocol: client supports [v4.channel.k8s.io, v3.channel.k8s.io, v2.channel.k8s.io, channel.k8s.io], server accepts [v4.channel.k8s.io v3.channel.k8s.io v2.channel.k8s.io channel.k8s.io]

### Further technical details

- Include the version of the packages you are using: Running using ReverseProxy.Code.Sample of latest master as of 10/10/2020
- The platform (Linux/macOS/Windows) Windows

Tratcher:
Fixes #467

I'm extending our current WebSockets mitigation to include support for SPDY because that's used by some kubernetes tools.

I also took this chance to reorganize the upgrade logic so it's checked closer to where it's used, and add a little more error handling. That makes it clearer that we only use it to determine the request version.

### DIFF
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -119,12 +119,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
                 var requestAborted = context.RequestAborted;
 
                 // :: Step 1: Create outgoing HttpRequestMessage
-                var upgradeFeature = context.Features.Get<IHttpUpgradeFeature>();
-                var isUpgradeRequest = (upgradeFeature?.IsUpgradableRequest ?? false)
-                    // Mitigate https://github.com/microsoft/reverse-proxy/issues/255, IIS considers all requests upgradeable.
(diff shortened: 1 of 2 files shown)


alnikola (approved):
LGTM

PR #579, opened 2020-11-26
[SF] LabelsParser supports all basic *Options

### LINKED ISSUE #557: [SF] Parse all supported *Options in LabelsParser
`LabelsParser` handles only a very limited set of YARP configuration `*Options`. It should be extended to support everything else that is compatible to Service Fabric (e.g. `LoadBalancingOptions`, `SessionAffinityOptions`, etc.).

alnikola:
It adds support for `LoadBalancingOptions`, `SessionAffinityOptions`, `ProxyHttpRequestOptions`, `PassiveHealthCheckOptions` to `LabelsParser`. This PR also fixes casing of `YARP.Backend.Healthcheck` label's key to be `YARP.Backend.HealthCheck`.

Fixes #557


src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -106,37 +107,78 @@ internal static List<ProxyRoute> BuildRoutes(Uri serviceName, Dictionary<string,
...
+                {
+                    Enabled = GetLabel(labels, "YARP.Backend.SessionAffinity.Enabled", false),
+                    Mode = GetLabel<string>(labels, "YARP.Backend.SessionAffinity.Mode", null),
+                    FailurePolicy = GetLabel<string>(labels, "YARP.Backend.SessionAffinity.FailurePolicy", null),
+                    Settings = sessionAffinitySettings
+                },
+                HttpRequest = new ProxyHttpRequestOptions
+                {
+                    Timeout = ToNullableTimeSpan(GetLabel<double?>(labels, "YARP.Backend.HttpRequest.Timeout", null)),
+                    Version = Version.TryParse(GetLabel<string>(labels, "YARP.Backend.HttpRequest.Version", null), out var version) ? version : null,
```
Tratcher, line 148: Shouldn't these all be using Parse instead of TryParse? We don't want to hide invalid config.

Tratcher: Version won't parse "2", it requires at least "2.0".
https://github.com/microsoft/reverse-proxy/blob/150e34c37b24d280e6c52c57a392b39ded4dcc1c/src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs#L37

Tratcher (approved):

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
alnikola: All of the above is fixed now.

PR #585, opened 2020-11-30
Consistently use InvariantCulture for doubles

MihaZupan:
We're using invariant culture when parsing the rate limit for health checks, but serializing it with the current culture (`.` vs `,` for floats).
This leads to a spceific test failing on machines with a different culture.

### DIFF
--- a/test/ReverseProxy.Tests/Service/HealthChecks/TransportFailureRateHealthPolicyTests.cs
+++ b/test/ReverseProxy.Tests/Service/HealthChecks/TransportFailureRateHealthPolicyTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Options;
 using Microsoft.ReverseProxy.Abstractions;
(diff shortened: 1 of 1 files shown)


test/ReverseProxy.Tests/Service/HealthChecks/TransportFailureRateHealthPolicyTests.cs
```diff
@@ -206,7 +207,7 @@ private HttpContext GetFailedRequestContext(ProxyError error)
         private ClusterInfo GetClusterInfo(string id, int destinationCount, double? failureRateLimit = null, TimeSpan? reactivationPeriod = null)
         {
             var metadata = failureRateLimit != null
-                ? new Dictionary<string, string> { { TransportFailureRateHealthPolicyOptions.FailureRateLimitMetadataName, failureRateLimit.ToString() } }
+                ? new Dictionary<string, string> { { TransportFailureRateHealthPolicyOptions.FailureRateLimitMetadataName, failureRateLimit?.ToString(CultureInfo.InvariantCulture) } }
```
ManickaP, line 210:
```suggestion
                ? new Dictionary<string, string> { { TransportFailureRateHealthPolicyOptions.FailureRateLimitMetadataName, failureRateLimit.ToString(CultureInfo.InvariantCulture) } }
```
It's in the `failureRateLimit != null` branch. Or am I missing something?

ManickaP (approved):
LGTM

test/ReverseProxy.Tests/Service/HealthChecks/TransportFailureRateHealthPolicyTests.cs
MihaZupan: It would have to be `.Value.ToString` in that case since it's `Nullable<double>` without that ToString overload. `?` seems cleaner (shorter).

Issue #531, opened 2020-11-10
Inconsistent use of nullable fields in config


Tratcher:
The config classes like Cluster are inconsistent about the use of nullable fields.
https://github.com/microsoft/reverse-proxy/blob/6defd11081d37c0d59b34b67cdabff240c9b8e84/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ProxyHttpClientOptions.cs#L14-L18

This makes it hard for the various parts of the implementation to tell if they're being given a default value or if the user actually specified that value. Since many parts of the implementations are supposed to be replaceable components, it would help if they could distinguish so they could provide their own defaults while still respecting explicit user config.

Proposal: Make most/all config fields nullable and only specify the default in the implementations.

karelz:
Triage: Affects base object model, therefore v1

karelz:
Should be fairly straightforward to do ... we would welcome help here.

PR #596, opened 2020-12-03
Make config fields nullable

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/531


Tratcher (commented):
I had meant for this to cover the config model types consumed through the IProxyConfigProvider (Cluster, ClusterProxyHttpClientOptions, etc.), I hadn't considered the `IOptions<T>` types like ActiveHealthCheckMonitorOptions. Thanks for being thorough 😁.

Having reviewed those IOptions types now, it looks like most of them were already intended to provide fallback defaults for fields from the config model. These types are also used directly by implementations, where an alternate implementation would likely provide its own options type. I think it's fine for these to keep their non-nullable default values. Please revert changes for these options and their consumers:
- ActiveHealthCheckMonitorOptions
- ConsecutiveFailuresHealthPolicyOptions
- TransportFailureRateHealthPolicyOptions

Tratcher (approved):

Tratcher:
Thanks

PR #597, opened 2020-12-03
Change IAffinityFailurePolicy to public

### LINKED ISSUE #595: Add support for custom affinity failure policies

Change ```IAffinityFailurePolicy``` from internal to public.

The [documentation](https://microsoft.github.io/reverse-proxy/articles/session-affinity.html#affinity-failure-policy) seems to imply that custom failure policies should be possible, however because the interface is internal it's not possible to actually implement any additional policies.  

In our app we need to execute a 302 redirect to recover from an affinity failure.  A custom failure policy seems like a good fit for this.  I tried it and it seems to work fine.

I would be very happy to submit a PR containing this very small change, but I understand that you may also want to review the API before committing to it...

hughesjon:
Allows implementation of custom failure policies, which fixes #595


Tratcher (approved):

Tratcher:
Thanks

PR #600, opened 2020-12-07
Make load-balancing extensible

### LINKED ISSUE #106: Add more extensible/custom load balancers

### LINKED ISSUE #229: Component-specific data storage
There is a need for the storage where different proxy components can store data specific to their internal logic.

At the very least this storage must provide the following capabilities:
1.	Storing mutable and immutable data
2.	Fast and thread safe retrieval of an entity by key efficiently operating in a highly concurrent environment
3.	Thread safe operations to add, update and delete data entities

Optionally it should also provide:
1.	Persistent storage preserving data between proxy process restarts
2.	Data changes notifications

Examples of the storage use cases:
1.	Destination affinity key lookup table for session affinity feature
2.	Load balancing counters

See also #87

MihaZupan:
Fixes #106
Fixes #229

Replace the internal `LoadBalancingMode`, `ILoadBalancer`, `LoadBalancer` with a public `ILoadBalancingPolicy` interface.

Following the pattern used with session affinity, `LoadBalancingMiddleware` resolves available policies from DI and select the current one per-invoke based on cluser config.

`LoadBalancingOptions` (a struct wrapping the `LoadBalancingMode` enum) is replaced by a string `LoadBalancingPolicy`.

The 5 existing load balancing strategies are implemented as separate `ILoadBalancingPolicy` types. Power of two choices remains the default.

ToDo: docs on load balancing


src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/LoadBalancingConstants.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.ReverseProxy.Abstractions.ClusterDiscovery.Contract
+{
+    /// <summary>
+    /// Names of built-in load balancing policies.
+    /// </summary>
+    public static class LoadBalancingConstants
+    {
+        public static class Policies
```
MihaZupan, line 11: Should this be just `LoadBalancingPolicies`?

alnikola: I think it could be shorten to this. The nested structure of `SessionAffinityConstants` is motivated by the presence of affinitization failures handling policies. It seems we don't need them for load balancing.

docs/docfx/articles/service-fabric-int.md
```diff
@@ -73,7 +73,7 @@ YARP integration is enabled and configured per each SF service. The configuratio
 These are the supported parameters:
 - `YARP.Enable` - indicates whether the service opt-ins to serving traffic through YARP. Default `false`
 - `YARP.EnableDynamicOverrides` - indicates whether application parameters replacement is enabled on the service. Default `false`
-- `YARP.Backend.LoadBalancing.Mode` - configures YARP load balancing mode. Optional parameter
+- `YARP.Backend.LoadBalancingPolicy` - configures YARP load balancing policy. Optional parameter
```
alnikola, line 76: Do we really need to combine `LoadBalancing` and `Policy`? I'd prefer to keep the old pattern i.e. `LoadBalancing.Policy`.

src/ReverseProxy/Service/LoadBalancing/PowerOfTwoChoicesLoadBalancingPolicy.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public DestinationInfo PickDestination(HttpContext context, IReadOnlyList<DestinationInfo> availableDestinations)
+        {
+            // Pick two, and then return the least busy. This avoids the effort of searching the whole list, but
+            // still avoids overloading a single endpoint.
+            var endpointCount = availableDestinations.Count;
+            var random = _randomFactory.CreateRandomInstance();
+            var firstEndpoint = availableDestinations[random.Next(endpointCount)];
+            var secondEndpoint = availableDestinations[random.Next(endpointCount)];
+            return (firstEndpoint.ConcurrencyCounter.Value <= secondEndpoint.ConcurrencyCounter.Value) ? firstEndpoint : secondEndpoint;
```
alnikola, line 31: I believe `Endpoint` name is a leftover from some past version. Let's rename it to `Destination` here also to make it consistent.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -305,8 +303,7 @@ private void UpdateRuntimeClusters(IList<Cluster> newClusters)
                                         timeout: newCluster.HealthCheck?.Active?.Timeout,
                                         policy: newCluster.HealthCheck?.Active?.Policy,
                                         path: newCluster.HealthCheck?.Active?.Path ?? string.Empty)),
-                                new ClusterLoadBalancingOptions(
-                                    mode: newCluster.LoadBalancing?.Mode ?? default),
+                                loadBalancingPolicy: newCluster.LoadBalancingPolicy,
```
alnikola, line 306: Do we really want to completely get rid of a separate type `LoadBalancingOptions`? I'd keep it for now, but just rename `Mode` to `Policy`.

test/ReverseProxy.Tests/Middleware/LoadBalancerMiddlewareTests.cs
```diff
@@ -1,155 +1,196 @@
...
         [Fact]
         public void Constructor_Works()
         {
-            Create<LoadBalancingMiddleware>();
+            CreateMiddleware();
         }
 
         [Fact]
-        public async Task Invoke_Works()
+        public async Task PickDestination_UnsupportedMode_Throws()
```
alnikola, line 41: Shouldn't it be named `UnsupportedPolicy`?

test/ReverseProxy.Tests/Service/Proxy/LoadBalancerTests.cs
```diff
@@ -2,9 +2,13 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Linq;
```
alnikola, line 5: Wouldn't it be better to rename the file to something like `LoadBalancingPoliciesTests`?

alnikola (commented):
Looks good overall, besides a few comments.

docs/docfx/articles/service-fabric-int.md
Tratcher: These labels should mirror the config structure as much as possible.

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -146,7 +146,7 @@ private Cluster CreateCluster(IConfigurationSection section)
             var cluster = new Cluster
             {
                 Id = section.Key,
-                LoadBalancing = CreateLoadBalancingOptions(section.GetSection(nameof(Cluster.LoadBalancing))),
+                LoadBalancingPolicy = CreateLoadBalancingOptions(section.GetSection(nameof(Cluster.LoadBalancingPolicy))),
```
Tratcher, line 149: You can remove this method and read the config directly.
`section[nameof(Cluster.LoadBalancingPolicy)],`

src/ReverseProxy/Service/LoadBalancing/FirstLoadBalancingPolicy.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.ReverseProxy.Service.LoadBalancing
+{
+    internal sealed class FirstLoadBalancingPolicy : ILoadBalancingPolicy
+    {
+        public string Name => LoadBalancingConstants.Policies.First;
+
+        public DestinationInfo PickDestination(HttpContext context, IReadOnlyList<DestinationInfo> availableDestinations)
+        {
+            return availableDestinations[0];
```
Tratcher, line 17: This should at least do an assert check that availableDestinations.Length > 0.

src/ReverseProxy/Service/LoadBalancing/ILoadBalancingPolicy.cs
```diff
@@ -2,22 +2,28 @@
...
+        /// <summary>
+        ///  A unique identifier for this load balancing policy. This will be referenced from config.
+        /// </summary>
+        public string Name { get; }
+
         /// <summary>
         /// Picks a destination to send traffic to.
         /// </summary>
         // TODO: How to ensure retries pick a different destination when available?
         DestinationInfo PickDestination(
```
Tratcher, line 25: Thinking out loud: I think we'll want to pass more state here like the route and cluster data, though I'm not sure which type to pass in. Algorithms with options will likely pull those from the cluster metadata. That would mean passing ClusterConfig or IReverseProxyFeature. IReverseProxyFeature might be best since I think that's where we want to expose the route data in the future. Yes they could pull the IReverseProxyFeature from the HttpContext, but that's not discoverable.

What do you think?

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: I think we should get rid of it. The policies will all have their own unique options, probably exposed through metadata. Are there any options you expect to be shared?

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/LoadBalancingConstants.cs
MihaZupan: I agree, renamed.

src/ReverseProxy/Service/LoadBalancing/PowerOfTwoChoicesLoadBalancingPolicy.cs
MihaZupan: Changed.

test/ReverseProxy.Tests/Service/Proxy/LoadBalancerTests.cs
MihaZupan: Renamed

src/ReverseProxy/Service/LoadBalancing/FirstLoadBalancingPolicy.cs
MihaZupan: Added

```c#
if (availableDestinations.Count == 0)
{
    return null;
}
```

checks to the built-in policies.

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/LoadBalancingPolicies.cs
```diff
@@ -1,33 +1,37 @@
...
+    public static class LoadBalancingPolicies
     {
         /// <summary>
-        /// Select two random endpoints and then select the one with the least assigned requests.
-        /// This avoids the overhead of LeastRequests and the worst case for Random where it selects a busy endpoint.
-        /// </summary>
-        PowerOfTwoChoices,
-        /// <summary>
-        /// Select the endpoint with the least assigned requests. This requires examining all nodes.
+        /// Select the first endpoint without considering load. This is useful for dual endpoint fail-over systems.
```
Tratcher, line 12: Ha, these also still say "endpoint"

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
alnikola: Nothing specific at the moment. I suggested that based on my previous experience with implementing things like session affinity and health checks where we had a dedicated types for the common settings affecting all policies. Thus, I imagined it can be a case here as well, but I do agree it might be stretched here.

alnikola (approved):
LGTM

Tratcher:
> ToDo: docs on load balancing

Can you complete the docs as part of this PR?

src/ReverseProxy/Service/LoadBalancing/ILoadBalancingPolicy.cs
MihaZupan: We should definitely keep the `availableDestinations` parameter.
If we're planning on adding more info onto `IReverseProxyFeature`, then that should be what we include for discoverability.

Do you think pointing out the `GetRequiredProxyFeature` extension in load balancing docs would help with discoverability enough to stick with `HttpContext`'s "everything"?

It doesn't seem like we can reason on if/what info custom policies will rely on.
None of the built-in do (except RoundRobin, but that's our choice for state storage), and none of the ones in #107 do.
As you mention, policies with options would likely need that info.

Tratcher: That's a good point that none of the proposed protocols have this requirement. We can wait to see if it comes up.

MihaZupan:
> Can you complete the docs as part of this PR?

There shouldn't be any big conflicting code changes coming in to block a merge, so I will push the docs to this PR.

Tratcher:
Also check the samples and docs:
https://github.com/microsoft/reverse-proxy/blob/e4426b93621a4f0e4c55c626236a21404efd264e/samples/ReverseProxy.Config.Sample/appsettings.json#L23-L25

https://github.com/microsoft/reverse-proxy/blob/f25992421a76dd4affb62a87e9f65487bc14f0b5/docs/docfx/articles/proxyhttpclientconfig.md#L102

https://github.com/microsoft/reverse-proxy/blob/dd9ac52be32892508e58f4067b61cf213e61d745/docs/docfx/articles/service-fabric-int.md#L76

MihaZupan:
@Tratcher can you PTAL at the added docs?

Thanks!

test/ReverseProxy.Tests/Middleware/LoadBalancerMiddlewareTests.cs
```diff
@@ -1,155 +1,196 @@
...
-                {
-                    AvailableDestinations = new List<DestinationInfo>() { destination1, destination2 },
-                    ClusterConfig = cluster1.Config
-                });
-            httpContext.Features.Set(cluster1);
+            var sut = CreateMiddleware();
 
-            var sut = Create<LoadBalancingMiddleware>();
+            var ex = await Assert.ThrowsAsync<ArgumentException>(async () => await sut.Invoke(context));
+            Assert.Equal($"No {typeof(ILoadBalancingPolicy)} was found for the id {PolicyName}. (Parameter 'id')", ex.Message);
```
Tratcher, line 53: Sorry, I just noticed this test and realized that ConfigValidator wasn't updated to check for valid load balancing polcies. Compare to the session affinity validation:
https://github.com/microsoft/reverse-proxy/blob/37ed6a93f3ea54cc412471ca85ac9b3e38aca749/src/ReverseProxy/Service/Config/ConfigValidator.cs#L295-L298

Tratcher (approved):
The doc looks nice and simple.

I did just realize we missed updating the ConfigValidator. Should be a simple fix.

test/ReverseProxy.Tests/Middleware/LoadBalancerMiddlewareTests.cs
MihaZupan: I assume https://github.com/microsoft/reverse-proxy/pull/600/commits/c4693ff6e5399f1eff003fc79eab0f99dbe4650a is sufficient? I manually verified that the validation works.

Tratcher: Yes, that looks fine. Please add one positive and negative unit test.

PR #607, opened 2020-12-09
Discoverer cleans up caches by removing expired entries before each iteration

alnikola:
Discoverer calls a new `ICachedServiceFabricCaller.CleanUpExpired` method before each discovery cycle to remove expired cached entries. This PR also renames `IServiceFabricCaller` to `ICachedServiceFabricCaller`.

### DIFF
--- a/src/ReverseProxy.ServiceFabric/ServiceDiscovery/Discoverer.cs
+++ b/src/ReverseProxy.ServiceFabric/ServiceDiscovery/Discoverer.cs
@@ -25,7 +25,7 @@ namespace Microsoft.ReverseProxy.ServiceFabric
         public static readonly string HealthReportProperty = "DynamicConfig";
 
         private readonly ILogger<Discoverer> _logger;
-        private readonly IServiceFabricCaller _serviceFabricCaller;
+        private readonly ICachedServiceFabricCaller _serviceFabricCaller;
         private readonly IServiceExtensionLabelsProvider _serviceFabricExtensionConfigProvider;
         private readonly IConfigValidator _configValidator;
         private readonly IOptionsMonitor<ServiceFabricDiscoveryOptions> _optionsMonitor;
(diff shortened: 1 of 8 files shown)


src/ReverseProxy.ServiceFabric/ServiceDiscovery/Discoverer.cs
```diff
@@ -53,6 +53,8 @@ public async Task<(IReadOnlyList<ProxyRoute> Routes, IReadOnlyList<Cluster> Clus
             // Take a snapshot of current options and use that consistently for this execution.
             var options = _optionsMonitor.CurrentValue;
 
+            _serviceFabricCaller.CleanUpExpired();
```
Tratcher, line 56: Why does this need to be an explicit call? Could the relevant cache cleanup happen implicitly when calling GetApplicationListAsync, GetServiceListAsync, etc.? It looks like those are only called once per discovery, so the cleanup wouldn't happen multiple times. It looks like the cleanup should be called about here:
https://github.com/microsoft/reverse-proxy/blob/084085f6377c2ef236bf2d65a23453352f94387b/src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/CachedServiceFabricCaller.cs#L164

Also, isn't the caching an implementation detail of the IServiceFabricCaller? Before this new CleanUpExpired method the caller didn't know or care if the results were cached. If the cache cleanup can be implicit, then I don't see a need for this new method or the renames.

alnikola: >  It looks like those are only called once per discovery

As far as I see it, all methods but `GetApplicationListAsync` is called multiple times in each discovery iteration. Meaning, one call to `GetApplicationListAsync` returns **K applications**, then `GetServiceListAsync` is called once for each of them that is **K times**. Each `GetServiceListAsync` returns **M services**, then `GetPartitionListAsync` is called once per each of them that is **M times**. And so on.

> Also, isn't the caching an implementation detail of the IServiceFabricCaller? Before this new CleanUpExpired method the caller didn't know or care if the results were cached.

That was correct previously, so I **renamed** the interface to `ICachedServiceFabricCaller` to make it clear that it provides caching capabilities. I do agree it doesn't look ideal, but it seems as the only way to clean up the caches once per discovery as explained above.

Tratcher (approved):

PR #611, opened 2020-12-10
String allocation-free label key matching in LabelsParser

### LINKED ISSUE #586: [SF] Label parsing is inefficient
The service fabric label parsing code is really inefficient.
https://github.com/microsoft/reverse-proxy/blob/23a572bc0b54c4975869786d5e003159efdc2cb9/src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs#L80

The pattern `$"{thisRoutePrefix}.Foo."` is used repeatedly and it creates a lot of the same garbage strings over and over again. The new Transforms support extends the problem (https://github.com/microsoft/reverse-proxy/pull/532).

SF polls for changes, so this code could be run quite frequently and evaluate many services. Review this whole code path for ways to reduce allocations.

alnikola:
LablesParser doesn't allocate strings while trying to match label keys against known parameter names.

Fixes #586

### DIFF
--- a/src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
+++ b/src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
@@ -35,6 +35,9 @@ namespace Microsoft.ReverseProxy.ServiceFabric
         /// </summary>
         private static readonly Regex _allowedTransformNamesRegex = new Regex(@"^\[\d\d*\]$");
 
+        // Look for route IDs
+        private const string RoutesLabelsPrefix = "YARP.Routes.";
+
         internal static TValue GetLabel<TValue>(Dictionary<string, string> labels, string key, TValue defaultValue)
         {
             if (!labels.TryGetValue(key, out var value))
@@ -43,14 +46,19 @@ namespace Microsoft.ReverseProxy.ServiceFabric
             }
             else
             {
-                try
-                {
-                    return (TValue)TypeDescriptor.GetConverter(typeof(TValue)).ConvertFromString(value);
-                }
-                catch (Exception ex) when (ex is ArgumentException || ex is FormatException || ex is NotSupportedException)
-                {
-                    throw new ConfigException($"Could not convert label {key}='{value}' to type {typeof(TValue).FullName}.", ex);
-                }
+                return ConvertLabelValue<TValue>(key, value);
+            }
+        }
+
+        private static TValue ConvertLabelValue<TValue>(string key, string value)
+        {
+            try
+            {
+                return (TValue)TypeDescriptor.GetConverter(typeof(TValue)).ConvertFromString(value);
+            }
+            catch (Exception ex) when (ex is ArgumentException || ex is FormatException || ex is NotSupportedException)
+            {
+                throw new ConfigException($"Could not convert label {key}='{value}' to type {typeof(TValue).FullName}.", ex);
             }
         }
 
@@ -59,8 +67,6 @@ namespace Microsoft.ReverseProxy.ServiceFabric
         {
             var backendId = GetClusterId(serviceName, labels);
 
-            // Look for route IDs
-            const string RoutesLabelsPrefix = "YARP.Routes.";
             var routesNames = new HashSet<string>();
             foreach (var kvp in labels)
             {
@@ -86,15 +92,17 @@ namespace Microsoft.ReverseProxy.ServiceFabric
             var routes = new List<ProxyRoute>();
             foreach (var routeName in routesNames)
             {
-                var thisRoutePrefix = $"{RoutesLabelsPrefix}{routeName}";
+                string hosts = null;
+                string path = null;
+                int? order = null;
                 var metadata = new Dictionary<string, string>();
                 var headerMatches = new Dictionary<string, RouteHeader>();
                 var transforms = new Dictionary<string, IDictionary<string, string>>();
                 foreach (var kvp in labels)
                 {
-                    if (kvp.Key.StartsWith($"{thisRoutePrefix}.Metadata.", StringComparison.Ordinal))
+                    if (ContainsKey(routeName, "Metadata.", kvp.Key, out var keyNameEnd))
                     {
-                        metadata.Add(kvp.Key.Substring($"{thisRoutePrefix}.Metadata.".Length), kvp.Value);
+                        metadata.Add(kvp.Key.Substring(keyNameEnd), kvp.Value);
                     }
                     else if (kvp.Key.StartsWith($"{thisRoutePrefix}.MatchHeaders.", StringComparison.Ordinal)) 
                     {
@@ -143,9 +151,9 @@ namespace Microsoft.ReverseProxy.ServiceFabric
                             throw new ConfigException($"Invalid header matching property '{propertyName}', only valid values are Name, Values, IsCaseSensitive and Mode.");
                         }
                     }
-                    else if (kvp.Key.StartsWith($"{thisRoutePrefix}.Transforms.", StringComparison.Ordinal)) 
+                    else if (ContainsKey(routeName, "Transforms.", kvp.Key, out keyNameEnd))
                     {
-                        var suffix = kvp.Key.Substring($"{thisRoutePrefix}.Transforms.".Length);
+                        var suffix = kvp.Key.Substring(keyNameEnd);
                         var transformNameLength = suffix.IndexOf('.');
                         if (transformNameLength == -1)
                         {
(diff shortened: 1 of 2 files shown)


src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -233,5 +249,37 @@ private static IReadOnlyList<string> SplitHosts(string hosts)
         {
             return hosts?.Split(',').Select(h => h.Trim()).Where(h => h.Length > 0).ToList();
         }
+
+        private static bool ContainsKey(string routeName, string expectedKeyName, string actualKey, out int keyNameEnd)
+        {
+            keyNameEnd = -1;
+            var prefixStart = actualKey.IndexOf(RoutesLabelsPrefix, StringComparison.Ordinal);
```
Tratcher, line 256: Shouldn't this be a StartsWith check? We only expect the prefix at the beginning.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -233,5 +249,37 @@ private static IReadOnlyList<string> SplitHosts(string hosts)
...
+            if (prefixStart < 0)
+            {
+                return false;
+            }
+
+            var routeNameStart = actualKey.IndexOf(routeName, prefixStart + RoutesLabelsPrefix.Length, StringComparison.Ordinal);
```
Tratcher, line 262: Can you use span or StringSegment to slice off the the previously matched part and do incremental StartsWith checks rather than IndexOf? IndexOf isn't a great fit because we always know where the thing we're looking for should be. StartsWith would also avoid false positives if the search value appeared later in the key.

span or StringSegment would also let you make the out parameter the remaining text without allocating. No need to depend on index.

Tratcher (commented):
Progress, but it needs more span (or equivalent).

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
alnikola: Fixed.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
alnikola: I replaced `IndexOf` with `Slice` and `StartsWith`, but didn't get the point about returning the remaining text. All callers of `ContainsKey` use the returned `keyNameEnd` to get a substring from the beginning of the key, not the remaining part (though, some of them process the reminder, too). Do you mean getting rid of all `Substring` calls there?

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -289,5 +305,36 @@ private static IReadOnlyList<string> SplitHosts(string hosts)
...
+            }
+
+            var routeNamePart = actualKey.AsSpan().Slice(RoutesLabelsPrefix.Length);
+            if (!routeNamePart.StartsWith(routeName, StringComparison.Ordinal))
+            {
+                return false;
+            }
+
+            var routeNameEnd = RoutesLabelsPrefix.Length + routeName.Length;
+            if (actualKey.Length == routeNameEnd || actualKey[routeNameEnd] != '.')
```
Tratcher, line 324: Why are you going back to the actual key here rather than continuing from the routeNamePart?
```suggestion
            var keyPart = routeNamePart.Slice(routeName.Length);
            if (!keyPart.StartsWith(".", StringComparison.Ordinal))
            {
                return false;
            }
            keyPart = keyPart.Slice(1);
```

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -289,5 +305,36 @@ private static IReadOnlyList<string> SplitHosts(string hosts)
...
+                return false;
+            }
+
+            var keyPart = routeNamePart.Slice(routeName.Length + 1);
+            if (!keyPart.StartsWith(expectedKeyName, StringComparison.Ordinal))
+            {
+                return false;
+            }
+
+            keyNameEnd = routeNameEnd + 1 + expectedKeyName.Length;
```
Tratcher, line 335: Here you can slice one more time to return (out) the remainder instead of the index. That will let the caller do ToString on the remainder as needed instead of Substring(index) on the original. It will also eliminate more index math here.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
alnikola: I just realized that I misinterpreted the code calling Substring, so you were right above that it deals with the remainder only not the beginning. Then, I do agree it can be replaced with calling either `ToString` or `Slice` on the remainder that should reduce the number of allocations.

alnikola: Done.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -59,8 +67,6 @@ internal static List<ProxyRoute> BuildRoutes(Uri serviceName, Dictionary<string,
         {
             var backendId = GetClusterId(serviceName, labels);
 
-            // Look for route IDs
-            const string RoutesLabelsPrefix = "YARP.Routes.";
             var routesNames = new HashSet<string>();
             foreach (var kvp in labels)
```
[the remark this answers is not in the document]
alnikola: Caching added.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -86,19 +92,21 @@ internal static List<ProxyRoute> BuildRoutes(Uri serviceName, Dictionary<string,
...
-                var thisRoutePrefix = $"{RoutesLabelsPrefix}{routeName}";
+                string hosts = null;
+                string path = null;
+                int? order = null;
                 var metadata = new Dictionary<string, string>();
                 var headerMatches = new Dictionary<string, RouteHeader>();
                 var transforms = new Dictionary<string, IDictionary<string, string>>();
                 foreach (var kvp in labels)
                 {
-                    if (kvp.Key.StartsWith($"{thisRoutePrefix}.Metadata.", StringComparison.Ordinal))
```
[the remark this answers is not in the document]
alnikola: Route prefix is checked only once now at the beginning of the loop.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
alnikola: It was kept as is from the original `IndexOf`-based version because a simple char comparison `actualKey[routeNameEnd] != '.'` looked a little bit faster than `StartsWith`. Replaced with your suggestion to make it consistent with the rest of the method.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
alnikola: Done.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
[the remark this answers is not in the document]
alnikola: Route name is now also checked only once.

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -59,53 +68,73 @@ internal static List<ProxyRoute> BuildRoutes(Uri serviceName, Dictionary<string,
...
-            // Look for route IDs
-            const string RoutesLabelsPrefix = "YARP.Routes.";
-            var routesNames = new HashSet<string>();
+            var routesNames = new Dictionary<StringSegment, string>();
             foreach (var kvp in labels)
             {
                 if (kvp.Key.Length > RoutesLabelsPrefix.Length && kvp.Key.StartsWith(RoutesLabelsPrefix, StringComparison.Ordinal))
                 {
-                    var suffix = kvp.Key.Substring(RoutesLabelsPrefix.Length);
+                    var suffix = new StringSegment(kvp.Key, RoutesLabelsPrefix.Length, kvp.Key.Length - RoutesLabelsPrefix.Length);
```
Tratcher, line 76:
```suggestion
                    var suffix = new StringSegment(kvp.Key).Subsegment(RoutesLabelsPrefix.Length);
```

src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -140,53 +169,60 @@ internal static List<ProxyRoute> BuildRoutes(Uri serviceName, Dictionary<string,
...
-                    RouteId = $"{Uri.EscapeDataString(backendId)}:{Uri.EscapeDataString(routeName)}",
+                    RouteId = $"{Uri.EscapeDataString(backendId)}:{Uri.EscapeDataString(routeNamePair.Value)}",
                     Match =
                     {
                         Hosts = SplitHosts(hosts),
                         Path = path,
                         Headers = headerMatches.Count > 0 ? headerMatches.Select(hm => hm.Value).ToArray() : null
                     },
-                    Order = GetLabel(labels, $"{thisRoutePrefix}.Order", DefaultRouteOrder),
+                    Order = order ?? DefaultRouteOrder,
```
Tratcher, line 225: DefaultRouteOrder is also null, so it looks like we don't need that constant anymore.
```suggestion
                    Order = order,
```

Tratcher (approved):

Issue #583, opened 2020-11-29
HTTP/2 connection to YARP with HTTP/1.1 to backend fails


devbeard:
### Describe the bug
I have set up a sample application with routing from https://localhost:5001 to a docker container that echoes all received headers. The container is running on http://localhost:8080. It responds with a connection=keep-alive, causing f.ex. curl to complain as this is not valid in the http/2 context.

```
* http2 error: Invalid HTTP header field was received: frame type: 1, stream: 1, name: [connection], value: [keep-alive]
* HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1)
```

**Question:** Is this a supported scenario - i.e. going from a secured, HTTP/2 channel with a HTTP/1.1 backend? Is it the client  (which has no knowledge of this) or proxy that is responsible of adhering to the protocol change?

### To Reproduce
Run some image on a non-HTTPS binding. Example:
`docker run -p 8080:8080 brndnmtthws/nginx-echo-headers`

Run a simple YARP instance pointing all traffic to http://localhost:8080
`dotnet run ReverseProxySample`

From WSL or similar, do a curl
`curl --insecure -v https://localhost:5001`

### Further technical details

- Include the version of the packages you are using
`1.0.0-preview.7.*`

- The platform (Linux/macOS/Windows)
`Linux and Windows`

Tratcher:
Yes HTTP/2 -> Http/1.1 is supported, but there are certainly gaps like this we need to address. This was likely caused by a gap in Kestrel that wasn't enforcing protocol specific headers like this, it was just fixed in the .NET 6.0 branch (https://github.com/dotnet/aspnetcore/pull/24543). We'll need to work around it in YARP for .NET 3.1 and 5.0.

The relevant code is here:
https://github.com/microsoft/reverse-proxy/blob/5291514a487f8ccb3866d94b541674c964367db8/src/ReverseProxy/Service/Proxy/HttpProxy.cs#L708-L713

Also related: https://github.com/microsoft/reverse-proxy/issues/439

Tratcher:
Mitigation: Response header transforms can remove unwanted headers.
https://microsoft.github.io/reverse-proxy/articles/transforms.html#responseheader

```
{
  "ResponseHeader": "Connection",
  "Set": "",
  "When": "Always"
}
```

devbeard:
Thanks @Tratcher. I'll give the workaround a go.

samsp-msft:
[...]

Issue #594, opened 2020-12-03
Separate IHttpProxy and Transforms


Tratcher:
IHttpProxy is getting much more attention/usage than most other parts of YARP. Long term we're interested in moving that component into ASP.NET Core. However, IHttpProxy also brings in the entire transforms infrastructure which it's not clear if people using IHttpProxy directly need, or if they could use a simplified model (e.g. callbacks).

It's going to be a pretty significant design change. It would require at least three callbacks for the different stages of the request:
- Request creation (multiple steps)
  - Method, Version, url parts
  - HttpContent creation
  - Request headers (split across the request and the content)
- Request trailers (future)
- Response headers
- Response trailers

I'm not sure how incremental the callback logic could be, or if each callback would have to duplicate all of the default logic it was replacing. In the simplest case the default logic runs first and then the transforms try to fix up the HttpRequestMessage afterwards, but this results in redundant work like dis-assembling and re-assembling the uri for every modification, or removing headers that you don't want forwarded, when it would have been better not to copy those headers in the first place.

I'll do some experiments and see what it would take to separate these two.

samsp-msft:
Significant API break, so needs to be in 1.0

Tratcher:
@davidfowl here's a first draft I have working at the IHttpProxy layer:
```
    public record HttpTransforms
    {
        public static readonly HttpTransforms Empty = new HttpTransforms();

        /// <summary>
        /// Indicates if all request headers should be copied to the outgoing HttpRequestMessage before invoking
        /// OnRequest. Disable this if you want to manually control which headers are copied.
        /// Note this copy excludes certain protocol headers like HTTP/2 psuedo headers (":authority").
        /// </summary>
        public bool CopyRequestHeaders { get; init; } = true;

        /// <summary>
        /// Indicates if all response headers should be copied from the received HttpResponseMessage before invoking
        /// OnResponse. Disable this if you want to manually control which headers are copied.
        /// Note this copy excludes certain protocol headers like `Transfer-Encoding: chunked`.
        /// </summary>
        public bool CopyResponseHeaders { get; init; } = true;

        /// <summary>
        /// Indicates if all response trailers should be copied from the received HttpResponseMessage before invoking
        /// OnResponseTrailers. Disable this if you want to manually control which headers are copied.
        /// </summary>
        public bool CopyResponseTrailers { get; init; } = true;

        /// <summary>
        /// A callback that is invoked prior to sending the proxied request. All HttpRequestMessage fields are
[...]

PR #612, opened 2020-12-11
Relayer transfroms

Tratcher:
#594 IHttpProxy is getting much more attention/usage than other parts of YARP. Long term we're interested in moving that component into ASP.NET Core. However, IHttpProxy also brings in the entire transforms infrastructure which it's not clear if people using IHttpProxy directly need, or if they could use a simplified model (e.g. callbacks).

Here's the proposed new transforms model for IHttpProxy:
```
    public record HttpTransforms
    {
        public static readonly HttpTransforms Empty = new HttpTransforms();

        /// <summary>
        /// Indicates if all request headers should be copied to the outgoing HttpRequestMessage before invoking
        /// OnRequest. Disable this if you want to manually control which headers are copied.
        /// Note this copy excludes certain protocol headers like HTTP/2 psuedo headers (":authority").
        /// </summary>
        public bool CopyRequestHeaders { get; init; } = true;

        /// <summary>
        /// Indicates if all response headers should be copied from the received HttpResponseMessage before invoking
        /// OnResponse. Disable this if you want to manually control which headers are copied.
        /// Note this copy excludes certain protocol headers like `Transfer-Encoding: chunked`.
        /// </summary>
        public bool CopyResponseHeaders { get; init; } = true;

        /// <summary>
        /// Indicates if all response trailers should be copied from the received HttpResponseMessage before invoking
        /// OnResponseTrailers. Disable this if you want to manually control which headers are copied.
        /// </summary>
        public bool CopyResponseTrailers { get; init; } = true;

        /// <summary>
        /// A callback that is invoked prior to sending the proxied request. All HttpRequestMessage fields are
        /// initialized except RequestUri, which will be initialized after the callback if no value is provided.
        /// The string parameter represents the destination URI prefix that should be used when constructing the RequestUri.
        /// The headers will be copied before the callback if CopyRequestHeaders is enabled.
        /// </summary>
        public Func<HttpContext, HttpRequestMessage, string, Task> OnRequest { get; init; }

        /// <summary>
        /// A callback that is invoked when the proxied response is received. The status code and reason phrase will be copied
        /// to the HttpContext.Response before the callback is invoked, but may still be modified there. The headers will be
        /// copied to HttpContext.Response.Headers before the callback if CopyResponseHeaders is enabled.
        /// </summary>
        public Func<HttpContext, HttpResponseMessage, Task> OnResponse { get; init; }

        /// <summary>
        /// A callback that is invoked after the response body to modify trailers, if supported. The trailers will be
        /// copied to the HttpContext.Response before the callback if CopyResponseTrailers is enabled.
        /// </summary>
        public Func<HttpContext, HttpResponseMessage, Task> OnResponseTrailers { get; init; }
    }
```

The idea is that everything should just work but you can get in and tweak things as needed. The higher layer uses these callbacks to hook in and run about the same structured transforms logic it was doing before. There's a little code duplication around header copying to to investigate, and some tests that need cleaning up.

### DIFF
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -88,14 +88,14 @@ namespace Microsoft.ReverseProxy.Service.Proxy
             HttpContext context,
             string destinationPrefix,
             HttpMessageInvoker httpClient,
-            HttpTransforms transforms,
+            HttpTransformer transformer,
             RequestProxyOptions requestOptions)
         {
             _ = context ?? throw new ArgumentNullException(nameof(context));
             _ = destinationPrefix ?? throw new ArgumentNullException(nameof(destinationPrefix));
             _ = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
 
-            transforms ??= HttpTransforms.Default;
+            transformer ??= HttpTransformer.Default;
 
             // HttpClient overload for SendAsync changes response behavior to fully buffered which impacts performance
             // See discussion in https://github.com/microsoft/reverse-proxy/issues/458
@@ -117,7 +117,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
 
                 // :: Step 1-3: Create outgoing HttpRequestMessage
                 var (destinationRequest, requestContent) = await CreateRequestMessageAsync(
-                    context, destinationPrefix, transforms, requestOptions, isStreamingRequest, requestAborted);
+                    context, destinationPrefix, transformer, requestOptions, isStreamingRequest, requestAborted);
 
                 // :: Step 4: Send the outgoing request using HttpClient
                 HttpResponseMessage destinationResponse;
@@ -173,7 +173,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
 
                 // :: Step 5: Copy response status line Client ◄-- Proxy ◄-- Destination
                 // :: Step 6: Copy response headers Client ◄-- Proxy ◄-- Destination
-                await CopyResponseStatusAndHeadersAsync(destinationResponse, context, transforms);
+                await CopyResponseStatusAndHeadersAsync(destinationResponse, context, transformer);
 
                 // :: Step 7-A: Check for a 101 upgrade response, this takes care of WebSockets as well as any other upgradeable protocol.
                 if (destinationResponse.StatusCode == HttpStatusCode.SwitchingProtocols)
@@ -202,7 +202,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
                 }
 
                 // :: Step 8: Copy response trailer headers and finish response Client ◄-- Proxy ◄-- Destination
-                await CopyResponseTrailingHeadersAsync(destinationResponse, context, transforms);
+                await CopyResponseTrailingHeadersAsync(destinationResponse, context, transformer);
 
                 if (isStreamingRequest)
                 {
@@ -243,7 +243,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
         }
 
         private async Task<(HttpRequestMessage, StreamCopyHttpContent)> CreateRequestMessageAsync(HttpContext context, string destinationPrefix,
-            HttpTransforms transforms, RequestProxyOptions requestOptions, bool isStreamingRequest, CancellationToken requestAborted)
+            HttpTransformer transformer, RequestProxyOptions requestOptions, bool isStreamingRequest, CancellationToken requestAborted)
         {
             // "http://a".Length = 8
             if (destinationPrefix == null || destinationPrefix.Length < 8)
@@ -279,7 +279,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
             destinationRequest.Content = requestContent;
 
             // :: Step 3: Copy request headers Client --► Proxy --► Destination
-            await transforms.TransformRequestAsync(context, destinationRequest, destinationPrefix);
+            await transformer.TransformRequestAsync(context, destinationRequest, destinationPrefix);
 
             // Allow someone to custom build the request uri, otherwise provide a default for them.
             var request = context.Request;
@@ -434,13 +434,13 @@ namespace Microsoft.ReverseProxy.Service.Proxy
             context.Response.StatusCode = StatusCodes.Status502BadGateway;
         }
 
-        private static Task CopyResponseStatusAndHeadersAsync(HttpResponseMessage source, HttpContext context, HttpTransforms transforms)
+        private static Task CopyResponseStatusAndHeadersAsync(HttpResponseMessage source, HttpContext context, HttpTransformer transformer)
         {
             context.Response.StatusCode = (int)source.StatusCode;
             context.Features.Get<IHttpResponseFeature>().ReasonPhrase = source.ReasonPhrase;
 
             // Copies headers
-            return transforms.TransformResponseAsync(context, source);
+            return transformer.TransformResponseAsync(context, source);
         }
 
         private async Task HandleUpgradedResponse(HttpContext context, HttpResponseMessage destinationResponse,
@@ -567,10 +567,10 @@ namespace Microsoft.ReverseProxy.Service.Proxy
             ResetOrAbort(context, isCancelled: responseBodyCopyResult == StreamCopyResult.Canceled);
         }
 
-        private static Task CopyResponseTrailingHeadersAsync(HttpResponseMessage source, HttpContext context, HttpTransforms transforms)
+        private static Task CopyResponseTrailingHeadersAsync(HttpResponseMessage source, HttpContext context, HttpTransformer transformer)
         {
             // Copies trailers
-            return transforms.TransformResponseTrailersAsync(context, source);
+            return transformer.TransformResponseTrailersAsync(context, source);
         }
 
 
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -88,8 +88,8 @@ namespace Microsoft.ReverseProxy.Service.Proxy
             HttpContext context,
             string destinationPrefix,
             HttpMessageInvoker httpClient,
-            HttpTransformer transformer,
-            RequestProxyOptions requestOptions)
+            RequestProxyOptions requestOptions,
+            HttpTransformer transformer)
         {
             _ = context ?? throw new ArgumentNullException(nameof(context));
             _ = destinationPrefix ?? throw new ArgumentNullException(nameof(destinationPrefix));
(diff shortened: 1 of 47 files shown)


Tratcher:
@davidfowl Why abstract + Default instead of having the defaults directly in a virtual base class? As a way of communicating to people that they must derive from it to change anything?

src/ReverseProxy/Service/Proxy/HttpTransforms.cs
```diff
@@ -0,0 +1,116 @@
...
+                // Note that trailers, if any, should already have been declared in Proxy's response
+                // by virtue of us having proxied all response headers in step 6.
+                CopyResponseHeaders(proxyResponse.TrailingHeaders, outgoingTrailers);
+            }
+
+            return Task.CompletedTask;
+        }
+
+
+        private static void CopyResponseHeaders(HttpHeaders source, IHeaderDictionary destination)
```
davidfowl, line 103: Should we make this protected static or is it simple enough to copy around

src/ReverseProxy/Service/Proxy/HttpTransforms.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Net.Http.Headers;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.Primitives;
+using Microsoft.ReverseProxy.Utilities;
+
+namespace Microsoft.ReverseProxy.Service.Proxy
+{
+    public class HttpTransforms
```
davidfowl, line 15: - HttpTransfomer
- ProxyTransfomer
- HttpProxyTransformer
- Some noun...

src/ReverseProxy/Service/Proxy/IHttpProxy.cs
```diff
@@ -25,7 +25,7 @@ public interface IHttpProxy
             HttpContext context,
             string destinationPrefix,
             HttpMessageInvoker httpClient,
-            Transforms transforms,
+            HttpTransforms transforms,
```
davidfowl, line 28: Should we have an overload where transforms are optional and the default thing happens?

src/ReverseProxy/Service/RuntimeModel/Transforms/RequestHeaderTransform.cs
```diff
@@ -15,8 +16,10 @@ public abstract class RequestHeaderTransform
...
         /// </summary>
         /// <param name="context">The original <see cref="HttpContext"/> for accessing other state.</param>
-        /// <param name="values">The original header value(s).</param>
-        /// <returns>The transformed result.</returns>
-        public abstract StringValues Apply(HttpContext context, StringValues values);
+        /// <param name="request">The outgoing <see cref="HttpRequestMessage"/> for reference.</param>
+        /// <param name="values">The original header value(s) from the incoming request, or StringValues.Empty if the header
+        /// was not present.</param>
+        /// <returns>The transformed result, or StringValues.Empty if the header should be suppressed.</returns>
+        public abstract StringValues Apply(HttpContext context, HttpRequestMessage request, StringValues values);
```
davidfowl, line 23: nit: request -> proxyRequest

src/ReverseProxy/Service/RuntimeModel/Transforms/Transforms.cs
```diff
@@ -37,26 +38,170 @@ public class Transforms
...
+            }
+        }
+
+        private static void RunRemainingResponseTransforms(HttpResponseMessage response, HttpContext context, IHeaderDictionary destination,
+            IReadOnlyDictionary<string, ResponseHeaderTransform> transforms, HashSet<string> transformsRun)
+        {
+            // Run any transforms that weren't run yet.
+            foreach (var (headerName, transform) in transforms) // TODO: What about multiple transforms per header? Last wins?
+            {
+                if (!(transformsRun?.Contains(headerName) ?? false))
```
davidfowl, line 195: this code is still ugly.

src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.ReverseProxy.Utilities
+{
+    internal static class RequestUtilities
+    {
+        // TODO: this list only contains "Transfer-Encoding" because that messes up Kestrel. If we don't need to add any more here then it would be more efficient to
+        // check for the single value directly. What about connection headers?
+        internal static readonly HashSet<string> ResponseHeadersToSkip = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
davidfowl, line 19: This may as well be an array.

src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// <param name="destinationPrefix">The scheme, host, port, and possibly path base for the destination server.</param>
+        /// <param name="path">The path to append.</param>
+        /// <param name="query">The query to append</param>
+        internal static Uri MakeDestinationAddress(string destinationPrefix, PathString path, QueryString query)
+        {
+            var builder = new StringBuilder(destinationPrefix);
+            if (path.HasValue)
+            {
+                // When PathString has a value it always starts with a '/'. Avoid double slashes when concatenating.
+                if (builder.Length > 0 && builder[^1] == '/')
```
davidfowl, line 36: +1 points for using range syntax

src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        /// <summary>
+        /// Appends the given path and query to the destination prefix while avoiding duplicate '/'.
+        /// </summary>
+        /// <param name="destinationPrefix">The scheme, host, port, and possibly path base for the destination server.</param>
+        /// <param name="path">The path to append.</param>
+        /// <param name="query">The query to append</param>
+        internal static Uri MakeDestinationAddress(string destinationPrefix, PathString path, QueryString query)
+        {
+            var builder = new StringBuilder(destinationPrefix);
```
davidfowl, line 32: Allocations in the hot path.

davidfowl:
> @davidfowl Why abstract + Default instead of having the defaults directly in a virtual base class? As a way of communicating to people that they must derive from it to change anything? Would a protected constructor accomplish the same?

Yea you're right.

src/ReverseProxy/Service/Proxy/HttpTransforms.cs
Tratcher: If you're going to customize TransformResponseAsync or TransformResponseTrailersAsync enough not to call the base implementation then you probably need to customize this code too. That's what happened for the structured transforms derivation.

src/ReverseProxy/Service/Proxy/HttpTransforms.cs
Tratcher: In the context of the other parameters, `HttpTransformer` seems the most consistent:
```C#
        Task ProxyAsync(
            HttpContext context,
            string destinationPrefix,
            HttpMessageInvoker httpClient,
            HttpTransfomer transformer,
            RequestProxyOptions requestOptions);
```

src/ReverseProxy/Service/Proxy/IHttpProxy.cs
Tratcher: HttpTransforms.Default covers providing a default implementation. The HttpProxy implementation also accepts null and uses HttpTransforms.Default internally. Are you concerned about the discoverability? I'll add something to the doc comments for starters.

We usually try to avoid overloads on the main interface. What about an extension method?

src/ReverseProxy/Utilities/RequestUtilities.cs
Tratcher: I expect this logic to get more interesting when we fix https://github.com/microsoft/reverse-proxy/issues/583. I'll hold off changes until then.

src/ReverseProxy/Utilities/RequestUtilities.cs
Tratcher: VS recommended it for me 😁.

src/ReverseProxy/Utilities/RequestUtilities.cs
Tratcher: Yes, but it's trying to save you from even more allocations 😁. Let's revisit this on its own when get get to profiling.

Tratcher:
@davidfowl I think everybody else is out now, consider yourself the sole reviewer.

src/ReverseProxy/Service/Proxy/IHttpProxy.cs
davidfowl: Yep

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -278,59 +268,30 @@ public HttpProxy(ILogger<HttpProxy> logger, IClock clock)
...
-            Log.Proxying(_logger, targetUrl);
-            var targetUri = new Uri(targetUrl, UriKind.Absolute);
-            return new HttpRequestMessage(HttpUtilities.GetHttpMethod(transformContext.Method), targetUri)
-            {
-                Version = transformContext.Version,
-#if NET
-                VersionPolicy = transformContext.VersionPolicy,
-#endif
-            };
+            // TODO: What if they replace the HttpContent object? That would mess with our tracking and error handling.
```
Kahbazi, line 290: Is it possible to create a class which inherits from `HttpRequestMessage` and pass that to `HttpTransformer` and in that class override the `HttpContent` property, so if user replace that we could wrap the `HttpContent` in our own new `HttpContent` which handles and tracks the error.

```csharp
internal class CustomHttpRequestMessage : HttpRequestMessage
{
    private HttpContent _content;

    public new HttpContent Content
    {
        get
        {
            return _content;
        }
        set
        {
            if (value is not CustomeHttpContent)
            {
                _content = new CustomeHttpContent(value);
            }
            else
            {
                _content = value;
            }
        }
    }
}

internal class CustomeHttpContent : HttpContent
{
    private readonly HttpContent _inner;

    public CustomeHttpContent(HttpContent inner)
    {
        _inner = inner;
    }

    protected override Task SerializeToStreamAsync(Stream stream, TransportContext context)
    {
        // track and handle the error here
    }
}

```

davidfowl (approved):

davidfowl:
Do we still want to (eventually) provide an API that makes this easy to do without overriding the IHttpProxy? Something you can do in middleware?

Tratcher:
> Do we still want to (eventually) provide an API that makes this easy to do without overriding the IHttpProxy? Something you can do in middleware?

That's where #60 comes in.

Issue #621, opened 2020-12-15
Support W3C Baggage header


Kahbazi:
### What should we add or change to make your life better?
Add an options in `ActivityPropagationHandler` to choose whether it will use `Correlation-Context` or `baggage` header, or maybe both! 

Related to https://github.com/dotnet/aspnetcore/pull/28328.

### Why is this important to you?
To follow the current standard.

karelz:
Triage: We might need to send both values (old one and new one). We can do it earlier than Runtime (which has separate issue).

karelz:
@Kahbazi do you want to take the work?

Kahbazi:
> @Kahbazi do you want to take the work?

@karelz Yes, I do.

Issue #164, opened 2020-05-14
Allow grouping and inheriting route config


Tratcher:
The config model can get very redundant if you need to define the same property with the same value on many routes. E.g. the Host, BackendId, Metadata, auth, etc..

It may be possible to allow a config structure where routes can be grouped together and inherit some properties (backendid, Prority, metadata, auth, Host, etc.).

samsp-msft:
Do you have an example? What will be different between the routes? Is it just the path to match? In which case would having AND / OR functionality in the paths enable a single route to match a wider set of URIs?

Tratcher:
The example given in #134 was Host. If you have 20 routes with different paths but all with the same Host then it would be convenient to group those together so you only have to define the Host once.

- Group
  - Host: example.com
    - Route1: /foo
    - Route2: /bar
   ...


> would having AND / OR functionality in the paths enable a single route to match a wider set of URIs?

You mean something like `Route: /foo OR /bar`? Routing already allows a parameter matching syntax like `/{paramter}/baz` that I'd expect to handle many of these cases.

Tratcher:
Another request: https://github.com/microsoft/reverse-proxy/issues/122#issuecomment-650671377

Have a shared authorization policy for a group of routes.

Tratcher:
IProxyConfigFilter has proven to be a useful approach to this problem, letting people provide a basic config and to programmatically fill in bulk data.
https://github.com/microsoft/reverse-proxy/blob/ddae8bf0438d5b431845eff82700e145f756175f/testassets/ReverseProxy.Config/CustomConfigFilter.cs#L14

Tratcher:
Alternate proposal for the json based config: If someone defines a route/cluster named "defaults" this would not be rendered into its own route/cluster instance, it would be used to fill in default values for other routes/clusters.
- Pros: 
  - You can enable a setting in one place and have it apply to all routes/clusters
  - You can still override the default per route/cluster
- Cons:
  - This is top level, it's not as granular as the grouping proposals discussed above
  - This wouldn't support any value templating or conditionals like you can do with IProxyConfigFilter
- Unknowns:
  - Would this convention need to be introduced before 1.0? Adding it afterward might cause conflicts for anyone that was already using those names for actual routes/clusters.
  - Would this also make sense for the code model? Why wouldn't you do it yourself when generating the routes/clusters?

samsp-msft:
[...]

PR #624, opened 2020-12-15
Support baggage header in ActivityPropagationHandler

Kahbazi:
I've added an enum for `ActivityContextHeaders` to support `baggage`, `CorrelationContext` or both.
I also shared the enum between the options model and runtime model which I'm not sure if it's ok. 

Fixes https://github.com/microsoft/reverse-proxy/issues/621

### DIFF
--- a/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ActivityContextHeaders.cs
+++ b/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ActivityContextHeaders.cs
@@ -1,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.ReverseProxy.Abstractions
+{
+    [Flags]
+    public enum ActivityContextHeaders
+    {
+        None = 0,
+        Baggage = 1,
+        CorrelationContext = 2,
+        BaggageAndCorrelationContext = Baggage | CorrelationContext
+    }
+}
--- a/src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
+++ b/src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
@@ -8,6 +8,7 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.ReverseProxy.Abstractions;
 
 namespace Microsoft.ReverseProxy.Telemetry
 {
@@ -18,12 +19,16 @@ namespace Microsoft.ReverseProxy.Telemetry
     {
         private const string RequestIdHeaderName = "Request-Id";
         private const string CorrelationContextHeaderName = "Correlation-Context";
+        private const string BaggageHeaderName = "baggage";
 
         private const string TraceParentHeaderName = "traceparent";
         private const string TraceStateHeaderName = "tracestate";
 
-        public ActivityPropagationHandler(HttpMessageHandler innerHandler) : base(innerHandler)
+        private readonly ActivityContextHeaders _activityContextHeaders;
+
+        public ActivityPropagationHandler(ActivityContextHeaders activityContextHeaders, HttpMessageHandler innerHandler) : base(innerHandler)
         {
+            _activityContextHeaders = activityContextHeaders;
         }
 
         protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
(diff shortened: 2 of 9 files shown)


Tratcher:
@MihaZupan can you be in charge of reviewing and merging this PR (after break)?

src/ReverseProxy/Telemetry/ActivityPropagationHandler.cs
```diff
@@ -18,12 +19,16 @@ public sealed class ActivityPropagationHandler : DelegatingHandler
     {
         private const string RequestIdHeaderName = "Request-Id";
         private const string CorrelationContextHeaderName = "Correlation-Context";
+        private const string BaggageHeaderName = "Baggage";
```
MihaZupan, line 22:
```suggestion
        private const string BaggageHeaderName = "baggage";
```

> In order to increase interoperability across multiple protocols and encourage successful integration, implementations SHOULD keep the header name lowercase.

we should use the lowercase header name as suggested by the spec

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ActivityContextHeaders.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+
+namespace Microsoft.ReverseProxy.Abstractions
+{
+    [Flags]
+    public enum ActivityContextHeaders
+    {
+        Baggage = 1,
+        CorrelationContext = 2,
+        BaggageAndCorrelationContext = 3
```
MihaZupan, line 13:
```suggestion
        BaggageAndCorrelationContext = Baggage | CorrelationContext
```

nit

src/ReverseProxy/Service/RuntimeModel/ClusterProxyHttpClientOptions.cs
```diff
@@ -48,12 +53,13 @@ public bool Equals(ClusterProxyHttpClientOptions other)
...
                    MaxConnectionsPerServer == other.MaxConnectionsPerServer &&
-                   PropagateActivityContext == other.PropagateActivityContext;
+                   PropagateActivityContext == other.PropagateActivityContext &&
+                   ActivityContextHeaders == other.ActivityContextHeaders;
         }
 
         public override int GetHashCode()
         {
-            return HashCode.Combine(SslProtocols, DangerousAcceptAnyServerCertificate, ClientCertificate, MaxConnectionsPerServer);
+            return HashCode.Combine(SslProtocols, DangerousAcceptAnyServerCertificate, ClientCertificate, MaxConnectionsPerServer, ActivityContextHeaders);
```
MihaZupan, line 62: We should include `PropagateActivityContext` in the hash code as well, likely a harmless oversight before

MihaZupan (commented):
Change LGTM with some small comments.

One thing to note is that `PropagateActivityContext` isn't really needed anymore - we could add `None = 0` to `ActivityContextHeaders` and remove `PropagateActivityContext`.

Tratcher:
This will need to be rebased.

Tratcher:
@MihaZupan @Kahbazi is there a reason ActivityPropagationHandler isn't written as a request transform? It might have been tough with the preview7 transforms infrastructure, but it should be compatible with the current infrastructure.

Kahbazi:
> @MihaZupan @Kahbazi is there a reason ActivityPropagationHandler isn't written as a request transform? It might have been tough with the preview7 transforms infrastructure, but it should be compatible with the current infrastructure.

I think the initial attempt was to add `DiagnosticHandler`, but now it makes more sense to have it as a transform.

So should I close this PR and open a new one for adding it as a transform after #703 got merged?

Tratcher:
There are some trade-offs.
- Transforms are configured per-route, not cluster.
- Structured transforms aren't available when using IHttpProxy directly.

Does activity propagation make sense as a route specific configuration, or does it make more sense as part of the cluster config? Will people using IHttpProxy directly want built-in activity propagation?

Kahbazi:
> Does activity propagation make sense as a route specific configuration, or does it make more sense as part of the cluster config?

I can't think of a scenario where activity propagation is needed only for a specific route. I always want it for all routes of a cluster, but then again I feel the same about forwarded transform and keeping original host header. Perhaps this makes more sense after implementing group for route config (#164)

> Will people using IHttpProxy directly want built-in activity propagation?

Since the default behavior of `HttpClient` has activity propagation and also events for diagnostic source, I think people expect the same behavior with `IHttpProxy`.

Tratcher:
> Since the default behavior of `HttpClient` has activity propagation and also events for diagnostic source, I think people expect the same behavior with `IHttpProxy`.

Ok, that rules out using transforms for this. Let's continue with the PR.

Tratcher (approved):

MihaZupan (approved):
LGTM

PR #626, opened 2020-12-16
Update proxyhttpclientconfig.md

sleemer:
fix: DangerousAcceptAnyServerCertificate instead of ValidateRemoteCertificate.

PR #628, opened 2020-12-16
Update proxyhttpclientconfig.md (#626)

alnikola:
fix: DangerousAcceptAnyServerCertificate instead of ValidateRemoteCertificate.


Tratcher (approved):

alnikola:
Hmm, it automatically deleted `release/docs` branch for some reason. Is it OK?

Tratcher:
Restore it. There's a rule that does that for all or branches merged to master. We should figure out how to suppress it for release branches.

Tratcher:
I've added branch protection rules for `release/docs` similar to the ones on master. That suppressed the delete-branch button, so it might suppress the auto-delete branch as well.

Tratcher:
Side note: remember to make merge commits (rather than squash-and-merge or rebase) when merging between long lived branches.

PR #629, opened 2020-12-16
Add dotnet6 feed for benchmarks

sebastienros:


Tratcher (approved):
Do we still need the dotnet5 feed too?

sebastienros:
Since 5.0 packages are on nuget.org I believe we don't. Updating the file.

Issue #588, opened 2020-12-02
[SF] Rename Host to Hosts


alnikola:
`YARP.Routes.<routeName>.Host` label of the SF service manifest has singular name `Host`, but it can contain a string with multiple host names. It would be better to rename it to the plural form 'Host**s**'.

Tratcher:
It looks like the code was fixed in the initial PR and only the docs were wrong.

PR #635, opened 2020-12-17
Change Host to Hosts in doc

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/588


Tratcher:
When this was originally filed this wasn't just a doc issue, the "Host" name is used in the source code as well. I'm not seeing that anymore, but can you also check yourself in case I missed it?

Kahbazi:
I checked the source. Looks like it's already been changed to `Hosts`.

Tratcher (approved):

Tratcher:
Thanks for double checking.

PR #636, opened 2020-12-18
Use ValueStopwatch to reduce allocation

Kahbazi:
`ValueStopwatch` could replace `Stopwatch` to reduce allocation.
Since `ValueStopwatch` is an internal type in `ReverseProxy` project, I added it to `ReverseProxy.ServiceFabric` as a linked file.


Tratcher (approved):

Tratcher:
Thanks

PR #637, opened 2020-12-19
Fix the link to the Code of Conduct

Kahbazi:
Or we could rename the file so it has the same name as it has in ASP.Net Core repository.


Tratcher:
Thanks. This format matches other repos in the Microsoft org so I think we'll keep it that way.

Issue #416, opened 2020-09-14
Documentation for swapping authentication types


crismsx:
A common scenario I'm running into is to proxy a request for the sole reason of swapping the authentication token. 

In particular, the use case involves the swapping of 2 OAuth2 tokens. The client has access to one IdentityServer and the resource server to a different IdentityServer. The proxy in between is the only component that has access to both IdentityServers and is able to swap the access tokens. The swap should work only for specific (configured) routes.

Consider adding documentation that covers this authentication scenario.

Tratcher:
See this discussion: https://github.com/microsoft/reverse-proxy/discussions/375

As this is a flow that's likely require a lot of customization I don't know that we'll be able to provide more than general guidance and docs. We'll have to wait and see what the most common patterns are.

karelz:
Triage: We can make it an example of adding a header based on the incoming user - likely related to #60

macsux:
We've been using HTTP Headers to do Auth/Authz at the proxy and propagate identity downstream as HTTP headers. This implies there is implicit transport security between proxy and final destination provided by the infrastructure, but this is now increasingly common with SDN, K8S, and use of reverse proxies in sidecar scenarios. I believe this is also a model that is applied by Azure AppService (they auto inject HTTP module to reconstruct .NET Principal from authentication headers). Consider offering this out of the box as it greatly simplifies downstream services processing identity without having to deal with any security blocks. We've been using this to offload Kerberos validation to the sidecar. The identity is extracted and 2 headers are sent downstream:

X-Identity: someuser\domain
X-Roles: User,SomeGroup
Alternatively repackaging identity into JWT can provide greater flexibility as it allows for the much richer expression of principal than plaintext headers above. For this option, it would be useful to have the ability to send both signed and unsigned tokens. Unsigned JWT tokens would lessen the burden of dealing with managing JWK that proxy would need to expose and downstream having to configure.

Signed JWT should also be supported by exposing a JWK url at the proxy. This is to facilitate scenarios where security between proxy and destination cannot be enforced at the transport infrastructure level. Also, some JWT libraries don't support unsigned JWT tokens.

[...]

PR #646, opened 2021-01-05
Refactor structured transforms

Tratcher:
This takes the work from #612 and applies it to structured transforms in preparation for #60 and #416. It incorporates feedback about the transforms being too structured, unable to arbitrarily manipulate the request/response, rename headers, do async work, etc.

The original design:
- Distinguished between request and request header transforms.
- Applied transforms while copying request/response headers. This was done for efficiency but required a complex indexing system and limited you to one transform rule per header name.
- Transforms are sync
- Shared transform logic for response headers and trailers, individual transforms only look at header values passed to them.

The new design:
- Unified request and request header transforms.
- Headers are copied before transforms are applied (if enabled). Transforms are then applied to the result. This is a bit less efficient since you need to copy the value, remove it, and then replace it, but it only affects headers with transforms, not every header. It's also a bit more work to write transforms against HttpRequestMessage/HttpContent but I've added some helper methods to smooth this out (TakeHeader, AddHeader).
- Transforms are more powerful now, they can affect the whole request rather than just a specific header/field.
- Transforms are async
- Split response header and trailer transforms since transforms now need to deal with the source and destination of their data.

I know this is a lot of churn. The most interesting changes to review are:
- [TransformBuilder](https://github.com/microsoft/reverse-proxy/pull/646/files#diff-0fb89f2699f9c7d3143b84021d8d836b2fcd1b1ec0a0223a28bc432d2b3592f8)
- [StructuredTransformer](https://github.com/microsoft/reverse-proxy/pull/646/files#diff-113396a09a3cd8e1fa7f4c65d14d79c3b7804dc84f8188f356dcf108da29fee8)
- [RequestTransform](https://github.com/microsoft/reverse-proxy/pull/646/files#diff-2e296d77a68dc839ce72d7838e78ddeb2eeb2d07ee7de1892a61e06b56c95e10)
- [ResponseTransform](https://github.com/microsoft/reverse-proxy/pull/646/files#diff-c3965c2ef1f5d65223fe80dd76f549d2c549d807f6bfdafdafc7d0ca58b8dc10)


src/ReverseProxy/Service/RuntimeModel/Transforms/RequestTransformContext.cs
```diff
@@ -1,29 +1,34 @@
...
         /// applying header transforms.
         /// </summary>
-        public HttpRequestMessage ProxyRequest { get; internal set; }
+        public HttpRequestMessage ProxyRequest { get; init; }
+
+        /// <summary>
+        /// Gets or sets if the request headers have been copied from the HttpRequest to the HttpRequestMessage and HttpContent.
+        /// Transforms use this when searching for the current value of a header they should operate on.
+        /// </summary>
+        public bool HeadersCopied { get; set; }
```
Kahbazi, line 31: Why isn't this also `init`? I can't think of a scenario if a transform would want to change this property.

Kahbazi (approved):

src/ReverseProxy/Service/RuntimeModel/Transforms/StructuredTransformer.cs
```diff
@@ -20,192 +20,137 @@ namespace Microsoft.ReverseProxy.Service.RuntimeModel.Transforms
...
-            Dictionary<string, RequestHeaderTransform> requestHeaderTransforms,
-            Dictionary<string, ResponseHeaderTransform> responseHeaderTransforms,
-            Dictionary<string, ResponseHeaderTransform> responseTrailerTransforms)
+        internal StructuredTransformer(bool? copyRequestHeaders, bool? copyResponseHeaders, bool? copyResponseTrailers,
+            IList<RequestTransform> requestTransforms,
+            IList<ResponseTransform> responseTransforms,
+            IList<ResponseTrailersTransform> responseTrailerTransforms)
         {
             ShouldCopyRequestHeaders = copyRequestHeaders;
+            ShouldCopyResponseHeaders = copyResponseHeaders;
```
Kahbazi, line 32: We could set the value in constructor only once and change the `bool?` to `bool`.
```suggestion
            ShouldCopyResponseHeaders = copyResponseHeaders.GetValueOrDefault(true);
```

src/ReverseProxy/Service/RuntimeModel/Transforms/RequestTransformContext.cs
Tratcher: We worked with a customer recently that was building their own proxy with Kestrel and HttpClient and they ran into header encoding issues in both directions. Such issues could be dealt with here by disabling the default copy logic and providing a transform that handled the encoding issues while copying the headers. That transform would set HeadersCopied when it was done so subsequent transforms would know where to look for headers.

src/ReverseProxy/Service/RuntimeModel/Transforms/RequestTransform.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// then the HttpContent, and falling back to the HttpContext only if
+        /// <see cref="RequestTransformContext.HeadersCopied"/> is not set.
+        /// This ordering allows multiple transforms to mutate the same header.
+        /// </summary>
+        /// <param name="headerName">The name of the header to take.</param>
+        /// <returns>The requested header value, or StringValues.Empty if none.</returns>
+        public static StringValues TakeHeader(RequestTransformContext context, string headerName)
+        {
+            var existingValues = StringValues.Empty;
+            if (context.ProxyRequest.Headers.TryGetValues(headerName, out var values))
```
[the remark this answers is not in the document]
Tratcher: I don't expect that to be a problem. The standard approach for adding a header is to try adding it to the HttpRequestMessage and fall back to the HttpContent if that fails. Even if that were reversed there shouldn't be duplicate headers. These HttpRequest/ResponseMessages aren't created by arbitrary user code, they're created by HttpProxy or HttpClient so the way they add headers should be predictable.

src/ReverseProxy/Service/RuntimeModel/Transforms/RequestTransform.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    /// <summary>
+    /// The base class for request transforms.
+    /// </summary>
+    public abstract class RequestTransform
+    {
+        /// <summary>
+        /// Transforms any of the available fields before building the outgoing request.
+        /// </summary>
+        public abstract Task ApplyAsync(RequestTransformContext context);
```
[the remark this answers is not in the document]
Tratcher: I tried that but ValueTask is not well supported on 3.1. I'm not sure it matters as much for Tasks without T results, we're able to return the same CompletedTask when it's not async.

src/ReverseProxy/Service/RuntimeModel/Transforms/RequestTransform.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        /// <param name="headerName">The name of the header to take.</param>
+        /// <returns>The requested header value, or StringValues.Empty if none.</returns>
+        public static StringValues TakeHeader(RequestTransformContext context, string headerName)
+        {
+            var existingValues = StringValues.Empty;
+            if (context.ProxyRequest.Headers.TryGetValues(headerName, out var values))
+            {
+                context.ProxyRequest.Headers.Remove(headerName);
+                existingValues = values.ToArray();
```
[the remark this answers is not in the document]
Tratcher: I'll try casting for now and see if there are cases later that break it.

src/ReverseProxy/Service/RuntimeModel/Transforms/StructuredTransformer.cs
Tratcher: I'd left it this way so I could test TransformBuilder for the cases where no config was provided vs an explicit negative config given. If it shows up in a trace then I'll remove it and find another way to test the builder.

src/ReverseProxy/Service/RuntimeModel/Transforms/RequestTransform.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Removes and returns the current header value by first checking the HttpRequestMessage,
+        /// then the HttpContent, and falling back to the HttpContext only if
+        /// <see cref="RequestTransformContext.HeadersCopied"/> is not set.
+        /// This ordering allows multiple transforms to mutate the same header.
+        /// </summary>
+        /// <param name="headerName">The name of the header to take.</param>
+        /// <returns>The requested header value, or StringValues.Empty if none.</returns>
+        public static StringValues TakeHeader(RequestTransformContext context, string headerName)
```
Tratcher, line 29: Would these Take and Add methods make more sense as instance methods on the RequestTransformContext?

alnikola (approved):
LGTM

davidfowl (approved):

Issue #439, opened 2020-09-24
Do not flow the Connection header?


Tratcher:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.10

The Connection header is per hop and shouldn't be forwarded through a proxy. Note most of the spec concerns here apply to traditional forward proxies, but they might still be useful for a reverse proxy. E.g. Forwarding a `Connection: close` header would allow the client to terminate the proxy's connection to the destination and negatively impact performance.

Note for WebSockets we do need to forward the `Connection: Upgrade` header, so we can't ban Connection outright.

karelz:
Triage: Motivation protect performance - we should filter out `Connection: close` (we don't have more at the moment).

Tratcher:
Keep-Alive should also be removed from requests. We already remove it from HTTP/2 responses.

PR #659, opened 2021-01-12
Remove connection-specific headeres for HTTP/2 and HTTP/3

### LINKED ISSUE #583: HTTP/2 connection to YARP with HTTP/1.1 to backend fails

MihaZupan:
Fixes #583

YARP workaround for https://github.com/dotnet/aspnetcore/pull/24543 that was merged into 6.0 in Kestrel.


src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -14,11 +14,26 @@ namespace Microsoft.ReverseProxy.Utilities
...
     {
-        // TODO: this list only contains "Transfer-Encoding" because that messes up Kestrel. If we don't need to add any more here then it would be more efficient to
-        // check for the single value directly. What about connection headers?
-        internal static readonly HashSet<string> ResponseHeadersToSkip = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        internal static bool ShouldSkipResponseHeader(string headerName, bool isHttp2OrGreater)
         {
-            HeaderNames.TransferEncoding
+            if (isHttp2OrGreater)
+            {
+                // This check can be dropped in .NET 6.0+ as these headers will be removed by Kestrel
```
Tratcher, line 21: It's probably better to keep the check even after the kestrel fix. It should be more efficient to avoid adding the header than to add and remove it.

Tratcher (commented):
This handles h2 response protocol correctness but there are more scenarios to consider. https://tools.ietf.org/html/rfc7230#section-6.1 says we should remove the Connection header from all requests and responses, and all headers referenced by that (e.g. forwarding `Connection: close` would be bad for proxy scenarios). However, removing `Connection: Upgrade` would break WebSockets.

Tratcher:
We also need to test what happens for requests. Does HttpClient filter out these headers on h2/3?

MihaZupan:
This PR specifically fixes the Http2 client => Http11 backend scenario described in #583.

Should the connection header (#439) be dealt-with in the same PR?

Tratcher:
This PR should also cover the request version of the #583 scenario.

Do some reading about the Connection header and decide if you want to handle that at the same time or as a separate issue/PR.

MihaZupan:
On the request side, HttpClient drops these headers:
https://github.com/dotnet/runtime/blob/45be6750943a043c15f2ce5e4700e48e63b5c7c0/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/Http2Connection.cs#L1143

Tratcher:
That's good. HttpClient doesn't skip the Keep-Alive header?

Tratcher (approved):
The Keep-Alive question can be investigated separately.

PR #662, opened 2021-01-13
Make the config object model read-only

### LINKED ISSUE #179: Use freezable pattern for contract types instead of deep cloning

Tratcher:
Fixes #179

Today Cluster, ProxyRoute, etc. are mutable objects used to build the initial data model. Because these are mutable we clone them several times to prevent anyone from modifying them when we don't expect it, such as before calling user callbacks.

To avoid extra copies we want to make the input object model immutable but still have it be easy to construct. Here I've experimented with using C# 9 [record types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#record-types) to achieve that. A record type is an immutable class with automatically generated methods like Clone and Equals. Using `init` only properties makes the objects still easy to construct. The new `with` operator allows making mutated clones of records so we can still manipulate the objects when we need to but not worry about someone else changing them unexpectedly.

There's a lot of churn-in-place here, but I'm happy with the results. All DeepClones were removed and I was able to remove many redundant classes from RuntimeModel that were only there as read-only copies of the config model.


samples/ReverseProxy.Code.Sample/CustomConfigFilter.cs
```diff
@@ -36,16 +36,16 @@ public Task ConfigureClusterAsync(Cluster cluster, CancellationToken cancel)
             return Task.CompletedTask;
         }
 
-        public Task ConfigureRouteAsync(ProxyRoute route, CancellationToken cancel)
+        public Task<ProxyRoute> ConfigureRouteAsync(ProxyRoute route, CancellationToken cancel)
```
davidfowl, line 39: ValueTask

src/ReverseProxy/Abstractions/Config/IProxyConfigFilter.cs
```diff
@@ -23,6 +23,6 @@ public interface IProxyConfigFilter
         /// Allows modification of a route configuration.
         /// </summary>
         /// <param name="route">The ProxyRoute instance to configure.</param>
-        Task ConfigureRouteAsync(ProxyRoute route, CancellationToken cancel);
+        Task<ProxyRoute> ConfigureRouteAsync(ProxyRoute route, CancellationToken cancel);
```
davidfowl, line 26: ValueTask

src/ReverseProxy/Abstractions/Config/ITransformBuilder.cs
```diff
@@ -16,11 +16,11 @@ public interface ITransformBuilder
         /// Validates that each transform is known and has the expected parameters. All transforms are validated and
         /// so all errors can be reported.
         /// </summary>
-        IList<Exception> Validate(IList<IDictionary<string, string>> transforms);
+        IList<Exception> Validate(IReadOnlyList<IReadOnlyDictionary<string, string>> transforms);
```
davidfowl, line 19: This looks crazy 😄

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyMatch.cs
```diff
@@ -3,52 +3,40 @@
...
 using System.Collections.Generic;
-using System.Linq;
 using Microsoft.ReverseProxy.Utilities;
 
 namespace Microsoft.ReverseProxy.Abstractions
 {
     /// <summary>
     /// Describes the matching criteria for a route.
     /// </summary>
-    public class ProxyMatch : IDeepCloneable<ProxyMatch>
```
davidfowl, line 14: Implement IEqutable and override Equals? I believe Records do this automagically but we want to override that logic here.

davidfowl:
I like it!

samples/ReverseProxy.Config.Sample/ReverseProxy.Config.Sample.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFrameworks>net5.0;netcoreapp3.1</TargetFrameworks>
     <OutputType>Exe</OutputType>
     <RootNamespace>Microsoft.ReverseProxy.Sample</RootNamespace>
+    <LangVersion>9.0</LangVersion>
```
WeihanLi, line 7: Maybe we could move this into root `Directory.Build.props` so that no need to set `LangVersion` for every project

samples/ReverseProxy.Code.Sample/CustomConfigFilter.cs
Tratcher: I tried that previously and it was very cumbersome on 3.1. I can take another look, but do we have any good examples of usage for 3.1?

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyMatch.cs
Tratcher: Records cannot override Equals, that's their most annoying feature so far. Who uses IEqutable?

src/ReverseProxy/Abstractions/Config/ITransformBuilder.cs
Tratcher: Yeah. We'll see if I can sort it out as part of #60 (coming soon).

samples/ReverseProxy.Code.Sample/CustomConfigFilter.cs
davidfowl: The helper methods to create them aren't in 3.1 but it shouldn't be a big change (just new it up instead).

src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
```diff
@@ -7,108 +11,107 @@ namespace Microsoft.ReverseProxy.Service
...
+        /// Clones the ProxyRoute and adds the transform.
         /// </summary>
-        public static void AddTransformPathSet(this ProxyRoute proxyRoute, PathString path)
+        /// <returns>The cloned route with the new transform.</returns>
+        public static ProxyRoute WithTransform(this ProxyRoute proxyRoute, IReadOnlyDictionary<string, string> transform)
         {
-            proxyRoute.Transforms ??= new List<IDictionary<string, string>>();
+            var transforms = proxyRoute.Transforms == null ?
+                new List<IReadOnlyDictionary<string, string>>()
+                : new List<IReadOnlyDictionary<string, string>>(proxyRoute.Transforms);
```
Kahbazi, line 21:
```suggestion
            List<IReadOnlyDictionary<string, string>> transforms;
            if (proxyRoute.Transforms == null)
            {
                transforms = new List<IReadOnlyDictionary<string, string>>();
            }
            else
            {
                transforms = new List<IReadOnlyDictionary<string, string>>(proxyRoute.Transforms.Count + 1);
                transforms.AddRange(proxyRoute.Transforms);
            }
```
Does this suggestion worth the effort to prevent the possible Resizing the underlying array of the list?

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyMatch.cs
ManickaP: You can override `Equals`, you cannot only override `Clone`.
This compiles and works as expected:
```C#
public record ProxyMatch
{
    public IReadOnlyList<string> Headers { get; init; }

    public virtual bool Equals(ProxyMatch other) => Headers.Count == other.Headers.Count;
}
```

src/ReverseProxy/Abstractions/Config/IProxyConfigFilter.cs
ManickaP, line 26: Why added return value only here? What about `ConfigureClusterAsync`? Their usage is very similar so why alter only one of them?

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
```diff
@@ -12,69 +12,53 @@ namespace Microsoft.ReverseProxy.Abstractions
...
-                ClusterId = ClusterId,
-                AuthorizationPolicy = AuthorizationPolicy,
-                CorsPolicy = CorsPolicy,
-                Metadata = Metadata?.DeepClone(StringComparer.OrdinalIgnoreCase),
-                Transforms = Transforms?.Select(d => new Dictionary<string, string>(d, StringComparer.OrdinalIgnoreCase)).ToList<IDictionary<string, string>>(),
-            };
-        }
-
-        internal static bool Equals(ProxyRoute proxyRoute1, ProxyRoute proxyRoute2)
+        public static bool Equals(ProxyRoute proxyRoute1, ProxyRoute proxyRoute2)
```
ManickaP, line 61: Could we get rid of these static `Equals` and replace them with the record implementation of `IEquatable`, overriding wherever we need custom behavior?

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -164,41 +165,44 @@ private Cluster CreateCluster(IConfigurationSection section)
 
         private static ProxyRoute CreateRoute(IConfigurationSection section)
         {
-            var route = new ProxyRoute
+            return new ProxyRoute
```
ManickaP, line 168: Expression body maybe?

```suggestion
            => new ProxyRoute
```

src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
```diff
@@ -47,6 +48,16 @@ internal static Version ReadVersion(this IConfiguration configuration, string na
             return children.ToDictionary(s => s.Key, s => s.Value, StringComparer.OrdinalIgnoreCase);
         }
 
+        internal static IReadOnlyDictionary<string, string> ReadOnlyStringDictionary(this IConfigurationSection section)
```
ManickaP, line 51: Do we really need this? Can we just wrap the result of `ReadStringDictionary` in `new ReadOnlyDictionary<string, string>`?
Or at least this very method could reuse `ReadStringDictionary`.
And it should be named `ReadReadOnlyStringDictionary` if we're to follow naming in this class.

Scratch that. What if we create a different extension:
```C#
public static class ReadOnlyExtensions
{
    public static IReadOnlyDictionary<TKey, TValue> AsReadOnly<TKey, TValue>(IDictionary<TKey, TValue> dictionary)
        => dictionary != null ? new ReadOnlyDictionary<TKey, TValue>(dictionary) : null;
}
```
And just patch the calling place with `.AsReadOnly()`

List already has such method: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.asreadonly?view=net-5.0

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -164,41 +165,44 @@ private Cluster CreateCluster(IConfigurationSection section)
...
-        private static IList<IDictionary<string, string>> CreateTransforms(IConfigurationSection section)
+        private static IReadOnlyList<IReadOnlyDictionary<string, string>> CreateTransforms(IConfigurationSection section)
         {
             if (section.GetChildren() is var children && !children.Any())
             {
                 return null;
             }
 
-            return children.Select(s => s.GetChildren().ToDictionary(d => d.Key, d => d.Value, StringComparer.OrdinalIgnoreCase)).ToList<IDictionary<string, string>>();
+            return children.Select(s => new ReadOnlyDictionary<string, string>(s.GetChildren().ToDictionary(d => d.Key, d => d.Value, StringComparer.OrdinalIgnoreCase)))
```
ManickaP, line 188: Can we split this line a bit, I'm getting lost in the long expression. 
And why aren't we calling the `ReadStringDictionary` here?

src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
```diff
@@ -7,108 +11,107 @@ namespace Microsoft.ReverseProxy.Service
...
-            proxyRoute.Transforms.Add(new Dictionary<string, string>
+            return proxyRoute with { Transforms = transforms };
+        }
+
+        /// <summary>
+        /// Clones the route and adds the transform which sets the request path with the given value.
+        /// </summary>
+        public static ProxyRoute WithTransformPathSet(this ProxyRoute proxyRoute, PathString path)
+        {
+            return proxyRoute.WithTransform(new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
```
ManickaP, line 33: When initializing from the config we went an extra mile and every Dictionary/List properly wrapped in their ReadOnly counterpart. But here we're forgoing that and just passing Dictionary itself.
So either we do care that everything is properly readonly and nobody cannot even cast the instance to non-readonly counterpart. Or we can quite substantially simplify the config reading. 
Or is there any other argument why is it here more lax?

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -238,14 +238,13 @@ private async Task<(IList<ProxyRoute>, IList<Exception>)> VerifyRoutesAsync(IRea
...
-                // Don't modify the original
-                var route = r.DeepClone();
+                var route = r;
 
                 try
                 {
                     foreach (var filter in _filters)
                     {
-                        await filter.ConfigureRouteAsync(route, cancellation);
+                        route = await filter.ConfigureRouteAsync(route, cancellation);
```
ManickaP, line 247: We do this change for route, but not for cluster. Why?

ManickaP (commented):
Not a full pass through the change. I'll continue on Monday.

src/ReverseProxy/Abstractions/Config/IProxyConfigFilter.cs
Tratcher: I haven't done the cluster half of the object model yet. I'll get back to that next week. Don't worry about a detailed review yet, I just wanted to check for any major design issues.

src/ReverseProxy/Abstractions/RouteDiscovery/Contract/ProxyRoute.cs
Tratcher: I'll give IEquatable a try.

There are a few scenarios where one or the other is null and the static API makes it easier to contain that logic. We'll see if how much that affects the usage with IEquatable.

src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
Tratcher: ReadStringDictionary should go away when I finish converting the cluster part of the object model. We'll see what's left to streamline then.

src/ReverseProxy/Service/Config/ProxyRouteTransformExtensions.cs
Tratcher: There is some awkwardness around IDictionary vs IReadOnlyDictionary vs Dictionary. Dictionary implements both interfaces, but IDictionary does not implement IReadOnlyDictionary and requires wrapping. Many of the config and linq APIs only return an IDictionary.

Dictionary is easy to use here and we don't have to worry about it being modified because nobody else ever has access to it without downcasting or other tricks. I can try a ReadOnlyDictionary to see if that's as easy.

ManickaP:
Will this change have any effect on the runtime model (`RouteInfo`, `RouteConfig`) or not? Or this this just to get rid of all the `DeepClone` stuff?

Tratcher:
> Will this change have any effect on the runtime model (`RouteInfo`, `RouteConfig`) or not? Or this this just to get rid of all the `DeepClone` stuff?

The main goal is to get rid of the DeepClone code. I don't expect many changes to the runtime model, but we may be able to remove some redundancy.

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -164,41 +165,44 @@ private Cluster CreateCluster(IConfigurationSection section)
...
+                .ToList<IReadOnlyDictionary<string, string>>().AsReadOnly();
         }
 
-        private static void InitializeProxyMatch(ProxyMatch proxyMatch, IConfigurationSection section)
+        private static ProxyMatch CreateProxyMatch(IConfigurationSection section)
         {
             if (!section.Exists())
             {
-                return;
+                return new ProxyMatch();
```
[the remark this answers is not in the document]
Tratcher: I'm going to return null here instead. A route without any match criteria is invalid regardless, but that check happens later in the config validator.

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
Tratcher: Only a few of these are concise enough for expression bodies and I'd rather use a consistent style.

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ActiveHealthCheckOptions.cs
```diff
@@ -8,62 +8,56 @@ namespace Microsoft.ReverseProxy.Abstractions
     /// <summary>
     /// Active health check options.
     /// </summary>
-    public sealed class ActiveHealthCheckOptions
+    public sealed record ActiveHealthCheckOptions : IEquatable<ActiveHealthCheckOptions>
```
ManickaP, line 11: I don't think we need to manually derive from `IEquatable`, that should `record` do automatically.
```suggestion
    public sealed record ActiveHealthCheckOptions
```

Tratcher:
> > Will this change have any effect on the runtime model (`RouteInfo`, `RouteConfig`) or not? Or this this just to get rid of all the `DeepClone` stuff?
> 
> The main goal is to get rid of the DeepClone code. I don't expect many changes to the runtime model, but we may be able to remove some redundancy.

There was actually a lot of redundancy on ClusterConfig I was able to remove. There were lots of Cluster* types that were read-only copies original config that we didn't need anymore.

src/ReverseProxy/Service/RuntimeModel/ClusterConfig.cs
```diff
@@ -23,49 +21,22 @@ public sealed class ClusterConfig
     {
         public ClusterConfig(
```
ManickaP, line 22: Could we even get rid of `ClusterConfig`? Maybe not necessarily as part of this PR, but in general.

ManickaP (approved):
Look cleaner with the records!

src/ReverseProxy/Microsoft.ReverseProxy.csproj
```diff
@@ -6,7 +6,6 @@
     <OutputType>Library</OutputType>
     <RootNamespace>Microsoft.ReverseProxy</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <LangVersion>9.0</LangVersion>
```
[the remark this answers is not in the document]
Tratcher: It was moved to the root Directory.Build.props file so it's shared by all projects.

src/ReverseProxy/Service/RuntimeModel/ClusterConfig.cs
Tratcher: I think we still need it. It serves as an atomic unit of state that's updated on the ClusterInfo and snapshotted with requests.

PR #700, opened 2021-01-28
Disable Hosting.Diagnostics logging

MihaZupan:
Results in removing the overhead associated with creating an Activity per request.

Diff should be around 13%.

Note: The HttpClient scenario doesn't suffer from this as it's not calling `Host.CreateDefaultBuilder`, but creating the builder manually.


alnikola (approved):
LGTM

PR #703, opened 2021-01-29
Make transforms extensible

### LINKED ISSUE #60: Custom request/response modifications via code transforms

Tratcher:
Fixes #60

Problem:
Developers want to manipulate the requests and responses in arbitrary ways, and to have different sets of transforms apply to different routes. Given routes are loaded from an abstract source, how can they customize those routes as or after they're loaded?

Proposed design:
I've broken the problem up into two parts, providing transforms purely from code vs. providing parametrized transforms that integrate with config.

A) Code-only transforms:
A developer is able to check each route as its created and decides if they should add a transform to it. They can inspect any field on the ProxyRoute to make this decision. This is implemented via one or more ITransformProviders registered in DI, with come convenience methods in Startup for simple Func scenarios. The context has parameterized extension methods for adding known transforms, or the developer can new up their own transform and add it.

```
            services.AddReverseProxy()
                .LoadFromMemory(routes, clusters)
                .AddTransforms(transformBuilderContext =>
                {
                    // For each route+cluster pair decide if we want to add transforms, and if so, which?
                    // This logic is re-run each time a route is rebuilt.

                    transformBuilderContext.AddPathPrefix("/prefix");

                    // Only do this for routes that require auth.
                    if (string.Equals("token", transformBuilderContext.Route.AuthorizationPolicy))
                    {
                        // See the next example
                    }
                });
```

This approach makes it easy to add custom code based transforms without defining additional classes.
```C#
                        transformBuilderContext.AddRequestTransform(async transformContext =>
                        {
                            var user = transformContext.HttpContext.User;
                            var tokenService = transformContext.HttpContext.RequestServices.GetRequiredService<TokenService>();
                            var token = await tokenService.GetAuthTokenAsync(user);
                            transformContext.ProxyRequest.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token);
                        });
```

B) Config based transforms:
Developers that want to integrate more deeply with the config can implement an ITransformFactory. This has methods Validate and Build. The config validator will iterate through the available factories until it finds one that matches the given transform and validates or fails the given parameters. Unknown transforms will cause the config to be rejected as invalid. Later the route builder will use the factories to Build transform instances for that route.

Factories allow for config parameter validation and can support multiple transform types. In the draft so far I've moved all of the path transforms to one factory as these share some logic.


src/ReverseProxy/Abstractions/Config/ITransformFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    public interface ITransformFactory
+    {
+        /// <summary>
+        /// Checks if the given transform values match a known transform, and if those values have any errors.
+        /// </summary>
+        /// <param name="context">The context to add any generated errors to.</param>
+        /// <param name="transformValues">The transform values to validate.</param>
+        /// <returns>True if this factory matches the given transform, otherwise false.</returns>
+        public bool Validate(TransformValidationContext context, IReadOnlyDictionary<string, string> transformValues);
```
Kahbazi, line 19: Is this intentional?
```suggestion
        bool Validate(TransformValidationContext context, IReadOnlyDictionary<string, string> transformValues);
```

src/ReverseProxy/Abstractions/Config/ITransformFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// Checks if the given transform values match a known transform, and if so, generates a transform and
+        /// adds it to the context. This can throw if the transform values are invalid.
+        /// </summary>
+        /// <param name="context">The context to add any generated transforms to.</param>
+        /// <param name="transformValues">The transform values to use as input.</param>
+        /// <returns>True if this factory matches the given transform, otherwise false.</returns>
+        public bool Build(TransformBuilderContext context, IReadOnlyDictionary<string, string> transformValues);
```
Kahbazi, line 28:
```suggestion
        bool Build(TransformBuilderContext context, IReadOnlyDictionary<string, string> transformValues);
```

src/ReverseProxy/Service/Config/TransformBuilder.cs
```diff
@@ -549,23 +488,69 @@ internal StructuredTransformer BuildInternal(IReadOnlyList<IReadOnlyDictionary<s
...
+
+            // Let the app add any more transforms it wants.
+            foreach (var transformFitler in _filters)
+            {
+                transformFitler.Apply(transformBuilderContext);
+            }
+
             // Suppress the host by default
-            if (!useOriginalHost ?? true)
+            if (transformBuilderContext.UseOriginalHost ?? true)
```
Kahbazi, line 533:
```suggestion
            if (!transformBuilderContext.UseOriginalHost ?? true)
```

src/ReverseProxy/Abstractions/Config/ITransformFactory.cs
Tratcher: Hmm, that used to cause a compiler error... Maybe they relaxed that in C#9 or one of the new compilers?

Removed.

Tratcher:
Marking this as ready for review as there should only be docs work left.

alnikola (approved):
LGTM

Tratcher:
Updated with docs. I've noticed that the new `record` types like ProxyRoute aren't handled well by the doc tools so I had to simplify some links to fix them. In most cases we had to link to the class instead of a specific member.

docs/docfx/articles/transforms.md
```diff
@@ -31,8 +31,14 @@ X-Forwarded-Proto: http
 X-Forwarded-Host: IncomingHost:5000
 ```
 
-## Configuration
-Transforms are defined on [ProxyRoute.Transforms](xref:Microsoft.ReverseProxy.Abstractions.ProxyRoute.Transforms) and can be bound from the `Routes` sections of the config file. As with other route properties, these can be modified and reloaded without restarting the proxy. A transform is configured using one or more key-value string pairs.
+Transforms fall into a few categories: request, response, and response trailers. Request and response body transforms are not supported by YARP but you can write middleware to do this. Request trailers are not supported because they are not supported by the underlying HttpClient.
+
+
+Transforms can be added to routes either through configuration or pragmatically.
```
Kahbazi, line 37: Is this a typo?
```suggestion
Transforms can be added to routes either through configuration or programmatically.
```

Tratcher: Yes :-), that's what I get for using spell check.

docs/docfx/articles/transforms.md
```diff
@@ -96,9 +102,42 @@ Here is an example of common transforms:
...
+## From Code
+
+Transforms can be added to routes pragmatically by calling the [AddTransforms](xref:Microsoft.Extensions.DependencyInjection.ReverseProxyServiceCollectionExtensions) method.
+
+`AddTransforms` can be called from `Startup.ConfigureServices` to provide a callback for configuring transforms. This callback is invoked each time a route is built or rebuilt and allows the developer to inspect the [ProxyRoute](xref:Microsoft.ReverseProxy.Abstractions.ProxyRoute) information and conditionally add transforms for it.
+
+The `AddTransforms` callback provides a [TransformBuilderContext](xref:Microsoft.ReverseProxy.Abstractions.Config.TransformBuilderContext) where transforms can be added or configured. Most transforms provide `TransformBuilderContext` extension methods to make them easier to add. These are extensions documented below with the individual transform descriptions.
 
-### Request
+The `TransformBuilderContext` also includes an `IServiceProvider` for access to any needed services.
```
Kahbazi, line 117: I think there's a problem accessing scoped service in transforms. We are only passing the only instance of `IServiceProvider` and not `IServiceScope.ServiceProvider` which is created in every request.
I think right now the scoped services would behave like singleton and also is not the same with the instance provided by the `IServiceScope.ServiceProvider` in the same request.

Tratcher: Building transforms is an application level event and only has access to application services. Transforms that need access to request scoped services would resolve them from the HttpContext each request.

docs/docfx/articles/transforms.md
```diff
@@ -96,9 +102,42 @@ Here is an example of common transforms:
...
 All configuration entries are treated as case-insensitive, though the destination server may treat the resulting values as case sensitive or insensitive such as the path.
 
-Transforms fall into a few categories: request, response, and response trailers. Request and response body transforms are not supported by YARP but you can write middleware to do this. Request trailers are not supported because they are not supported by the underlying HttpClient.
+The details for these transforms are covered later in this document.
+
+Developers that want to integrate their custom transforms with the `Transforms` section of configuration can do so using [ITransformFactory](#itransformfactory) described below.
+
+## From Code
+
+Transforms can be added to routes pragmatically by calling the [AddTransforms](xref:Microsoft.Extensions.DependencyInjection.ReverseProxyServiceCollectionExtensions) method.
```
Kahbazi, line 111:
```suggestion
Transforms can be added to routes programmatically by calling the [AddTransforms](xref:Microsoft.Extensions.DependencyInjection.ReverseProxyServiceCollectionExtensions) method.
```

heatherkovich69 (approved):

PR #715, opened 2021-02-05
Log ProxyError with LogLevel.Error

grounzero:
Prevent suppressing error messages when minimum log level is set above `LogLevel.Information`

### DIFF
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -643,7 +643,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
                 "Proxying to {targetUrl}");
 
             private static readonly Action<ILogger, ProxyError, string, Exception> _proxyError = LoggerMessage.Define<ProxyError, string>(
-                LogLevel.Information,
+                LogLevel.Error,
                 EventIds.ProxyError,
                 "{error}: {message}");
 


Tratcher:
While we do need to tune these log levels, we were intentional about not marking some of these conditions as error logs. We want the error logs to highlight errors that are likely caused by the proxy itself such as configuration issues.

ProxyErrors like the ones logged here are routine business for the proxy, often caused by the client or the destination, and we don't want to over-report them.

grounzero:
Ahh, I see what you mean.  To de-clutter I had my minimum loglevel set as warning though I found in this case it was eating up a bit too much.

Would it be better if the log level could be configured in `ProxyHttpClientOptions` with something like "ErrorLogLevel"?

```
        "HttpClient": {
          "MaxConnectionsPerServer": 50,
	  "ErrorLogLevel": "Information",
          "HttpRequest": {
            "Version": "2",
            "VersionPolicy": "RequestVersionOrLower",
            "Timeout": "00:00:30"
          }
        },
```

Tratcher:
That shouldn't be needed, you should already be able to configure this from the logging end by selectively enabling just that feature area. Something like we do in the templates here:
https://github.com/dotnet/aspnetcore/blob/1a2723ded40a4204cb7b7f262b0e1388d7607fde/src/ProjectTemplates/Web.ProjectTemplates/content/WebApi-CSharp/appsettings.json#L49-L50
If our feature areas don't have the right granularity then that's certainly worth addressing.

PR #718, opened 2021-02-08
Enable unit testing of IActiveHealthCheckPolicy without InternalsVisibleTo

### LINKED ISSUE #705: Idea: Make it possible to unit test a custom IActiveHealthCheckPolicy

I've implemented a custom `IActiveHealthCheckPolicy`.  I've started trying to write some unit tests for this policy, but it seems to be very difficult because `ClusterInfo` is sealed and only has an internal constructor.  I don't think just un-sealing or making the constructor public would solve my problem, because it looks like I'm going to keep running into the same problem - constructing a `ClusterInfo` requires an `IDestinationManager` - which is also internal, as is its only implementation.

I'd like to be able to test my code.  I'm happy to do the work to make a PR to make things more testable, but I would appreciate some guidance on the best way to achieve that goal.

hughesjon:
This will resolve #705.

Makes just enough public to reproduce the unit tests for the built in policy - https://github.com/microsoft/reverse-proxy/blob/66f39b675e05c9b2490cf49f817c33060c9b02e8/test/ReverseProxy.Tests/Service/HealthChecks/ConsecutiveFailuresHealthPolicyTests.cs#L17

### DIFF
--- a/src/ReverseProxy/Service/Management/DestinationManager.cs
+++ b/src/ReverseProxy/Service/Management/DestinationManager.cs
@@ -5,7 +5,7 @@ using Microsoft.ReverseProxy.RuntimeModel;
 
 namespace Microsoft.ReverseProxy.Service.Management
 {
-    internal sealed class DestinationManager : ItemManagerBase<DestinationInfo>, IDestinationManager
+    public sealed class DestinationManager : ItemManagerBase<DestinationInfo>, IDestinationManager
     {
         protected override DestinationInfo InstantiateItem(string itemId)
         {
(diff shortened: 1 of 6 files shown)


Tratcher:
Thanks for jumping in. What you've done makes sense in the context of #705, but I'm not going to merge it immediately. We want to be very deliberate about the public surface area so I want to go over this with the team to see if this new surface area makes sense end-to-end.

One possible alternative is that we change what's passed into IActiveHealthCheckPolicy.

Tratcher:
We're going to start a series of API reviews to decide what we want the public surface area to be for 1.0. As part of that we'll be looking at issues like this and work how to make the right things public.

I'm going to close this for now pending that API review.

PR #721, opened 2021-02-08
Avoid computing the AbsoluteUri if logging is disabled

MihaZupan:
Gives us a bit over 1% RPS

This is the only place where we're using AbsoluteUri, so this change avoids the overhead of computing and storing the string / extra info inside of Uri.


Tratcher (approved):

Kahbazi:
Or can we pass the `Uri` instead of `string` to the log?

```csharp
private static readonly Action<ILogger, Uri, Exception> _proxying = LoggerMessage.Define<Uri>(
                LogLevel.Information,
                EventIds.Proxying,
                "Proxying to {targetUrl}");
```

MihaZupan:
Afaik that would end up serializing `Uri.ToString()` which is different from `AbsoluteUri`

Kahbazi:
Ah, I see!

Issue #733, opened 2021-02-10
Stash LoadBalancingPolicy on ClusterConfig


Kahbazi:
Currently the `ILoadBalancePolicy` is get from a dictionary on each request in `LoadBalancingMiddleware`.
https://github.com/microsoft/reverse-proxy/blob/5cfbc328dec97048d751e2e3b396ec26993c7bd1/src/ReverseProxy/Middleware/LoadBalancingMiddleware.cs#L54

Can we stash a `ILoadBalancePolicy` on `ClusterConfig` when building a cluster to improve the performance?

Tratcher:
That's an interesting idea. @MihaZupan @alnikola has the load balancing policy lookup shown up in the CPU profiles at all?

Kahbazi:
The same thing also could be done for `IPassiveHealthCheckPolicy`.
https://github.com/microsoft/reverse-proxy/blob/5cfbc328dec97048d751e2e3b396ec26993c7bd1/src/ReverseProxy/Middleware/PassiveHealthCheckMiddleware.cs#L40-L42

karelz:
Triage: We should first prove it shows up on profiles to justify the complexity.
Per @MihaZupan dictionary lookups are cheaper in .NET 5 as well and it didn't show up in @alnikola profiles so far ...

MihaZupan:
A few Dictionary improvements I had in mind (looks like they all made it in 5.0):
https://github.com/dotnet/coreclr/pull/27195
https://github.com/dotnet/coreclr/pull/27299
https://github.com/dotnet/runtime/pull/36252

karelz:
Triage: Closing - the work in #734 did not bring the expected gains. Not worth it.

PR #734, opened 2021-02-10
Stash ILoadBalancingPolicy in ClusterConfig

Kahbazi:
Move the selection of `ILoadBalancingPolicy` from `LoadBalancingMiddleware` to `ProxyConfigManager` so that we are only searching a dictionary once when the endpoints are creating instead of every requests.

Currently when the number of destinations are one or zero, we won't check the `ILoadBalancingPolicy`. In order to skip these checks on each requests, `FirstLoadBalancingPolicy` is being selected when number of destinations are one or zero, regardless of load balancing policy for that cluster.

Fixes https://github.com/microsoft/reverse-proxy/issues/733

### DIFF
--- a/src/ReverseProxy/Middleware/LoadBalancingMiddleware.cs
+++ b/src/ReverseProxy/Middleware/LoadBalancingMiddleware.cs
@@ -19,17 +19,14 @@ namespace Microsoft.ReverseProxy.Middleware
     internal class LoadBalancingMiddleware
     {
         private readonly ILogger _logger;
-        private readonly IDictionary<string, ILoadBalancingPolicy> _loadBalancingPolicies;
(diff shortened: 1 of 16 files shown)


Tratcher (changes requested):
Thanks for looking into this.

These lookups have not shown up as problematic in the analyses we've done. If you can show the improvement in a CPU profile or RPS improvement then we're more likely to consider it.

The design change isn't strictly beneficial either as it increases the API coupling across components. If the perf improvement were measurable we'd consider that an OK tradeoff.

Please close this if you're not able to get the perf data to go with it.

Kahbazi:
I used BenchmarkDotNet for profiling and the improvement was negligible. I'm closing this PR.

PR #739, opened 2021-02-14
Avoid ReasonPhrase overhead when using default values

MihaZupan:
If `ReasonPhrase` is set, AspNetCore will end up allocating the byte array for it, even if the value is the same as the cached one. This PR avoids setting the field if the value is the default (cached).

https://github.com/dotnet/aspnetcore/blob/e1bda79d87aa712a0b053824b4d31afa2c66cdce/src/Servers/Kestrel/Core/src/Internal/Http/ReasonPhrases.cs#L233

This can be improved in AspNetCore as well.

Saves a `byte[]` and a few `string` allocations / request.


Tratcher:
> This can be improved in AspNetCore as well.

Please file an issue in AspNetCore for that.

Tratcher (approved):

MihaZupan:
Opened https://github.com/dotnet/aspnetcore/pull/30189

Issue #675, opened 2021-01-19
Support EnableMultipleHttp2Connections 


Tratcher:
https://github.com/microsoft/reverse-proxy/blob/e4426b93621a4f0e4c55c626236a21404efd264e/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ProxyHttpClientOptions.cs#L22-L23

This is unblocked for .NET 5.0.

karelz:
Triage: We should have that as people will hit perf problems without it.
@Kahbazi any interest to pick this up?

Kahbazi:
@karelz Yes, I will take this.

PR #748, opened 2021-02-16
Add EnableMultipleHttp2Connections to ProxyHttpClientOptions

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/675

### DIFF
--- a/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ProxyHttpClientOptions.cs
+++ b/src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ProxyHttpClientOptions.cs
@@ -43,8 +43,14 @@ namespace Microsoft.ReverseProxy.Abstractions
         /// </summary>
         public ActivityContextHeaders? ActivityContextHeaders { get; init; }
 
-        // TODO: Add this property once we have migrated to SDK version that supports it.
-        //public bool? EnableMultipleHttp2Connections { get; init; }
+#if NET
+        /// <summary>
+        /// Gets or sets a value that indicates whether additional HTTP/2 connections can
+        //  be established to the same server when the maximum number of concurrent streams
+        //  is reached on all existing connections.
+        /// </summary>
+        public bool? EnableMultipleHttp2Connections { get; init; }
+#endif
 
         /// <inheritdoc />
         public bool Equals(ProxyHttpClientOptions other)
@@ -58,6 +64,9 @@ namespace Microsoft.ReverseProxy.Abstractions
                 && CertEquals(ClientCertificate, other.ClientCertificate)
                 && DangerousAcceptAnyServerCertificate == other.DangerousAcceptAnyServerCertificate
                 && MaxConnectionsPerServer == other.MaxConnectionsPerServer
+#if NET
+                && EnableMultipleHttp2Connections == other.EnableMultipleHttp2Connections
+#endif
                 && ActivityContextHeaders == other.ActivityContextHeaders;
         }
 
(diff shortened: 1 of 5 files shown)


Tratcher (approved):

Tratcher:
Quick as usual, thanks 😁

samsp-msft:
@MihaZupan - can you rerun the h2 perf tests now this is incorporated.

MihaZupan:
Reran the tests and this was the reason for the h1 vs h2 difference.

Note: `-single` means without setting the `EnableMultipleHttp2Connections`

| load                   | `win-h2-https` | `win-h2-h2-single` |          | `win-h2-h2` |         |
| ---------------------- | ------------ | ---------------- | -------- | --------- | ------- |
| Mean latency (us)      |        4.782 |           13.805 | +188,71% |     4.867 |  +1,79% |
| Requests               |      797.972 |          276.323 |  -65,37% |   783.509 |  -1,81% |
| Requests/sec           |       54.362 |           18.717 |  -65,57% |    52.339 |  -3,72% |
| Requests/sec (max)     |       63.942 |           27.504 |  -56,99% |    72.236 | +12,97% |
| Read throughput (MB/s) |        13,72 |             4,76 |  -65,29% |     13,47 |  -1,81% |

| load                   | `lin-h2-https` | `lin-h2-h2-single` |          | `lin-h2-h2` |         |
| ---------------------- | ------------ | ---------------- | -------- | --------- | ------- |
| Mean latency (us)      |        4.066 |           14.913 | +266,81% |     6.326 | +55,59% |
| Requests/sec           |       62.974 |           17.165 |  -72,74% |    40.487 | -35,71% |
| Requests/sec (max)     |       70.212 |           31.087 |  -55,72% |    54.874 | -21,85% |
| Read throughput (MB/s) |        15,36 |             4,21 |  -72,58% |      9,89 | -35,61% |


While enabling `EnableMultipleHttp2Connections` for the benchmarks makes complete sense, should we just make multiple connections the default? @Tratcher @samsp-msft

Tratcher:
EnableMultipleHttp2Connections wasn't enabled by default in HttpClient due to compatibility concerns, right? Is YARPs environment sufficiently different to avoid those?

samsp-msft:
I think that was because we didn't know whether the server(s) would support multiple connections from the same client - its quite likely that will be blocked when facing the internet, but not for server to server or proxy to server scenarios.

I think we can enable by default for YARP.

samsp-msft:
Based on the bolded below

| load                   | `lin-h2-https` | `lin-h2-h2-single` |          | `lin-h2-h2` |         |
| ---------------------- | ------------ | ---------------- | -------- | --------- | ------- |
| Mean latency (us)      |        4.066 |           14.913 | +266,81% |     6.326 | +55,59% |
| Requests               |      942.108 |          257.287 |  -72,69% |   606.141 | -35,66% |
| Requests/sec           |       **62.974** |           17.165 |  -72,74% |    **40.487** | -35,71% |
| Requests/sec (max)     |       70.212 |           31.087 |  -55,72% |    54.874 | -21,85% |
| Read throughput (MB/s) |        15,36 |             4,21 |  -72,58% |      9,89 | -35,61% |

It seems like there is still some discrepancy in the h2 implementation on Linux compared to https. This is not showing for the windows results. We should drill into this to find out why.

MihaZupan:
Afaik the concern originally was that it's technically against the spec to open more than 1 HTTP/2 connection (as the server should just allow more streams).

Tratcher:
I recall there were specific servers brought up that were known to reject multiple connections as a protocol violation, but I don't see that recorded in the original issue or PR.

At this point we can opt-in by default and see what feedback we get. It will be harder to opt-in by default later.

PR #753, opened 2021-02-17
Remove *.pfx from gitignore

MihaZupan:
When running crank benchmarks using a local YARP build, crank will respect the gitignores when uploading the files.

Having this `.pfx` in it means we're not copying the `testCert.pfx` to the perf machines, breaking some runs (https, h2).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -232,7 +232,6 @@ ClientBin/
 *.dbmdl
 *.dbproj.schemaview
 *.jfm
-*.pfx
 *.publishsettings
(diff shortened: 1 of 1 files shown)


ManickaP:
I guess this is related: https://github.com/dotnet/crank/issues/232

MihaZupan:
Yeah semi-related, but this is an issue on Yarp's side that we're using an ignore for something that was already tracked before.

Tratcher (approved):

PR #761, opened 2021-02-18
Creating a basic sample with annotations

samsp-msft:


samples/Basic YARP Sample/MyStartup.cs
```diff
@@ -0,0 +1,49 @@
...
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        {
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+            // Enable endpoint routing, required for the reverse proxy
+            app.UseRouting();
+            
+            // TODO: should this be removed: app.UseAuthorization();
```
samsp-msft, line 40: Do we need this in the sample?

samples/Basic YARP Sample/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace BasicYARPSample
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            // Create a Kestrel web server, and tell it to use the MyStartup class
+            // for the service configuration
+            var myHostBuilder = Host.CreateDefaultBuilder(args);
```
samsp-msft, line 15: This is a bit more verbose, but is less alien to somebody new than the .Build().Run() that is usually here.

samples/Basic YARP Sample/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+            myHostBuilder.ConfigureWebHostDefaults(webHostBuilder =>
```
samsp-msft, line 16: I am in 2 minds about the lambda verses a second method, both are a bit odd from a 101 standpoint.

samples/Basic YARP Sample/appsettings.json
```diff
@@ -0,0 +1,56 @@
+{
+  // Base URLs the server listens on, must be configured independently of the routes below
```
samsp-msft, line 2: Our config files support comments, so lets use them to explain what is going on.

samples/Basic YARP Sample/appsettings.json
```diff
@@ -0,0 +1,56 @@
...
+  "Urls": [
```
samsp-msft, line 3: Not strictly necessary, but makes it more obvious how to change what the server is listening on.

samples/Basic YARP Sample/appsettings.json
```diff
@@ -0,0 +1,56 @@
...
+    "http://localhost:5000",
+    "https://localhost:5001"
+  ],
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      // Uncomment to hide diagnostic messages
```
samsp-msft, line 10: Kind of nice to have this as on by default as it shows what is going on with the proxy, and why things may fail.

samsp-msft:
@davidfowl @Tratcher - We don't have an obvious basic starting point. The goal is to make it seem as less alien as possible for people who are not deeply immersed in the ASP.NET core ecosystem.

samples/Basic YARP Sample/MyStartup.cs
```diff
@@ -0,0 +1,49 @@
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace BasicYARPSample
+{
+    // Provides a configuration for the kestrel server with reverse proxy middleware
+    public class MyStartup
```
Tratcher, line 10: Startup is a very standard convention in AspNet. Changing it is likely to cause confusion when comparing to other samples.
```suggestion
    public class Startup
```

samples/Basic YARP Sample/MyStartup.cs
```diff
@@ -0,0 +1,49 @@
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace BasicYARPSample
+{
+    // Provides a configuration for the kestrel server with reverse proxy middleware
```
Tratcher, line 9:
```suggestion
    // Sets up the ASP.NET application with the reverse proxy enabled.
```

samples/Basic YARP Sample/MyStartup.cs
```diff
@@ -0,0 +1,49 @@
...
+        public MyStartup(IConfiguration configuration)
+        {
+            // Default configuration comes from AppSettings.json file in project/output
+            Configuration = configuration;
+        }
+
+        public IConfiguration Configuration { get; }
+
+        // This method gets called by the runtime. Use this method to add capabilities to
+        // the web server via services in the DI container.
```
Tratcher, line 21:
```suggestion
        // the web application via services in the DI container.
```

samples/Basic YARP Sample/MyStartup.cs
```diff
@@ -0,0 +1,49 @@
...
+        public void ConfigureServices(IServiceCollection services)
+        {
+            // Add the reverse proxy to capability to the server
+            var ProxyBuilder = services.AddReverseProxy();
+            // Initialize the reverse proxy from the "ReverseProxy" section of configuration
+            ProxyBuilder.LoadFromConfig(Configuration.GetSection("ReverseProxy"));
```
Tratcher, line 27: Style:
```suggestion
            var proxyBuilder = services.AddReverseProxy();
            // Initialize the reverse proxy from the "ReverseProxy" section of configuration
            proxyBuilder.LoadFromConfig(Configuration.GetSection("ReverseProxy"));
```

samples/Basic YARP Sample/MyStartup.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline for each request
```
Tratcher, line 30:
```suggestion
        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline that handles requests
```

samples/Basic YARP Sample/MyStartup.cs
Tratcher: If they enable auth on any of their routes they'll need this, but that's covered in the docs.

samples/Basic YARP Sample/appsettings.json
```diff
@@ -0,0 +1,56 @@
...
+      // "Microsoft": "Warning",
+      "Microsoft.Hosting.Lifetime": "Information"
+    }
+  },
+  "ReverseProxy": {
+    // Defines the routes the proxy will listen on
```
Tratcher, line 16: This is ambiguous with the urls given above.
```suggestion
    // Routes tell the proxy which requests to forward.
```

samples/Basic YARP Sample/appsettings.json
```diff
@@ -0,0 +1,56 @@
...
+        }
+      },
+      "cluster2": {
+        "Destinations": {
+          "first_destination": {
+            "Address": "https://localhost:10000"
+          },
+          "another_destination": {
+            "Address": "https://localhost:20000"
+          }
```
Tratcher, line 51: They won't have these servers, do we want examples that will fail?

samples/Basic YARP Sample/appsettings.json
```diff
@@ -0,0 +1,56 @@
...
+      {
+        // matches /something/* and routes to 2 localhost addresses
+        "RouteId": "route2",
+        "ClusterId": "cluster2",
+        "Match": {
+          "Path": "/something/{*any}"
+        }
+      }
+    ],
+    // Defines the clusters that routes can route to
```
Tratcher, line 35:
```suggestion
    // Clusters tell the proxy where to forward requests and how.
```

samples/Basic YARP Sample/appsettings.json
```diff
@@ -0,0 +1,56 @@
...
+    "Clusters": {
+      "example": {
+        "Destinations": {
+          "example.com": {
+            "Address": "http://www.example.com/"
+          }
+        }
+      },
+      "cluster2": {
```
Tratcher, line 44: Show load balancing config or session affinity as a sample cluster option?

samples/Basic YARP Sample/Program.cs
Tratcher: FYI There's tooling for EF and testing that depend on the CreateHostBuilder pattern.

Can you keep the CreateHostBuilder method? The parts you've expanded were already in Main.

samples/Basic YARP Sample/MyStartup.cs
```diff
@@ -0,0 +1,49 @@
...
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+            // Enable endpoint routing, required for the reverse proxy
+            app.UseRouting();
+            
+            // TODO: should this be removed: app.UseAuthorization();
+
+            // Configure EndPoints to use the reverse proxy
```
Tratcher, line 42:
```suggestion
            // Register the reverse proxy routes
```

samples/Basic YARP Sample/appsettings.json
```diff
@@ -0,0 +1,56 @@
+{
+  // Base URLs the server listens on, must be configured independently of the routes below
+  "Urls": [
+    "http://localhost:5000",
+    "https://localhost:5001"
+  ],
```
Tratcher, line 6: Urls is one string with semicolons, not an array.
```suggestion
  "Urls": "http://localhost:5000;https://localhost:5001",
```

samples/Basic YARP Sample/Properties/launchSettings.json
```diff
@@ -0,0 +1,14 @@
+﻿{
+  "$schema": "http://json.schemastore.org/launchsettings.json",
+  "profiles": {
+    "YARP Proxy Sample": {
+      "commandName": "Project",
+      "launchBrowser": true,
```
Tratcher, line 6: With the Urls set in the appsettings.json rather than the launchSettings.json, the tooling doesn't know which url to launch. I don't think this applies to the command line scenarios anyways.
```suggestion
      "launchBrowser": false,
```

Tratcher (commented):
The goal makes sense, especially with the extra comments you added. A few of the changes like the Startup class name seem more likely to add confusion than remove it.

samples/Basic YARP Sample/Basic YARP Sample.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
```
Tratcher, line 1: Are spaces in the folder and project name a good idea? They tend to cause havoc on the command line.

samples/Basic YARP Sample/Program.cs
davidfowl: Why would they use EF Core in the YARP process?

samsp-msft: +1 - also that's less likely to be the customer for this type of sample. Sure somebody could be doing an advanced proxy that is pulling routes from a db, but that customer is more likely to use one of the other samples.

samples/Basic YARP Sample/Basic YARP Sample.csproj
```diff
@@ -0,0 +1,11 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.ReverseProxy" Version="1.0.0-preview.*" />
```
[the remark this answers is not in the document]
samsp-msft: Do we have a todo in the release checklist to update this in the samples? If so then fine.
Also there is so little in this sample that is likely to conflict.

samples/Basic YARP Sample/MyStartup.cs
samsp-msft: IMHO both are a really bad name, as its not really startup, its more of a configuration. ServiceInit or something would be better.

samples/Basic YARP Sample/appsettings.json
samsp-msft: I tried a single string and it threw, I think logically its an array. The array works!.

samples/Basic YARP Sample/Properties/launchSettings.json
```diff
@@ -0,0 +1,14 @@
+﻿{
+  "$schema": "http://json.schemastore.org/launchsettings.json",
```
samsp-msft, line 2: If launchsettings can have a schema file - does one exist for asp.net core, or can/should we create one for YARP that has all of the config options in it so that those using VS (and possibly code) get intellisense?

samples/Basic YARP Sample/appsettings.json
samsp-msft: What would you suggest instead?

samples/Basic YARP Sample/Basic YARP Sample.csproj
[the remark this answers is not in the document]
Tratcher: Our in-repo samples don't use packages references, they use project references. We do have TODO in the release check-list to update the getting-started guide that uses a versioned package reference. This sample could be added to the list.
https://microsoft.github.io/reverse-proxy/articles/getting_started.html
https://github.com/microsoft/reverse-proxy/blob/main/docs/operations/Release.md#consume-nupkg

Long term I expect this would be better off in its own repo so people can acquire it directly.

samples/Basic YARP Sample/MyStartup.cs
Tratcher: Granted, but let's stick with the standards where can.

samples/Basic YARP Sample/appsettings.json
Tratcher: How about `https://example.com/foo` and ``https://example.com/bar`?

samples/Basic YARP Sample/appsettings.json
Tratcher: I just confirmed `"Urls": "https://localhost:5004;http://localhost:5002",` works. When I try the array format you have here it's ignored and it uses the default values (which also happen to be 5000 and 5001).

You're right an array would make more sense in the JSON scenario, but this key was designed mainly for use on command line and environment variables.

samples/Basic YARP Sample/Properties/launchSettings.json
Tratcher: That's a good question, I know we just added a schema for appsettings.json. If we need to create one then that should be owned by ASP.NET, not YARP. I'll file an issue in aspnetcore.

Tratcher: Oh, this is already a registered schema...
https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/launchsettings.json

samsp-msft: How can we get appsettings.json schema to include YARP?

Tratcher: That would take a contribution to [here](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/appsettings.json). That might be worth it when yarp becomes more stable, until then maybe we should provide our own?

Tratcher: That seems like something we'd invest in when working on the pre-built exe scenario.

samples/BasicYarpSample/appsettings.json
```diff
@@ -0,0 +1,54 @@
...
+          }
+        }
+      },
+      "cluster2": {
+        "Destinations": {
+          "first_destination": {
+            "Address": "https://contoso.com"
+          },
+          "another_destination": {
+            "Address": "https://bing.com:20000"
```
Tratcher, line 47:
```suggestion
            "Address": "https://bing.com"
```

samples/BasicYarpSample/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Extensions.Hosting;
+
+
+namespace BasicYARPSample
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            // Create a Kestrel web server, and tell it to use the MyStartup class
```
karelz, line 13:
```suggestion
            // Create a Kestrel web server, and tell it to use the Startup class
```

samples/BasicYarpSample/appsettings.json
```diff
@@ -0,0 +1,54 @@
...
+      {
+        // Matches anything and routes it to www.example.com
+        "RouteId": "AnExample",
+        "ClusterId": "example",
+        "Match": {
+          "Path": "{**catch-all}"
+        }
+      },
+      {
+        // matches /something/* and routes to 2 localhost addresses
```
karelz, line 24:
```suggestion
        // matches /something/* and routes to 2 external addresses
```

karelz (changes requested):

samples/Basic YARP Sample/Basic YARP Sample.csproj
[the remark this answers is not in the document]
samsp-msft: Added to the release instructions

Tratcher (approved):
Let's discuss the other samples in sync-up tomorrow.

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+This sample reads its configuration from the [appsettings.json](appsettings.json) file which defines 2 routes and clusters:
+
+* *AnExample* - this route has a path of *{\*\*catch-all}* which means that it will match any path, unless there is a another more specific route.
+   It routes to a cluster named *example* which has a single destination of http://example.com
+* *route2* - this route matches a path of */something/{\*any}* so will match any path that begins with "/something/". 
+As its a more specific route, it will match before the route above, even though its listed second. 
+This routes to a cluster named "cluster2" with 2 destinations. 
+It will load balance between those destinations using a Power of two choices algorithm. 
+That algorithm is best with more than 2 choices, but shows how to specify an algorithm in config.
```
karelz, line 19:
```suggestion
    As its a more specific route, it will match before the route above, even though it is listed second. 
    This routes to a cluster named *cluster2* with 2 destinations. 
    It will load balance between those 2 destinations using a Power of two choices algorithm. 
    That algorithm is best with more than 2 choices, but shows how to specify an algorithm in config.
```

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+* [Program.cs](program.cs) - Provides the main entrypoint for .NET which uses an WebHostBuilder to initialize the Sestrel server which listens for http requests. Typically, this file does not need to be modified for any proxy scenarios.
+* [Startup.cs](startup.cs) - Provides a class that is used to configure and control how http requests are handled by Kestrel. In this sample, it does the bare minimum of:
+  * Adding proxy functionality to the service
+  * Specifying that the proxy configuration will come from the config file (altrenatively it could be specified via code)
+  * Telling Kestrel to use its routing service, to register the routes from YARP into its routing table, and use YARP to handle those requests.
+* [appsettings.json](appsettings.json) - The configuration file for the .NET app, including sections for Kestrel, logging and the YARP proxy configuration. 
+* [Properties/launchsettings.json](Properties/launchsettings.json) - a configuration file use by Visual Studio to tell it how to start the app when debugging.
```
karelz, line 33:
```suggestion
- [BasicYarpSample.csproj](BasicYarpSample.csproj) - A C# project file (conceptually similar to a make file) that tells it to target the .NET 5 runtime, and to reference the proxy library from [nuget](https://www.nuget.org/packages/Microsoft.ReverseProxy/) (.NET's package manager).
- [Program.cs](Program.cs) - Provides the main entrypoint for .NET which uses an WebHostBuilder to initialize the Kestrel server which listens for http requests. Typically, this file does not need to be modified for any proxy scenarios.
- [Startup.cs](Startup.cs) - Provides a class that is used to configure and control how http requests are handled by Kestrel. In this sample, it does the bare minimum of:
  - Adding proxy functionality to the service.
  - Specifying that the proxy configuration will come from the config file (altrenatively it could be specified via code).
  - Telling Kestrel to use its routing service, to register the routes from YARP into its routing table, and use YARP to handle those requests.
- [appsettings.json](appsettings.json) - The configuration file for the .NET app, including sections for Kestrel, logging and the YARP proxy configuration. 
- [Properties/launchsettings.json](Properties/launchsettings.json) - A configuration file used by Visual Studio to tell it how to start the app when debugging.
```

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Visual Studio Code
+* Download and install Visual Studio Code (free) from https://code.visualstudio.com/ - versions are available for Windows, Linux and MacOS.
+* Download and install the .NET SDK from https://dotnet.microsoft.com/download if not already installed. Versions are available for Windows, Linux and MacOS.
+* Open the folder for the sample in VS Code (File->Open Folder).
+* Press F5 to debug, or Ctrl + F5 to run the sample without debugging.
+
+### Visual Studio
+
+* Download and install Visual Studio from https://visualstudio.microsoft.com/ - versions are available for Windows & MacOS including a free community edition.
```
karelz, line 53:
```suggestion
* Download and install Visual Studio from https://visualstudio.microsoft.com/ - versions are available for Windows and MacOS, including a free community edition.
```

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+* Open the project file.
+* Press F5 to debug, or Ctrl + F5 to run the sample without debugging.
+
+## Things to try
+ * Change the ports the proxy listens on using the URLs property in configuration or on the cmd line
+ * Change the routes and destinations used by the proxy
+ * A web server sample is available in the [sampleserver](../sampleserver) folder. It will output the request headers as part of the response body so they can be examined with a browser. 
+   * The urls the server listens to can be changed on the cmd line, so that multiple instances can be run. 
+     eg ```dotnet run ../SampleServer --Urls "http://localhost:10000;https://localhost:10010"```
```
karelz, line 62:
```suggestion
- Change the ports the proxy listens on using the URLs property in configuration or on the command line.
- Change the routes and destinations used by the proxy.
- A web server sample is available in the [SampleServer](../SampleServer) folder. It will output the request headers as part of the response body so they can be examined with a browser.
    - The URLs the server listens to can be changed on the command line, so that multiple instances can be run. 
     eg ```dotnet run ../SampleServer --Urls "http://localhost:10000;https://localhost:10010"```
```

karelz (approved):
Formatting, casing and spelling changes only

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
+# Basic YARP Sample
+
+This sample shows how to consume the YARP Library to produce a simple reverse proxy server. 
+The proxy server is implemented as middleware for the ASP.NET Core "Kestrel" web server. Kestrel provides the front end for the proxy, which listens for http requests, and then passes them to the proxy for paths that the proxy has registered for. The proxy handles the requests by:
```
Tratcher, line 4:
```suggestion
The proxy server is implemented a plugin component for ASP.NET Core applications. ASP.NET Core servers like Kestrel provide the front end for the proxy by listening for http requests and then passing them to the proxy for paths that the proxy has registered. The proxy handles the requests by:
```

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+- *AnExample* - this route has a path of *{\*\*catch-all}* which means that it will match any path, unless there is another more specific route.
+   It routes to a cluster named *example* which has a single destination of http://example.com
+- *route2* - this route matches a path of */something/{\*any}* which means that it will match any path that begins with "/something/". 
+    As its a more specific route, it will match before the route above, even though it is listed second. 
+    This routes to a cluster named *cluster2* with 2 destinations. 
+    It will load balance between those 2 destinations using a Power of two choices algorithm. 
+    That algorithm is best with more than 2 choices, but shows how to specify an algorithm in config.
+
+**Note:** The destination addresses used in the sample are using DNS names rather than IP addresses, this is so that the sample can be run and used without further changes. In a typical deployment, the destination servers should be specified with protocol, IP & ports, such as "https://123.4.5.6:7890"
```
Tratcher, line 21: There is no requirement to use IPs over host names, that's up to the deployment. I think most of our current customers use host names.

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+The proxy will listen to HTTP requests on port 5000, and HTTPS on port 5001. These are changable via the URLs property in config, and can be limited to just one protocol if required.
+
+## Files
+- [BasicYarpSample.csproj](BasicYarpSample.csproj) - A C# project file (conceptually similar to a make file) that tells it to target the .NET 5 runtime, and to reference the proxy library from [nuget](https://www.nuget.org/packages/Microsoft.ReverseProxy/) (.NET's package manager).
+- [Program.cs](Program.cs) - Provides the main entrypoint for .NET which uses an WebHostBuilder to initialize the Kestrel server which listens for http requests. Typically, this file does not need to be modified for any proxy scenarios.
+- [Startup.cs](Startup.cs) - Provides a class that is used to configure and control how http requests are handled by Kestrel. In this sample, it does the bare minimum of:
```
Tratcher, line 28:
```suggestion
- [Program.cs](Program.cs) - Provides the main entrypoint for .NET which uses an WebHostBuilder to initialize the server which listens for http requests. Typically, this file does not need to be modified for any proxy scenarios.
- [Startup.cs](Startup.cs) - Provides a class that is used to configure and control how http requests are handled by the server. In this sample, it does the bare minimum of:
```

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+  - Adding proxy functionality to the service.
```
Tratcher, line 29:
```suggestion
  - Adding proxy functionality to the services collection.
```

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+  - Specifying that the proxy configuration will come from the config file (altrenatively it could be specified via code).
+  - Telling Kestrel to use its routing service, to register the routes from YARP into its routing table, and use YARP to handle those requests.
```
Tratcher, line 31:
```suggestion
  - Telling ASP.NET to use its routing service, to register the routes from YARP into its routing table, and use YARP to handle those requests.
```

samples/BasicYarpSample/README.md
```diff
@@ -0,0 +1,62 @@
...
+- [Properties/launchsettings.json](Properties/launchsettings.json) - A configuration file used by Visual Studio to tell it how to start the app when debugging.
+
+## Getting started
+
+### Command line
+
+* Download and install the .NET SDK (free) from https://dotnet.microsoft.com/download if not already installed. Versions are available for Windows, Linux and MacOS.
+* Clone or extract a zip of the sample files.
+* Use ```dotnet run``` either within the sample folder or passing in the path to the .csproj file to start the server.
+* File change notification is used for the appsettings.config file so changes can be made on the fly.
```
Tratcher, line 42:
```suggestion
* File change notification is used for the appsettings.json file so changes can be made on the fly.
```

Tratcher (commented):
Clarifying terms.

samples/BasicYarpSample/README.md
samsp-msft: Really?

PR #771, opened 2021-02-20
Task => ValueTask

MihaZupan:
Removes some `Task` allocations (3 out of 83 objects or around 4% of allocated bytes).

They are only converted to tasks in `HandleUpgradedResponse`.

### DIFF
--- a/src/ReverseProxy/Service/Proxy/HttpProxy.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpProxy.cs
@@ -243,7 +243,7 @@ namespace Microsoft.ReverseProxy.Service.Proxy
             }
         }
 
-        private async Task<(HttpRequestMessage, StreamCopyHttpContent)> CreateRequestMessageAsync(HttpContext context, string destinationPrefix,
+        private async ValueTask<(HttpRequestMessage, StreamCopyHttpContent)> CreateRequestMessageAsync(HttpContext context, string destinationPrefix,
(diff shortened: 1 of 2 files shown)


Tratcher (approved):

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -482,8 +482,8 @@ private static Task CopyResponseStatusAndHeadersAsync(HttpResponseMessage source
 
             using var abortTokenSource = CancellationTokenSource.CreateLinkedTokenSource(longCancellation);
 
-            var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, _clock, abortTokenSource.Token);
-            var responseTask = StreamCopier.CopyAsync(isRequest: false, destinationStream, clientStream, _clock, abortTokenSource.Token);
+            var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, _clock, abortTokenSource.Token).AsTask();
```
davidfowl, line 485: How much did we actually save...

davidfowl:
I'd like to see before and after profiles for changes like these. You can use the VS memory profiler. This looks like to added some allocations back with the AsTask for the use of WhenAny

MihaZupan:
These `ValueTask`s are only converted to `Task`s in the `HandleUpgradedResponse` case - we save on the regular request path

davidfowl:
Great! I still think we should get in the habit of showing the before/after.

Issue #663, opened 2021-01-13
Improve TransformRequestAsync, TransformResponseAsync & TransformResponseTrailersAsync in non-async path


Kahbazi:
Currently all request and response transformers are sync. Since the transformers are in the hot path, can we improve `TransformRequestAsync`, `TransformResponseAsync` & `TransformResponseTrailersAsync` by removing the async-await in the main method and return a `Task` directly?

This is what I have in mind:

```csharp
public Task TransformResponse()
{
    if (ShouldCopyResponseHeaders.GetValueOrDefault(true))
    {
        var task = base.TransformResponseAsync(httpContext, proxyResponse);
        if (!task.IsCompletedSuccessfully)
        {
            return TransformResponseWithAwait(task);
        }
    }

    if (ResponseTransforms.Count == 0)
    {
        return Task.CompletedTask;
    }

    var transformContext = new ResponseTransformContext()
    {
        HttpContext = httpContext,
        ProxyResponse = proxyResponse,
        HeadersCopied = ShouldCopyResponseHeaders.GetValueOrDefault(true),
    };

    for (int i = 0; i < ResponseTransforms.Count; i++)
    {
        var task = ResponseTransforms[i].ApplyAsync(transformContext);

        if (!task.IsCompletedSuccessfully)
        {
            return TransformResponseWithAwait(task, i, ResponseTransforms);
        }
    }

    return Task.CompletedTask;
}
```

Tratcher:
It adds some complexity, but it might be worth it.  I'd wait until after we've worked out a few more parts of the object model like #60.

karelz:
Triage: Do experiment, profile before and after.

Kahbazi:
I will take this task. What's the best way to profile the changes?

MihaZupan:
If you point me to a buildable branch, I can run the change through our perf infra.

Kahbazi:
@MihaZupan Here's the changes https://github.com/microsoft/reverse-proxy/pull/775.

PR #775, opened 2021-02-22
Add a non async path for transforms

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/663

### DIFF
--- a/src/ReverseProxy/Service/RuntimeModel/Transforms/StructuredTransformer.cs
+++ b/src/ReverseProxy/Service/RuntimeModel/Transforms/StructuredTransformer.cs
@@ -63,12 +63,52 @@ namespace Microsoft.ReverseProxy.Service.RuntimeModel.Transforms
         /// </summary>
         internal IList<ResponseTrailersTransform> ResponseTrailerTransforms { get; }
 
-        public override async Task TransformRequestAsync(HttpContext httpContext, HttpRequestMessage proxyRequest, string destinationPrefix)
+        public override Task TransformRequestAsync(HttpContext httpContext, HttpRequestMessage proxyRequest, string destinationPrefix)
         {
             if (ShouldCopyRequestHeaders.GetValueOrDefault(true))
             {
-                await base.TransformRequestAsync(httpContext, proxyRequest, destinationPrefix);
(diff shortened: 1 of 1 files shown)


MihaZupan:
Did several runs, seems to be a measurable improvement of around 0.5 - 1 %.

MihaZupan:
Could we reduce some code duplication by extracting the shared portion after `ShouldCopyRequestHeaders` into a separate method?

Kahbazi:
> Could we reduce some code duplication by extracting the shared portion after `ShouldCopyRequestHeaders` into a separate method?

I'm not sure how to refactor it. The types are different, and each method is accessing a different property.

Tratcher:
@MihaZupan did it help with allocations any?

MihaZupan:
> did it help with allocations any?

No - since these methods never actually "go async", the state machines stay as structs and don't allocate.


Seeing as the performance difference is small (almost within the margin of error), I don't think it is worth duplicating all this logic.
Thank you for looking into this potential optimization @Kahbazi, but I don't think we'll be taking this change unless we see a bigger perf difference / find a way to avoid complicating the logic too much.

Did you by any chance run a microbenchmark for just the transformer to see a perf difference there?

Kahbazi:
> Did you by any chance run a microbenchmark for just the transformer to see a perf difference there?

No. I will run one and post the result.

Kahbazi:
Here's a benchmark of `TransformRequestAsync` with the default transformers which are removing the incoming host header and X-Forwarded headers.

|  Method |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
| Current | 3.210 us | 0.0450 us | 0.0399 us | 0.5684 |     - |     - |   1.74 KB |
|     New | 2.894 us | 0.0267 us | 0.0208 us | 0.5531 |     - |     - |    1.7 KB |

samsp-msft:
Triage: Thank you for the effort, but based on the numbers we don't think this change is worth taking for the amount of complexity it introduces.

PR #780, opened 2021-02-23
Avoid looking up default ReasonPhrases on HTTP/2+

MihaZupan:


src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -439,10 +439,13 @@ private static Task CopyResponseStatusAndHeadersAsync(HttpResponseMessage source
         {
             context.Response.StatusCode = (int)source.StatusCode;
 
-            // Don't explicitly set the field if the default reason phrase is used
-            if (source.ReasonPhrase != ReasonPhrases.GetReasonPhrase((int)source.StatusCode))
+            if (!ProtocolHelper.IsHttp2OrGreater(context.Request.Protocol))
```
Tratcher, line 442: Do you want to check the HttpResponseMessage.Version as well?

MihaZupan: I didn't add that as response transforms for an HTTP/2 request to the backend could set the reason phrase on `HttpResponseMessage`, which could then be returned in an HTTP/1.1 response

Tratcher (approved):

Issue #782, opened 2021-02-23
Active and Passive health checks should have default policies


Tratcher:
When enabling active or passive health checks today the Policy field must be specified. These fields should be optional and a default used. Compare to how we provide defaults for load balancing and session affinity.
```JSON
        "HealthCheck": {
          "Active": {
            "Enabled": "true",
            "Interval": "00:00:10",
            "Timeout": "00:00:10",
            "Policy": "ConsecutiveFailures",
            "Path": "/api/health"
          },
          "Passive": {
            "Enabled": "true",
            "Policy": "TransportFailureRate",
            "ReactivationPeriod": "00:05:00"
          }
        },
```

Identified when trying to simplify the config samples.

PR #783, opened 2021-02-23
Set default policy for passive and active health check

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/782

### DIFF
--- a/src/ReverseProxy/Service/Config/ConfigValidator.cs
+++ b/src/ReverseProxy/Service/Config/ConfigValidator.cs
@@ -370,11 +370,7 @@ namespace Microsoft.ReverseProxy.Service
 
             var activeOptions = cluster.HealthCheck.Active;
             var policy = activeOptions.Policy;
-            if (string.IsNullOrEmpty(policy))
-            {
-                errors.Add(new ArgumentException($"Active health policy name is not set on the cluster '{cluster.Id}'"));
-            }
-            else if (!_activeHealthCheckPolicies.ContainsKey(policy))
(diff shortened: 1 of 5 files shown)


src/ReverseProxy/Service/Config/ConfigValidator.cs
```diff
@@ -370,11 +370,7 @@ private void ValidateActiveHealthCheck(IList<Exception> errors, Cluster cluster)
 
             var activeOptions = cluster.HealthCheck.Active;
             var policy = activeOptions.Policy;
-            if (string.IsNullOrEmpty(policy))
-            {
-                errors.Add(new ArgumentException($"Active health policy name is not set on the cluster '{cluster.Id}'"));
-            }
-            else if (!_activeHealthCheckPolicies.ContainsKey(policy))
+            if (!string.IsNullOrEmpty(policy) && !_activeHealthCheckPolicies.ContainsKey(policy))
```
Tratcher, line 373: You still want to check that the default policy is present since that's what we'll use later. The same for Passive. Compare to load balancing:
https://github.com/microsoft/reverse-proxy/blob/9d1a8d73d70d22046dacfca57811939ad4ec8a39/src/ReverseProxy/Service/Config/ConfigValidator.cs#L297-L306
This is mainly to catch if we break something later.

Tratcher (commented):
I'll wait for a review from @alnikola since I don't think he's seen the original issue yet.

alnikola (approved):
LGTM

Tratcher:
Thanks

PR #784, opened 2021-02-24
Reorganize samples

Tratcher:
The samples should be:
- Simple, focused on specific scenarios
- Runnable. I've changed most to using a package dependency so they don't have to build the whole repo.
  - nuget couldn't resolve our new SF and Telemetry packages from the dotnet feeds for some reason. Need to follow-up.
- Should proxy to live endpoints (e.g. example.com).
- Have lots of comments and a readme - TODO: @samsp-msft can you help with that in a follow-up PR?
- In the future these may move to their own repo(s) to make them easier to consume.

Non-sample dev projects and benchmarks have been moved to the testassets folder.

No product changes.


samsp-msft (approved):
Scanned and looks good, any niggles can be fixed as subsequent updates.

samsp-msft:
I have commented out the addition of samples to the build so that it no longer fails.

Issue #788, opened 2021-02-25
DestinationInfo points to a DestinationConfig which does not have all the data of the Abstractions.Destination Record


samsp-msft:
### Describe the bug
From RuntimeModel.DestinationInfo I want to access the metadata collection specified in config. The Config property is pointing to a DestinationConfig which is intended as snapshot of the config object, but it doesn't have all the properties of the Abstractions.Destination object used by config.

Now that config objects are records, the immutable clone is unnecessary, and the Config should just point directly at the Abstractions.Destination object. 

This same pattern can be applied to ClusterInfo/ClusterConfig and RouteInfo/RouteConfig.

This will get rid of 3 extra types, and cloning at runtime.

Tratcher:
The RouteConfig and ClusterConfig have dynamically constructed state like transforms and the HttpClient. DestinationConfig is the only one without extra state.

samsp-msft:
Why is ClusterConfig.Options called Options?

Tratcher:
> Why is ClusterConfig.Options called Options?

@samsp-msft Calling it `Cluster` because it used a `Cluster` type seemed redundant.

samsp-msft:
State, Snapshot, Config would all be better names :-)

PR #793, opened 2021-02-26
Store instance of Destination in DestinationConfig

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/788

### DIFF
--- a/src/ReverseProxy/Service/RuntimeModel/DestinationConfig.cs
+++ b/src/ReverseProxy/Service/RuntimeModel/DestinationConfig.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
+using Microsoft.ReverseProxy.Abstractions;
 
 namespace Microsoft.ReverseProxy.RuntimeModel
 {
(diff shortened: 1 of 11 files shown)


Tratcher:
This is consistent with ClusterConfig. The open question is if DestinationConfig serves any purpose or if it can be eliminated? It does not track any additional state. I'll bring this up in our Tuesday meeting and get back to you.

Kahbazi:
> This is consistent with ClusterConfig. The open question is if DestinationConfig serves any purpose or if it can be eliminated? It does not track any additional state. I'll bring this up in our Tuesday meeting and get back to you.

Maybe it could be used for custom load balancing policies that need state per destination.

Tratcher:
> > This is consistent with ClusterConfig. The open question is if DestinationConfig serves any purpose or if it can be eliminated? It does not track any additional state. I'll bring this up in our Tuesday meeting and get back to you.
> 
> Maybe it could be used for custom load balancing policies that need state per destination.

When we looked at doing something similar for passive health checks we found that using ConditionalWeakTable to store the data locally worked better.
https://github.com/microsoft/reverse-proxy/blob/1fb29bbe27ff316f06996bc08eabc17cb0e79282/src/ReverseProxy/Service/HealthChecks/TransportFailureRateHealthPolicy.cs#L34

src/ReverseProxy/Service/RuntimeModel/DestinationConfig.cs
```diff
@@ -37,5 +42,10 @@ public DestinationConfig(string address, string health)
         /// Endpoint accepting active health check probes.
         /// </summary>
         public string Health { get; }
```
Tratcher, line 44: Why keep these properties if they're duplicates from the Destination?

Kahbazi: I will remove them.

Tratcher (approved):

Tratcher:
We discussed this briefly but didn't come to any conclusions. We'll revisit it when we do a fine grained API review.

PR #795, opened 2021-02-26
Updating benchmarks certificate

sebastienros:


sebastienros:
This one is not expired. Will also match the certificate for other benchmarks.

sebastienros:
However I am not even sure it's used anywhere. Where is it configured in yarp?

Tratcher:
Here?
https://github.com/microsoft/reverse-proxy/blob/3bcabe79afc7687c05db35be077dc2e10283b959/testassets/BenchmarkApp/Program.cs#L23

sebastienros:
@Tratcher thanks, I was looking in the startup and in the config.

Tratcher (approved):

PR #796, opened 2021-02-26
Code sample cleanup

samsp-msft:


samples/ReverseProxy.Code.Sample/InMemoryConfigProvider.cs
```diff
@@ -18,30 +23,42 @@ public static IReverseProxyBuilder LoadFromMemory(this IReverseProxyBuilder buil
...
+        /// <summary>
+        /// Implementation of the IProxyConfigProvider.GetConfig method to supply the current snapshot of configuration
+        /// </summary>
+        /// <returns>An immutable snapshot of the current configuration state</returns>
         public IProxyConfig GetConfig() => _config;
 
+        /// <summary>
+        /// Swaps the config state with a new snapshot of the configuration, then signals the change
+        /// </summary>
         public void Update(IReadOnlyList<ProxyRoute> routes, IReadOnlyList<Cluster> clusters)
```
samsp-msft, line 49: The assumption is that the routes and cluster lists here are immutable, which is not necessarily the case. Should the collections be cloned to ensure that they actually are?

samples/ReverseProxy.Code.Sample/Startup.cs
```diff
@@ -34,54 +70,54 @@ public void ConfigureServices(IServiceCollection services)
...
-                        var someCriteria = false; // MeetsCriteria(context);
-                        if (someCriteria)
-                        {
-                            // Here we check available destinations for the current cluster and pick one using custom criteria.
-                            var availableDestinationsFeature = context.Features.Get<IReverseProxyFeature>();
-                            var destination = availableDestinationsFeature.AvailableDestinations[0]; // PickDestination(availableDestinationsFeature.Destinations);
-                            // Load balancing will no-op if we've already reduced the list of available destinations to 1.
-                            availableDestinationsFeature.AvailableDestinations = destination;
-                        }
+                //Todo: Replace with a lookup of metadata - but not currently exposed correctly here
```
samsp-msft, line 114: Todo: Replace when #788 and #793 are resolved.

samples/ReverseProxy.Code.Sample/InMemoryConfigProvider.cs
Tratcher: I was using the IReadOnlyList contract to communicate the immutability. You're right that someone could mess with the underlying collection, but you can only protect people from themselves so much. This is a class the developer is meant to adapt to their purposes so I would leave that design choice up to them.

samsp-msft: Is there an extension method or something that does this as a simple CloneAsReadOnly() or similar?

Tratcher: I don't know of one that guarantees a deep copy. Most conversion methods short circuit if a type cast is sufficient.

samples/ReverseProxy.Code.Sample/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+  This sample shows the routes and destinations being created in code, and then passed to an in-memory provider. The role of the in-memory provider is to give change notifications to YARP for when the config has been changed and needs to be updated. YARP uses a snapshot model for its configuration, so that changes are applied as an atomic action, that will apply to subsequent requests after the change is applied. Existing requests that are already being processed will be completed using the configuration snapshot from the time that they were recieved.
+
+  The ```IProxyConfig``` interface implemented in [InMemoryConfigProvider.cs](InMemoryConfigProvider.cs) includes a change token which is used to signal when a batch of changes to the configuration is complete, and the proxy should take a snapshot and update its internal configuration. Part of the snapshot procesisng is to create an optimized route table in ASP.NET, which can be a CPU intensive operation, for that reason we don't recommend signalling for updates more than once per second.
```
Tratcher, line 11:
```suggestion
  The ```IProxyConfig``` interface implemented in [InMemoryConfigProvider.cs](InMemoryConfigProvider.cs) includes a change token which is used to signal when a batch of changes to the configuration is complete, and the proxy should take a snapshot and update its internal configuration. Part of the snapshot processing is to create an optimized route table in ASP.NET, which can be a CPU intensive operation, for that reason we don't recommend signaling for updates more than once per 15 seconds. 
```

samples/ReverseProxy.Code.Sample/README.md
```diff
@@ -0,0 +1,52 @@
...
+- ## Custom pipeline step
+
+  YARP uses a pipeline model for the stages involved in processing each request:
+
+  - Mapping the request path to a route and cluster of destinations
+  - Pre-assigning servers based on existing session affinity headers in the request 
+  - Filtering the destination list for servers that are not healthy
+  - Load balancing between the remaining servers based on load etc
+  - Storing session affinity if applicable
+  - Transforming headers if required
```
Tratcher, line 22:
```suggestion
  - Transforming the request and response if required
```

samples/ReverseProxy.Code.Sample/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Key Files
+
+The following files are key to implementing the features described above:
+
+- ### [Startup.cs](Startup.cs)
+  Provides the initialization routines for ASP.NET and the reverse proxy.
+
+  ```ConfigureServices``` is called once and sets up the proxy passing in the InMemoryConfig provider instance. The sample routes and clusters definitions are created as part of this initialization. The config provider instance is used for the lifetime of the proxy.
+
+  ```Configure``` is called for each request to setup the request pipeline. As an additional step is added, the proxy pipeline is configured here.
```
Tratcher, line 38:
```suggestion
  ```Configure``` is called once at startup to setup the request pipeline. As an additional step is added, the proxy pipeline is configured here. 
```

Tratcher (approved):

PR #804, opened 2021-03-01
Direct sample cleanup

samsp-msft:
Added Readme


samples/ReverseProxy.Direct.Sample/Startup.cs
```diff
@@ -39,16 +41,23 @@ public void Configure(IApplicationBuilder app, IHttpProxy httpProxy)
                 UseCookies = false
             });
 
+            // Setup our own Header transform class
```
Tratcher, line 44:
```suggestion
            // Setup our own request transform class
```

samples/ReverseProxy.Direct.Sample/Startup.cs
```diff
@@ -58,8 +67,23 @@ public void Configure(IApplicationBuilder app, IHttpProxy httpProxy)
             });
         }
 
+        /// <summary>
+        /// Custom Header transformation
```
Tratcher, line 71:
```suggestion
        /// Custom request transformation
```

samples/ReverseProxy.Direct.Sample/Startup.cs
```diff
@@ -11,10 +11,11 @@
 using Microsoft.ReverseProxy.Service.Proxy;
 using Microsoft.ReverseProxy.Service.RuntimeModel.Transforms;
 
-namespace Microsoft.ReverseProxy.Sample
+namespace YARP.Sample
 {
     /// <summary>
-    /// ASP .NET Core pipeline initialization.
+    /// ASP.NET Core pipeline initialization showing how to use IHttpProxy to directly handle proxying requests
+    /// skipping the intermediary stages of YARP.
```
Tratcher, line 18:
```suggestion
    /// ASP.NET Core pipeline initialization showing how to use IHttpProxy to directly handle proxying requests.
    /// With this approach you are responsible for destination discovery, load balancing, and related concerns.
```

samples/ReverseProxy.Direct.Sample/Startup.cs
```diff
@@ -39,16 +41,23 @@ public void Configure(IApplicationBuilder app, IHttpProxy httpProxy)
...
             var transformer = new CustomTransformer(); // or HttpTransformer.Default;
             var requestOptions = new RequestProxyOptions { Timeout = TimeSpan.FromSeconds(100) };
 
             app.UseRouting();
             app.UseEndpoints(endpoints =>
             {
+                // Normally for YARP we would call endpoints.MapReverseProxy() which would use YARP's pipeline to figure out a destination based
+                // on configuration and the stages in the pipeline.
+                // If you have alternatives to those steps, you can skip them and just tell the proxy which destination endpoint to use. It uses the 
+                // HttpClient to forward the request and response, calling into the supplied header transformer.
```
Tratcher, line 54: It's hard to say which usage model of YARP is more 'Normal', I think we currently have more customers using IHttpProxy than anything else. This comparison also assumes you're familiar with the other model, where we don't make the opposite comparison in the other samples.
```suggestion
                // When using IHttpProxy for direct proxying you are responsible for routing, destination discovery, load balancing, affinity, etc..
                // For an alternate example that includes those features see BasicYarpSample.
```

samples/ReverseProxy.Direct.Sample/README.md
```diff
@@ -0,0 +1,48 @@
+# YARP Direct Proxy Example
+
+The YARP Proxy uses a pipeline architecture, with multiple stages. These include:
+- Integrating with ASP.NET routing
+- Mapping from a route to a cluster of destinations
+- Filtering the destinations based on health status and affinity
+
+Some customers who have previously built a custom proxy solution have implementations of these concepts, but are coming to YARP for its support of HTTP/2, gRPC and modern protocols that are more complex to implement than HTTP/1.1. 
+They already have implementations of routing etc that work with their other infrastructure, that they don't want to have to replace or rewrite. 
+The goal with YARP is to provide a highly customizable proxy - so its very componentized. To make it easier to integrate YARP into these scenarios, the component that proxies requests is exposed via the HttpProxy class, which can be called directly, and has few dependencies on the rest of YARP's infrastructure.
```
Tratcher, line 10: This is being described in contrast to the other model rather than talking about this one directly. I think we can consolidate.

```suggestion
This example is for applications that have their own means of routing, load balancing, affinity, etc. and only need to forward a specific request to a specific destination. IHttpProxy provides that functionality, including HTTP/2, gRPC, and WebSockets support, and includes a callback pattern for transforming requests and responses as they are proxied.
```

samples/ReverseProxy.Direct.Sample/README.md
```diff
@@ -0,0 +1,48 @@
...
++-------------------+           +-------------------+           +-------------------+
+|      Client       |  ──(a)──► |      Proxy        |  ──(b)──► |    Destination    |
+|                   | ◄──(d)──  |                   | ◄──(c)──  |                   |
++-------------------+           +-------------------+           +-------------------+
+```
+
+(a) and (b) show the *request* path, going from the client to the destination.
+(c) and (d) show the *response* path, going from the destination back to the client.
+
+Normal proxying comprises the following steps:
```
Tratcher, line 24: This is a pretty in-depth description of the inner workings of HttpProxy. Should this be added to https://github.com/microsoft/reverse-proxy/blob/main/docs/docfx/articles/direct-proxying.md?

I'd expect the sample readme to focus more on usage and link to the concept doc for more depth.

samples/ReverseProxy.Direct.Sample/README.md
```diff
@@ -0,0 +1,48 @@
...
+| 6 | Copy response status line | Client ◄-- Proxy ◄-- Destination |
+| 7 | Copy response headers | Client ◄-- Proxy ◄-- Destination |
+| 8.1 | Check for a 101 upgrade response, this takes care of WebSockets as well as any other upgradeable protocol. | |
+| 8.1.1 | Upgrade client channel | Client ◄--- Proxy ◄--- Destination |
+| 8.1.2 | Copy duplex streams and return | Client ◄--► Proxy ◄--► Destination |
+| 8.2 | Copy (normal) response body | Client ◄-- Proxy ◄-- Destination |
+| 9 | Copy response trailer headers and finish response | Client ◄-- Proxy ◄-- Destination |
+| 10 | Wait for completion of step 2: copying request body | Client --► Proxy --► Destination |
+
+To enable control over steps 4 and 7, the HttpProxy operation takes a HttpTransformer which controls which headers are copied from the request to the response and vice-versa. Your implementation can decide which headers are copied, modify them, or insert additional headers as required.
```
Tratcher, line 42:
```suggestion
To enable control over mapping request and response fields and headers between the client and destination, the HttpProxy.ProxyAsync method takes a HttpTransformer. Your implementation can modify the request url, method, protocol version, response status code, or decide which headers are copied, modify them, or insert additional headers as required.
```

samples/ReverseProxy.Direct.Sample/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+**Note:** When using the YARP pipeline, YARP has some [default header transforms](https://microsoft.github.io/reverse-proxy/articles/transforms.html), such as adding ```X-Forwarded-For```, when using the HttpProxy class directly these transforms are not performed. You have full control of the transforms performed, and need to account for this in your HttpTransformer implementation.
```
Tratcher, line 44:
```suggestion
**Note:** When using the HttpProxy class directly there are no transforms included by default, you have full control of the transforms in your HttpTransformer implementation. The alternate YARP pipeline model (see BasicYarpSample), has some [default header transforms](https://microsoft.github.io/reverse-proxy/articles/transforms.html), such as adding ```X-Forwarded-For``` and removing the original Host header.
```

docs/docfx/articles/direct-proxying.md
```diff
@@ -28,6 +28,8 @@ It does not include:
 
 ## Example
 
+See [ReverseProxy.Direct.Sample](https://github.com/microsoft/reverse-proxy/tree/main/samples/ReverseProxy.Direct.Sample) as a pre-built sample, or use the steps below.
```
Tratcher, line 31: This should link to the release/docs branch so the sample and doc stay in sync.
```suggestion
See [ReverseProxy.Direct.Sample](https://github.com/microsoft/reverse-proxy/tree/release/docs/samples/ReverseProxy.Direct.Sample) as a pre-built sample, or use the steps below.
```

samples/ReverseProxy.Direct.Sample/Startup.cs
samsp-msft: Our internal guys are probably not normal and the main usage.

samples/ReverseProxy.Direct.Sample/README.md
samsp-msft: I don't see a problem with the info going into both. I plagiarized some of the content from the source.

samples/ReverseProxy.Direct.Sample/README.md
samsp-msft: The vast majority of the docs, samples etc are going to be based on that model. I think we need to contrast that with the direct approach. I will rework your suggestion into the content.

samples/ReverseProxy.Direct.Sample/Startup.cs
Tratcher: Fair, but a lot of external customers are using IHttpProxy as well.

samples/ReverseProxy.Direct.Sample/README.md
```diff
@@ -0,0 +1,44 @@
+# YARP Direct Proxy Example
+
+Some customers who have an existing custom proxy for HTTP/1.1 are looking at YARP for a solution to handle more complex requests, such as HTTP/2, gRPC, WebSockets in future QUIC and HTTP/3. These pplications have their own means of routing, load balancing, affinity, etc. and only need to forward a specific request to a specific destination. To make it easier to integrate YARP into these scenarios, the component that proxies requests is exposed via IHttpProxy which can be called directly, and has few dependencies on the rest of YARP's infrastructure.
```
Tratcher, line 3:
```suggestion
Some customers who have an existing custom proxy for HTTP/1.1 are looking at YARP for a solution to handle more complex requests, such as HTTP/2, gRPC, WebSockets in future QUIC and HTTP/3. These applications have their own means of routing, load balancing, affinity, etc. and only need to forward a specific request to a specific destination. To make it easier to integrate YARP into these scenarios, the component that proxies requests is exposed via IHttpProxy which can be called directly, and has few dependencies on the rest of YARP's infrastructure. 
```

samples/ReverseProxy.Direct.Sample/Startup.cs
```diff
@@ -39,16 +41,21 @@ public void Configure(IApplicationBuilder app, IHttpProxy httpProxy)
...
             app.UseEndpoints(endpoints =>
             {
+                // When using IHttpProxy for direct proxying you are responsible for routing, destination discovery, load balancing, affinity, etc..
+                // For an alternate example that includes those features see BasicYarpSample.
                 endpoints.Map("/{**catch-all}", async httpContext =>
                 {
                     await httpProxy.ProxyAsync(httpContext, "https://example.com", httpClient, requestOptions, transformer);
                     var errorFeature = httpContext.Features.Get<IProxyErrorFeature>();
+                    
+                    //Check if the proxy operation was successful
```
Tratcher, line 58:
```suggestion
                    // Check if the proxy operation was successful
```

Tratcher (approved):

PR #807, opened 2021-03-03
State flag for initial health check probe being done.

### LINKED ISSUE #306: LifeCycle Management: state flag when YARP is truly ready to handle requests
This is a corollary to #23, but for startup rather than shutdown.

Have an event for when YARP is up and running, and is ready for requests. This needs to be when the config is loaded, and the backends have been pinged for the destinations in the initial config.

YARP will be pinged by health checks for higher up systems, and needs to be able to respond affirmatively if its really ready to respond to requests.

ManickaP:
Adds a state flag `InitialDestinationsProbed` to `IActiveHealthCheckMonitor` that can be regularly polled. It becomes, and stays for the rest of the app lifetime, `true` once all the destinations have been probed, regardless their availability/response.

Fixes #306

cc: @davidni @Tratcher


src/ReverseProxy/Service/HealthChecks/IActiveHealthCheckMonitor.cs
```diff
@@ -12,6 +12,16 @@ namespace Microsoft.ReverseProxy.Service.HealthChecks
     /// </summary>
     public interface IActiveHealthCheckMonitor
     {
+        /// <summary>
+        /// Whether the initial round of active health check passed, regardless the results.
```
Tratcher, line 16:
```suggestion
        /// Whether the initial round of active health checks have run, regardless of the results.
```

src/ReverseProxy/Service/HealthChecks/IActiveHealthCheckMonitor.cs
```diff
@@ -12,6 +12,16 @@ namespace Microsoft.ReverseProxy.Service.HealthChecks
...
+        /// </summary>
+        /// <returns>
+        /// <c>false</c> until the initial round of health check requests has been processed.
+        /// <c>true</c> when all the initially configured destinations has been queried, regardless their availability or returned status code.
```
Tratcher, line 20:
```suggestion
        /// <c>true</c> when all the initially configured destinations have been queried, regardless their availability or returned status code.
```

Tratcher (approved):
Looks good to me, but wait for David to see it before merging.

src/ReverseProxy/Service/HealthChecks/IActiveHealthCheckMonitor.cs
```diff
@@ -12,6 +12,16 @@ namespace Microsoft.ReverseProxy.Service.HealthChecks
...
     {
+        /// <summary>
+        /// Whether the initial round of active health checks have run, regardless of the results.
+        /// </summary>
+        /// <returns>
+        /// <c>false</c> until the initial round of health check requests has been processed.
+        /// <c>true</c> when all the initially configured destinations have been queried, regardless their availability or returned status code.
+        /// The property stays <c>true</c> for the rest of the proxy process lifetime.
+        /// </returns>
+        public bool InitialDestinationsProbed { get; }
```
Kahbazi, line 23:
```suggestion
        bool InitialDestinationsProbed { get; }
```

Kahbazi, line 23: nit:
```suggestion
        bool InitialDestinationsProbed { get; }
```

samsp-msft:
@davidfowl

src/ReverseProxy/Service/HealthChecks/IActiveHealthCheckMonitor.cs
```diff
@@ -12,6 +12,16 @@ namespace Microsoft.ReverseProxy.Service.HealthChecks
     /// </summary>
     public interface IActiveHealthCheckMonitor
     {
+        /// <summary>
+        /// Whether the initial round of active health checks have run, regardless of the results.
```
davidfowl, line 16:
```suggestion
        /// Gets a value that determines whether the initial round of active health checks have run, regardless of the results.
```

davidfowl:
How is this flag used in practice? Where is it polled?

test/ReverseProxy.Tests/Service/HealthChecks/ActiveHealthCheckMonitorTests.cs
```diff
@@ -408,11 +656,19 @@ private ClusterInfo GetClusterInfo(string id, string policy, bool activeCheckEna
...
+        private Mock<HttpMessageInvoker> GetHttpClient(Task<HttpResponseMessage> task = null, Action cancellation = null)
         {
             var httpClient = new Mock<HttpMessageInvoker>(() => new HttpMessageInvoker(new Mock<HttpMessageHandler>().Object));
             httpClient.Setup(c => c.SendAsync(It.IsAny<HttpRequestMessage>(), It.IsAny<CancellationToken>()))
-                .ReturnsAsync((HttpRequestMessage m, CancellationToken c) => new HttpResponseMessage(HttpStatusCode.OK) { Version = m.Version });
+                .Returns((HttpRequestMessage m, CancellationToken c) =>
+                {
+                    if (cancellation != null)
+                    {
+                        c.UnsafeRegister(_ => cancellation(), null);
```
davidni, line 667: nit: Is there a strong reason to not just use `Register` here?

davidni (approved):
LGTM

test/ReverseProxy.Tests/Service/HealthChecks/ActiveHealthCheckMonitorTests.cs
ManickaP: No, there isn't. I check the other occurrences, which are all `Register`, and changed this place as well.

ManickaP:
@davidfowl:
> How is this flag used in practice? Where is it polled?

This is a question rather for @davidni.

However, if you're missing sample usage or a functional test or better documentation of this feature, please say so directly and I'll add it.

Tratcher:
@ManickaP how about showing it here in this sample:
https://github.com/microsoft/reverse-proxy/blob/cee49090bc214c2ab3617837f32e47e350935925/testassets/ReverseProxy.Code/Controllers/HealthController.cs#L22

I think the idea is that a controller like this would inject the service into the constructor and return a 503 until the new property became true.

testassets/ReverseProxy.Code/Controllers/HealthController.cs
```diff
@@ -19,7 +30,7 @@ public class HealthController : ControllerBase
         public IActionResult CheckHealth()
         {
             // TODO: Implement health controller, use guid in route.
-            return Ok();
+            return _healthCheckMonitor.InitialDestinationsProbed ? Ok() : StatusCode(503);
```
Tratcher, line 33: 👍 That's what I expected.

:shipit:

Issue #587, opened 2020-12-01
Expose route metadata so that we can read it from a custom middleware


goenning:
### What should we add or change to make your life better?

Expose route metadata so that we can read it from a middleware to build custom logic based on it

### Why is this important to you?

I'm looking to perform body transformation which will vary per route, so the idea is to store a hint on the metadata and use that on a middleware. Example:

```json
{
    "RouteId": "route1",
    "ClusterId": "cluster1",
    "Match": {
        "Hosts": [
            "example.com"
        ]
    },
    "Metadata": {
        "RequestBodyTransform": "Tranformation1",
    }
}
```

Startup.cs
```csharp
endpoints.MapReverseProxy(proxyPipeline =>
{
    proxyPipeline.UseMiddleware<MyMiddleware>();
});
```

MyMiddleware.cs
```csharp
public async Task Invoke(HttpContext context)
{
    var route = context.GetProxyRoute();
    if (route.Metadata["RequestBodyTransform"] == "Tranformation1")
    {
        ...
    }

    await _next(context);
}
```

Ref: https://github.com/microsoft/reverse-proxy/discussions/576

samsp-msft:
Triage: needs some design.

PR #814, opened 2021-03-06
Design review feedback

Tratcher:
This addresses design feedback from a review session we did this week. The focus was on middleware extensibility scenarios and ensuring a good public API surface. Some highlights:
- We've removed some of the short circuits that would happen when entering the proxy pipeline, such as if there were no healthy destinations. The pipeline now executes so you can customize what happens.
  - We also removed short circuits from load balancing. Error responses are now primarily returned from the end of the pipeline.
- Naming of properties on IReverseProxyFeature.
- Route data is now included on IReverseProxyFeature. Fixes https://github.com/microsoft/reverse-proxy/issues/587

### DIFF
--- a/src/ReverseProxy/Middleware/DestinationInitializerMiddleware.cs
+++ b/src/ReverseProxy/Middleware/DestinationInitializerMiddleware.cs
@@ -29,39 +29,20 @@ namespace Microsoft.ReverseProxy.Middleware
             var routeConfig = context.GetRequiredRouteConfig();
 
             var cluster = routeConfig.Cluster;
+            // TODO: Validate on load https://github.com/microsoft/reverse-proxy/issues/797
             if (cluster == null)
             {
-                Log.NoClusterFound(_logger, routeConfig.Route.RouteId);
-                context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
-                return Task.CompletedTask;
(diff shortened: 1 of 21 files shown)


src/ReverseProxy/Middleware/DestinationInitializerMiddleware.cs
```diff
@@ -75,42 +56,10 @@ private static class Log
                 EventIds.NoClusterFound,
                 "Route '{routeId}' has no cluster information.");
 
-            private static readonly Action<ILogger, string, string, Exception> _clusterDataNotAvailable = LoggerMessage.Define<string, string>(
-                LogLevel.Information,
-                EventIds.ClusterDataNotAvailable,
```
Kahbazi, line 80: Does the event Ids also need to be removed?
https://github.com/microsoft/reverse-proxy/blob/9f568f1875eb4485cd9d6b58b85fccb9224b5658/src/ReverseProxy/EventIds.cs#L14

Tratcher: Removed.

MihaZupan (approved):
The change (including teapots) LGTM

alnikola (approved):
LGTM

Issue #25, opened 2020-03-24
Review host matching validation


Tratcher:
The RouteValidator uses a strict regex to check the host matching pattern.  I have my doubts that this level of validation is even needed. The source is developer input, not external input, and it's only used in the route matcher (HostAttribute). We should be careful about second guessing what the HostMatcherPolicy can handle. For example the regex doesn't allow IDNs (Unicode hosts). 

Preferably we'd use a validation helper provided directly by the framework so they'd be in sync.

mifopen:
Is there any workaround for IDN support? I need to route requests based on IDN host

Tratcher:
@mifopen IDN support should be possible, we just haven't tried it yet. Are you willing to do a little experimentation?

First we need to figure out which format the HostMatcherPolicy requires, unicode or punycode. I think it's unicode from looking at the code:
https://github.com/dotnet/aspnetcore/blob/27610b0901535ce4b71832d7f3adf708f0c689c7/src/Http/Routing/src/Matching/HostMatcherPolicy.cs#L333
https://github.com/dotnet/aspnetcore/blob/27610b0901535ce4b71832d7f3adf708f0c689c7/src/Http/Http/src/Internal/DefaultHttpRequest.cs#L122
https://github.com/dotnet/aspnetcore/blob/27610b0901535ce4b71832d7f3adf708f0c689c7/src/Http/Http.Abstractions/src/HostString.cs#L158-L162

Second, I think this regex validation needs to be removed to allow for unicode characters:
https://github.com/microsoft/reverse-proxy/blob/5291514a487f8ccb3866d94b541674c964367db8/src/ReverseProxy/Service/Config/ConfigValidator.cs#L139

Tratcher:
#817 removed the offending regex because it was preventing hosts with ports. We still need to add some tests for IDNs, host:port, etc..

karelz:
Triage: We need the IDN part which is still missing.

PR #817, opened 2021-03-08
remove hosts regex validator from configvalidator

faix:
Removing the hosts name regex validation as the hostname is not coming from user input.

#25

### DIFF
--- a/src/ReverseProxy/Service/Config/ConfigValidator.cs
+++ b/src/ReverseProxy/Service/Config/ConfigValidator.cs
@@ -24,30 +24,6 @@ namespace Microsoft.ReverseProxy.Service
 {
     internal class ConfigValidator : IConfigValidator
     {
-        // TODO: IDN support. How strictly do we need to validate this anyways? This is app config, not external input.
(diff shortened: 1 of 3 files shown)


Tratcher:
Looks like this test also needs to be removed:
https://github.com/microsoft/reverse-proxy/blob/2221f4094f55a7a533cc1df79cef45208267a76a/test/ReverseProxy.Tests/Service/Config/ConfigValidatorTests.cs#L147
https://dev.azure.com/dnceng/public/_build/results?buildId=1027358&view=ms.vss-test-web.build-test-results-tab&runId=31906434&resultId=100470&paneView=debug

and this one updated:
https://github.com/microsoft/reverse-proxy/blob/f126615bcd4101110ed4467cb560201ece480fe1/test/ReverseProxy.Tests/Service/Management/ProxyConfigManagerTests.cs#L280

faix:
yup, was just looking into the CI builds.

Tratcher (approved):

Tratcher:
Thanks.

PR #819, opened 2021-03-09
Unsafe header enumeration POC

MihaZupan:
Proof of concept change.

Actual enumeration change is f067ffd.

1774495 is used to avoid fetching header values that we were gonna throw away anyway.
As the header values are lazily initialized & parsed, these show up after bypassing validation.

The change yields improvements of:
- 4-7 % throughput (RPS)
- 25 % less allocated bytes (5730 => 4328)
- 34 % less allocated objects (76 => 50)

### DIFF
--- a/src/ReverseProxy/Service/Proxy/HttpTransformer.cs
+++ b/src/ReverseProxy/Service/Proxy/HttpTransformer.cs
@@ -1,10 +1,16 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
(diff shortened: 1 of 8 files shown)


MihaZupan:
For reference, allocations for 10 000 HTTPS-HTTPS requests on a .Net 6.0 nightly SDK.

Before
image


After
image

samsp-msft:
Triage: This has shown its purpose, but we unfortunately can't take this based on hackiness of the change.

Issue #811, opened 2021-03-04
Rebranding checklist


Tratcher:
We plan to re-brand this project to be fully independent. This includes updating the:
- [x] Project and package names Microsoft -> Yarp
   - Yarp.ReverseProxy
   - Yarp.ReverseProxy.ServiceFabric
- [x] Namespaces Microsoft -> Yarp
  -  Yarp.ReverseProxy
- [ ] ~~Some APIs like MapReverseProxy~~ (nevermind)
- [ ] Publish the new packages
- [x] Docs
- [x] Samples: Some samples reference published packages for stability.

Things that are not covered by this issue but will likely change later:
- License
- File copyright headers
- CLA process
- Binary signing
- Security issue reporting process (?)
- Github Org - yarpproxy is reserved
- Repo name

samsp-msft:
We are looking at submitting YARP to the CNCF, as a sandbox project. They use the apache license rather than MIT. So it would be changing to apache if accepted.

PR #832, opened 2021-03-12
Rebrand namespaces and docs 

Tratcher:
Contributes to #811

There should be no functional changes here, only branding in namespaces, docs, and the session affinity cookie name.

I standardized all the samples to use the `Yarp.Sample` namespace.


samsp-msft (approved):
Did a scan through. LGTM. As it builds I assume it would have caught any big problems.
What happens to names for ILogger, do notes in the sample need to include Yarp.ReverseProxy in config for logging?

Tratcher:
> What happens to names for ILogger, do notes in the sample need to include Yarp.ReverseProxy in config for logging?

You're right that this changes the logs so they now come from the Yarp namespace. I think that's OK for our samples, they end up using the default app logging config. For Development (appsettings.Develoment.json) they default to Debug, and for production (appsettings.json) they default to Information.

The appsettings in BasicYarpSample could use an update though. It doesn't have an appsettings.Development.json so it will always use Info logs. Some of the comments are out of date too.

PR #835, opened 2021-03-12
Update package dependent samples 

Tratcher:
Contributes to #811

I've updates these samples temporarily using packages from our daily builds on the dotnet5 feed. Normally we'd wait for a nuget.org release, but there are more changes here than usual that I wanted done in time for the release. We should plan to push the newly branded packages to nuget next week.


Tratcher:
I just noticed that I'd missed the session affinity header so I've included that here.

MihaZupan (approved):

PR #838, opened 2021-03-14
Make IsExternalInit public

### LINKED ISSUE #837: Issues targeting .NET Core 3.1
When creating direct proxy and setting timeout
var requestOptions = new RequestProxyOptions() { Timeout = TimeSpan.FromSeconds(100) };

The error is  CS8400 Feature 'init-only setters' is not available in C# 8.0. Please use language version 9.0 or greater.

This is not an option for MVC projects targeting LTS 3.1
Please consider compatible constructor.

Thank you

Tratcher:
C# 9 features like init only properties weren't officially supported on .NET Core 3.1, though it can be used with the 5.0 SDK and a copy of the IsExternalInit class (which we were told by Jared to make public). 

cc: @davidni we'll see if this being public means that you don't have to add your own copy.

### DIFF
--- a/src/ReverseProxy/Utilities/IsExternalInit.cs
+++ b/src/ReverseProxy/Utilities/IsExternalInit.cs
@@ -1,8 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#if !NET
 namespace System.Runtime.CompilerServices
 {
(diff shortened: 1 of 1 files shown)


Tratcher:
Triage: Do not add the constructors until we find someone blocked without them.

MihaZupan (approved):

PR #840, opened 2021-03-15
Let IProxyConfigProvider.GetConfig() return ValueTask<IProxyConfig>.

pengweiqhca:


Tratcher:
This code path was intentionally not async for two reasons:
1) The first config load intentionally blocks app startup to ensure that routes are available for the first request.
https://github.com/microsoft/reverse-proxy/blob/d1dc9ab0374ea141f4a4dca934e7ef6cd9858c50/src/ReverseProxy/Configuration/ReverseProxyIEndpointRouteBuilderExtensions.cs#L71
(Admittedly this code path could use some cleanup for consistency).
2) IProxyConfigProvider is supposed to do all subsequent config reloads in the background and only signal the IProxyConfig.ChangeToken when a new config is available.

Both of these scenarios expect GetConfig to complete synchronously or block.

I'm going to close this PR. In the future please open issues to discuss design changes before submitting a PR.

Issue #200, opened 2020-05-22
Proxy can act as a Kubernetes Ingress Controller


galvesribeiro:
Hello folks!

First of all, great work! Congratz for the first preview release! It is indeed something .Net Community was lagging behind and had for too long to deal with clumsy things like nginx.

So, I'm coming to a point where I'd like to make a POC with YARP for a controlled small set of microservices in Kubernetes and that comes to a need to make it an ingestion controller.

With that in mind, I would like to ask if the Kube integration is something you guys plan to have on this repo, on some other under MSFT umbrella, or if it is something that the community will have to implemente for their own needs.

If it is something MSFT will (or already is) develop, ok, all we have todo is wait. If that is something that community will have to build, is that something that would be acceptable on this repo as a PR? If we should go to this PR, I'de like to use this issue to discuss the approach we should take to contribute it.

Thanks! Keep it going!

cwoolum:
I was just thinking it would be pretty cool to have an operator for this. I don't think it would make sense in this repo though. Also, I don't see there even being a docker image based on this repo since this is only a library. 

The operator would probably want to provide guidance on building an image.. maybe a Dockerfile?

davidfowl:
I think it should be in this repository as another project.

galvesribeiro:
Yeah, I think that too. The other project can contain the Dockerfile and whatever is required for the controller to run.

samsp-msft:
Acting as a K8s ingestion controller is something we have been thinking about and is on the backlog, but we don't yet know where it would fit in the timeline. 

The first step is what are your requirements?

galvesribeiro:
@samsp-msft Well, the current proxy functionality that the framework has in the current preview, has everything we need for now.

What would be generally required for people using Kubernetes is an ingress controller that can be configured using kubernetes config objects, like YAML, and that can spin up/down pods that will handle the traffic and at same time, provide configuration to those pods in a transparent way so whenever you change the config of the ingress deployment, it would "replicate" the config to the running proxy pods.

I'll dig into the specifics until this weekend and will post back here some sort of proposal.

Will keep in touch.

davidfowl:
We have a prototype of Kestrel as an ingress written by @jkotalik so we have a pretty good idea of what's required to integrate YARP.

jkotalik:
Yeah, I was tempted to port that over soon as a starting piece that can be improved upon. Code is here: https://github.com/jkotalik/kestrel-ingress

[...]

PR #842, opened 2021-03-17
Adds IngressController to Yarp

jkotalik:
Fixes https://github.com/microsoft/reverse-proxy/issues/200

This is a basic implementation of a Kubernetes ingress and ingress controller.

Relevant subfolders are mostly in src/ReverseProxy.KuberenetesController and src/ReverseProxy.WebApp.

I left most of the OperatorFramework subfolder structure unchanged as we want to reorganize it.

Key follow ups:
- [ ] General cleanup (fixing docker files, yaml, etc. that I haven't quite gotten to)
- [ ] Documentation (getting started, configuration, similar to 
- [ ] Architecture documentation (https://microsoft.github.io/reverse-proxy/articles/service-fabric-int.html)
- [ ] Tutorials (for building your own image to be used in k8s, using a built in one)


jkotalik:
cc @karelz @samsp-msft @Tratcher @davidfowl

Tratcher:
What's the content in /samples/KuberenetesIngress for? Is that apps that you would publish to k8 behind the new ingress?

General structure suggestions:
/src/ReverseProxy.KubernetesController -> /src/ReverseProxy.Kubernetes.Controller
/src/ReverseProxy.KubernetesProtocol -> /src/ReverseProxy.Kubernetes.Protocol
/src/ReverseProxy.WebApp -> Split this up:
- The LoadFromMessages implementation moves to /src/ReverseProxy.Kubernetes.Protocol
- Program and Startup move to something like /samples/KuberenetesIngress/ReverseProxy.Kubernetes.Ingress (some day we'll ship a pre-built exe for this, but we'll start with a sample you clone.)

jkotalik:
> What's the content in /samples/KuberenetesIngress for? Is that apps that you would publish to k8 behind the new ingress?

Yes

src/ReverseProxy.Kubernetes.Ingress/Services/Receiver.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Yarp.ReverseProxy.Abstractions;
+using Yarp.ReverseProxy.Configuration;
+using Yarp.ReverseProxy.Service;
+using Yarp.ReverseProxy.Kubernetes.Protocol;
+
+namespace Yarp.ReverseProxy.Kubernetes.Ingress.Services
+{
+    public class Receiver : BackgroundHostedService
```
Tratcher, line 24: The Receiver looks like standard equipment, people wouldn't be customizing this right? Can it move over to the Protocol package and be included in LoadFromMessages too?

Startup is the main content I expect in the Ingress app, that's where people would do most customization.

src/ReverseProxy.Kubernetes.Ingress/Services/Receiver.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                await _limiter.WaitAsync(cancellationToken);
+
+                Logger.LogInformation("Connecting with {ControllerUrl}", _options.ControllerUrl);
+
+                try
+                {
+                    using var client = new HttpClient();
```
Tratcher, line 58: Why is the client inside the outer while loop? Move up to line ~49?

src/ReverseProxy.Kubernetes.Ingress/Services/Receiver.cs
```diff
@@ -0,0 +1,95 @@
...
+                    var request = new HttpRequestMessage();
+                    request.RequestUri = new Uri(_options.ControllerUrl);
+                    request.Method = HttpMethod.Get;
+                    using var response = await client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+                    using var stream = await response.Content.ReadAsStreamAsync();
```
Tratcher, line 63:
```suggestion
                    using var stream = await client.GetStreamAsync(_options.ControllerUrl, cancellationToken);
```

src/ReverseProxy.Kubernetes.Ingress/Startup.cs
```diff
@@ -0,0 +1,50 @@
...
+            _configuration = configuration;
+        }
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.Configure<ReceiverOptions>(_configuration.Bind);
+            services.AddHostedService<Receiver>();
+            services.AddReverseProxy()
+                .LoadFromMessages();
```
Tratcher, line 31:
```suggestion
                .LoadFromMessages(receiverOptions => _configuration.Bind(receiverOptions));
```

src/ReverseProxy.Kubernetes.Protocol/MessageConfigProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Extensions.Primitives;
+using Yarp.ReverseProxy.Abstractions;
+using Yarp.ReverseProxy.Service;
+
+namespace Yarp.ReverseProxy.Kubernetes.Protocol
+{
+    public class MessageConfigProvider : IProxyConfigProvider, IUpdateConfig
```
Tratcher, line 12: Receiver and MessageConfigProvider can be combine, and IUpdateConfig removed.

src/ReverseProxy.Kubernetes.Protocol/MessageConfigProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        private volatile MessageConfig _config;
+
+        public MessageConfigProvider()
+        {
+            _config = new MessageConfig(null, null);
+        }
+
+        public IProxyConfig GetConfig() => _config;
```
Tratcher, line 21: It's recommended that GetConfig block if it's called before the first config becomes available. That will prevent the ingress from fully starting and accepting requests until the config is available. Otherwise you may 404 the first requests.

src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -0,0 +1,183 @@
...
+
+                    foreach (var port in subset.Ports ?? Enumerable.Empty<V1EndpointPort>())
+                    {
+                        if (!MatchesPort(port, service?.Port))
+                        {
+                            continue;
+                        }
+
+                        var protocol = context.Options.Https ? "https" : "http";
+                        var uri = $"{protocol}://{address.Ip}:{port.Port}";
```
Tratcher, line 94: IPv6? it will need square brackets.

src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -0,0 +1,183 @@
...
+                foreach (var address in subset.Addresses ?? Enumerable.Empty<V1EndpointAddress>())
+                {
+                    if (!clusters.ContainsKey(service?.Name))
+                    {
+                        clusters.Add(service.Name, new ClusterTrasfer());
+                    }
+
+                    var ip = address.Ip;
+                    var cluster = clusters[service.Name];
+                    cluster.ClusterId = ip;
```
Tratcher, line 84: The cluster id is the destination ip?

In theory there should be many routes -> few clusters -> many destinations.

This looks more like one route -> one cluster -> one destination?

Tratcher (commented):
I want to talk to you about the object model mapping in YarpParser. Everything else is code cleanup suggestions. I don't expect to review OperatorFramework in this PR, we can follow up on that seperately.

src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
jkotalik: It's a follow up I have, but very low priority. K8s doens't use IPv6 unless you opt in and it's extremely uncommon.

src/ReverseProxy.Kubernetes.Protocol/MessageConfigProvider.cs
jkotalik: Hm, what if there is no k8s ingress available? Are you saying that we should still wait on an empty config before accepting requests?

Otherwise I think that makes sense.

Tratcher: You mean if the controller isn't available? At that point it might make sense for the ingress to fail in some way. Throwing from here would be the most blunt solution.

src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -0,0 +1,183 @@
...
+                    foreach (var port in subset.Ports ?? Enumerable.Empty<V1EndpointPort>())
+                    {
+                        if (!MatchesPort(port, service?.Port))
+                        {
+                            continue;
+                        }
+
+                        var protocol = context.Options.Https ? "https" : "http";
+                        var uri = $"{protocol}://{address.Ip}:{port.Port}";
+                        cluster.Destinations[uri] = new Destination()
```
jkotalik, line 95: nit: don't recreate destination each time.

src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -0,0 +1,183 @@
...
+                        clusters.Add(service.Name, new ClusterTrasfer());
+                    }
+
+                    var ip = address.Ip;
+                    var cluster = clusters[service.Name];
+                    cluster.ClusterId = ip;
+
+                    foreach (var port in subset.Ports ?? Enumerable.Empty<V1EndpointPort>())
+                    {
+                        if (!MatchesPort(port, service?.Port))
```
jkotalik, line 88: Different ports for a route to the same IP need to be in different clusters (all destinations in a cluster are equal).

src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -0,0 +1,183 @@
...
+            {
+                subsets = endpoints.SingleOrDefault(x => x.Name == service?.Name).Subsets;
+            }
+
+            // make sure cluster is present
+            foreach (var subset in subsets ?? Enumerable.Empty<V1EndpointSubset>())
+            {
+                foreach (var address in subset.Addresses ?? Enumerable.Empty<V1EndpointAddress>())
+                {
+                    if (!clusters.ContainsKey(service?.Name))
```
jkotalik, line 77: name should be cluster:port

Tratcher:
Looking pretty good, and only one minor build error left. Let me know when you're ready for a final review.

jkotalik:
Go ahead.

src/ReverseProxy.Kubernetes.Protocol/Yarp.ReverseProxy.Kubernetes.Protocol.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
Tratcher, line 4: Is there anything preventing these projects from supporting netcoreapp3.1? If not please add the 3.1 TFM for consistency.

If it would take more work then only summarize the blockers here for future reference.

Tratcher (approved):

galvesribeiro:
@jkotalik awesome work!

Quick question... Is the Operator Framework something that is planned to live as a standalone repo published as a Nuget so people can create operators based on it? Or is it something built purposefuly for Yarp?

I'm building an operator for Orleans right now and would love to use this if that is reusable.

Thanks!

jkotalik:
> Quick question... Is the Operator Framework something that is planned to live as a standalone repo published as a Nuget so people can create operators based on it? Or is it something built purposefuly for Yarp?

Long term the goal is to make it standalone in some capacity (separate repo, package, etc). It isn't specific to YARP.

We don't have a timeline on when it would be available on nuget, but we aim to do so.

galvesribeiro:
Ok understood. Hopefuly it won't be long! 😄 

For now will use it as a submodule. 

Thanks!

galvesribeiro:
@jkotalik Jost to give you a feedback. 

I was able to create an operator for Microsoft Orleans and it worked just fine for what we wanted. Unfortunately, we will not be able to release it until we have the Operator Framework out as a NuGet (at least in preview).

Will keep an eye on it, thanks!

StevenTCramer:
@jkotalik @Tratcher Is there any documentation on using Yarp as a K8s Ingress Controller?

Issue #742, opened 2021-02-15
Dispose SchedulerEntry.Timer in case it's already been added


Kahbazi:
Should the `Timer` of `SchedulerEntry` get disposed in case the entry already exists in the dictionary?
https://github.com/microsoft/reverse-proxy/blob/9f568f1875eb4485cd9d6b58b85fccb9224b5658/src/ReverseProxy/Service/Management/EntityActionScheduler.cs#L62-L66

karelz:
Triage: Looks like we should call it. Should be easy to do -- anyone interested?

PR #865, opened 2021-03-23
Dispose the timer

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/742


MihaZupan:
This `TryAdd` will never return false - we're trying to add a new object instance, so it can't be present in the dictionary already.

We should instead assert this
```c#
var added = _entries.TryAdd(...);
Debug.Assert(added);
```

MihaZupan (approved):
Thanks!

PR #866, opened 2021-03-23
Allow skipping the response body 

### LINKED ISSUE #821: Interrupt and replace proxy response
It would be useful to be able to cancel sending the proxy response back and optionally send a different response.

I need to watch 302 Forward headers from the proxy response, and if they match certain criteria, send back a 200 text/html response with a notice and the link so that if the user did actually intend to navigate there they need to click the link. I got this working currently by doing the response in `TransformResponseAsync`, relying on the proxy failing with "response has already started" error. It works but is certainly not a great solution.

Tratcher:
Fixes #821

When using IHttpProxy there are scenarios where you want to discard the proxy response (errors, retry, rewrite, etc.). Because IHttpProxy.ProxyAsync is only one API there's no opportunity to interrupt it.

This change modifies one of the transform callbacks to allow short circuiting the response proxying. The status code will have been copied already but is easy for a developer to overwrite. The developer can control if headers are copied by calling base or not. This is a cheep way to give some control over the flow without breaking ProxyAsync up into a more complex API. Users that want to substitute or rewrite the response would be able to do so within the response transform callback.

A similar solution is provided in the higher layer for structured transforms. The main difference at that layer is you can have multiple transforms so the flag is moved to the shared context.

See https://github.com/microsoft/reverse-proxy/issues/821#issuecomment-800651857 for alternatives and their tradeoffs.


Tratcher:
*Ping*

MihaZupan (approved):
Looks good.

Should we extend the comment to make it clear the user is free to write different headers/content to the context / retry the proxy operation if they suppress the response body?

src/ReverseProxy/Service/Proxy/HttpTransformer.cs
```diff
@@ -67,7 +67,8 @@ public virtual Task TransformRequestAsync(HttpContext httpContext, HttpRequestMe
         /// </summary>
         /// <param name="httpContext">The incoming request.</param>
         /// <param name="proxyResponse">The response from the destination.</param>
-        public virtual Task TransformResponseAsync(HttpContext httpContext, HttpResponseMessage proxyResponse)
+        /// <returns>A bool indicating if the response should be proxied to the client or not.</returns>
```
Tratcher, line 70:
```suggestion
        /// <returns>A bool indicating if the response should be proxied to the client or not. A derived implementation 
        /// that returns false may send an alternate response inline or return control to the caller for it to retry, respond, 
        /// etc.</returns>
```

Issue #799, opened 2021-02-27
Use StringBuilderPooledObjectPolicy instead of StringBuilder in RequestHeaderForwardedTransform


Kahbazi:

davidfowl:
Don't use either? Maybe ValueStringBuilder?

Kahbazi:
https://github.com/dotnet/runtime/blob/e51470d6142784f7989d946fd3f4a667d5e743a6/src/libraries/Common/src/System/Text/ValueStringBuilder.cs#L11
`ValueStringBuilder` is internal right now. Is it ok if the implementation is copied to yarp?

Tratcher:
That seems bigger than we'd want to copy, though we might only need some of it. Can you check how much we'd need?

PR #867, opened 2021-03-23
Use ValueStringBuilder for forwarded transform

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/799

### DIFF
--- a/src/ReverseProxy/Service/RuntimeModel/Transforms/RequestHeaderForwardedTransform.cs
+++ b/src/ReverseProxy/Service/RuntimeModel/Transforms/RequestHeaderForwardedTransform.cs
@@ -52,11 +52,11 @@ namespace Yarp.ReverseProxy.Service.RuntimeModel.Transforms
 
             var httpContext = context.HttpContext;
 
-            var builder = new StringBuilder();
-            AppendProto(httpContext, builder);
-            AppendHost(httpContext, builder);
-            AppendFor(httpContext, builder);
-            AppendBy(httpContext, builder);
+            var builder = new ValueStringBuilder();
+            AppendProto(httpContext, ref builder);
+            AppendHost(httpContext, ref builder);
+            AppendFor(httpContext, ref builder);
+            AppendBy(httpContext, ref builder);
             var value = builder.ToString();
 
             var existingValues = TakeHeader(context, ForwardedHeaderName);
@@ -73,7 +73,7 @@ namespace Yarp.ReverseProxy.Service.RuntimeModel.Transforms
             return default;
         }
 
-        private void AppendProto(HttpContext context, StringBuilder builder)
+        private void AppendProto(HttpContext context, ref ValueStringBuilder builder)
         {
             if (ProtoEnabled)
             {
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Utilities/ValueStringBuilder.cs
```diff
@@ -0,0 +1,125 @@
+using System;
+using System.Buffers;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace Yarp.ReverseProxy.Utilities
+{
+    internal ref partial struct ValueStringBuilder
```
Tratcher, line 8: Add a comment with a permanent link to the original.

src/ReverseProxy/Utilities/ValueStringBuilder.cs
```diff
@@ -0,0 +1,125 @@
+using System;
```
Tratcher, line 1: License header.

src/ReverseProxy/Service/RuntimeModel/Transforms/RequestHeaderForwardedTransform.cs
```diff
@@ -178,12 +178,12 @@ private void AppendNode(IPAddress ipAddress, int port, NodeFormat format, String
             if (addPort)
             {
                 builder.Append(':');
-                builder.Append(port);
+                builder.Append(port.ToString());
```
Tratcher, line 181: InvariantCulture

davidfowl: This allocation seems unnecessary.

Tratcher: True, but ValueStringBuilder doesn't support ints.

It looks like StringBuilder calls into Number.Formatting:
https://github.com/dotnet/runtime/blob/79ae74f5ca5c8a6fe3a48935e85bd7374959c570/src/libraries/System.Private.CoreLib/src/System/Number.Formatting.cs#L830

MihaZupan: In Uri code we do
```c#
const int MaxUshortLength = 5;
bool success = port.TryFormat(dest.AppendSpan(MaxUshortLength), out int charsWritten);
Debug.Assert(success);
dest.Length -= MaxUshortLength - charsWritten;
```

src/ReverseProxy/Utilities/ValueStringBuilder.cs
```diff
@@ -0,0 +1,144 @@
...
+                RawChars[pos] = s[0];
+                _pos = pos + 1;
+            }
+            else
+            {
+                AppendSlow(s);
+            }
+        }
+
+        private void AppendSlow(string s)
```
MihaZupan, line 73:
```suggestion
```

This part can be removed as the `Length == 1` case really _isn't_ common. We should fix this in the runtime too.

Tratcher (approved):

Tratcher:
Thanks

Issue #801, opened 2021-03-01
Transforms and direct proxying


leastprivilege:
### What should we add or change to make your life better?
I am using this approach for a specialized scenario:

https://microsoft.github.io/reverse-proxy/articles/direct-proxying.html

Still I'd like to re-use your transforms "library". But I don't see an easy connection between the two. Am I missing something? Would a "standalone" transforms-builder be useful where I can pass the result in `ProxyAsync`?

### Why is this important to you?
I don't want to re-invent the wheel ;)

Tratcher:
I have been wondering about this disparity. Which transforms did you have in mind? In a related discussion (https://github.com/microsoft/reverse-proxy/discussions/768) we found that many of the transforms provided minimal functionality and only existed to map config to code.

It is possible to re-use the structured transform infrastructure today, I think the interesting question is how much of it do you need and how hard is it to re-use those parts?

The transform types themselves are all public and can be new'd up and executed directly. That might make sense if you're only looking for one or two. If you want to run many then it might help if we made the [StructuredTransformer](https://github.com/microsoft/reverse-proxy/blob/45d063e06d88f4aace4e34f7ec8e1b50f16f53dc/src/ReverseProxy/Service/RuntimeModel/Transforms/StructuredTransformer.cs#L18-L26) class public?

The ITransformBuilder exists to map config based rules to their associated transform classes. That doesn't seem applicable for the IHttpProxy scenario?

leastprivilege:
e.g. the forwarded headers...

Yes - the chaining of transforms should be public and independent of the config plumbing. 

Maybe same goes for the HTTP message invoker "factory" (IOW - returning a message invoke based on a name).

Tratcher:
> Maybe same goes for the HTTP message invoker "factory" (IOW - returning a message invoke based on a name).

Yeah, some folks have asked about deriving from the internal ProxyHttpClientFactory. Would you want to use that together with IHttpProxy? The CreateClient [inputs](https://github.com/microsoft/reverse-proxy/blob/45d063e06d88f4aace4e34f7ec8e1b50f16f53dc/src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientContext.cs#L14) are a little weird, but not prohibitive. I'll fork that out to its own issue.

Tratcher:
Here's what it would look like if we made just the StructuredTransformer class public.

            var requestTransforms = new List<RequestTransform>();
[...]

PR #869, opened 2021-03-24
Make ProxyHttpClientFactory public and extensible 

### LINKED ISSUE #805: Make ProxyHttpClientFactory public and extensible
Today when someone wants to customize their IProxyHttpClientFactory they have to copy a lot of code from the internal ProxyHttpClientFactory to avoid some common pitfalls.
https://microsoft.github.io/reverse-proxy/articles/proxyhttpclientconfig.html#custom-iproxyhttpclientfactory

Find a way to make the existing factory more extensible so they don't have to copy it.

One idea would be to have a built in extension something like this:
```c#
            services.AddReverseProxy()
                .CustomizeHttpClient((context, handler) => handler.MaxConnectionsPerServer = 25)
```

Note there have also been asks for using IProxyHttpClientFactory together with IHttpProxy.
https://github.com/microsoft/reverse-proxy/issues/801#issuecomment-788200333

### LINKED ISSUE #843: Sample/Docs for using a DelegatingHandler

Please add a sample or documentation for using a DelegatingHandler with ProxyHttpClient.
DelegatingHandler was mentioned in #137, but I can't find any further information.

I have a Single-Page Application (in Blazor) with a Backend-For-Frontend (BFF). I am using cookie authentication on the BFF and all requests from the SPA get forwarded by the BFF to an API that expects a bearer token.
Currently, I am using ProxyKit together with the [IdentityModel.AspNetCore](https://github.com/IdentityModel/IdentityModel.AspNetCore/) package.
The BFF takes the token that is stored in the cookie and adds it to the authorization header. If the token is expired, it gets automatically renewed. This is done via the `UserAccessTokenHandler`.

In ProxyKit it is very easy to configure it:
```cs
services.AddAccessTokenManagement();
services.AddProxy((clientBuilder) =>
{
    // adds the access token to all forwarded requests,
    // and refreshes the access token if needed.
    clientBuilder.AddUserAccessTokenHandler();
});
```

I understand that it's not recommended to use the `HttpClient` because it buffers responses by default. But it would be nice to have a similar way to register delegating handlers.
Another option would be to rewrite the logic and use a transformation instead of the delegating handler. But then I would have to maintain it by myself and I think it's better to reuse what's already there and don't reinvent it.

So, for now, I implemented a custom `IProxyHttpClientFactory`, but I am not sure if this is the preferred way of doing it:

```cs
public class UserAccessTokenProxyHttpClientFactory : IProxyHttpClientFactory
{
    private readonly UserAccessTokenHandler _userAccessTokenHandler;

    public UserAccessTokenProxyHttpClientFactory(UserAccessTokenHandler userAccessTokenHandler)
    {
        _userAccessTokenHandler = userAccessTokenHandler;

        var socketsHttpHandler = new SocketsHttpHandler
        {
            UseProxy = false,
            AllowAutoRedirect = false,
            UseCookies = false
        };

        _userAccessTokenHandler.InnerHandler = socketsHttpHandler;
    }

    public HttpMessageInvoker CreateClient(ProxyHttpClientContext context)
    {
        if (context.OldClient != null && context.NewOptions == context.OldOptions)
        {
            return context.OldClient;
        }

        return new HttpMessageInvoker(_userAccessTokenHandler, disposeHandler: false);
    }
}
```

My full sample can be found here: https://github.com/berhir/BlazorWebAssemblyCookieAuth

Tratcher:
Fixes #805 Make ProxyHttpClientFactory public and extensible
Fixes #843 (@berhir) Enable adding HttpMessageHandler middleware (e.g. DelegatingHandler)
Contributes to #801 Allow using ProxyHttpClientFactory for direct IHttpProxy scenarios

ProxyHttpClientFactory is now public so people can: 
A) Create it directly when using IHttpProxy
B) Derive from it to customize the SocketHttpHandler settings without completely re-implementing IProxyHttpClientFactory
C) Derive from it to enable wrapping SocketHttpHandler with DelegatingHandler middleware.

There's also a new extension for simple scenarios:
```C#
            services.AddReverseProxy()
                .ConfigureClient((context, handler) =>
                {
                    handler.Expect100ContinueTimeout = TimeSpan.FromMilliseconds(300);
                })
```

### DIFF
--- a/src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs
+++ b/src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs
@@ -6,6 +6,7 @@ using System.Net;
 using System.Net.Http;
 using System.Security.Cryptography.X509Certificates;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Yarp.ReverseProxy.Abstractions;
 using Yarp.ReverseProxy.Telemetry;
 
@@ -14,9 +15,14 @@ namespace Yarp.ReverseProxy.Service.Proxy.Infrastructure
     /// <summary>
     /// Default implementation of <see cref="IProxyHttpClientFactory"/>.
     /// </summary>
-    internal class ProxyHttpClientFactory : IProxyHttpClientFactory
+    public class ProxyHttpClientFactory : IProxyHttpClientFactory
     {
         private readonly ILogger<ProxyHttpClientFactory> _logger;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ProxyHttpClientFactory"/> class.
+        /// </summary>
+        public ProxyHttpClientFactory() : this(NullLogger<ProxyHttpClientFactory>.Instance) { }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ProxyHttpClientFactory"/> class.
@@ -35,7 +41,6 @@ namespace Yarp.ReverseProxy.Service.Proxy.Infrastructure
                 return context.OldClient;
             }
 
-            var newClientOptions = context.NewOptions;
             var handler = new SocketsHttpHandler
             {
                 UseProxy = false,
@@ -46,6 +51,34 @@ namespace Yarp.ReverseProxy.Service.Proxy.Infrastructure
                 // NOTE: MaxResponseHeadersLength = 64, which means up to 64 KB of headers are allowed by default as of .NET Core 3.1.
             };
 
+            ConfigureHandler(context, handler);
+
+            var middleware = WrapHandler(context, handler);
+
+            Log.ProxyClientCreated(_logger, context.ClusterId);
+
+            return new HttpMessageInvoker(middleware, disposeHandler: true);
+        }
+
+        /// <summary>
+        /// Checks if the options have changed since the old client was created. If not then the
+        /// old client will be re-used. Re-use can avoid the latency of creating new connections.
+        /// </summary>
+        protected virtual bool CanReuseOldClient(ProxyHttpClientContext context)
+        {
+            return context.OldClient != null && context.NewOptions == context.OldOptions;
+        }
+
+        /// <summary>
+        /// Allows configuring the <see cref="SocketsHttpHandler"/> instance. The base implementation
+        /// applies settings from <see cref="ProxyHttpClientContext.NewOptions"/>.
+        /// <see cref="SocketsHttpHandler.UseProxy"/>, <see cref="SocketsHttpHandler.AllowAutoRedirect"/>,
+        /// <see cref="SocketsHttpHandler.AutomaticDecompression"/>, and <see cref="SocketsHttpHandler.UseCookies"/>
+        /// are disabled prior to this call.
+        /// </summary>
+        protected virtual void ConfigureHandler(ProxyHttpClientContext context, SocketsHttpHandler handler)
+        {
+            var newClientOptions = context.NewOptions;
             if (newClientOptions.SslProtocols.HasValue)
             {
                 handler.SslOptions.EnabledSslProtocols = newClientOptions.SslProtocols.Value;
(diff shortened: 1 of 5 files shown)


src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs
```diff
@@ -14,10 +15,15 @@ namespace Yarp.ReverseProxy.Service.Proxy.Infrastructure
...
     /// </summary>
-    internal class ProxyHttpClientFactory : IProxyHttpClientFactory
+    public class ProxyHttpClientFactory : IProxyHttpClientFactory
     {
         private readonly ILogger<ProxyHttpClientFactory> _logger;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ProxyHttpClientFactory"/> class.
+        /// </summary>
+        public ProxyHttpClientFactory() : this(NullLogger<ProxyHttpClientFactory>.Instance) { }
```
Kahbazi, line 25: Is there any IoC container that choose the constructors based on the order? If so maybe this new constructor should be be moved after `public ProxyHttpClientFactory(ILogger<ProxyHttpClientFactory> logger)`?

vdjuric: Why is this parameterless constructor needed?

Maybe it would be better to make this constructor private or protected. Public static factory method could be added, for example CreateProxyHttpClientFactoryWithoutLogging()

Tratcher: > Why is this parameterless constructor needed?

This is an optional part of the PR. It makes it easier for people to new this up directly or to derive from it. A factory would help people new it up, but not derive from it.

vdjuric: As @Kahbazi mentioned, there may be issues when resolving such instance via DI if using services.AddSingleton<IProxyHttpClientFactory, ProxyHttpClientFactory>()

You would then probably have to manually construct IProxyHttpClientFactory instance in DI registration delegate:

services.AddSingleton<IProxyHttpClientFactory>(provider => {
    return new ProxyHttpClientFactory(...constructor parameters...);
});

vdjuric: OK sorry, just tried how it behaves in .NET 5 ... It seems Microsoft DI tries to resolve constructor with most parameters :)

Kahbazi: @Tratcher @vdjuric Does other implementations of Microsoft DI abstraction like Autofac also resolve constructor with most parameters?

Tratcher: The only implementation I've seen complaints about is ActivatorUtilities which picks the first constructor.

Kahbazi: Ah thanks! I knew something was picking the first constructor but I couldn't remember what it was. 😄

src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs
```diff
@@ -46,6 +51,34 @@ public HttpMessageInvoker CreateClient(ProxyHttpClientContext context)
                 // NOTE: MaxResponseHeadersLength = 64, which means up to 64 KB of headers are allowed by default as of .NET Core 3.1.
             };
 
+            ConfigureHandler(context, handler);
+
+            var middleware = WrapHandler(context, handler);
+
+            Log.ProxyClientCreated(_logger, context.ClusterId);
+
+            return new HttpMessageInvoker(middleware, disposeHandler: true);
```
Kahbazi, line 60: Is this `HttpMessageInvoker` ever get disposed?

Tratcher: No, there's no way of telling when it's no longer in use, except that it will get GC'd.

testassets/ReverseProxy.Code/Startup.cs
```diff
@@ -52,6 +52,10 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddReverseProxy()
                 .LoadFromMemory(routes, clusters)
+                .ConfigureClient((context, handler) =>
+                {
+                    handler.Expect100ContinueTimeout = TimeSpan.FromMilliseconds(300);
```
alnikola, line 57: Why do we need this?

alnikola (approved):
LGTM

testassets/ReverseProxy.Code/Startup.cs
Tratcher: `Expect100ContinueTimeout` or `ConfigureClient`?
- ConfigureClient this lets users change basic settings without implementing a new class.
- Expect100ContinueTimeout is only used here has an example.

PR #879, opened 2021-03-27
Adding WebProxy support for outgoing requests

vdjuric:
Hi,

I added initial WebProxy support for outgoing requests. This is useful for enterprise environments which frequently use internal web proxies.

All tests are successfully passing.

Kind regards,
Vladimir Djurić
https://vladimir.djuric.si/


src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/ProxyHttpClientOptions.cs
```diff
@@ -44,6 +44,11 @@ public sealed record ProxyHttpClientOptions
         /// </summary>
         public ActivityContextHeaders? ActivityContextHeaders { get; init; }
 
+        /// <summary>
+        /// Optional web proxy used when communicating with the destination server. 
+        /// </summary>
+        public System.Net.IWebProxy WebProxy { get; init; }
```
Tratcher, line 50: Hmm, this would add another non-serializable type to the config model. We've been trying to limit that. Is there a reason we need to directly expose IWebProxy here? Would it be possible to expose the raw values from WebProxyConfigData here and then build the proxy instance in the client factory?

vdjuric: Yes, of course. I have done it this way to mimic ProxyHttpClientOptions.ClientCertificate implementation. I will remove IWebProxyConfigLoader and construct web proxy instance in client factory.

In future, we could maybe expose IWebProxy factory so user could have more power when generating web proxy instance. Similar pattern could be applied to ProxyHttpClientOptions.ClientCertificate implementation.

Thank you for your feedback!

vdjuric: Hi,

I just commited newest version. All tests are passing.

Tratcher: > I have done it this way to mimic ProxyHttpClientOptions.ClientCertificate implementation.

Yeah, ClientCertificate is the other problematic property that we haven't figured out what to do with.

vdjuric: @MihaZupan If one will be able to derive from ProxyHttpClientFactory, register new implementation (as singleton) via dependency injection and override ConfigureHandler, then WebProxyOptions can be simplified if needed (by removing BypassOnLocal and UseDefaultCredentials properties).

Although, it may be easier for end-user to configure WebProxy directly from config.

I am neutral here :)

vdjuric: @Tratcher For certificate loading I would create simple configuration structure with properties:
-X509FindType
-findValue
-validOnly

This would be used for X509Certificate2Collection.Find method: https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2collection.find?view=net-5.0

For more advanced scenarios one would have to derive from ProxyHttpClientFactory and override ConfigureHandler.

MihaZupan: Right, I'm thinking in terms of how commonly such settings would be changed vs. mirroring the SocketsHttpHandler surface.

Tratcher: UseDefaultCredentials seems like it will be a common need. I don't know how frequently BypassOnLocal will be used, but if we already have two options then a third is harmless.

Tratcher (approved):
Looks good. I only have a few remaining cosmetic comments.

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/WebProxyOptions.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// true to bypass the proxy for local addresses; otherwise, false.
+        /// If null, default value for <seealso cref="System.Net.WebProxy"/> implementation will be used.
+        /// </summary>
+        public bool? BypassOnLocal { get; init; }
+
+        /// <summary>
+        /// Controls whether the <seealso cref="System.Net.CredentialCache.DefaultCredentials"/> are sent with requests.
+        /// If null, default value for <seealso cref="System.Net.WebProxy"/> implementation will be used.
```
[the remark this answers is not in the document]
vdjuric: Thanks!

I have included internal constants with default values in WebProxyOptions. I am not relying on default values from WebProxy implementation (but currently, they should be same).

Should we leave BypassOnLocal and UseDefaultCredentials properties as nullable bool? or should we change them to non-nullable bool ?

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -330,10 +332,23 @@ private ProxyHttpClientOptions CreateProxyHttpClientOptions(IConfigurationSectio
...
 
+        private static WebProxyOptions TryGetWebProxyOptions(IConfigurationSection webProxyConfig)
+        {
+            if (webProxyConfig == null || !webProxyConfig.Exists())
+            {
+                return null;
+            }
+
+            var options = new WebProxyOptions();
+            webProxyConfig.Bind(options);
```
Kahbazi, line 348: `Bind` is not used for other options, each property is read separately. Does this also need to change for consistency?

vdjuric: Additional question: since Bind is using reflection, is this acceptable?

Tratcher: Fair point. The biggest reason not to use bind is that it messes up the linker (the setters get deleted because it can't tell they're in use). It would be better for consistency to avoid Bind regardless.

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/WebProxyOptions.cs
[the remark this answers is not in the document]
Tratcher: We prefer nullables in the config so an extended implementation can choose its own defaults.

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/WebProxyOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// </summary>
+        public Uri Address { get; init; }
+
+        /// <summary>
+        /// true to bypass the proxy for local addresses; otherwise, false.
+        /// If null, default value will be used: false
+        /// </summary>
+        public bool? BypassOnLocal { get; init; }
+
+        internal const bool BypassOnLocalDefaultValue = false;
```
Tratcher, line 21:
```suggestion
```

The consumer (ProxyHttpClientFactory) gets to decide what the actual defaults are. The defaults in the doc comments reflect the defaults used by the default ProxyHttpClientFactory.

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
vdjuric: Thank you! This should be now fixed.

src/ReverseProxy/Abstractions/ClusterDiscovery/Contract/WebProxyOptions.cs
vdjuric: Thank you! Defaults should be now fixed :)

src/ReverseProxy/Service/Proxy/Infrastructure/ProxyHttpClientFactory.cs
```diff
@@ -85,6 +92,35 @@ public HttpMessageInvoker CreateClient(ProxyHttpClientContext context)
...
+            }
+
+            if (webProxyOptions.BypassOnLocal.HasValue)
+            {
+                webProxy.BypassProxyOnLocal = webProxyOptions.BypassOnLocal.Value;
+            }
+            else
+            {
+                webProxy.BypassProxyOnLocal = false; //use default value
+            }
```
Kahbazi, line 119:
```suggestion
            webProxy.UseDefaultCredentials = webProxyOptions.UseDefaultCredentials.GetValueOrDefault(false);

            webProxy.BypassProxyOnLocal = webProxyOptions.BypassOnLocal.GetValueOrDefault(false);
```

vdjuric: Thank you!

src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
```diff
@@ -23,6 +23,16 @@ internal static class ConfigurationReadingExtensions
             return configuration[name] is string value ? TimeSpan.Parse(value, CultureInfo.InvariantCulture) : null;
         }
 
+        internal static Uri ReadUri(this IConfiguration configuration, string name)
+        {
+            if (configuration[name] is string value && Uri.TryCreate(value, UriKind.Absolute, out var parsedUri))
```
Tratcher, line 28: We've intentionally avoided the Try APIs in the config reader because we want to actively report errors. Use the normal constructor and let it throw if the value is invalid.

Tratcher (approved):
One comment about validation, but otherwise this looks good.

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -320,6 +320,8 @@ private ProxyHttpClientOptions CreateProxyHttpClientOptions(IConfigurationSectio
                 }
             }
 
+            var webProxy = TryGetWebProxyOptions(section.GetSection(nameof(ProxyHttpClientOptions.WebProxy)));
```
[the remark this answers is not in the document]
vdjuric: With newest commit, I moved web proxy parsing inline to bo more aligned with other parsing (minor cosmetic changes can be still adjusted if needed).

src/ReverseProxy/Utilities/ConfigurationReadingExtensions.cs
vdjuric: This makes sense. I changed this in newest commit.

Tratcher:
Thanks

PR #881, opened 2021-03-29
Handle `Expect:100-continue` when destination chooses not to accept request body

### LINKED ISSUE #617: Test 100-continue scenarios

davidni:
Contributes #617


Tratcher (approved):

Tratcher:
Thanks.

I'll leave #617 open for us to add more automated tests here.

PR #882, opened 2021-03-29
Disallow punycode host names in config

### LINKED ISSUE #25: Review host matching validation

MihaZupan:
HostMatcherPolicy will accept host names in punycode, but AspNetCore always tries to match against the unicode form of the request host.

This change gives Yarp users a warning in this case, where the entry would effectively get ignored.

I tested that Yarp works with `ünicode`, but not `xn--nicode-2ya` as the host.

Fixes #25

### DIFF
--- a/src/ReverseProxy/Service/Config/ConfigValidator.cs
+++ b/src/ReverseProxy/Service/Config/ConfigValidator.cs
@@ -121,7 +121,7 @@ namespace Yarp.ReverseProxy.Service
 
             foreach (var host in hosts)
             {
-                if (string.IsNullOrWhiteSpace(host))
+                if (string.IsNullOrEmpty(host))
                 {
                     errors.Add(new ArgumentException($"Empty host name has been set for route '{routeId}'."));
                 }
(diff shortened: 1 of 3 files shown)


src/ReverseProxy/Service/Config/ConfigValidator.cs
```diff
@@ -121,9 +121,13 @@ private static void ValidateHost(IList<Exception> errors, IReadOnlyList<string>
 
             foreach (var host in hosts)
             {
-                if (string.IsNullOrEmpty(host))
+                if (string.IsNullOrWhiteSpace(host))
```
Tratcher, line 124: Avoid OrWhiteSpace, it does some strange unicode checks. We can revisit this if we find whitespace values are a common problem.

Tratcher (commented):
Please update the API docs on ProxyMatch.Hosts to cover this scenario. Also mention that ports and wildcards are allowed.

MihaZupan:
Updated the summary.

Also added c51f05d to avoid getting the `Route requires Hosts or Path specified` error for empty hosts.

Tratcher (approved):

Issue #762, opened 2021-02-19
Simplify Config Values


rposener:
### What should we add or change to make your life better?
Make config values simpler.  I deployed as test to Linux Web App in Azure, and quickly discovered that appsettings (e.g. ReverseProxy__Clusters__clientapi__Destinations__clientapi1__Address) kind of thing did not work.  It got cut-off and doesn't read the entire setting name.  A simpler Config setup as default would be nice.  Easy enough to work-around by setting my own config and using a fluent setup.

### Why is this important to you?
Just seems like it's a simper experience as default.

Tratcher:
You're trying to provide the config via environment variables? I can see how that'd get unwieldy. That structure was designed more fore appsettings.json files.

karelz:
Triage: Our config is way too complicated for full environment variables support (the flattening is rather ugly as you pointed out).
To replace destinations at runtime (post deployment), you should be able to use `IProxyConfigFilter` to replace symbolic names with real addresses of destinations.
We have a sample for using `IProxyConfigFilter`: https://github.com/microsoft/reverse-proxy/blob/main/samples/ReverseProxy.Config.Sample/CustomConfigFilter.cs

We should add doc for this.

samsp-msft:
The scenario of being able to more easily access and change config for app service hosting is a good one, and the current story is not great.

I think the improvement would be to update the config filter sample to fetch a value from a custom key, and then apply that to records in the proxy configuration. The problem is that would require there to be logic to know which values to pull from where to be replaced.

I am wondering if the sample could/should show a regex pattern match of looking for {{key}} or similar in the appsettings.json for  destinations or hosts in routes, and then do a lookup for the value?

@tratcher - I initially thought that the cancellation token was for the Filter to be able to say that the value is invalid, but it seems like its to cancel the async operation. Is there something like the token on a custom config provider to say that the values are now stale?

karelz:
Triage: Environment variables can't change during process runtime.

samsp-msft:
Resolved by #883 which shows how to use a filter to replace values in config as its loaded.

PR #883, opened 2021-03-30
Config filter sample update

samsp-msft:
Renamed config sample to make it clearer what it does.
Added logic to look at destination addresses, and if they match a pattern, treat those as redirects to environment variables.


samsp-msft:
> This makes for a good config filter demo, but I don't want to loose our general config demo.

Hmm, I must confess that I didn't realize that it was supposed to be a config demo 😺 - this is partly due to the lack of documentation, but also the config file in this sample did not seem to be particularly special.

We probably need to rethink what a config sample would look like, and how that would relate to the pretty exhaustive config that is now part of the config docs. I think whatever the sample includes, it needs to be much more fully documented! We should probably start with a list of requirements/goals for a config sample and then see which sample those fit in, or does it warrant one of its own.

Tratcher:
Putting the config that you commented in https://github.com/microsoft/reverse-proxy/blob/main/docs/docfx/articles/configfiles.md#all-config-properties into a config sample seems like a good approach. I'd want the config filter sample to have a much more targeted config file.

Tratcher:
Missing a using?
`samples/ReverseProxy.ConfigFilter.Sample/CustomConfigFilter.cs(27,64): error CS0103: (NETCORE_ENGINEERING_TELEMETRY=Build) The name 'StringComparer' does not exist in the current context`

samples/ReverseProxy.ConfigFilter.Sample/appsettings.json
```diff
@@ -0,0 +1,82 @@
...
+        "SessionAffinity": {
+          "Enabled": "true"
+        },
+        "HealthCheck": {
+          "Passive": {
+            "Enabled": "true",
+            "Policy": "TransportFailureRate",
+            "ReactivationPeriod": "00:05:00"
+          }
+        },
```
Tratcher, line 62: Trim out any optional config not related to the config filter, it's distracting.
```suggestion
```

samples/ReverseProxy.ConfigFilter.Sample/appsettings.json
```diff
@@ -0,0 +1,82 @@
...
+          {
+            "RequestHeader": "foo0",
+            "Append": "bar"
+          },
+          {
+            "ResponseHeader": "foo",
+            "Append": "bar",
+            "When": "Always"
+          }
+        ]
```
Tratcher, line 47:
```suggestion
```

Tratcher (approved):

Tratcher:
Looks like this is ready to merge.

samsp-msft:
> Missing a using?
> `samples/ReverseProxy.ConfigFilter.Sample/CustomConfigFilter.cs(27,64): error CS0103: (NETCORE_ENGINEERING_TELEMETRY=Build) The name 'StringComparer' does not exist in the current context`

Yep, you suggested the change that added the comparer, that I accepted without merging locally 😺

Tratcher:
#762

PR #886, opened 2021-03-31
Remove ItemManager classes, make ClusterInfo ctor public

### LINKED ISSUE #315: ItemManagerBase has some inefficiencies
### Describe the problem
The ItemManagerBase updates the signal when ever an item changes and that results in a recreation of the collection:

https://github.com/microsoft/reverse-proxy/blob/6fb3d3b6bbd5525755231a2e2f48cc4c77c84526/src/ReverseProxy/Service/Management/ItemManagerBase.cs#L94

This happens in response to any changes to routes, destinations and clusters. What's worse is that the entire list changes any item any item in that list changes. That will trigger a signal update per item added/removed.

### LINKED ISSUE #513: Simplify or replace the item manager classes
Since #505 is removing signals, we think that significantly reduces the usefulness of the IItemManager group of classes as well. They're used exclusively by the ProxyConfigManager and we anticipate their use can be replaced by ConcurrentDictionary or IImmutableDictionary.

### LINKED ISSUE #705: Idea: Make it possible to unit test a custom IActiveHealthCheckPolicy

Tratcher:
Fixes #315 Improve ItemManager perf
Fixes #513 Replace ItemManager
Fixes #705 Make ClusterInfo ctor and Destinations public to improve testability cc: @hughesjon

The ItemManager classes (RouteManager, ClusterManager, DestinationManager) were abstractions that were no longer being taken advantage of. They've been removed in favor of ConcurrentDictionary's.

In the process I've made ClusterInfo's constructor and Destinations property public to address testability concerns in ActiveHealthChecks that take these as inputs.


src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -109,7 +109,7 @@ private void Initialize()
 
         private void CreateEndpoints()
         {
-            var routes = _routeManager.GetItems();
+            var routes = _routes.Values;
```
davidfowl, line 112: Is the intent to get a snapshot here? If so we should add a comment because .Values locks the world and copies the list.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -152,7 +152,7 @@ public async Task<EndpointDataSource> InitialLoadAsync()
             }
 
             // Initial active health check is run in the background.
-            _ = _activeHealthCheckMonitor.CheckHealthAsync(_clusterManager.GetItems());
+            _ = _activeHealthCheckMonitor.CheckHealthAsync(_clusters.Values);
```
davidfowl, line 155: Is the intent to get a snapshot here? If so we should add a comment because .Values locks the world and copies the list.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -325,101 +325,136 @@ private async Task<(IList<Cluster>, IList<Exception>)> VerifyClustersAsync(IRead
...
+
+                    foreach (var listener in _clusterChangeListeners)
+                    {
+                        listener.OnClusterAdded(newCluster);
+                    }
+                }
             }
 
-            foreach (var existingCluster in _clusterManager.GetItems())
+            foreach (var existingCluster in _clusters.Values)
```
davidfowl, line 404: You can safely enumerate pairs without snapshotting here.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -325,101 +325,136 @@ private async Task<(IList<Cluster>, IList<Exception>)> VerifyClustersAsync(IRead
...
+                        Config = new DestinationConfig(incomingDestination.Value),
+                    };
+                    var added = destinationManager.TryAdd(newDestination.DestinationId, newDestination);
+                    Debug.Assert(added);
+                    changed = true;
+                }
             }
 
-            foreach (var existingDestination in destinationManager.GetItems())
+            foreach (var existingDestination in destinationManager.Values)
```
davidfowl, line 457: Same here

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -430,64 +465,69 @@ private void UpdateRuntimeDestinations(IReadOnlyDictionary<string, Destination>
...
+                    };
+                    var added = _routes.TryAdd(newRoute.RouteId, newRoute);
+                    Debug.Assert(added);
+                    changed = true;
+                    Log.RouteAdded(_logger, newRoute.RouteId);
+                }
             }
 
-            foreach (var existingRoute in _routeManager.GetItems())
+            foreach (var existingRoute in _routes.Values)
```
davidfowl, line 518: Same here

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -430,64 +465,69 @@ private void UpdateRuntimeDestinations(IReadOnlyDictionary<string, Destination>
...
-                    // because `IRouteManager.GetItems` returns a copy of the list of routes.
+                    // because _routeManager.Values returns a copy of the list of routes.
                     //
-                    // NOTE 2: Removing the route from `IRouteManager` is safe and existing
+                    // NOTE 2: Removing the route from _routeManager is safe and existing
                     // ASP .NET Core endpoints will continue to work with their existing behavior since
                     // their copy of `RouteConfig` is immutable and remains operational in whichever state is was in.
                     Log.RouteRemoved(_logger, existingRoute.RouteId);
-                    _routeManager.TryRemoveItem(existingRoute.RouteId);
+                    var removed = _routes.TryRemove(existingRoute.RouteId, out var _);
```
davidfowl, line 529: This is also safe with concurrent dictionary

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -44,7 +51,11 @@ public ClusterConfig Config
             internal set => _config = value ?? throw new ArgumentNullException(nameof(value));
         }
 
-        internal IDestinationManager DestinationManager { get; }
+        /// <summary>
+        /// All of the destinations associated with this cluster. This collection is populated by the configuration system
+        /// and should only be directly modified in a test environment.
+        /// </summary>
+        public ConcurrentDictionary<string, DestinationInfo> Destinations { get; } = new(StringComparer.OrdinalIgnoreCase);
```
davidfowl, line 58: Public API?

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
```diff
@@ -106,7 +117,9 @@ private void UpdateDynamicStateInternal(bool force)
                 try
                 {
                     var healthChecks = _config?.Options.HealthCheck;
-                    var allDestinations = DestinationManager.Items;
+                    // Values already makes a copy of the collection, downcast to avoid making a second copy.
+                    // https://github.com/dotnet/runtime/blob/e164551f1c96138521b4e58f14f8ac1e4369005d/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs#L2145-L2168
+                    var allDestinations = (IReadOnlyList<DestinationInfo>)Destinations.Values;
```
davidfowl, line 122: Doesn't shit copy again?

Tratcher: This is the one place we do want an explicit snapshot of the collection. The snapshot is stored in AllDestinations and a filtered copy is stored as healthyDestinations.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
Tratcher: Yes, this is public to allow for testing 3rd party implementations of some of our public interfaces like #705.

Tratcher:
Rebased and removed most uses of CD.Values.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
davidfowl: fixed the type 🤣

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
davidfowl: Mutable as well?

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
davidfowl: Doesn't this get called from the request thread?

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
Tratcher: So you can populate the destinations for your tests.

src/ReverseProxy/Service/RuntimeModel/ClusterInfo.cs
Tratcher: Passive health checks queue this call to the threadpool to avoid blocking the request:
https://github.com/microsoft/reverse-proxy/blob/28f748a0836f826b1f0ccd5404e0e23873fb744e/src/ReverseProxy/Service/HealthChecks/DestinationHealthUpdater.cs#L67

Active health checks don't take place on request threads.

davidfowl: This is going to copy on read which is going to be a performance problem. If this is a scenario then we'll need to go back to dictionaries that snapshot on bulk updates.

davidfowl: Snapshot the destinations per cluster manually.

Tratcher: Updated

davidfowl: Very nice

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -33,8 +34,9 @@ internal class ProxyConfigManager : EndpointDataSource, IDisposable
         private readonly object _syncRoot = new object();
         private readonly ILogger<ProxyConfigManager> _logger;
         private readonly IProxyConfigProvider _provider;
-        private readonly IClusterManager _clusterManager;
-        private readonly IRouteManager _routeManager;
+        private readonly IEnumerable<IClusterChangeListener> _clusterChangeListeners;
```
davidfowl, line 37: Capture this as an array

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -62,8 +63,7 @@ internal class ProxyConfigManager : EndpointDataSource, IDisposable
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
             _provider = provider ?? throw new ArgumentNullException(nameof(provider));
-            _clusterManager = clusterManager ?? throw new ArgumentNullException(nameof(clusterManager));
-            _routeManager = routeManager ?? throw new ArgumentNullException(nameof(routeManager));
+            _clusterChangeListeners = clusterChangeListeners ?? throw new ArgumentNullException(nameof(clusterChangeListeners));
```
davidfowl, line 66:
```suggestion
            _clusterChangeListeners = (clusterChangeListeners as IClusterChangeListener[]) ?? clusterChangeListeners?.ToArray() ?? throw new ArgumentNullException(nameof(clusterChangeListeners));
```

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
davidfowl, line 37:
```suggestion
        private readonly IClusterChangeListener[] _clusterChangeListeners;
```

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -152,7 +152,8 @@ public async Task<EndpointDataSource> InitialLoadAsync()
             }
 
             // Initial active health check is run in the background.
-            _ = _activeHealthCheckMonitor.CheckHealthAsync(_clusterManager.GetItems());
+            // Directly enumerate the ConcurrentDictionary to limit locking and copying.
+            _ = _activeHealthCheckMonitor.CheckHealthAsync(_clusters.Select(pair => pair.Value));
```
davidfowl, line 156: _clusters.Values?

davidfowl (approved):
Small nit. I think we should review the public APIs here.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: Didn't you just ask me not to use _clusters.Values here because it locks and copies, and the enumerator would limit both?

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
Tratcher: Do I also need to change the foreach to a for loop, or is the array enumerator automatically better?

Tratcher:
> Small nit. I think we should review the public APIs here.

@davidfowl We're string API reviews on Tuesday if you want to join us.

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
davidfowl: foreach gets lowered by c# to a for loop https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEBMBGAsAKBQZgAJ1CBhQgb10JuKJQBZCBZAClIG0BdQgEwhgQAlJWq1xAMwD2COBADGAC1YA3CAkLzCASwB2fAcLHiaVHCYuaAdABEpASRishAbmMmAvu6/na7gsRMdo7O7mbiPh5AA==

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
davidfowl: I thought we wanted to snapshot here. Is this code trying to enumerate the dictionary without snapshotting? I didn't realize that.

MihaZupan (approved):

PR #888, opened 2021-04-01
Add WebProxy to ServiceFabric

Kahbazi:


src/ReverseProxy.ServiceFabric/ServiceDiscovery/Util/LabelsParser.cs
```diff
@@ -317,8 +317,13 @@ internal static Cluster BuildCluster(Uri serviceName, Dictionary<string, string>
                     EnableMultipleHttp2Connections = GetLabel<bool?>(labels, "YARP.Backend.HttpClient.EnableMultipleHttp2Connections", null),
                     RequestHeaderEncoding = !string.IsNullOrEmpty(requestHeaderEncodingLabel) ? Encoding.GetEncoding(requestHeaderEncodingLabel) : null,
 #endif
+                    WebProxy = new WebProxyOptions
```
Tratcher, line 320: https://github.com/microsoft/reverse-proxy/pull/879/files#diff-c36308ec27c72a71f1ddd3d9c179b2714b9483195035238efc8c13f7d6040af6R77-R83

The client factory uses the presence of this type to enable proxies, which might not be the best idea. Maybe that should check for a non-null Address field instead?

Kahbazi: The non-null Address is being check here.
https://github.com/microsoft/reverse-proxy/pull/879/files#diff-c36308ec27c72a71f1ddd3d9c179b2714b9483195035238efc8c13f7d6040af6R97-R100
Is this enough?

Tratcher: Good catch, nevermind.

Tratcher (approved):

Tratcher:
Thanks

PR #890, opened 2021-04-05
Middleware docs and helper API cleanup

### LINKED ISSUE #114: Proxy can be customized with middleware to enable advanced scenarios
From looking at the source code I can see that you can add custom middlewares to the reverse proxy pipeline but I dont see any documentation outlining how to do it.
I think many of the people looking at the project is because they need to do some customization at the pipeline level and they want to avoid having to do it in another language like Lua in the case of Kong. Having this documentation would be very useful.

### LINKED ISSUE #878: Unit testing with GetRequiredCluster()
We have a middleware that reads some metadata from the cluster configuration. In previous versions of YARP we did this by getting the `IReverseProxyFeature` from the `HttpContext` directly. Preview 10 introduced a `GetRequiredCluster()` extension method on the `HttpContext` so I figured we might as well use that. The change was easy enough, but unfortunately my unit tests broke. That's because `GetRequiredCluster()` relies on the `Endpoint` being set on the `HttpContext` which we can set using `SetEndpoint()`. However, the endpoint also needs to have `RouteConfig` metadata which I'm unable to create in my unit tests since it requires a `ClusterInfo` instance and the constructor for that class is `internal`. 

Ideally I would like to be able to test my middleware without having to fetch the `IReverseProxyFeature` directly in my code. Or perhaps I'm just missing something that would allow me to do this?

Tratcher:
Fixes #114 - Docs for customizing functionality through ASP.NET middleware.
Fixes #878 - Testing middleware: Re-implements middleware helpers like GetRequiredCluster to fetch their information from the IReverseProxy feature rather than the endpoint, only DestinationInitializerMiddleware accesses the endpoint directly to create the feature. I've also moved these extensions to the same namespace as HttpContext.

cc: @jmezach

### DIFF
--- a/docs/docfx/articles/middleware.md
+++ b/docs/docfx/articles/middleware.md
@@ -1,0 +1,145 @@
+# Middleware
+
+## Introduction
+
+ASP.NET Core uses a [middleware pipeline](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/) to divide request processing into discrete steps. These can be added, removed, and ordered as needed to suite an applications needs. ASP.NET Core middleware can also be used customize proxy specific behavior.
+
+## Defaults
+The [getting started](getting_started.md) sample shows the following Configure method. This sets up a middleware pipeline with development tools, routing, and proxy configured endpoints (`MapReverseProxy`).
+
+```C#
+public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+{
+    if (env.IsDevelopment())
+    {
+        app.UseDeveloperExceptionPage();
+    }
+    
+    // Enable endpoint routing, required for the reverse proxy
+    app.UseRouting();
+    // Register the reverse proxy routes
+    app.UseEndpoints(endpoints => 
+    {
+        endpoints.MapReverseProxy(); 
+    }); 
+} 
+```
+
+The parmeterless [MapReverseProxy](xref:Microsoft.AspNetCore.Builder.ReverseProxyIEndpointRouteBuilderExtensions.MapReverseProxy) overload includes all standard proxy middleware for [session affinity](session-affinity.md), [load balancing](load-balancing.md), [passive health checks](dest-health-checks.md), and the final proxying of the request. Each of these checks the configuration of the matched route, cluster, and destination and performs their task accordingly.
+
+## Adding Middleware
+
+Middleware added to your application pipeline will see the request in different states of processing depending on where the middleware is added. Middleware added before `UseRouting` will see all requests and can manipulate them before any routing takes place. Middleware added between `UseRouting` and `UseEndpoints` can call `HttpContext.GetEndpoint()` to check which endpoint routing matched the request to (if any), and use any metadata that was associated with that endpoint. This is how [Authentication, Authorization](authn-authz.md) and [CORS](cors.md) are handled.
+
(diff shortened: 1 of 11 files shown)


docs/docfx/articles/middleware.md
```diff
@@ -0,0 +1,158 @@
...
+
+Middleware MUST check `HttpResponse.HasStarted` before modifying response fields after calling `next()`. If the response has already started being sent to the client then the middleware can no longer modify it (except maybe Trailers). [Transforms](transforms.md) can be used to inspect and suppress unwanted responses. Otherwise see the next note.
+
+Middleware should avoid interacting with the request or response bodies. Bodies are not buffered by default, so interacting with them can prevent them from reaching their destinations. While enabling buffering is possible, it's discouraged as it can add significant memory and latency overhead. Using a wrapped, streaming approach is recommended if the body must be examined or modified. See the [ResponseCompression](https://github.com/dotnet/aspnetcore/blob/24588220006bc164b63293129cc94ac6292250e4/src/Middleware/ResponseCompression/src/ResponseCompressionMiddleware.cs#L55-L73) middleware for an example.
+
+Middleware MUST NOT do any multi-threaded work on an individual request, `HttpContext` and its associated members are not thread safe.
```
samsp-msft, line 157: is thread safe the correct term? I wonder if this should say to use async, and only to access the context from the thread the async scheduler has assigned for that unit of work?

src/ReverseProxy/Middleware/HttpContextFeaturesExtensions.cs
```diff
@@ -17,6 +18,7 @@ public static class HttpContextFeaturesExtensions
         /// </summary>
         public static ClusterInfo GetRequiredCluster(this HttpContext context)
         {
+            // TODO: Retarget these to wrap GetRequiredProxyFeature
```
samsp-msft, line 21: todo for who?

samsp-msft (approved):

docs/docfx/articles/middleware.md
```diff
@@ -0,0 +1,158 @@
...
+    proxyPipeline.UseProxyLoadBalancing();
+    proxyPipeline.UsePassiveHealthChecks();
+});
+```
+
+By default this overload of `MapReverseProxy` only includes the minimal setup and proxying logic at the start and end of its pipeline. Middleware for session affinity, load balancing, and passive health checks are not included by default so that you can exclude, replace, or control their ordering with any additional middleware.
+
+## Custom Proxy Middleware
+
+Middlware inside the `MapReverseProxy` pipeline have access to all of the proxy data and state associated with a request (the route, cluster, destinations, etc.) through the [IReverseProxyFeature](xref:Yarp.ReverseProxy.Middleware.IReverseProxyFeature). This is available from `HttpContext.Features` or the extension method `HttpContext.GetRequiredProxyFeature()`.
```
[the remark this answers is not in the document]
Tratcher: Required in this case means if will throw if the feature is missing. It saves having to do null checks afterwards.

docs/docfx/articles/middleware.md
Tratcher: Thread safe is the lower level concern here. Sync or async, you can still only have one thread at a time working with the HttpContext.

src/ReverseProxy/Middleware/HttpContextFeaturesExtensions.cs
Tratcher: For me 😁, it's already done.
```suggestion
```

PR #894, opened 2021-04-06
Make StructuredTransformer public

leastprivilege:
As discussed here #801

### DIFF
--- a/src/ReverseProxy/Service/RuntimeModel/Transforms/StructuredTransformer.cs
+++ b/src/ReverseProxy/Service/RuntimeModel/Transforms/StructuredTransformer.cs
@@ -15,7 +15,7 @@ namespace Yarp.ReverseProxy.Service.RuntimeModel.Transforms
     /// <summary>
     /// Transforms for a given route.
     /// </summary>
-    internal class StructuredTransformer : HttpTransformer
+    public class StructuredTransformer : HttpTransformer
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


Tratcher:
While making StructuredTransformer public is sufficient short term, its user experience for creating the transforms is still pretty bad right?

What about the proposal from the issue?
```
            var transformBuilder = app.ApplicationServices.GetRequiredService<ITransformBuilder>();
            var transforms = transformBuilder.Create(transformBuilderContext =>
            {
                transformBuilderContext.AddXForwarded();
            });
```

We already have all those types and extensions, only the Create method would be new. I bring it up here because this proposal wouldn't use a public StructuredTransformer and I'd like to avoid a stopgap if possible.

leastprivilege:
What would `Create` return?

Tratcher:
The base type HttpTransformer.

alnikola:
There is now a related PR #913 introducing the `Create` method discussed above.

Tratcher:
Replaced by #913. @leastprivilege give that a try and let us know what you think.

leastprivilege:
Cool thanks!

Is there an ETA for the next preview?

Tratcher:
Nothing specific, though we are about due for our monthly preview. You can try the daily builds in the meantime:
https://github.com/microsoft/reverse-proxy/blob/main/docs/DailyBuilds.md

PR #896, opened 2021-04-07
Make RequestHeaderOriginalHost not conflict with custom hosts 

### LINKED ISSUE #859: RequestHeaderOriginalHost=false conflicts with RequestHeader "Host" Set
Regressed in preview 8
https://github.com/microsoft/reverse-proxy/discussions/858#discussioncomment-505443

The transform `"RequestHeaderOriginalHost": "false"` (where false is the default value) conflicts with someone trying to set a custom host via `{ "RequestHeader": "Host", "Set": "customHost" }`

If RequestHeaderOriginalHost is not set or is false, then a transform is added that removes the host header. The problem is that this is added and runs after your RequestHeader Set transform, so the Host gets set and then removed.

Workaround:
```
{ "RequestHeaderOriginalHost": "true" },
{
"RequestHeader": "Host",
"Set": "customHost"
}
```

Possible fix: Have RequestHeaderOriginalHost=false only remove the host from the proxy request if it matches the original one from the request.

Tratcher:
Fixes #859 cc: @mrob1600 

The host header for a proxy request can be set to one of three options:
A) The original request Host header
B) null, use the host from the RequestUri
C) A custom value that's different from both of the above

YARP defaults to B, but that means it must remove the host after all of the headers are copied. This removal breaks users that set a custom Host header via a transform.

I've re-worked how the host header is managed and added more tests to cover this scenario. A custom host should now always win, no matter what the other settings are.


alnikola (approved):
LGTM

Issue #634, opened 2020-12-17
Raise an error when reserved keywords for auth and CORS is used


Kahbazi:
### What should we add or change to make your life better?
There are multiple keywords for authorization and CORS policy, which yarp has special behavior for them. There could be an exception or at least error log if those keywords are actually a valid policy in ASP.Net Core authorization or CORS. 

### Why is this important to you?
Having a log will save time in troubleshooting why specific policy is not running.

Tratcher:
You mean here where we check for special values but otherwise pass it through to IAuthorizationPolicyProvider?
https://github.com/microsoft/reverse-proxy/blob/5758d7fa692af72c6ba07ba60044eb51df9d7af3/src/ReverseProxy/Service/Config/ConfigValidator.cs#L234-L250

You would want it to generate an error if "Default" or "Anonymous" was an actual registered auth policy?

I wonder how common a mistake that would be.

Kahbazi:
> You would want it to generate an error if "Default" or "Anonymous" was an actual registered auth policy what we ended up ignoring?

Yes.

> I wonder how common a mistake that would be.

It could be rare the cases that someone might add a policy named "default" instead of setting the `DefaultPolicy`.

stefannikolei:
Wouldn't it be good to have an Analyzer for that?

Tratcher:
@stefannikolei this is an error in data that would normally be pulled from outside the app at runtime. Analyzer's only apply during development.

karelz:
Triage: We should wait for demand. Does not seem to be too common ...

Kahbazi:
> Triage: We should wait for demand. Does not seem to be too common ...

I don't think any customer would demand this feature because it's just an error for preventing customer to ignore the policies.
Since this is labeled Help Wanted, do you accept PR for it?

Tratcher:
> Since this is labeled Help Wanted, do you accept PR for it?

Sure, though we'd prefer to focus on the Help Wanted items in YARP 1.0.0 first.
https://github.com/microsoft/reverse-proxy/labels/help%20wanted

PR #898, opened 2021-04-08
Raise error when a keyword is used for policy

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/634

### DIFF
--- a/src/ReverseProxy/Service/Config/ConfigValidator.cs
+++ b/src/ReverseProxy/Service/Config/ConfigValidator.cs
@@ -220,7 +220,7 @@ namespace Yarp.ReverseProxy.Service
                 var policy = await _authorizationPolicyProvider.GetPolicyAsync(authorizationPolicyName);
                 if (policy != null)
                 {
-                    errors.Add(new ArgumentException($"Authorization policy '{authorizationPolicyName}' is reserved for yarp and needs to be changed"));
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Service/Config/ConfigValidator.cs
```diff
@@ -217,11 +217,21 @@ private async ValueTask ValidateAuthorizationPolicyAsync(IList<Exception> errors
 
             if (string.Equals(AuthorizationConstants.Default, authorizationPolicyName, StringComparison.OrdinalIgnoreCase))
             {
+                var policy = await _authorizationPolicyProvider.GetPolicyAsync(authorizationPolicyName);
+                if (policy != null)
+                {
+                    errors.Add(new ArgumentException($"Authorization policy '{authorizationPolicyName}' is reserved for yarp and needs to be changed"));
```
Tratcher, line 223: This isn't actually a problem with the config, is it? Default and Anonymous are expected values in config. The problem is that a policy was defined in Startup using these reserved names. I'll try to clarify the comment:
```suggestion
                    errors.Add(new ArgumentException($"The application has registered an authorization policy named '{authorizationPolicyName}' that conflicts with the reserved authorization policy name used on this route. The registered policy name needs to be changed for this route to function."));
```
Is that clearer? If so let's use it for all of them.

Tratcher (approved):

Tratcher:
Thanks

PR #913, opened 2021-04-13
Add ITransformBuilder.Create method

Kahbazi:
- Is it ok that `Cluster` and `ProxyRoute` is null in context? or should there be another type and change all extensions method?
- Should there be any default transform in `Create` (e.g. X-Forwarded) ? 

Fixes https://github.com/microsoft/reverse-proxy/issues/801

### DIFF
--- a/testassets/ReverseProxy.Direct/Startup.cs
+++ b/testassets/ReverseProxy.Direct/Startup.cs
@@ -2,14 +2,14 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Collections.Generic;
 using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
-using Yarp.ReverseProxy.Middleware;
+using Yarp.ReverseProxy.Abstractions.Config;
+using Yarp.ReverseProxy.Service;
 using Yarp.ReverseProxy.Service.Proxy;
 using Yarp.ReverseProxy.Service.RuntimeModel.Transforms;
 
@@ -41,7 +41,17 @@ namespace Yarp.ReverseProxy.Sample
                 UseCookies = false
             });
 
-            var transformer = new CustomTransformer(); // or HttpTransformer.Default;
+            var transformBuilder = app.ApplicationServices.GetRequiredService<ITransformBuilder>();
+            var transformer = transformBuilder.Create(context =>
+            {
+                context.AddQueryRemoveKey("param1");
+                context.AddQueryValue("area", "xx2", false);
+                context.AddOriginalHost(false);
+            });
+
+            // or var transformer = new CustomTransformer();
+            // or var transformer = HttpTransformer.Default;
+
             var requestOptions = new RequestProxyOptions { Timeout = TimeSpan.FromSeconds(100) };
 
             app.UseRouting();
--- a/src/ReverseProxy/Abstractions/Config/ITransformBuilder.cs
+++ b/src/ReverseProxy/Abstractions/Config/ITransformBuilder.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using Yarp.ReverseProxy.Abstractions;
+using Yarp.ReverseProxy.Abstractions.Config;
 using Yarp.ReverseProxy.Service.Proxy;
 
 namespace Yarp.ReverseProxy.Service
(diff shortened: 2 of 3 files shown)


testassets/ReverseProxy.Direct/Startup.cs
```diff
@@ -41,7 +41,17 @@ public void Configure(IApplicationBuilder app, IHttpProxy httpProxy)
                 UseCookies = false
             });
 
-            var transformer = new CustomTransformer(); // or HttpTransformer.Default;
+            var transformBuilder = app.ApplicationServices.GetRequiredService<ITransformBuilder>();
+            var transformer = transformBuilder.Create(context =>
+            {
+                context.AddQueryRemoveKey("param1");
+                context.AddQueryValue("area", "xx2", false);
+                context.AddOriginalHost(false);
```
alnikola, line 49: It seems there is one RequestUri transform omitted which is present [in the current code](https://github.com/microsoft/reverse-proxy/blob/01af113e65f5553d8cad3410c0b224e21b0964d0/testassets/ReverseProxy.Direct/Startup.cs#L76).

testassets/ReverseProxy.Direct/Startup.cs
```diff
@@ -41,7 +41,17 @@ public void Configure(IApplicationBuilder app, IHttpProxy httpProxy)
                 UseCookies = false
             });
 
-            var transformer = new CustomTransformer(); // or HttpTransformer.Default;
```
alnikola, line 44: StructuredTransformer has more logic than a simple CustomTransfomer, thus I'm not sure if such a replacement is a right thing to do here. @Tratcher What do you think?

alnikola (commented):
At the first glance, the PR makes sense, but after a closer review I have got some concerns.

testassets/ReverseProxy.Direct/Startup.cs
Kahbazi: My intention was not to use this method as a replacement, but to only show the different ways that user could create transform.
Perhaps better comment on code would clear this up.

testassets/ReverseProxy.Direct/Startup.cs
Tratcher: The two samples don't need to be equivalent. Using the transform builder you modify individual parts of the uri and it assembles the whole afterwards.

Tratcher (commented):
> * Is it ok that `Cluster` and `ProxyRoute` is null in context? or should there be another type and change all extensions method?

Do any of today's transforms fail if cluster or route are null? While it is a bit odd, it seems a lot simpler to allow them to be null than to duplicate all of the extensions.

> Should there be any default transform in Create (e.g. X-Forwarded) ?

Yes, the ITransformBuilder should have consistent defaults between methods. That should only mean copying these two parts:
https://github.com/microsoft/reverse-proxy/pull/913/files#diff-0fb89f2699f9c7d3143b84021d8d836b2fcd1b1ec0a0223a28bc432d2b3592f8R143-R153

Or moving these parts to a shared method:
https://github.com/microsoft/reverse-proxy/pull/913/files#diff-0fb89f2699f9c7d3143b84021d8d836b2fcd1b1ec0a0223a28bc432d2b3592f8R143-R161

Kahbazi:
> Do any of today's transforms fail if cluster or route are null? While it is a bit odd, it seems a lot simpler to allow them to be null than to duplicate all of the extensions.

Nothing would fail right now. Is there a possibility there are some transform in this library that rely on cluster or route?
To prevent duplicating the extensions method, one class could inherit the other one.

Tratcher:
> > Do any of today's transforms fail if cluster or route are null? While it is a bit odd, it seems a lot simpler to allow them to be null than to duplicate all of the extensions.
> 
> Nothing would fail right now. Is there a possibility there are some transform in this library that rely on cluster or route?
> To prevent duplicating the extensions method, one class could inherit the other one.

Looking at RequestTransformContext, we don't actually pass the route or cluster to the transforms for execution, they're only on the TransformBuilderContext for potential use when doing setup. Here's one example:
https://github.com/microsoft/reverse-proxy/blob/38bddcdc7773e9e64837923390014badfb6c1fcc/src/ReverseProxy/Service/SessionAffinity/AffinitizeTransformProvider.cs#L47-L56

I think we can go with the current design and revisit it if we find an issue.

Tratcher (approved):

Tratcher:
Thanks

Tratcher:
I forgot about the default transforms for host and x-fowarded. Can you do a follow-up PR for those?

Kahbazi:
@tratcher I haven't add the default transforms to the create method yet. Don't you want that too?

Kahbazi:
> I forgot about the default transforms for host and x-fowarded. Can you do a follow-up PR for those?

Yes. I will send a PR.

PR #915, opened 2021-04-14
Only enable events/counters if there are listeners available

MihaZupan:
Removed the `AddProxyTelemetryListener`, `AddHttpTelemetryListener`, ... extension methods as they don't provide any benefit over just `AddTelemetryListeners` after this change. This is the only change to the public API, the rest are internal implementation details.

Added an abstract `EventListenerService` class as a base of all listeners.
It abstracts away the synchronization around enabling the EventSource based on parameters from the constructor.

The listener implementations now inspect the service collection for registered consumers and only enable the necessary `EventSource`s with the needed parameters. We will only enable events if a telemetry consumer is present & only enable event counters if a metrics consumer is present. Previously we would always enable events & metrics, even if only metrics were being used.

`EventSource` & `EventSourceListener` have a weird interaction where the `OnEventSourceCreated` may be called from the base constructor. That means parameters passed to the listener constructor can't control what happens in `OnEventSourceCreated`.
To work around that, the `EventSourceName` is taken from a property implemented by derived types instead of being a ctor argument.
Since only the ctor can decide on whether we need events/metrics, we delay calling `EnableEvents` until the ctor runs and inspects the service collection.
We avoid returning from `OnEventSourceCreated` before enabling the `EventSource` so that the first event isn't dropped.

### DIFF
--- a/src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
+++ b/src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
@@ -49,7 +49,6 @@ namespace Yarp.ReverseProxy.Telemetry.Consumption
         private EventSource _eventSource;
         private readonly object _syncObject = new();
         private readonly ManualResetEventSlim _initializedMre = _threadStaticInitializedMre;
-        private readonly int _ctorThreadId = Environment.CurrentManagedThreadId;
         private readonly bool _enableEvents;
         private readonly bool _enableMetrics;
 
@@ -91,10 +90,11 @@ namespace Yarp.ReverseProxy.Telemetry.Consumption
                 }
 
                 // Ensure that the constructor finishes before exiting this method (so that the first events aren't dropped)
-                // It's possible that we are executing as a part of the base ctor - check the Thread ID to avoid a deadlock
-                if (Environment.CurrentManagedThreadId != _ctorThreadId)
+                // It's possible that we are executing as a part of the base ctor - only block if we're running on a different thread
+                var mre = _initializedMre;
+                if (mre is not null && !ReferenceEquals(mre, _threadStaticInitializedMre))
                 {
-                    _initializedMre?.Wait();
+                    mre.Wait();
                 }
             }
         }
--- a/src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
+++ b/src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
@@ -1,0 +1,102 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace Yarp.ReverseProxy.Telemetry.Consumption
+{
+    internal abstract class EventListenerService<TService, TTelemetryConsumer, TMetricsConsumer> : EventListener, IHostedService
+    {
+        protected abstract string EventSourceName { get; }
+
+        protected readonly ILogger<TService> Logger;
+        protected readonly IServiceProvider ServiceProvider;
+        protected readonly IHttpContextAccessor HttpContextAccessor;
+
+        private EventSource _eventSource;
+        private readonly ManualResetEventSlim _initializedMre = new();
+        private readonly int _ctorThreadId = Environment.CurrentManagedThreadId;
+        private readonly bool _enableEvents;
+        private readonly bool _enableMetrics;
+
+        public EventListenerService(ILogger<TService> logger, IServiceProvider serviceProvider, IHttpContextAccessor httpContextAccessor, ServiceCollectionInternal services)
+        {
+            Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            ServiceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+            HttpContextAccessor = httpContextAccessor ?? throw new ArgumentNullException(nameof(httpContextAccessor));
+            _ = services ?? throw new ArgumentNullException(nameof(services));
+
+            _enableEvents = services.Services.Any(s => s.ServiceType == typeof(TTelemetryConsumer));
+            _enableMetrics = services.Services.Any(s => s.ServiceType == typeof(TMetricsConsumer));
+
+            lock (this)
+            {
+                if (_eventSource is not null)
+                {
+                    EnableEventSource();
+                }
+
+                // A different thread could be waiting for the ctor to finish, signal it that we're done
+                _initializedMre.Set();
+                _initializedMre = null;
+            }
+        }
+
+        protected override void OnEventSourceCreated(EventSource eventSource)
+        {
+            if (eventSource.Name == EventSourceName)
+            {
+                lock (this)
+                {
+                    _eventSource = eventSource;
+
+                    if (_initializedMre is null)
+                    {
+                        // Ctor already finished - enable the EventSource here
+                        EnableEventSource();
+                    }
+                }
+
+                // Ensure that the constructor finishes before exiting this method (so that the first events aren't dropped)
+                // It's possible that we are executing as a part of the base ctor - check the Thread ID to avoid a deadlock
+                if (Environment.CurrentManagedThreadId != _ctorThreadId)
+                {
+                    var mre = _initializedMre;
+                    if (mre is not null)
+                    {
+                        // This wait will generally be very short (waiting for the ctor on a different thread to finish)
+                        // If the ctor throws, the MRE will never be released - add a short timeout
+                        mre.Wait(TimeSpan.FromSeconds(15));
+                        mre.Dispose();
+                    }
+                }
+            }
+        }
+
+        private void EnableEventSource()
+        {
+            if (!_enableEvents && !_enableMetrics)
+            {
+                return;
+            }
+
+            var eventLevel = _enableEvents ? EventLevel.LogAlways : EventLevel.Critical;
+            var arguments = _enableMetrics ? new Dictionary<string, string> { { "EventCounterIntervalSec", MetricsOptions.Interval.TotalSeconds.ToString() } } : null;
+
+            EnableEvents(_eventSource, eventLevel, EventKeywords.None, arguments);
+            _eventSource = null;
+        }
+
+        public Task StartAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+
+        public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
+    }
+}
--- a/src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
+++ b/src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
@@ -3,11 +3,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.Tracing;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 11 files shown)


src/ReverseProxy.TelemetryConsumption/Sockets/SocketsEventListenerService.cs
```diff
@@ -170,22 +149,22 @@ private void OnEventCounters(EventWrittenEventArgs eventData)
...
-                if (previous is null || _serviceProvider is null)
+                if (previous is null)
                 {
                     return;
                 }
 
                 try
                 {
-                    foreach (var consumer in _serviceProvider.GetServices<ISocketsMetricsConsumer>())
+                    foreach (var consumer in ServiceProvider.GetServices<ISocketsMetricsConsumer>())
```
samsp-msft, line 159: @davidfowl - David this has to go back to the httpcontext and get the services each time. I'm wondering how efficient that is, and whether there should be an ability as part of services to register for change notifications, and cache the list locally?

MihaZupan: I'm working on a change that will minimize the DI overhead here without affecting the public API.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            ServiceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+            HttpContextAccessor = httpContextAccessor ?? throw new ArgumentNullException(nameof(httpContextAccessor));
+            _ = services ?? throw new ArgumentNullException(nameof(services));
+
+            _enableEvents = services.Services.Any(s => s.ServiceType == typeof(TTelemetryConsumer));
+            _enableMetrics = services.Services.Any(s => s.ServiceType == typeof(TMetricsConsumer));
+
+            lock (this)
```
alnikola, line 40: 1. Why do we need to lock inside the ctor? It's impossible to call it from more than one thread.
2. Please, don't lock on `this`. Use a dedicated field like `object _syncObject` instead.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
```diff
@@ -0,0 +1,102 @@
...
+                // Ensure that the constructor finishes before exiting this method (so that the first events aren't dropped)
+                // It's possible that we are executing as a part of the base ctor - check the Thread ID to avoid a deadlock
+                if (Environment.CurrentManagedThreadId != _ctorThreadId)
+                {
+                    var mre = _initializedMre;
+                    if (mre is not null)
+                    {
+                        // This wait will generally be very short (waiting for the ctor on a different thread to finish)
+                        // If the ctor throws, the MRE will never be released - add a short timeout
+                        mre.Wait(TimeSpan.FromSeconds(15));
```
alnikola, line 77: Adding timeouts to a thread synchronization logic is usually a sign that there is something wrong with the algorithm. Also, on slower platforms (e.g. slow CI agent VM), 15 seconds might not be enough.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
```diff
@@ -0,0 +1,102 @@
...
+                _initializedMre.Set();
+                _initializedMre = null;
+            }
+        }
+
+        protected override void OnEventSourceCreated(EventSource eventSource)
+        {
+            if (eventSource.Name == EventSourceName)
+            {
+                lock (this)
```
alnikola, line 57: Same issue with locking on `this`.

alnikola (commented):
`EventListenerService` initialization logic seems very suspicious.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
MihaZupan: The base constructor of this class (EventListener's ctor) registers the listener with other EventSources.
Since the base ctor runs before the derived one, `OnEventSourceCreated` can already be called from arbitrary threads while the derived ctor is still running. So the ctor itself will always run on a single thread, but we have to synchronize with `OnEventSourceCreated`.

I can create a dedicated object for locking, even though this is an internal object not exposed to the user.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
MihaZupan: This mre will be set as soon as the derived ctor finishes (so really quickly).
The case where the wait would time out is either:
- the ctor threw and never set the MRE
  - we could guard against that in the derived ctor, but we can't catch exceptions from the base ctor from the derived one
- the machine is so slow it takes 15sec to run a ctor
  - we could make this longer, the only difference is how long a crashing app takes to unblock other threads

If the ctor threw, the listener is broken anyway - there will be no value from any events.

The wait timing out does not actually pose a real risk.
The only reason it's here is so that the first event(s) will not be dropped - it's more so for keeping the tests stable.


There is a way to avoid this hack, but it would require hundreds of lines of boilerplate code across all listeners, so it didn't seem worth it.

MihaZupan: I came up with a slightly modified approach - the ctor is observed from a `Create` helper that always signals the MRE.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
MihaZupan: Added a dedicated object for this lock

alnikola: Understood. Looks good then.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
```diff
@@ -3,26 +3,52 @@
...
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 
 namespace Yarp.ReverseProxy.Telemetry.Consumption
 {
     internal abstract class EventListenerService<TService, TTelemetryConsumer, TMetricsConsumer> : EventListener, IHostedService
     {
+        // We need a way to signal to OnEventSourceCreated that the EventListenerService constructor finished
+        // OnEventSourceCreated may be called before we even reach the derived ctor (as it's exposed from the base ctor)
+        // Because of that, we can't assign the MRE as part of the ctor, we have to do it as part of the object initializer
```
alnikola, line 22: I believe you meant "as part of the type initializer"?

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
```diff
@@ -3,26 +3,52 @@
...
+        // But since the ctor itself may throw here, we need a way to observe the same MRE instance from outside the ctor
+        // We pull the MRE from a ThreadStatic that a ctor wrapper (Create) can observe
+        [ThreadStatic]
+        private static ManualResetEventSlim _threadStaticInitializedMre;
+
+        public static TEventListener Create<TEventListener>(IServiceProvider serviceProvider)
+            where TEventListener : EventListenerService<TService, TTelemetryConsumer, TMetricsConsumer>
+        {
+            _threadStaticInitializedMre = new();
```
alnikola, line 31: It looks like that only the thread which called of Create, ctor and `OnEventSourceCreated` methods will actually use `_threadStaticInitializedMre` because it's now thread static. Is it intended?

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
```diff
@@ -69,14 +94,7 @@ protected override void OnEventSourceCreated(EventSource eventSource)
                 // It's possible that we are executing as a part of the base ctor - check the Thread ID to avoid a deadlock
                 if (Environment.CurrentManagedThreadId != _ctorThreadId)
```
alnikola: Why do we need it now when `_initializedMre` is initialized from the thread static `_threadStaticInitializedMre`? `_initializedMre` will be null for all other threads.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
MihaZupan: Yes, the instace from `_threadStaticInitializedMre` is saved in the `_initializedMre` field.

Meaning that any calls on the same instance will share the same MRE, regardless of which thread they are running on.

In `OnEventSourceCreated`, however, we only want to wait on this instance's MRE if we're not running as part of the ctor - when the ThreadStatic instance is different from the field one.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
MihaZupan: I removed the `_ctorThreadId` field in 1910ae2 since we can compare with `_threadStaticInitializedMre` instead.

Looks like your review window hasn't picked up the change yet.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
MihaZupan: Right it's not an object initializer, it's the "whatever initialized fields before the instance ctor".

Looking at the C# spec that's `variable initializer` of a field. I'll update the comment.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
alnikola: What would happen if the same type of EventListnerService is created from multiple threads? They won't be synchronized, right? Is it expected?

alnikola:
Overall, this PR makes sense, but at the same time is overloaded with tricky thread synchronization logic. Not sure if it's the only way to achieve the goal.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
MihaZupan: The DI guarantees that only a single instance of a given HostedService type is created.

However, if someone creates multiple DI containers, they can end up with different instances of the type within the process.
This is still fine since `EventSource` and `EnableEvents` are thread-safe.

MihaZupan:
> Overall, this PR makes sense, but at the same time is overloaded with tricky thread synchronization logic. Not sure if it's the only way to achieve the goal.

Totally agree :)

It's unfortunate that this issue exists with the EventListener.
We have to synchronize somehow if we rely on any instance state within the event callbacks.
So far, the state were the `HttpContextAccesor`, `ServiceProvider` and `Logger` fields - and listeners had null guards against this race condition.
We will likely be adding more state in the future, so it's worth doing the synchronization properly once, in one place.

If there are simpler/better approaches I'll happily use those instead.

src/ReverseProxy.TelemetryConsumption/EventListenerService.cs
```diff
@@ -0,0 +1,120 @@
...
+        private readonly bool _enableMetrics;
+
+        public EventListenerService(ILogger<TService> logger, IServiceProvider serviceProvider, IHttpContextAccessor httpContextAccessor, ServiceCollectionInternal services)
+        {
+            Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            ServiceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+            HttpContextAccessor = httpContextAccessor ?? throw new ArgumentNullException(nameof(httpContextAccessor));
+            _ = services ?? throw new ArgumentNullException(nameof(services));
+
+            _enableEvents = services.Services.Any(s => s.ServiceType == typeof(TTelemetryConsumer));
```
davidfowl, line 62: Why do you need to scan the service collection rather than injecting `IEnumerable<TTelemetryConsumer>` or using defaults like `TTelemetryConsumer consumer = null` as a ctor arg and checking for null here.

MihaZupan: An implementation for a consumer could have the assumption that it will only be created as a part of a request. For example, assuming that HttpContextAccesor has a value in the ctor.

Checking like this avoids creating throw-away instances outside of a request.

davidfowl: I think we should avoid hacks like this and define how we want this to work.

MihaZupan: > hacks like this

Are you referring to manually inspecting the services? Why is that problematic?

> define how we want this to work

Which part?

davidfowl: Are these handler per request or singleton? Scanning the service collection is a nono.

MihaZupan: Consumers can theoretically be registered with any lifetime, but realistically only scoped/singleton would make sense (I would expect 90% case to be a single scoped consumer).

> Scanning the service collection is a nono. It's incomplete and doesn't have a full view of resolved services. Doesn't work in so many cases.

Do you mean that injecting `IEnumerable<TTelemetryConsumer>` would be different from looking at the `ServiceType`?

davidfowl: > Consumers can theoretically be registered with any lifetime, but realistically only scoped/singleton would make sense (I would expect 90% case to be a single scoped consumer).

Right so lets tell everyone that this is a singleton and not scoped or transient and make assumptions about that.

> Do you mean that injecting IEnumerable<TTelemetryConsumer> would be different from looking at the ServiceType?

Yes, looking at the service collection is incomplete and fragile.

MihaZupan: > Right so lets tell everyone that this is a singleton and not scoped or transient and make assumptions about that.

In the majority case it will be **scoped**.

davidfowl: That's our choice 😄. Very rarely is it safe to pick arbitrary lifetimes for services.

MihaZupan: The idea of correlating the events to a request only makes sense for scoped consumers.

> Yes, looking at the service collection is incomplete and fragile.

This surprises me, are there any issues opened around this?

davidfowl: > This surprises me, are there any issues opened around this?

It's not a issue, it's fundamental. For example if you choose to use a 3rd party DI container, the IServiceCollection will only have a subset of the registrations. The `IServiceCollection` acts as input to the underlying container's equivalent representation.

MihaZupan:
Closing in favour of #928

PR #923, opened 2021-04-19
API review, middleware naming

Tratcher:
API review feedback:
- Middleware naming
- Have MapReverseProxy use a derived IApplicationBuilder so that proxy middleware are only usable within that pipeline.
- Removed AffinityMiddlewareTestBase since there there is now only one affinity middleware.

The samples folder uses package references and will need to be updated later.


MihaZupan (approved):

alnikola (approved):
LGTM

PR #927, opened 2021-04-21
Per Request Metrics mk2

samsp-msft:
Implementation of per-request metrics based off events exposed by Yarp.ReverseProxy.Telemetry.Consumption. Implemented outside of the proxy as part of the metrics sample (which also gives the sample some meat).

This is a new branch/PR based on meeting on 4/20/2021.

### DIFF
--- a/samples/ReverseProxy.Metrics.Sample/HttpTelemetryConsumer.cs
+++ b/samples/ReverseProxy.Metrics.Sample/HttpTelemetryConsumer.cs
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Net.Http;
+using System.Threading;
+using Microsoft.AspNetCore.Http;
+using Yarp.ReverseProxy.Service.Proxy;
+using Yarp.ReverseProxy.Telemetry.Consumption;
+
+namespace Yarp.Sample
+{
+#if NET5_0_OR_GREATER
+
+    public sealed class HttpTelemetryConsumer : IHttpTelemetryConsumer
+    {
+        private readonly IHttpContextAccessor _httpContextAccessor;
+
+        public HttpTelemetryConsumer(IHttpContextAccessor httpContextAccessor)
+        {
+            _httpContextAccessor = httpContextAccessor;
+        }
+
+        public void OnRequestStart(DateTime timestamp, string scheme, string host, int port, string pathAndQuery, int versionMajor, int versionMinor, HttpVersionPolicy versionPolicy)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpRequestStartOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+
+        public void OnRequestStop(DateTime timestamp)
+        {
+            var metrics = PerRequestMetrics.Current;
+            Console.WriteLine($"requeststop {(timestamp - metrics.StartTime).Ticks}");
+        }
+
+        public void OnRequestFailed(DateTime timestamp)
+        {      
+        }
+
+        public void OnConnectionEstablished(DateTime timestamp, int versionMajor, int versionMinor)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpConnectionEstablishedOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+
+        public void OnRequestLeftQueue(DateTime timestamp, TimeSpan timeOnQueue, int versionMajor, int versionMinor)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpRequestLeftQueueOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+
+        public void OnRequestHeadersStart(DateTime timestamp)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpRequestHeadersStartOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+
+        public void OnRequestHeadersStop(DateTime timestamp)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpRequestHeadersStopOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+
+        public void OnRequestContentStart(DateTime timestamp)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpRequestContentStartOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+
+        public void OnRequestContentStop(DateTime timestamp, long contentLength)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpRequestContentStopOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+
+        public void OnResponseHeadersStart(DateTime timestamp)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpResponseHeadersStartOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+
+        public void OnResponseHeadersStop(DateTime timestamp)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpResponseHeadersStopOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+    }
+#endif
+}
--- a/samples/ReverseProxy.Metrics.Sample/HttpTelemetryConsumer.cs
+++ b/samples/ReverseProxy.Metrics.Sample/HttpTelemetryConsumer.cs
@@ -14,13 +14,6 @@ namespace Yarp.Sample
 
     public sealed class HttpTelemetryConsumer : IHttpTelemetryConsumer
     {
-        private readonly IHttpContextAccessor _httpContextAccessor;
-
-        public HttpTelemetryConsumer(IHttpContextAccessor httpContextAccessor)
-        {
-            _httpContextAccessor = httpContextAccessor;
-        }
-
         public void OnRequestStart(DateTime timestamp, string scheme, string host, int port, string pathAndQuery, int versionMajor, int versionMinor, HttpVersionPolicy versionPolicy)
         {
             var metrics = PerRequestMetrics.Current;
@@ -30,7 +23,7 @@ namespace Yarp.Sample
         public void OnRequestStop(DateTime timestamp)
         {
             var metrics = PerRequestMetrics.Current;
-            Console.WriteLine($"requeststop {(timestamp - metrics.StartTime).Ticks}");
+            metrics.HttpRequestContentStopOffset = (timestamp - metrics.StartTime).Ticks;
         }
 
         public void OnRequestFailed(DateTime timestamp)
--- a/samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
+++ b/samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
@@ -26,22 +26,22 @@ namespace Yarp.Sample
 
 
         // Offset Tics for each part of the proxy operation
-        public long RouteInvokeOffset { get; set; }
-        public long ProxyStartOffset { get; set; }
-        public long HttpRequestStartOffset { get; set; }
-        public long HttpConnectionEstablishedOffset { get; set; }
-        public long HttpRequestLeftQueueOffset { get; set; }
+        public float RouteInvokeOffset { get; set; }
+        public float ProxyStartOffset { get; set; }
+        public float HttpRequestStartOffset { get; set; }
+        public float HttpConnectionEstablishedOffset { get; set; }
+        public float HttpRequestLeftQueueOffset { get; set; }
 
-        public long HttpRequestHeadersStartOffset { get; set; }
-        public long HttpRequestHeadersStopOffset { get; set; }
-        public long HttpRequestContentStartOffset { get; set; }
-        public long HttpRequestContentStopOffset { get; set; }
+        public float HttpRequestHeadersStartOffset { get; set; }
+        public float HttpRequestHeadersStopOffset { get; set; }
+        public float HttpRequestContentStartOffset { get; set; }
+        public float HttpRequestContentStopOffset { get; set; }
 
-        public long HttpResponseHeadersStartOffset { get; set; }
-        public long HttpResponseHeadersStopOffset { get; set; }
-        public long HttpResponseContentStopOffset { get; set; }
+        public float HttpResponseHeadersStartOffset { get; set; }
+        public float HttpResponseHeadersStopOffset { get; set; }
+        public float HttpResponseContentStopOffset { get; set; }
 
-        public long ProxyStopOffset { get; set; }
+        public float ProxyStopOffset { get; set; }
 
         //Info about the request
         public ProxyError Error { get; set; }
@@ -56,6 +56,11 @@ namespace Yarp.Sample
         public string ToJson()
         {
             return JsonSerializer.Serialize(this, _jsonOptions);
+        }
+
+        public float CalcOffset(DateTime timestamp)
+        {
+            return (float)(timestamp - StartTime).TotalMilliseconds;
         }
     }
 }
--- a/samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
+++ b/samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
@@ -43,9 +43,6 @@ namespace Yarp.Sample
 
         public long ProxyStopOffset { get; set; }
 
-        public TimeSpan ProxyStopOffsetSpan => new TimeSpan(ProxyStopOffset);
-        public string ProxyStopOffsetString => ProxyStopOffsetSpan.ToString();
-
         //Info about the request
         public ProxyError Error { get; set; }
         public long RequestBodyLength { get; set; }
--- a/samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
+++ b/samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
@@ -1,0 +1,73 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Yarp.ReverseProxy.Service.Proxy;
+using System.Text.Json;
+
+namespace Yarp.Sample
+{
+    public class PerRequestMetrics
+    {
+        private static readonly AsyncLocal<PerRequestMetrics> _local = new AsyncLocal<PerRequestMetrics>();
+
+        /// <summary>
+        /// Factory to instantiate or restore the metrics from AsyncLocal storage
+        /// </summary>
+        public static PerRequestMetrics Current
+        {
+            get {
+                var current = _local.Value;
+                if ( current == null)
+                {
+                    current = new PerRequestMetrics();
+                    _local.Value = current;
+                }
+                return current;
+            }
+        }
+
+        // Ensure we are only fetched via the factory
+        private PerRequestMetrics() { }
+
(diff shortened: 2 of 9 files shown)


samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,68 @@
+#ReverseProxy.Metrics.Sample
+
+This sample demonstrates how to use the ReverseProxy.Telemetry.Consumption library to listen to telemetry data from YARP. In this case it uses the events to create a per-request data structure with detailed timings for each operation that takes place as part of the proxy operation.
+
+Internally YARP uses EventSource to collect telemetry events and metrics from a number of subsystems that are used toprocess the requests. The YARP telemetry library provides wrapper classes that collect these events metrics and make them available for Consumption. To listen for the metrics you register classes with DI that implement an interface for each subsystem. Event/metric listeners will only be created for the subsystems that you register for, as each registration has performance implications.
```
alnikola, line 5:
```suggestion
Internally YARP uses EventSource to collect telemetry events and metrics from a number of subsystems that are used to process the requests. The YARP telemetry library provides wrapper classes that collect these events metrics and make them available for Consumption. To listen for the metrics you register classes with DI that implement an interface for each subsystem. Event/metric listeners will only be created for the subsystems that you register for, as each registration has performance implications.
```

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,68 @@
...
+- **Kestrel** which is the web server that handles incomming requests. 
+
+  Events include:
+    - When the a request is started and stopped
+    
+  Metrics include:
+    - Connection Rate - how many connections are opened a second
+    - Total number of connections
+    - Number of TLS handshakes
+    - Incomming queue lengths
```
alnikola, line 27:
```suggestion
    - Incoming queue length
```

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,68 @@
...
+    - Number of request in flight
+    - Number of requests that have failed
+- **Kestrel** which is the web server that handles incomming requests. 
+
+  Events include:
+    - When the a request is started and stopped
+    
+  Metrics include:
+    - Connection Rate - how many connections are opened a second
+    - Total number of connections
```
alnikola, line 25: Is it the number of currently open connections or all the connections since the moment application started?

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+This sample demonstrates how to use the ReverseProxy.Telemetry.Consumption library to listen to telemetry data from YARP. In this case it uses the events to create a per-request data structure with detailed timings for each operation that takes place as part of the proxy operation.
+
+Internally YARP uses EventSource to collect telemetry events and metrics from a number of subsystems that are used toprocess the requests. The YARP telemetry library provides wrapper classes that collect these events metrics and make them available for Consumption. To listen for the metrics you register classes with DI that implement an interface for each subsystem. Event/metric listeners will only be created for the subsystems that you register for, as each registration has performance implications.
+
+The subsystems are:
+- **Proxy** which represents the overall proxy operation, and success or failure. 
+
+  Events include:
+    - When the proxy request is started and stopped
```
alnikola, line 11: Does it only cover request start/stop or it actually means from the request start to the response stop?

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,68 @@
...
+    - When the request/response bodies have been processed
+
+  Metrics include:
+    - Number of requests started
+    - Number of request in flight
+    - Number of requests that have failed
+- **Kestrel** which is the web server that handles incomming requests. 
+
+  Events include:
+    - When the a request is started and stopped
```
alnikola, line 21:
```suggestion
    - When a request is started and stopped
```

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,68 @@
...
+    
+  Metrics include:
+    - Connection Rate - how many connections are opened a second
+    - Total number of connections
+    - Number of TLS handshakes
+    - Incomming queue lengths
+
+- **Http** which is the HttpClient which makes outgoing requests to the destination servers.
```
alnikola, line 29: Could we have a better name than "Http" here? In example, "HttpClient"
```suggestion
- **HttpClient ** which is the an HTTP client which makes outgoing requests to the destination servers. 
```

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+  Events include:
+    - When Http connections are created
+    - When Http requests are queued and dequeued due to a lack of available connections
+    - When Headers are processed
+    - When Content starts being transferred. Stop events are not provided for content as it is async and completion is dependent on the consuming code.
+
+  Metrics include:
+    - Number of outgoing requests started
+    - Number of Requests failed
```
alnikola, line 39:
```suggestion
    - Number of requests failed
```

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,68 @@
...
+    - Number of active requests
```
alnikola, line 40: What is the difference from "Number of outgoing requests started"?

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,68 @@
...
+    - Number of outbound connections
+
+- **Sockets** which collects metrics about the amount of data send and received
```
alnikola, line 43:
```suggestion
- **Sockets** which collects metrics about the amount of data sent and received
```

samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
```diff
@@ -0,0 +1,73 @@
...
+        public long HttpRequestHeadersStartOffset { get; set; }
+        public long HttpRequestHeadersStopOffset { get; set; }
+        public long HttpRequestContentStartOffset { get; set; }
+        public long HttpRequestContentStopOffset { get; set; }
+
+        public long HttpResponseHeadersStartOffset { get; set; }
+        public long HttpResponseHeadersStopOffset { get; set; }
+        public long HttpResponseContentStopOffset { get; set; }
+
+        public long ProxyStopOffset { get; set; }
```
MihaZupan, line 54: We could leave all of these as `TimeSpan` that is easier to deal with/there is no ambiguity about what the value is

samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
```diff
@@ -0,0 +1,73 @@
...
+            get {
+                var current = _local.Value;
+                if ( current == null)
+                {
+                    current = new PerRequestMetrics();
+                    _local.Value = current;
+                }
+                return current;
+            }
+        }
```
MihaZupan, line 29:
```suggestion
        public static PerRequestMetrics Current => _local.Value ??= new PerRequestMetrics();
```

samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
```diff
@@ -0,0 +1,73 @@
...
+        public long ResponseBodyLength { get; set; }
+        public long RequestContentIops { get; set; }
+        public long ResponseContentIops { get; set; }
+        public string DestinationId { get; set; }
+        public string ClusterId { get; set; }
+        public string RouteId { get; set; }
+
+        public string ToJson()
+        {
+            return JsonSerializer.Serialize(this, new JsonSerializerOptions { WriteIndented = true });
```
MihaZupan, line 69: We should store the `JsonSerializerOptions` instance into a `private static readonly` and re-use it here (the options are used as a cache for everything).

samples/ReverseProxy.Metrics.Sample/PerRequestYarpMetricCollectionMiddleware.cs
```diff
@@ -0,0 +1,58 @@
...
+        public async Task InvokeAsync(HttpContext context)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.StartTime = DateTime.UtcNow;
+
+            // Call the next steps in the middleware, including the proxy
+            await _next(context);
+
+            // Called after the other middleware steps have completed
+            _logger.LogInformation("PerRequestMetrics: "+ metrics.ToJson());
```
MihaZupan, line 42: Depending on how realistic vs simple we want this sample to be, we could use

```c#
private static readonly Action<ILogger, PerRequestMetrics, Exception> _perRequestMetrics = LoggerMessage.Define<PerRequestMetrics>(
    LogLevel.Information,
    new EventId(123, "PerRequestMetrics"),
    "{metrics}");

// ...
_perRequestMetrics(logger, metrics, null);
```

samples/ReverseProxy.Metrics.Sample/HttpTelemetryConsumer.cs
```diff
@@ -0,0 +1,89 @@
...
+        public void OnRequestStart(DateTime timestamp, string scheme, string host, int port, string pathAndQuery, int versionMajor, int versionMinor, HttpVersionPolicy versionPolicy)
+        {
+            var metrics = PerRequestMetrics.Current;
+            metrics.HttpRequestStartOffset = (timestamp - metrics.StartTime).Ticks;
+        }
+
+        public void OnRequestStop(DateTime timestamp)
+        {
+            var metrics = PerRequestMetrics.Current;
+            Console.WriteLine($"requeststop {(timestamp - metrics.StartTime).Ticks}");
```
MihaZupan, line 33: Should this be saving the info as well?

samples/ReverseProxy.Metrics.Sample/ProxyTelemetryConsumer.cs
```diff
@@ -12,33 +13,59 @@ public sealed class ProxyTelemetryConsumer : IProxyTelemetryConsumer
     {
         private readonly IHttpContextAccessor _httpContextAccessor;
 
-        private DateTime _startTime;
-
         public ProxyTelemetryConsumer(IHttpContextAccessor httpContextAccessor)
         {
             _httpContextAccessor = httpContextAccessor;
```
MihaZupan, line 18: Also unused now

samples/ReverseProxy.Metrics.Sample/HttpTelemetryConsumer.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    public sealed class HttpTelemetryConsumer : IHttpTelemetryConsumer
+    {
+        private readonly IHttpContextAccessor _httpContextAccessor;
+
+        public HttpTelemetryConsumer(IHttpContextAccessor httpContextAccessor)
+        {
+            _httpContextAccessor = httpContextAccessor;
+        }
+
```
MihaZupan, line 23:
```suggestion
```

Unused

samples/ReverseProxy.Metrics.Sample/README.md
samsp-msft: Since the app started.

samples/ReverseProxy.Metrics.Sample/README.md
samsp-msft: AFAIK its from where the proxy code starts to end. It doesn't include routing and other ASP.NET infra that can occur before it.

samples/ReverseProxy.Metrics.Sample/README.md
samsp-msft: That is a problem for Miha, not this sample...

samples/ReverseProxy.Metrics.Sample/README.md
samsp-msft: One is a count of active, the other is a total count of all that have been opened, regardless of close state.

samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
samsp-msft: But timespan doesn't serialize or ToString() very well which is key to this kind of class. I did start with using double to store times as ms, but ticks actually seems easier to serialize.

samples/ReverseProxy.Metrics.Sample/README.md
MihaZupan: Correct, this is the time from start to finish within `HttpProxy`. The closest you can get to exact Start/Stop times are the Kestrel Start/Stop events.

samples/ReverseProxy.Metrics.Sample/HttpTelemetryConsumer.cs
Tratcher: Use ILogger instead of Console.

samples/ReverseProxy.Metrics.Sample/PerRequestYarpMetricCollectionMiddleware.cs
samsp-msft: I don't actually expect the output to be pushed via ILogger in reality. It was a good way to print the data to the console so that you can see it. I'll add a comment to that affect.

samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
Tratcher: What was the JsonSerializer output in this case, something like "23:59:59.999"?

MihaZupan: `System.Text.Json` is weird here:
```json
{
  "Long": 25200000000,
  "TimeSpan": {
    "Ticks": 25200000000,
    "Days": 0,
    "Hours": 0,
    "Milliseconds": 0,
    "Minutes": 42,
    "Seconds": 0,
    "TotalDays": 0.029166666666666667,
    "TotalHours": 0.7,
    "TotalMilliseconds": 2520000,
    "TotalMinutes": 42,
    "TotalSeconds": 2520
  }
}
```

Newtonsoft is a lot more reasonable
```json
{
  "Long": 25200000000,
  "TimeSpan": "00:42:00"
}
```

samsp-msft: "ProxyStopOffset": 2934452,
        "ProxyStopOffsetSpan": {
          "Ticks": 2934452,
          "Days": 0,
          "Hours": 0,
          "Milliseconds": 293,
          "Minutes": 0,
          "Seconds": 0,
          "TotalDays": 3.3963564814814817E-06,
          "TotalHours": 8.151255555555555E-05,
          "TotalMilliseconds": 293.4452,
          "TotalMinutes": 0.0048907533333333334,
          "TotalSeconds": 0.2934452
        },
        "ProxyStopOffsetString": "00:00:00.2934452",

Tratcher: Ah, https://github.com/dotnet/runtime/issues/29932

samples/ReverseProxy.Metrics.Sample/HttpTelemetryConsumer.cs
samsp-msft: It was debug code I didn't notice while scanning through.

samples/ReverseProxy.Metrics.Sample/PerRequestMetrics.cs
MihaZupan: That is a painful issue to read :/

samsp-msft: I could switch to a single which would probably pretty print and serialize well. Any precision lost will not be relevant.

samples/ReverseProxy.Metrics.Sample/Startup.cs
```diff
@@ -34,16 +34,31 @@ public void ConfigureServices(IServiceCollection services)
                 .LoadFromConfig(_configuration.GetSection("ReverseProxy"));
 
             services.AddHttpContextAccessor();
+
+            // Interface that collects general metrics about the proxy
             services.AddSingleton<IProxyMetricsConsumer, ProxyMetricsConsumer>();
+
+            // Registration of a listener to events for proxy telemetry 
             services.AddScoped<IProxyTelemetryConsumer, ProxyTelemetryConsumer>();
```
samsp-msft, line 42: @MihaZupan does this need to be scoped? I don't have per-instance state on the listeners, and got rid of the httpContext local, so I think these can be singleton now?

samples/ReverseProxy.Metrics.Sample/Startup.cs
```diff
@@ -34,16 +34,31 @@ public void ConfigureServices(IServiceCollection services)
...
             services.AddProxyTelemetryListener();
+
+            // Registration of a listener to events for HttpClient telemetry
+            // Note: this depends on changes implemented in .NET 5 
+#if NET5_0_OR_GREATER
```
samsp-msft, line 47: Should we just make these samples .NET 5+ and eliminate the 3.1 support and build errors without these guards?

MihaZupan: +1, we should drop 3.1 for samples

samples/ReverseProxy.Metrics.Sample/Startup.cs
MihaZupan: Yes these should be singleton.

samsp-msft:
Should be ready to merge.

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,69 @@
...
+    - Number of TLS handshakes
+    - Incomming queue length
+
+- **Http** which is the HttpClient which makes outgoing requests to the destination servers. 
+
+  Events include:
+    - When Http connections are created
+    - When Http requests are queued and dequeued due to a lack of available connections
+    - When Headers are processed
+    - When Content starts being transferred. Stop events are not provided for content as it is async and completion is dependent on the consuming code.
```
MihaZupan, line 36: > Stop events are not provided for content as it is async and completion is dependent on the consuming code.

What does this mean?

For response content, there are Start/Stop events fired by the Proxy telemetry.

samples/ReverseProxy.Metrics.Sample/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+  Metrics include:
+    - Number of outgoing requests started
+    - Number of requests failed
+    - Number of active requests
+    - Number of outbound connections
+
+- **Sockets** which collects metrics about the amount of data sent and received
+- **NameResolution** which collects metrics for DNS lookup of destinations
+
```
MihaZupan, line 46: Also **NetSecurity** for SslStream handshake events/session metrics

samples/ReverseProxy.Metrics.Sample/Startup.cs
```diff
@@ -34,16 +34,29 @@ public void ConfigureServices(IServiceCollection services)
...
         }
 
         /// <summary>
         /// This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         /// </summary>
         public void Configure(IApplicationBuilder app)
         {
+            // Custom middleware that collects and reports the proxy metrics
+            // Placed at the beginning so it is the first and last thing run for each request
+            app.UsePerRequestMetricCollection();
```
MihaZupan, line 58: Could alternatively use the Kestrel telemetry to get the most accurate timings possible. This would aviod the need for a middleware as Kestrel RequestStart/Stop are the first and last events to fire within the request scope.

MihaZupan (approved):

PR #932, opened 2021-04-22
React to ProxyMatch rename in docs

Kahbazi:


Tratcher:
@MihaZupan @Kahbazi These wildcard sample dependencies keep breaking the CI. While we're making breaking changes we should change them to use a full package version and update them manually as we make breaks.

@MihaZupan can you change them to use a full package version in a separate PR that also fixes the AddProxyTelemetryListener break?

Tratcher (approved):
Thanks @Kahbazi, I forgot to check the docs this time.

PR #933, opened 2021-04-22
React to Telemetry Consumption API change and add tests

MihaZupan:
Now that the build of #931 produced updated packages, react to the removal of `Add*TelemetryListener`.

Also adding tests for the helper APIs added in #928.


MihaZupan:
@Tratcher  Just saw your https://github.com/microsoft/reverse-proxy/pull/932#issuecomment-824923053

Will push the change here

MihaZupan:
Actually, this should be the last breaking change atm, so I'll put up a PR to update target versions after this is merged (since this is fixing a bug the would break the sample on the current package build).

PR #938, opened 2021-04-22
Adding a README for the samples folder

samsp-msft:
Note: We need to add "Tags" for the latest release and links for those into this doc. I think its as simple as 

```
git tag -m "Marking preview10 as the latest" latest v1.0.0-preview10
git push --tags
```
But we need to confirm, and I don't think that happens as part of a PR.

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -4,9 +4,9 @@
 
 The samples in this folder are in sync with the main branch for YARP. If there have been breaking changes to the API or configuration, they may not match what is published on Nuget. To avoid disappointment, if using the samples with the YARP library published to Nuget, please change to the branch to match the latest release or preview, either using the branch dropdown or these links:
 
-**[Samples folder in latest release/preview](tbd)**
+**[Samples folder in latest release/preview](https://github.com/microsoft/reverse-proxy/tree/release/latest)**
 
-**[Source zip for latest release, including samples](tbd)**
+**[Source zip for latest release, including samples](https://github.com/microsoft/reverse-proxy/releases/latest)**
 
 ----
 
--- a/samples/README.md
+++ b/samples/README.md
@@ -4,7 +4,7 @@
 
 The samples in this folder are in sync with the main branch for YARP. If there have been breaking changes to the API or configuration, they may not match what is published on Nuget. To avoid disappointment, if using the samples with the YARP library published to Nuget, please change to the branch to match the latest release or preview, either using the branch dropdown or these links:
 
-**[Samples folder in latest release/preview](https://github.com/microsoft/reverse-proxy/tree/release/latest)**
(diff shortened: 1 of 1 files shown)


samsp-msft:
@Tratcher

Tratcher:
The preview10 release is already tagged: https://github.com/microsoft/reverse-proxy/tree/v1.0.0-preview10

Tratcher:
Or do you mean creating a `latest` tag that we move? Tags aren't meant to be moved, it does weird things when syncing a repo.

samsp-msft:
Yes, have a latest tag.

Tratcher:
We can use a branch instead of a tag.

samsp-msft:
Oh, and then update the "latest" branch to match what the latest release is - that'll work.

Tratcher:
Here you go: https://github.com/microsoft/reverse-proxy/tree/release/latest

samsp-msft:
Is there a way to get a link to the latest release drop as well. I see that on other repos that I regularly use?

samsp-msft:
Links fixed.

samples/README.md
```diff
@@ -0,0 +1,27 @@
+# Samples
+
+**Warning: Breaking Changes**
+
+The samples in this folder are in sync with the main branch for YARP. If there have been breaking changes to the API or configuration, they may not match what is published on Nuget. To avoid disappointment, if using the samples with the YARP library published to Nuget, please change to the branch to match the latest release or preview, either using the branch dropdown or these links:
+
+**[Samples folder in latest release/preview](https://github.com/microsoft/reverse-proxy/tree/release/latest/samples)**
+
+**[Source zip for latest release, including samples](https://github.com/microsoft/reverse-proxy/releases/latest)**
```
Tratcher, line 9: This actually goes to the releases page (https://github.com/microsoft/reverse-proxy/releases), is that what you intended?

Tratcher (approved):

samsp-msft:
yes as we don’t have an assets link AFAIK, the release page has a link to download a zip for that release. Ideally we’d have a zip of just the samples - we can do that with their own repo, and also have branches that match each release and only for releases.

Please excuse any typo’s this message was fat fingered on my phone.
________________________________

samples/README.md
Tratcher:
```suggestion
**[Source zip for latest release, including samples](https://github.com/microsoft/reverse-proxy/releases)**
```

Tratcher (approved):

samples/README.md
samsp-msft: latest does nothing now, but from checking with other repos that release, that will just show a page with the latest release.

Tratcher: Ah. That make some sense.

Issue #924, opened 2021-04-20
Make internal classes sealed for consistency


Kahbazi:

karelz:
Triage: Nice to have. Perf wins would not be measurable most likely.
For consistency, it make sense. If we do it, we should ALSO add static analysis rule (.editorconfig) to add it into our coding guidelines.

PR #939, opened 2021-04-23
sealed internal class

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/924

### DIFF
--- a/src/OperatorFramework/src/Controller/Informers/ResourceInformer.cs
+++ b/src/OperatorFramework/src/Controller/Informers/ResourceInformer.cs
@@ -411,7 +411,7 @@ namespace Microsoft.Kubernetes.Controller.Informers
             }
         }
 
-        internal class Registration : IResourceInformerRegistration
+        internal sealed class Registration : IResourceInformerRegistration
         {
             private bool _disposedValue;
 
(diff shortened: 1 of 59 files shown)


karelz (approved):

src/OperatorFramework/src/Controller/Informers/ResourceInformer.cs
```diff
@@ -436,7 +436,7 @@ public Registration(ResourceInformer<TResource> resourceInformer, ResourceInform
 
             public Task ReadyAsync(CancellationToken cancellationToken) => ResourceInformer.ReadyAsync(cancellationToken);
 
-            protected virtual void Dispose(bool disposing)
+            private void Dispose(bool disposing)
```
Kahbazi, line 439: Should I revert the changes of this class or change the Dispose method so it is only one method?

karelz: I would revert it. OperatorFramework may live on its own eventually. No point in hyper-optimizing for things like these. Keeping the common dispose pattern is likely more future-proof.

MihaZupan:
I can't find any editorconfig options for this specific case.

Issue #940, opened 2021-04-23
Preview 11 release checklist


MihaZupan:
- [x] Ensure there's a release branch created (see [Branching](https://github.com/microsoft/reverse-proxy/blob/main/docs/operations/Branching.md))
- [x] Ensure the `Version.props` has the `PreReleaseVersionLabel` updated to the next preview
- [x] Identify and validate the build on the `microsoft-reverse-proxy-official` pipeline
  - https://dev.azure.com/dnceng/internal/_build/results?buildId=1104350&view=results
- [x] Release the build
  - https://www.nuget.org/packages/Yarp.ReverseProxy/1.0.0-preview.11.21223.5
- [x] Tag the commit
- [x] Draft release notes
- [x] Publish the docs
- [x] Publish release notes
- [x] Close the [old milestone](https://github.com/microsoft/reverse-proxy/milestones)
- [x] Announce on social media https://twitter.com/_MihaZupan/status/1385712228465352714
- [x] Set the preview branch to protected
- [x] Delete the [previous preview branch](https://github.com/microsoft/reverse-proxy/branches)

Fix branches:
- [x] Protect `release/1.0.0-preview11` and `release/latest`
- [x] Delete `release/1.0.0-preview10` and `release/docs`

Tratcher:
@samsp-msft is working on an new step for this list here: https://github.com/microsoft/reverse-proxy/pull/938
After "Tag the commit"
- [ ] Reset the `release/latest` branch to match the new preview branch.

This shouldn't require a PR or any new commits, it should be a simple reset forward. We likely don't need to make the branch protected.

MihaZupan:
Does it make sense to have both `release/latest` and `release/docs`? They should both have the same meaning now

Tratcher:
Hmm. In theory release/docs could be incrementally updated between releases, but I guess that wouldn't hurt the samples. @samsp-msft does merging these concepts make sense to you?

Tratcher:
@samsp-msft release/docs also targets the prior release and is only updated between releases if we find a bug in the docs. Doc bug fixes would not impact the samples.

samsp-msft:
ok, but lets make it release/latest as that's more obvious as to what it points at

Tratcher:
That seems workable. We'd have to replace `release/docs` in a few places like:
[...]

Issue #873, opened 2021-03-24
Operator Framework Packaging


Tratcher:
https://github.com/microsoft/reverse-proxy/tree/main/src/OperatorFramework includes 5 new packages that we don't anticipate owning long term. For preview11 and 1.0 we need to decide if we should directly ship these as public packages and APIs, or if we want to embed them as internal dependencies / APIs.

PR #941, opened 2021-04-23
Add IsPackable=false for Kubernetes pacakges on internal builds

MihaZupan:
#873 #940


Tratcher:
Do this for everything in https://github.com/microsoft/reverse-proxy/tree/main/src/OperatorFramework/src as well.

MihaZupan:
Added. Also changed indentation to match the 2-spaces we use everywhere else.

Tratcher (approved):
Perfect. Now preview11 should have the same artifacts as 10.

Issue #935, opened 2021-04-22
Breaking changes and samples, oh my


samsp-msft:
We have a problem that breaking changes to API & configuration can break samples. There are a couple of problems that need to be solved:
- Customers should be easily directed to samples that are correct for the latest release of YARP
- Customers should be able to build those samples pulling the appropriate package from nuget.org
- Sample authoring should be against the current state of the packages build from /src - so if you make API changes, then you need to update the sample to match

To aid with this, I think we need to make the following changes:
- [ ] Update the references in the sample to contain both a package reference and a project reference, with a conditional variable that selects which to use
- [ ] Add the variable to the build scripts so that it applies for builds of the complete YARP tree, or by the build lab
- [x] Add a latest tag to the releases so that we can link to just "latest" - this may be automatic but failing because we have the pre-release flag set.
- [x] Add a readme to samples folder pointing users to the versions for the most recent release
- [ ] Add a release output which is a zip of just the samples, along with any build.props files to ensure that they reference that version of the release, making it easier to get to the samples
  or
- [ ] publish the samples to a different project under YarpProxy that has a default branch for the most recent release

samsp-msft:
@MihaZupan @Tratcher @alnikola

karelz:
Triage:

Plan discussed:
- [x] Samples folder in repo will have project references
- [ ] Release branches will reference nuget packages -- need to be part of release checklist

PR #944, opened 2021-04-23
conditional refs

samsp-msft:
Conditional references that use nuget or local reference depending on where they are built. If you move the sample out from the enlistment it uses nuget.

This would fix part of #935


Tratcher:
That should work, with the disclaimer than the package reference will be out of date so the code might not actually build/work when copied out. We'll have to decide how frequently to update the package references. At least once after every release would be a good start.

MihaZupan:
Closing in favor of the manual approach of changing project references to package references on release https://github.com/microsoft/reverse-proxy/issues/935#issuecomment-840707950

PR #953, opened 2021-04-27
Fix SDL config and update CredScan suppression

MihaZupan:
We don't have a `PoliCheckExclusions` file in this repo. Fixing that config uncovers the issue that the benchmark app key has moved since we added the CredScan suppression for it.

### DIFF
--- a/.config/CredScanSuppressions.json
+++ b/.config/CredScanSuppressions.json
@@ -4,7 +4,7 @@
         {
             "_justification": "Legitimate key/cert used for testing",
             "file": [
-                "samples/BenchmarkApp/testCert.pfx",
+                "testassets/BenchmarkApp/testCert.pfx",
                 "test/TestCertificates/testCert.pfx",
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -184,5 +184,4 @@ stages:
         -TsaIterationPath $(_TsaIterationPath)
         -TsaRepositoryName "ReverseProxy"
         -TsaCodebaseName "ReverseProxy"
-        -TsaPublish $True
-        -PoliCheckAdditionalRunConfigParams @("UserExclusionPath < $(Build.SourcesDirectory)/eng/PoliCheckExclusions.xml")'
```
Tratcher, line 188: It might be better to create the PoliCheckExclusions file now in case we need to add exclusions later. Remembering how to re-add this config later will be harder.

MihaZupan: changed

Tratcher (approved):

PR #955, opened 2021-04-28
Route API renames

Tratcher:
From API review: For the main concepts in YARP (route, cluster, destination) we use three levels of object model. 
- ProxyRoute, Cluster, Destination - the raw config input, immutable records
- RouteConfig, ClusterConfig, DestinationConfig - A set of objects derived from the raw config like the HttpClient to use, structured transforms,  etc.. This is rebuilt each time config changes and swapped out atomically so that new requests get a new, coherent model.
- RouteInfo, ClusterInfo, DestinationInfo - The long lived objects that identify an route/cluster/destination across config reloads. These track ongoing state like counters and health.

The naming is inconsistent and not very descriptive. We came up with the following new names:
- Route/Cluster/DestinationConfig for the lowest level config input
- Route/Cluster/DestinationState for the derived configuration
- Route/Cluster/DestinationEntity for the long lived objects

In this PR I've only updated the route types to see if the new names work well.
- *Config seems to work well for the lowest level.
- *State seems like a misnomer for the middle level since a lot of the state is actually tracked by the upper level. We also have DestinationHealthState and ClusterDynamicState types that represent real state. The middle level objects actually represent derived configuration. RouteModel, RouteConstruct?
- *Entity is ok, but still vague. RouteInstance? RouteState?

Update:
- Route/Cluster/DestinationConfig for the lowest level config input
- Route/Cluster/DestinationModel for the derived configuration
- Route/Cluster/DestinationState for the long lived objects

I've still only done the Route renames for this PR. I'll do a follow-up PR for clusters and destinations so it's easier to review.

### DIFF
--- a/src/ReverseProxy/Service/DynamicEndpoint/ProxyEndpointFactory.cs
+++ b/src/ReverseProxy/Service/DynamicEndpoint/ProxyEndpointFactory.cs
@@ -27,7 +27,7 @@ namespace Yarp.ReverseProxy
 
         private RequestDelegate _pipeline;
 
-        public Endpoint CreateEndpoint(RouteConfig route, IReadOnlyList<Action<EndpointBuilder>> conventions)
+        public Endpoint CreateEndpoint(RouteState route, IReadOnlyList<Action<EndpointBuilder>> conventions)
         {
             var proxyRoute = route.ProxyRoute;
             var proxyMatch = proxyRoute.Match;
--- a/src/ReverseProxy/Service/DynamicEndpoint/ProxyEndpointFactory.cs
+++ b/src/ReverseProxy/Service/DynamicEndpoint/ProxyEndpointFactory.cs
@@ -27,33 +27,33 @@ namespace Yarp.ReverseProxy
 
         private RequestDelegate _pipeline;
 
-        public Endpoint CreateEndpoint(RouteState route, IReadOnlyList<Action<EndpointBuilder>> conventions)
+        public Endpoint CreateEndpoint(RouteModel route, IReadOnlyList<Action<EndpointBuilder>> conventions)
         {
-            var proxyRoute = route.ProxyRoute;
-            var proxyMatch = proxyRoute.Match;
+            var config = route.Config;
+            var match = config.Match;
 
             // Catch-all pattern when no path was specified
-            var pathPattern = string.IsNullOrEmpty(proxyMatch.Path) ? "/{**catchall}" : proxyMatch.Path;
+            var pathPattern = string.IsNullOrEmpty(match.Path) ? "/{**catchall}" : match.Path;
 
             var endpointBuilder = new RouteEndpointBuilder(
                 requestDelegate: _pipeline ?? throw new InvalidOperationException("The pipeline hasn't been provided yet."),
                 routePattern: RoutePatternFactory.Parse(pathPattern),
-                order: proxyRoute.Order.GetValueOrDefault())
+                order: config.Order.GetValueOrDefault())
             {
-                DisplayName = proxyRoute.RouteId
+                DisplayName = config.RouteId
             };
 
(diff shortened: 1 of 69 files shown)


samsp-msft:
How about Snapshot for the intermediary middle layer, and State for the long lived objects.

- RouteConfig
- RouteSnapshot
- RouteState

Tratcher:
Triage: 
RouteConfig
RouteModel
RouteState

src/ReverseProxy/Service/DynamicEndpoint/ProxyEndpointFactory.cs
```diff
@@ -27,33 +27,33 @@ internal sealed class ProxyEndpointFactory
 
         private RequestDelegate _pipeline;
 
-        public Endpoint CreateEndpoint(RouteConfig route, IReadOnlyList<Action<EndpointBuilder>> conventions)
+        public Endpoint CreateEndpoint(RouteModel route, IReadOnlyList<Action<EndpointBuilder>> conventions)
         {
-            var proxyRoute = route.ProxyRoute;
-            var proxyMatch = proxyRoute.Match;
+            var config = route.Config;
```
alnikola, line 32: Should we rename 'config' to 'route' here to be consistent?

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
```diff
@@ -76,18 +76,18 @@ void Action(EndpointBuilder endpointBuilder)
...
 
             void Action(EndpointBuilder endpointBuilder)
             {
-                var routeConfig = endpointBuilder.Metadata.OfType<RouteConfig>().Single();
+                var route = endpointBuilder.Metadata.OfType<RouteModel>().Single();
 
-                var cluster = routeConfig.Cluster?.Config.Options;
-                var proxyRoute = routeConfig.ProxyRoute;
+                var cluster = route.Cluster?.Config.Options;
+                var config = route.Config;
```
alnikola, line 88: `config` to `route` here as well?

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -494,22 +494,22 @@ private bool UpdateRuntimeRoutes(IList<ProxyRoute> incomingRoutes)
...
                         changed = true;
                         Log.RouteChanged(_logger, currentRoute.RouteId);
                     }
                 }
                 else
                 {
-                    var newConfig = BuildRouteConfig(incomingRoute, cluster);
-                    var newRoute = new RouteInfo(incomingRoute.RouteId)
+                    var newModel = BuildRouteModel(incomingRoute, cluster);
+                    var newRoute = new RouteState(incomingRoute.RouteId)
```
alnikola, line 510: `newRoute` to `newState`?

alnikola (approved):
LGTM with a few suggestions.

src/ReverseProxy/Service/DynamicEndpoint/ProxyEndpointFactory.cs
Tratcher: To limit verbose names I used 'config' 'match' 'model' etc. when working only with route objects like in this method, 'route' 'cluster' when only dealing with one object type like config,  and 'routeConfig' 'clusterState' when working with a mixture of objects.

I'll check the others to see if this is applied consistently.

src/ReverseProxy/Service/DynamicEndpoint/ReverseProxyConventionBuilder.cs
Tratcher: I'll change these to `routeModel`, `routeConfig`, `clusterConfig` to avoid confusion with the mixture of objects.

PR #959, opened 2021-04-29
Include symbols

MihaZupan:


alnikola (approved):
LGTM

Tratcher:
Description please, how does this change our publish flow and end user experience?

MihaZupan:
A pack command will produce an aditional `.snupkg` files per package.
The idea is that the user (IDE) has the option to fetch Yarp PDBs to aid in local debugging.

Regarding the publish flow, it should "just work". The extra "packages" are included under artifacts.
image

We should verify that they are getting published in the next release.

MihaZupan:
Though it seems our packages are already embedding pdbs ... I'll take a look. Idealy we should be publishing the symbols separately to reduce the package size

samsp-msft:
This may be a silly question, but as we ship the source and create a branch for each release, why do we need to publish symbols - if somebody has a problem with Yarp - wouldn't it be easiest to download the source and change the ref from a package to a project?

Tratcher:
VS can pull the symbols and sources for you automatically.

Switching from a package reference to a project reference is a lot more manual labor.

PR #963, opened 2021-04-30
API renames: Clusters

Tratcher:
A continuation of #955
Cluster ->ClusterConfig
ClusterConfig -> ClusterModel
ClusterInfo -> ClusterState

I'll do Destinations in a 3rd PR.

Also fixed #909, Id -> ClusterId to match RouteId and DestinationId.

PR #973, opened 2021-05-03
API renames: destinations

Tratcher:
Continuation of #963
Destination -> DestinationConfig
DestinationConfig -> DestinationModel
DestinationInfo -> DestinationState

I moved DestinationConfig up two folders because it was the only thing in that directory.


alnikola (approved):
LGTM

PR #977, opened 2021-05-05
Filter out disallowed connection-related request headers

### LINKED ISSUE #439: Do not flow the Connection header?

alnikola:
It removes most of the connection-related request headers received from client before proxy a request to a downstream to shield the YARP against malicious clients which can try to manipulate YARP to proxy HTTP connections. Headers required for WebSockets protocol upgrade are exempted from this policy.

Fixes #439


src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -294,6 +294,8 @@ public HttpProxy(ILogger<HttpProxy> logger, IClock clock)
             // :: Step 3: Copy request headers Client --► Proxy --► Destination
             await transformer.TransformRequestAsync(context, destinationRequest, destinationPrefix);
 
+            FilterConnectionHeaders(destinationRequest, isUpgradeRequest);
```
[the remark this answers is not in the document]
alnikola: @Tratcher Could you please share your opinion about the above?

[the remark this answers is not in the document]
alnikola: I moved the removal logic to HttpTransformer and added the logic to HttpProxy restoring connection upgrade headers for an upgradeable request.
**Note**: I didn't add separate unit tests for HttpTransformer because HttpProxy creates its instance by default. If you think they are still necessary I will add them.

src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -15,19 +15,24 @@ namespace Yarp.ReverseProxy.Utilities
     /// </summary>
     public static class RequestUtilities
     {
+        internal static bool ShouldSkipRequestHeader(string headerName)
+        {
+            return _invalidH2H3Headers.Contains(headerName);
```
Tratcher, line 20: This is a misleading now, we're removing these from HTTP/1.1 requests too, it just happens to be the same set. Rename it to _headersToExclude? The the explanation for why they're excluded should go in a code comment on that member.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -304,6 +304,28 @@ public HttpProxy(ILogger<HttpProxy> logger, IClock clock)
...
+            {
+                var hasConnectionUpgrade = false;
+                foreach (var headerValue in connectionValue)
+                {
+                    if (headerValue.Contains("upgrade"))
+                    {
+                        hasConnectionUpgrade = true;
+                        break;
+                    }
+                }
```
Tratcher, line 319: We want a more precise match on the upgrade value, and to preserve the original input whenever possible.
```suggestion
            if (isUpgradeRequest)
            {
                var connectionValues =  context.Request.Headers.GetCommaSeperatedValues(HeaderNames.Connection);
                string connectionUpgradeValue = null;
                foreach (var headerValue in connectionValues)
                {
                    if (headerValue.Equals("upgrade", StringComparison.OrdinalIgnoreCase))
                    {
                        connectionUpgradeValue = headerValue;
                        break;
                    }
                }
```

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -304,6 +304,28 @@ public HttpProxy(ILogger<HttpProxy> logger, IClock clock)
...
+
+                if (hasConnectionUpgrade && context.Request.Headers.TryGetValue(HeaderNames.Upgrade, out var upgradeValue))
+                {
+                    destinationRequest.Headers.TryAddWithoutValidation(HeaderNames.Connection, "upgrade");
```
Tratcher, line 323:
```suggestion
                if (connectionUpgradeValue != null && context.Request.Headers.TryGetValue(HeaderNames.Upgrade, out var upgradeValue))
                {
                    destinationRequest.Headers.TryAddWithoutValidation(HeaderNames.Connection, connectionUpgradeValue);
```

src/ReverseProxy/Utilities/RequestUtilities.cs
Tratcher: Move the pseudo header check here too:
```
// Filter out HTTP/2 pseudo headers like ":method" and ":path", those go into other fields.
                if (headerName.Length > 0 && headerName[0] == ':')
                {
                    continue;
                }
```

Tratcher: Please do something similar for the response, removing transport headers even for HTTP/1.1 and restoring those required for WebSockets.

alnikola: - Pseudo-headers filtering moved to the same method as connection ones
- Connection headers filtering for response added

src/ReverseProxy/Service/Proxy/HttpProxy.cs
alnikola: Done

src/ReverseProxy/Service/Proxy/HttpProxy.cs
alnikola: Done

Tratcher (approved):

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -294,14 +293,39 @@ public HttpProxy(ILogger<HttpProxy> logger, IClock clock)
...
             Log.Proxying(_logger, destinationRequest.RequestUri);
 
             // TODO: What if they replace the HttpContent object? That would mess with our tracking and error handling.
-            return (destinationRequest, requestContent);
+            return (destinationRequest, requestContent, isUpgradeRequest);
+        }
+
+        private static void RestoreUpgradeHeaders(HttpContext context, HttpRequestMessage destinationRequest, bool isUpgradeRequest)
+        {
+            if (isUpgradeRequest)
```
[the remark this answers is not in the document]
alnikola: Having `if` inside this kind of a function is just a style preference. I moved it to the caller since it seems more convenient for you.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -461,7 +485,24 @@ private static ValueTask<bool> CopyResponseStatusAndHeadersAsync(HttpResponseMes
             }
 
             // Copies headers
-            return transformer.TransformResponseAsync(context, source);
+            var result = await transformer.TransformResponseAsync(context, source);
+
+            RestoreUpgradeHeaders(context, source, isUpgradeRequest);
```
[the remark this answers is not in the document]
alnikola: Done.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -461,7 +485,24 @@ private static ValueTask<bool> CopyResponseStatusAndHeadersAsync(HttpResponseMes
...
+        private static void RestoreUpgradeHeaders(HttpContext context, HttpResponseMessage response, bool isUpgradeRequest)
+        {
+            if (isUpgradeRequest && response.Headers.TryGetValues(HeaderNames.Upgrade, out var upgradeValues))
+            {
+                var stringValues = StringValues.Empty;
+                foreach(var value in upgradeValues)
+                {
+                    stringValues = StringValues.Concat(stringValues, value);
+                }
+                context.Response.Headers.TryAdd(HeaderNames.Upgrade, stringValues);
```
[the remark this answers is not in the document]
alnikola: Done.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -117,7 +116,7 @@ public HttpProxy(ILogger<HttpProxy> logger, IClock clock)
                 var isStreamingRequest = isClientHttp2 && ProtocolHelper.IsGrpcContentType(context.Request.ContentType);
 
                 // :: Step 1-3: Create outgoing HttpRequestMessage
-                var (destinationRequest, requestContent) = await CreateRequestMessageAsync(
+                var (destinationRequest, requestContent, isUpgradeRequest) = await CreateRequestMessageAsync(
```
Tratcher, line 119: isUpgradeRequest is no longer used here.

src/ReverseProxy/Service/Proxy/HttpProxy.cs
```diff
@@ -447,7 +471,7 @@ private async Task HandleRequestFailureAsync(HttpContext context, StreamCopyHttp
             context.Response.StatusCode = StatusCodes.Status502BadGateway;
         }
 
-        private static ValueTask<bool> CopyResponseStatusAndHeadersAsync(HttpResponseMessage source, HttpContext context, HttpTransformer transformer)
+        private static async ValueTask<bool> CopyResponseStatusAndHeadersAsync(HttpResponseMessage source, HttpContext context, HttpTransformer transformer)
```
Tratcher, line 474: the async/await can be reverted to avoid extra perf overhead.

Tratcher (commented):
Minor cleanup items

src/ReverseProxy/Service/Proxy/HttpProxy.cs
alnikola: Opened the cleanup PR #1005

PR #983, opened 2021-05-06
Rename cluster sub options types to config

Tratcher:
Now that we've renamed the base object model to Route/Cluster/DestinationConfig, we're doing the same for their sub types that are currently named *Options
HealthCheckOptions -> HealthCheckConfig
ActiveHealthCheckOptions -> ActiveHealthCheckConfig
PassiveHealthCheckOptions -> PassiveHealthCheckConfig
ProxyHttpClientOptions -> ProxyHttpClientConfig
WebProxyOptions -> WebProxyConfig
RequestProxyOptions -> RequestProxyConfig
SessionAffinityOptions -> SessionAffinityConfig

Since I'm already renaming them, I'm also flattening their directory structure. I've moved non-config options types to their own relevant service folder/namespace.

I did fix one minor issue where some of the `bool?` properties were being defaulted to `false` instead of `null` by the config parsers.


src/ReverseProxy/Service/RuntimeModel/ClusterState.cs
```diff
@@ -127,11 +127,10 @@ private void UpdateDynamicStateInternal(bool force)
                         throw new InvalidOperationException("ProcessDestinationChanges must be called first.");
                     }
 
-                    if ((healthChecks?.Enabled).GetValueOrDefault())
```
Tratcher, line 130: healthChecks.Enabled was removed because this was the only pace it was used, and activeEnabled and passiveEnabled had to be checked independently anyways.

alnikola (approved):
LGTM

MihaZupan (approved):

Issue #352, opened 2020-07-28
Contracts should be serializable/deserializable using System.Text.Json


chandramouleswaran:
### Describe the bug
Contracts should be serializable/deserializable using System.Text.Json
public ProxyMatch Match { get; private set; } = new ProxyMatch();

### To Reproduce
Try to read the ReverseProxy configuration with System.Text.Json - Match object is created but has null values for all properties.

### Further technical details

- Include the version of the packages you are using - 1.0.0-preview.4.20373.1
- The platform (Linux/macOS/Windows) - Windows

samsp-msft:
Why are you trying to serialize the configuration? If what you are trying to do is to dynamically manage the routes, then that will be handled by a new API that is being built as part of #8.

Tratcher:
Though serialization may come in handy even in the #8 scenario. You've got to store the data in some format.

chandramouleswaran:
Having the feature you have pointed to is useful - in the meantime, we are serializing the dynamically discovered route and updating a JSON which is part of the config management system.

samsp-msft:
We should revisit after we have completed #8. Object Model is likely to change as part of design of #8, so we should reassess after design and see whether they are suitable for serialization.

Tratcher:
@jkotalik and @davidni, you are both successfully serializing and deserializing the models using System.Text.Json as part of your providers, correct?

The only values I'm aware of that wouldn't be serializable are:
X509Certificate2 ProxyHttpClientOptions.ClientCertificate 
Encoding ProxyHttpClientOptions.RequestHeaderEncoding

We should add a regression test for this scenario, and decide what to do about the two properties above.

Tratcher:
Other unsupported types I found in testing:
- TimeSpan - Used by health checks, request timeouts

PR #994, opened 2021-05-13
Remove client certificate config

Tratcher:
Contributes to #352 

There have been a number of concerns raised with this approach so we're going to remove it for now and reconsider it in the future.
- It's different between the json config model and the code config model
- it requires passwords in the json config
- There's no rotation support, you have to modify the cluster config to point at a new cert and trigger a reload
- Multiple config provider implementations are using System.Text.Json to serialize the code model across processes, but that doesn't make sense for X509Certificate2.
- Cert lifetime issues across config reloads. The current model requires a complex weak reference cache.

Recommended alternatives:
- Use the improved [ProxyHttpClientFactory](https://github.com/microsoft/reverse-proxy/pull/869) support to specify the client certificate in code. This way the app controls the cert lifetimes, credentials, etc..
- Certificate information can be stored in the cluster metadata collection and consumed in the ProxyHttpClientFactory.

Future proposals:
- We may add an indirection model where a cluster references a certificate by a string id that is used to resolve the certificate from a separate service that manages them. I'd expect ProxyHttpClientFactory to consume that service.


alnikola (approved):
LGTM

samsp-msft:
@Tratcher - Do we need to document how to do this now that we have removed the capability from config, or provide it in a sample (potentially the auth sample)?

Tratcher:
The workaround for directly configuring the SocketHttpHandler is here:
https://github.com/microsoft/reverse-proxy/blob/4f9e59dc6a196a42de6e6b5e34880328a3abef7f/docs/docfx/articles/proxyhttpclientconfig.md#configuring-the-http-client

Want me to update that doc to show setting a client certificate instead of 100-continue?

PR #995, opened 2021-05-13
Flow RequestHeaderEncoding as a string

Tratcher:
Contributes to #352

We have multiple providers that use System.Text.Json to serialize the config model across processes. The Encoding type doesn't serialize, so I've changed the config model to flow the encoding by name and resolve it to an instance later.


alnikola (approved):
LGTM

ManickaP (approved):
LGTM

PR #1001, opened 2021-05-15
Enable nullable on Yarp.ReverseProxy

### LINKED ISSUE #654: Enable Nullable Reference Type
Is there any plan to enable nullable reference type in v1.0?

Tratcher:
Contributes to #654

There are a few places in the config model where we indicate that required values are not nullable, even though someone could leave them out. Additionally, some are set just after construction. We declare them non-nullable so the user experience is good after the initial construction/validation step. C#10 adds a new `required` label we'll want to use in the future for many of these.


src/ReverseProxy/Abstractions/ClusterConfig.cs
```diff
@@ -15,46 +15,49 @@ public sealed record ClusterConfig
     {
         /// <summary>
         /// The Id for this cluster. This needs to be globally unique.
+        /// This field is required.
         /// </summary>
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
         public string ClusterId { get; init; }
```
WeihanLi, line 21: Maybe `public string ClusterId { get; init; }=null!;` is better? So that we could avoid `param warning` in our source code

pranavkm: Yup. In Asp.Net, we default initialize them with = `default!;`

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -60,7 +57,7 @@ public IProxyConfig GetConfig()
                 _subscription = ChangeToken.OnChange(_configuration.GetReloadToken, UpdateSnapshot);
                 UpdateSnapshot();
             }
-            return _snapshot;
+            return _snapshot!;
         }
 
         private void UpdateSnapshot()
```
pranavkm, line 63:
```suggestion
        [MemberNotNull(nameof(_snapshot))]
        private void UpdateSnapshot()
```

This should allow you to remove the forgiveness operator on line 60.

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
```diff
@@ -69,7 +66,7 @@ private void UpdateSnapshot()
             lock (_lockObject)
             {
                 Log.LoadData(_logger);
-                ConfigurationSnapshot newSnapshot = null;
+                ConfigurationSnapshot? newSnapshot = null;
```
pranavkm, line 69:
```suggestion
                ConfigurationSnapshot newSnapshot;
```

Does this work?

src/ReverseProxy/Service/Management/ProxyConfigManager.cs
```diff
@@ -89,22 +89,17 @@ public override IReadOnlyList<Endpoint> Endpoints
...
-                lock (_syncRoot)
+                if (_endpoints == null)
                 {
-                    if (_endpoints == null)
+                    lock (_syncRoot)
                     {
-                        CreateEndpoints();
+                        if (_endpoints == null)
+                        {
+                            CreateEndpoints();
```
pranavkm, line 98: You could annotate `CreateEndpoints` / `UpdateEndpoints` with `MemberNotNull(nameof(_endpoints))`

src/ReverseProxy/Configuration/ConfigurationConfigProvider.cs
Tratcher: Is there an alternative to MemberNotNull when cross compiled to 3.1?

pranavkm: You can compile the attributes in to the assembly. We do this for our ns2.0 projects: https://github.com/dotnet/aspnetcore/blob/main/src/Shared/Nullable/NullableAttributes.cs

alnikola (approved):
LGTM

PR #1008, opened 2021-05-19
Remove prohibited headers

alnikola:
Removes request and response headers designated as prohibited in the last review.

### DIFF
--- a/src/ReverseProxy/Utilities/RequestUtilities.cs
+++ b/src/ReverseProxy/Utilities/RequestUtilities.cs
@@ -42,7 +42,19 @@ namespace Yarp.ReverseProxy.Utilities
             HeaderNames.TransferEncoding,
             HeaderNames.KeepAlive,
             HeaderNames.Upgrade,
-            "Proxy-Connection"
+            "Proxy-Connection",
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -42,7 +42,19 @@ internal static bool ShouldSkipResponseHeader(string headerName)
...
+            "Proxy-Connection",
+            "Proxy-Authenticate",
+            "Proxy-Authentication-Info",
+            "Proxy-Authorization",
+            "Proxy-Features",
+            "Proxy-Instruction",
+            "Security-Scheme",
+            "ALPN",
+            "Close",
+#if NET
```
samsp-msft, line 54: With this .NET guard, does this mean we skip this stripping on 3.1?

Tratcher: Add
```
#else 
"Alt-Svc",
```

Tratcher (approved):

MihaZupan (approved):

PR #1009, opened 2021-05-19
Mark the ServiceFabric package as non-shipping

Tratcher:
We don't plan to ship the SF package with 1.0 so I've marked it as non-packable like we did for k8. We'll revisit post-1.0.

Downside: We won't ship any more previews of this package in the meantime.

Alternatives:
- Continue to ship previews and only disable this for the final 1.0 release.
- Categorize packages such that the ones we do want for 1.0 get a final release and the others stay on previews (how?).

Thoughts?


Tratcher:
Closing for now, there are other plans here.

PR #1015, opened 2021-05-21
DefaultAffinityKeyName is appended with CRC32 of cluster ID

### LINKED ISSUE #976: Affinity conflicts between clusters
The affinity flow looks like this:
request -> route -> cluster -> check affinity -> load balance -> proxy request -> establish affinity to destination -> proxy response

However, this can run into conflicts when you have multiple clusters.
request1 -> route1 -> cluster1 -> check affinity -> load balance -> proxy request -> establish affinity to destination1 -> proxy response
request2 -> route2 -> cluster2 -> check affinity [destination1 does not exist on this cluster] -> load balance -> proxy request -> establish affinity to destination2 [overrides affinity cookie] -> proxy response

And what's worse, `CookieSessionAffinityProviderOptions` is global, you can't even specify different cookie names per route.

alnikola:
DefaultAffinityKeyName is appended with CRC32 of cluster ID to avoid affinity key conflicts between clusters when a custom AffinityKeyName is not set.

Fixes #976


MihaZupan:
How come we are using CRC32 here as opposed to say SHA256? Then we don't have to think about the chance of a collision.

src/ReverseProxy/Utilities/Crc32.cs
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Yarp.ReverseProxy.Utilities
+{
+    internal class Crc32
```
MihaZupan, line 8: This implementation is weird - it looks like it was copy-pasted from C into dotnet/runtime without any thought.
For example `long` in C is `int` in C# - we're doing "CRC**32**" over 64bit values here

alnikola: The names is correct because it's using only 32 bits, but may be I can change the data type to `uint`.
It's **not** a blind copy, but simplified and improved version of [this original C# code](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/tests/System/IO/Compression/CRC.cs#L6).

alnikola:
> How come we are using CRC32 here as opposed to say SHA256?

There is a tradeoff between the chance of collision vs the header length. SHA256 would require 64 characters for hex representation whereas CRC32 needs only 8. It was [already discussed](https://github.com/microsoft/reverse-proxy/issues/976#issuecomment-842508550) that the chance of collision is reasonable small here.

MihaZupan:
If header length is a concern, we should be using more efficient encoding than base16 / shorten the prefix.


> the chance of collision is reasonable small here

A 32bit hash is not appropriate where collisions are not acceptable. In our case, a collision is an error condition.
The probability also isn't even that small since 32 bits of entropy isn't all that much.

alnikola:
Please, keep in mind, that just the fact of hash collision doesn't qualify as an error condition. The error happens only if **all** of the following statements are true:
- ClusterID CRC32 is the same for two or more clusters
- Session affinity enabled on two or more conflicting clusters
- Requests with the same affinity key go to two or more conflicting clusters

The probability of this is quite low.

MihaZupan:
This is a variation on the birthday problem, the probabilities are not negligible.
Taking the risk to save a dozen bytes on headers is not justified IMO.

src/ReverseProxy/Utilities/Crc32.cs
Tratcher:
```suggestion
    internal static class Crc32
```

src/ReverseProxy/Utilities/Crc32.cs
```diff
@@ -0,0 +1,60 @@
...
+        static Crc32()
+        {
+            _crcTable = MakeCrcTable();
+        }
+
+        // Update a running CRC with the bytes --the CRC
+        // should be initialized to all 1's, and the transmitted value
+        // is the 1's complement of the final running CRC (see the
+        // crc() routine below)).
+        public static uint UpdateCRC(uint crc, ReadOnlySpan<byte> buf)
```
Tratcher, line 22:
```suggestion
        private static uint UpdateCRC(uint crc, ReadOnlySpan<byte> buf)
```

Tratcher (commented):
What are the collision probabilities for CRC32? Something like clusters / 2^32? That seems like more than enough entropy given most apps will only have a few clusters, or hundreds for a large app.

MihaZupan:
> What are the collision probabilities for CRC32? Something like clusters / 2^32?

In general approximately `N^2 / 2^33` (yay birthday problem).

[probability table](https://preshing.com/20110504/hash-collision-probabilities/)

With non-manual config, we might see hundreds/thousands of clusters used.

As Alexander pointed out, you also have to take into account that this would only be a problem if a user ended up being routed to a pair of clusters with the collision. This probability can be countered by assuming that a non-trivial number of users will end up using the service.

The probability of a collision drops by `2^extraBits` times so we can easily choose a more reasonable number.
For example, changing
`X-Yarp-Proxy-Affinity-FFFFFFFF` to
`X-Yarp-Affinity-Base64Base64Ba`
decreases the chance of a collision by 4.503.599.627.370.496 (2^52) times while keeping the header length the same.
(first has 32 bits of entropy while second has 84)

I am not proposing we make the header 100 chars long, just that we bring the chances of random errors happening down from `will definitely happen somewhere at some point` to `meteor landing on your house` levels.

MihaZupan:
In either case, we can truncate the output of an existing hash implementation instead, to avoid bringing in logic for CRC32.

src/ReverseProxy/Service/SessionAffinity/ISessionAffinityProvider.cs
```diff
@@ -34,6 +34,7 @@ public interface ISessionAffinityProvider
         /// <param name="context">Current request's context.</param>
         /// <param name="config">Affinity config.</param>
         /// <param name="destination"><see cref="DestinationState"/> to which request is to be affinitized.</param>
-        void AffinitizeRequest(HttpContext context, SessionAffinityConfig config, DestinationState destination);
+        /// /// <param name="clusterId">Target cluster ID.</param>
```
Kahbazi, line 37:
```suggestion
        /// <param name="clusterId">Target cluster ID.</param>
```

alnikola:
OK, now I do agree with Miha that we need to increase the suffix length to reduce the collision probability. I will replace CRC32 with SHA256, but use only first 10 bytes which can be encoded in 14 characters. I will also shorten the DefaultCookieName and DefaultCustomHeaderName to be as follows:
```
DefaultCustomHeaderName = "X-Yarp-Affinity";
DefaultCookieName = ".Yarp.Affinity";
```
@Tratcher Does it look good to you?

Tratcher:
> The only reason why we're hashing at all here is to avoid large header names if clusterId is long?

The cluster name is private config data and the cookie/header name is given to the client in clear text, not a good combination. Also, the cluster name is likely to contain characters that aren't allowed in cookies or header names. Length is also a concern.

Size of the cookie is one concern, as well as cost to compute the value. The compute is mitigated by caching. The size looks like it can be mitigated by encoding.

SHA256 gets you 256 bits of data, or 32 bytes, or 4 characters in base64url encoding. At that point no truncation is needed.

>  I will also shorten the DefaultCookieName and DefaultCustomHeaderName to be as follows

👍 

Side note: In the nginx config the cookie name is a required parameter, there is no default, which would make this whole discussion moot. Is this added complexity worth it to support a default value?

MihaZupan:
> The cluster name is private config data and the cookie/header name is given to the client in clear text, not a good combination.

Int

alnikola:
>Is this added complexity worth it to support a default value?

The complexity is not large here especially if we just use the built-in SHA256. I believe having the default affinity key names is beneficial because it saves our customers' efforts on writing the config.

alnikola:
Replaced CRC32 with first 10 bytes of SHA256. Perf impact should be minimal because it is calculated once per cluster.

src/ReverseProxy/Service/SessionAffinity/BaseSessionAffinityProvider.cs
```diff
@@ -92,9 +93,18 @@ public virtual AffinityResult FindAffinitizedDestinations(HttpContext context, I
...
-        protected abstract void SetAffinityKey(HttpContext context, SessionAffinityConfig config, T unencryptedKey);
+        protected abstract void SetAffinityKey(HttpContext context, SessionAffinityConfig config, T unencryptedKey, string clusterId);
+
+        protected string GetDefaultKeyNameSuffix(string clusterId)
+        {
+            var bytes = Encoding.UTF8.GetBytes(clusterId);
+            using var hashAlgo = SHA256.Create();
+            var hash = hashAlgo.ComputeHash(bytes);
+            var hashString = Convert.ToBase64String(hash.AsSpan(0, 10));
+            return hashString.Replace('+', 'a').Replace('-', 'b').Replace('=', 'c');
```
MihaZupan, line 106: Base64 as the name suggests uses 6 bits/character.
The most efficient binary => base64 conversion is therefore achieved when using binaries with lengths multiple of 3.
As we're not concerned with encoding the full hash here, only keeping the deterministic property, you can work around this by truncating the base64 value instead of the underlying binary.

Changing
```c#
Convert.ToBase64String(hash.AsSpan(0, 10))
```
to
```c#
Convert.ToBase64String(hash.AsSpan(0, 12))[..14]
```
means you are now encoding 86 (14 * 6) bits of information with 14 characters, where as before you were encoding 80 bits (10 * 8) using 16 characters (= padding had to be applied).

If using 16 characters was intended, use
```diff
-Convert.ToBase64String(hash.AsSpan(0, 10));
+Convert.ToBase64String(hash.AsSpan(0, 12));
```
instead, which will use the same number of characters while encoding 16 more bits.

MihaZupan: Why is `hashString.Replace('+', 'a').Replace('-', 'b')` needed?

Replacing `.Replace('=', 'c')` will not be needed if using either of the above suggestions (truncating base64 / encoding 12 bytes)

alnikola: >Why is hashString.Replace('+', 'a').Replace('-', 'b') needed?

I'm not sure if they are allowed in cookie names, so replaced all non-alphanumeric characters used by base64 to safe here.

alnikola: These suffixes won't be decoded back anyways.

alnikola: >means you are now encoding 86 (14 * 6) bits of information with 14 characters, where as before you were encoding 80 bits (10 * 8) using 16 characters (= padding had to be applied).

I don't care for a correctness here, just need some more or less unique string.

alnikola: >Replacing .Replace('=', 'c') will not be needed if using either of the above suggestions (truncating base64 / encoding 12 bytes)

OK. I changed it to encode 12 bytes to avoid padding with `=`.

MihaZupan: > I'm not sure if they are allowed in cookie names, so replaced all non-alphanumeric characters used by base64 to safe here.

They are allowed, the replacements can be skipped safely.

alnikola: Replacing removed.

MihaZupan (commented):
The PR LGTM, only concern is whether we want to do this at all - https://github.com/microsoft/reverse-proxy/pull/1015#issuecomment-846453355

alnikola:
>In that case hashing the id is not appropriate either as it's just obscuring it

@MihaZupan Could you please elaborate on this? I didn't get why one-way hashing doesn't fit in here.

MihaZupan:
We're saying cluster IDs are secrets.
Exposing a SHA256 of a secret is still leaking data - it can be brute forced.

Effectively the same reason why leaking hashed passwords is still a security concern.

alnikola:
>We're saying cluster IDs are secrets.

The secrecy level can be different. Cluster names are not credentials. Moreover, we are exposing only first 12 bytes, so it's impossible to reliable reconstruct the original string.

MihaZupan:
> The secrecy level can be different. Cluster names are not credentials.

I agree, but the issue here is more of a "secret" vs "not secret".
If it's a secret, obfuscating it doesn't make its disclosure ok.

> Moreover, we are exposing only first 12 bytes, so it's impossible to reliable reconstruct the original string.

You could say the same about any fixed-size digest - there will always be infinitely many inputs that hash to the same input.
Leaking 12/20/x number of bytes instead of 32 is not much better.

alnikola:
>I agree, but the issue here is more of a "secret" vs "not secret".

It's not all or nothing. Reconstructing the original ClusterID from first 12 bytes on SHA256 hash will require enormous computation power, but the result will have very little value for a potential attacker. We should be reasonable when assessing security concerns.

MihaZupan:
> Reconstructing the original ClusterID from first 12 bytes on SHA256 hash will require enormous computation power

To crack every possible ID? Yes, virtually impossible.

To crack a majority of reasonable ones? Rather trivial - a single consumer GPU can perform billions of SHA256 hashes a second.

> but the result will have very little value for a potential attacker. We should be reasonable when assessing security concerns.

If we're saying that is acceptable, we're at security through obscurity and should just not bother.

src/ReverseProxy/Service/SessionAffinity/AffinitizeTransform.cs
```diff
@@ -25,11 +25,11 @@ public AffinitizeTransform(ISessionAffinityProvider sessionAffinityProvider)
...
         {
             var proxyFeature = context.HttpContext.GetReverseProxyFeature();
-            var options = proxyFeature.Cluster.Config.SessionAffinity;
+            var cluster = proxyFeature.Cluster.Config;
             // The transform should only be added to routes that have affinity enabled.
-            Debug.Assert(options?.Enabled ?? true, "Session affinity is not enabled");
+            Debug.Assert(cluster.SessionAffinity?.Enabled ?? true, "Session affinity is not enabled");
             var selectedDestination = proxyFeature.ProxiedDestination!;
-            _sessionAffinityProvider.AffinitizeRequest(context.HttpContext, options!, selectedDestination);
+            _sessionAffinityProvider.AffinitizeRequest(context.HttpContext, selectedDestination, cluster);
```
Tratcher, line 32: Please revert any unnecessary parameter re-ordering in this PR. No need to break implementers more than we need to.

Tratcher (commented):
Let's settle the outstanding questions in the design meeting tomorrow.

src/ReverseProxy/Service/SessionAffinity/AffinitizeTransform.cs
alnikola: Ah, OK. I just wanted to make them consistent. Because `FindAffinitizedDestinations` had `SessionAffinityConfig` as the last argument, so I decided to move it on `AffinitizeRequest`, too.

Tratcher: Ah, I assumed it was a side-effect of the refactoring. Consistency is a good enough reason to keep what you have since the API's already been broken.

karelz:
Triage: We should remove default and document the header has to be set and unique across clusters.

alnikola:
Obsolete. The issue is fixed by #1027.

PR #1016, opened 2021-05-21
Fix KubernetesIngress sample backend path in .sln

MihaZupan:
Currently VS is failing to load the project because of the path mismatch.


alnikola (approved):
LGTM

Tratcher (approved):

PR #1026, opened 2021-05-26
add cipher info to tls parser

### LINKED ISSUE #1023: Expose cipher suites from TlsFrameInfo

We've had requests to expand the information in TlsFrameInfo to include the cipher suites sent from the client. 

The ability to reject the connection based on the cipher suites is important to some reverse proxies.

cc @Tracher @wfurt

wfurt:
this is basic implementation to cover  #1023.
Adding support for server e.g. negotiated cipher should not be difficult but I was not sure if it is needed. 
We would probably add ushort/TlsCipherSuite instead of exposing it as array of size 1.

I did chat with @davidfowl about array pools but I'm not sure if that is really needed. I decided to expose the set as Memory so one can play with that later if needed. 

Note that this PR exposes the raw set e.g. can have values outside of TlsCipherSuite enum. 
That is for example 65413 in the test e.g. GOST or 255 e.g. pseudo renegotiation cipher from rfc5746

cc; @Tratcher @davidfowl 

fixes  #1023

### DIFF
--- a/src/ReverseProxy/Utilities/TlsFrameHelper.cs
+++ b/src/ReverseProxy/Utilities/TlsFrameHelper.cs
@@ -107,6 +107,7 @@ namespace Yarp.ReverseProxy.Utilities.Tls
             ServerName = 0x1,
             ApplicationProtocol = 0x2,
             Versions = 0x4,
+            CipherSuites = 0x8,
         }
 
         [Flags]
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Utilities/TlsFrameHelper.cs
```diff
@@ -397,6 +406,10 @@ private static bool TryParseClientHello(ReadOnlySpan<byte> clientHello, ref TlsF
             // Skip SessionID (max size 32 => size fits in 1 byte)
             p = SkipOpaqueType1(p);
 
+            if (options == ProcessingOptions.All || (options & ProcessingOptions.CipherSuites) == ProcessingOptions.CipherSuites)
```
Tratcher, line 409: I hear Enum.HasFlags is a lot better than it used to be?

Tratcher (approved):

src/ReverseProxy/Utilities/TlsFrameHelper.cs
wfurt: I don't know @tratcher. We have other places like this so if we want to I think we should make complete sweep as separate PR.

davidfowl: It is. It's JIT optimized now.

davidfowl (approved):

Issue #797, opened 2021-02-27
Require that routes have matching clusters


Tratcher:
Today routes and clusters are [validated](https://github.com/microsoft/reverse-proxy/blob/d9eeb548817364c7199acadd177f699781091a48/src/ReverseProxy/Service/Management/ProxyConfigManager.cs#L208-L209) individually before officially loading the model. However, this step does not validate if every route references a real cluster. When the model is rendered later it's possible that a route will be created that has no matching cluster.

Today this is allowed and then checks are done in middleware to short circuit routes without a cluster.
https://github.com/microsoft/reverse-proxy/blob/d9eeb548817364c7199acadd177f699781091a48/src/ReverseProxy/Middleware/DestinationInitializerMiddleware.cs#L32-L37

I reviewed this with @davidni and he said the expected scenarios for this (a more dynamic pipeline) never materialized.

Proposal:
- Check that each route references a valid cluster before accepting the config as valid.
- Remove the associated checks from middleware.

Kahbazi:
While we are enforcing the validation, how about to also throw an error if a Cluster has no Destination?

Tratcher:
> While we are enforcing the validation, how about to also throw an error if a Cluster has no Destination?

I'm not sure, destinations can be much more ephemeral. E.g. You may create the route and cluster, first, and then incrementally add destinations as they come online.

samsp-msft:
From an extensibility perspective, I would prefer the validation to be done as late as possible, preferably at pipeline stages that need data. That way people can add their own stages to the pipeline that may add/remove or otherwise change the data attached to an individual request.

samsp-msft:
We should discuss this in the context of extensibility around cluster selection.

PR #1027, opened 2021-05-26
Remove default AffinityKeyName values

### LINKED ISSUE #976: Affinity conflicts between clusters

alnikola:
Removes default `AffinityKeyName` values which are customer affinity header name and affinity cookie name. Now, the user must always explicitly set `SessionAffinityConfig.AffinityKeyName` property.

Fixes #976


src/ReverseProxy/Service/SessionAffinity/CustomHeaderSessionAffinityProvider.cs
```diff
@@ -30,7 +28,7 @@ protected override string GetDestinationAffinityKey(DestinationState destination
 
         protected override (string? Key, bool ExtractedSuccessfully) GetRequestAffinityKey(HttpContext context, SessionAffinityConfig config)
         {
-            var customHeaderName = config.AffinityKeyName ?? DefaultCustomHeaderName;
+            var customHeaderName = config.AffinityKeyName!;
```
MihaZupan, line 31: I assume we want to make `AffinityKeyName` on `SessionAffinityConfig` non-nullable now? That would also avoid using `!` everywhere.

alnikola: I tried this, but it didn't work due to a delayed initialization. Let me check it again, may be I manage to do it.

alnikola: OK. I did it.

alnikola: Hmm, there is issue with JSON serialization. It fails with message

>"Deserialization of reference types without parameterless constructor is not supported."

MihaZupan: We don't have ctors on other records with required properties - only using the `!` operator on the type itself

Tratcher: The pattern we've used elsewhere for required paramters is like this:
https://github.com/microsoft/reverse-proxy/blob/d6d8374664bd86b5be9a75be212bf9911df4292f/src/ReverseProxy/Abstractions/DestinationConfig.cs#L19

And then validating its presence in the config validator.

Adding constructors breaks serialization.

src/ReverseProxy/Abstractions/SessionAffinityConfig.cs
```diff
@@ -10,6 +10,12 @@ namespace Yarp.ReverseProxy.Abstractions
     /// </summary>
     public sealed record SessionAffinityConfig
     {
+
+        public SessionAffinityConfig(string affinityKeyName)
+        {
+            AffinityKeyName = affinityKeyName;
+        }
+
```
Tratcher, line 18: Remove, this breaks seralization.

src/ReverseProxy/Abstractions/SessionAffinityConfig.cs
```diff
@@ -33,7 +39,7 @@ public sealed record SessionAffinityConfig
         /// This value should be unique across clusters to avoid affinity conflicts.
         /// https://github.com/microsoft/reverse-proxy/issues/976
         /// </summary>
-        public string? AffinityKeyName { get; init; }
+        public string AffinityKeyName { get; init; }
```
Tratcher, line 42: Update the comment to indicate this is required.
```suggestion
        public string AffinityKeyName { get; init; } = default!;
```

src/ReverseProxy/Service/SessionAffinity/CustomHeaderSessionAffinityProvider.cs
alnikola: Done.

src/ReverseProxy/Abstractions/SessionAffinityConfig.cs
alnikola: Done

src/ReverseProxy/Abstractions/SessionAffinityConfig.cs
```diff
@@ -32,8 +33,9 @@ public sealed record SessionAffinityConfig
         /// The provider will give its own default if no value is set.
         /// This value should be unique across clusters to avoid affinity conflicts.
         /// https://github.com/microsoft/reverse-proxy/issues/976
+        /// This property must be set right after creating a new <see cref="SessionAffinityConfig"/> instance.
         /// </summary>
-        public string? AffinityKeyName { get; init; }
+        public string AffinityKeyName { get; init; } = default!;
```
MihaZupan, line 38: Nit: In other places we've just said "This field is required. "

https://github.com/microsoft/reverse-proxy/blob/d6d8374664bd86b5be9a75be212bf9911df4292f/src/ReverseProxy/Abstractions/ClusterConfig.cs#L16-L20

src/ReverseProxy/Service/Config/ConfigValidator.cs
```diff
@@ -329,6 +329,11 @@ private void ValidateSessionAffinity(IList<Exception> errors, ClusterConfig clus
                 errors.Add(new ArgumentException($"No matching IAffinityFailurePolicy found for the affinity failure policy name '{affinityFailurePolicy}' set on the cluster '{cluster.ClusterId}'."));
             }
 
+            if (string.IsNullOrEmpty(cluster.SessionAffinity.AffinityKeyName))
+            {
+                errors.Add(new ArgumentException($"Affinity key name set on the cluster '{cluster.ClusterId}' must not be null."));
+            }
```
MihaZupan, line 335: Not part of this PR, but this is another place where we could be doing more validation (making sure the Key is valid for use as header/cookie name).

Semi-related: #797 (early validation)

MihaZupan (approved):

src/ReverseProxy/Abstractions/SessionAffinityConfig.cs
alnikola: Fixed

PR #1030, opened 2021-05-27
Don't drop content-specific headers on requests without a body

### LINKED ISSUE #618: Investigate content headers being dropped when content-length is 0
In theory you can have a POST request with a Content-Length: 0, but still have other content headers like Content-Type which would be used for routing the request at the destination. 

In today's HttpProxy logic we avoid creating the HttpContent on the request if there's no body since HttpClient will automatically create a chunked body for any HttpContent without a specified length (and a servers treat requests like GETs with a chunked body as invalid). We may need to special case zero length bodies and provide an empty content object so that we can still forward headers like Content-Type (would HttpClient allow that?).

MihaZupan:
Fixes #618

Headers defined as `HttpHeaderType.Content` in [KnownHeaders.cs](https://github.com/dotnet/runtime/blob/c90debb5dc8ee85840bbef1a44b8502ab169f47b/src/libraries/System.Net.Http/src/System/Net/Http/Headers/KnownHeaders.cs#L28-L61) may only be present in the `HttpContentHeaders` collection.
To avoid dropping such headers when there is no request body (e.g. `Content-Length` is 0), we inject a dummy 0-length content object to hold the headers.

### DIFF
--- a/src/ReverseProxy/Utilities/RequestUtilities.cs
+++ b/src/ReverseProxy/Utilities/RequestUtilities.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Net.Http;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Primitives;
@@ -147,6 +148,13 @@ namespace Yarp.ReverseProxy.Utilities
                     Debug.Assert(added.GetValueOrDefault(), $"A header was dropped; {headerName}: {string.Join(", ", headerValues)}");
                 }
             }
+
+#if DEBUG
+            if (request.Content is EmptyHttpContent content && content.Headers.TryGetValues(HeaderNames.ContentLength, out var contentLength))
+            {
+                Debug.Assert(contentLength.Single() == "0", "An actual content should have been set");
+            }
+#endif
         }
     }
 }
--- a/src/ReverseProxy/Utilities/RequestUtilities.cs
+++ b/src/ReverseProxy/Utilities/RequestUtilities.cs
@@ -3,10 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Net.Http;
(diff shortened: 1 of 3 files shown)


test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs
```diff
@@ -448,42 +448,72 @@ public async Task ProxyAsync_UpgradableRequestFailsToUpgrade_ProxiesResponse()
...
                 {
                     Assert.Equal(new Version(2, 0), request.Version);
                     Assert.Equal(method, request.Method.Method, StringComparer.OrdinalIgnoreCase);
 
-                    Assert.Null(request.Content);
+                    // When adding content-specific headers, we will inject an EmptyHttpContent if no other content is present
+                    if (headers.Any())
+                    {
+                        Assert.NotNull(request.Content);
+                        Assert.Equal("EmptyHttpContent", request.Content.GetType().Name);
```
alnikola, line 498: Could we compare actual types here?
```suggestion
                        Assert.Equal(typeof(EmptyHttpContent), request.Content.GetType());
```

src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -101,23 +123,38 @@ internal static void AddHeader(HttpRequestMessage request, string headerName, St
                 string headerValue = value;
                 if (!request.Headers.TryAddWithoutValidation(headerName, headerValue))
                 {
+                    if (request.Content is null && _contentHeaders.Contains(headerName))
+                    {
+                        request.Content = new EmptyHttpContent();
```
Kahbazi, line 128: Is it possible to stash one instance of 
`EmptyHttpContent` and reuse that?

MihaZupan: It serves as a container for `HttpContentHeaders` that may change between requests

alnikola (approved):
LGTM

Tratcher (approved):

Issue #118, opened 2020-04-27
Design streaming request detection


davidni:
Streaming requests (e.g. duplex gRPC request) require special handling by YARP (at a minimum, e.g. remove minimum request data rate enforcement).

Today we check for gRPC content type, but this is both too broad (not all gRPC requests require streaming semantics) as well as too narrow (other protocols than gRPC on top of HTTP/2 could also require streaming semantics).

Some proposals were discussed  in https://github.com/microsoft/reverse-proxy/pull/80#discussion_r416156407

samsp-msft:
I am assuming the problem here is potential DoS attacks - keep opening connections and doing very little with each of them.
In the non proxy case of a .NET gRPC back end, presumably there is something that is looking at the type of gRPC message and changing the timeouts? Can we rely on that for the proxy - that we look for the back end server to make a determination and to close the connection if the backend rejects it?
If a client has a long running, but low traffic connection, such as the connections for mobile notifications, does that cause limits on the server or resource starvation?

rynowak:
> e.g. remove minimum request data rate enforcement

If the data rate is the problem blocking this, how is special casing gRPC not just a gaping security hole? 🤔 If an attacker can get around by the data rate enforcement by just using gRPC then that seems like it makes every gRPC service vulnerable right?

davidfowl:
This is why signalr sends pings in its protocol.

davidni:
> If an attacker can get around by the data rate enforcement by just using gRPC then that seems like it makes every gRPC service vulnerable right?

@rynowak not just gRPC. This discussion also applies to HTTP/1.1 upgraded connections. See docs for [KestrelServerLimits.MinRequestBodyDataRate](https://github.com/dotnet/aspnetcore/blob/da1a721003bf8c5affcfa5883cb301d5de803277/src/Servers/Kestrel/Core/src/KestrelServerLimits.cs#L266-L277):
```cs
/// ...
/// This limit has no effect on upgraded connections which are always unlimited.
/// ..
public MinDataRate MinRequestBodyDataRate { get; set; }
```

Tratcher:
To summarize the discussion from #80:

gRPC streaming and non-streaming requests look identical to the proxy, they use all the same headers. The backend app only knows which endpoints are supposed to be streaming because it's told that when building the endpoint. The backend app already disables request body data rates and size limits for streaming requests (just like websockets), but it still enforces them for non-streaming requests. 

[...]

PR #1032, opened 2021-05-28
Doc gRPC scenarios 

### LINKED ISSUE #952: Expose gRPC service in YARP
### Describe the bug
I've created a [sample solution](https://github.com/mohsenShakiba/GrpcyarpTest) that contains three projects:
- GrpcClient
- GrpcService
- YarpService

What I want is to expose the GrpcService Through the YarpService so that the client can send gRPC requests to YARP instead of GrpcService.
Now when I run the GrpcClient the request fails with a 502 response.
What am I doing wrong? am I missing a piece of configuration?
Is it even possible to expose gRPC endpoints in YARP?

### To Reproduce

- Clone the https://github.com/mohsenShakiba/GrpcyarpTest repo
- Run GrpcService with command: `dotnet run -p ./GrpcService/`
- Run YarpService with command: `dotnet run -p ./YarpService/`
- Run GrpcClient with command: `dotnet run -p ./GrpcClient/`

### Further technical details
GrpcService listens on:
```c#
options.ListenAnyIP(6000, listenOptions => { listenOptions.Protocols = HttpProtocols.Http2; });
options.ListenAnyIP(7000, listenOptions => { listenOptions.Protocols = HttpProtocols.Http1; });
```
And the YarpService also listens on:
```c#
options.ListenAnyIP(7500, listenOptions => { listenOptions.Protocols = HttpProtocols.Http1AndHttp2; });
options.ListenAnyIP(8000, listenOptions => { listenOptions.Protocols = HttpProtocols.Http2; });
```
Also I have a cluster that points to HTTP/2 grpc service
```json
"cluster1": {
  "Destinations": {
    "cluster2/destination1": {
      "Address": "http://localhost:6000/"
    }
  }
},
```

Tratcher:
Fixes #952
Contributes to #118

There's a lot of interest in YARP for proxying gRPC so we want an explicit doc calling it out. It doesn't require much special handling unless you want to do it without TLS.

### DIFF
--- a/docs/docfx/articles/grpc.md
+++ b/docs/docfx/articles/grpc.md
@@ -2,11 +2,13 @@
 
 ## Introduction
 
-[gRPC](https://grpc.io/) is a common protocol used to exchange messages between clients and servers. It's based on top of HTTP (mainly HTTP/2) and can be proxied through YARP. While YARP doesn't need to be aware of the message formats (proto files), you do need to make sure the right protocols are enabled for the incoming server and outgoing client.
+[gRPC](https://grpc.io/) is a language agnostic, high-performance Remote Procedure Call (RPC) framework. It's built on top of HTTP/2 and can be proxied through YARP. While YARP doesn't need to be aware of the gRPC messages, you do need to make sure the right HTTP protocol is enabled. gRPC requires HTTP/2 and gRPC calls will fail if YARP isn't correctly configured to send and receive HTTP/2 requests.
 
-## Configure the Incoming Server
+## Configure YARP's Incoming Protocols
 
-gRPC requires HTTP/2 for most scenarios. HTTP/1.1 and HTTP/2 are enabled by default on ASP.NET Core servers but https (TLS) is required to negotiate HTTP/2. HTTP/2 over http (non-TLS) is only supported on Kestrel and requires specific settings.  For details see [here](https://docs.microsoft.com/aspnet/core/grpc/aspnetcore#server-options).
+gRPC requires HTTP/2 for most scenarios. HTTP/1.1 and HTTP/2 are enabled by default on ASP.NET Core servers (YARP's front end) but they require https (TLS) for HTTP/2 so YARP needs to be listening on a `https://` URL.
+
+HTTP/2 over http (non-TLS) is only supported on Kestrel and requires specific settings.  For details see [here](https://docs.microsoft.com/aspnet/core/grpc/aspnetcore#server-options).
 
 This shows configuring Kestrel to use HTTP/2 over http (non-TLS):
 ```json
@@ -22,7 +24,7 @@ This shows configuring Kestrel to use HTTP/2 over http (non-TLS):
 }
 ```
 
-## Configure the Outgoing Client
+## Configure YARP's Outgoing Protocols
 
 YARP automatically negotiates HTTP/1.1 or HTTP/2 for outgoing proxy requests, but only for https (TLS). HTTP/2 over http (non-TLS) requires additional settings. Note outgoing protocols are independent of incoming ones. E.g. https can be used for the incoming connection and http for the outgoing one, this is called TLS termination. See [here](proxyhttpclientconfig.md#httprequest) for configuration details.
 
@@ -43,4 +45,9 @@ This shows configuring the outgoing proxy request to use HTTP/2 over http. Note 
 
 ## gRPC-Web
 
-[gRPC-Web](https://grpc.io/docs/platforms/web/basics/) is a version of gRPC that's compatible with HTTP/1.1. It can be proxied by YARP's default configuration without any special considerations.
+[gRPC-Web](https://grpc.io/docs/platforms/web/basics/) is an alternative wire-format for gRPC that's compatible with HTTP/1.1.
+
+* [`application/grpc`](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) - gRPC over HTTP/2 is how gRPC is typically used.
+* [`application/grpc-web`](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) - gRPC-Web modifies the gRPC protocol to be compatible with HTTP/1.1. gRPC-Web can be used in more places. gRPC-Web can be used by browser apps and in networks without complete support for HTTP/2. Two advanced gRPC features are not supported: client streaming and bidirectional streaming.
+
+gRPC-Web can be proxied by YARP's default configuration without any special considerations.
--- a/docs/docfx/articles/grpc.md
+++ b/docs/docfx/articles/grpc.md
@@ -1,0 +1,46 @@
+# Proxing gRPC
+
+## Introduction
+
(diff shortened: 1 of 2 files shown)


docs/docfx/articles/grpc.md
```diff
@@ -0,0 +1,46 @@
+# Proxing gRPC
+
+## Introduction
+
+[gRPC](https://grpc.io/) is a common protocol used to exchange messages between clients and servers. It's based on top of HTTP (mainly HTTP/2) and can be proxied through YARP. While YARP doesn't need to be aware of the message formats (proto files), you do need to make sure the right protocols are enabled for the incoming server and outgoing client.
+
+## Configure the Incoming Server
```
alnikola, line 7: I've a bit stumbled upon the terms "incoming server" and "outgoing client" because didn't get at the first glance that it's about YARP's components. Could we somehow clarify that?

docs/docfx/articles/grpc.md
```diff
@@ -0,0 +1,46 @@
+# Proxing gRPC
+
+## Introduction
+
+[gRPC](https://grpc.io/) is a common protocol used to exchange messages between clients and servers. It's based on top of HTTP (mainly HTTP/2) and can be proxied through YARP. While YARP doesn't need to be aware of the message formats (proto files), you do need to make sure the right protocols are enabled for the incoming server and outgoing client.
```
JamesNK, line 5:
```suggestion
[gRPC](https://grpc.io/) is a language agnostic, high-performance Remote Procedure Call (RPC) framework. It's built on top of HTTP/2 and can be proxied through YARP. While YARP doesn't need to be aware of the gRPC messages, you do need to make sure the right HTTP protocol is enabled. gRPC requires HTTP/2 and gRPC calls will fail if YARP isn't correctly configured to send and receive HTTP/2 requests.
```

JamesNK: IMO keep things simple by just talking about standard gRPC over HTTP/2. That's what most people use. Mention gRPC-Web under its own section.

docs/docfx/articles/grpc.md
```diff
@@ -0,0 +1,46 @@
...
+    "cluster1/destination1": {
+      "Address": "http://localhost:6000/"
+    }
+  }
+},
+```
+
+## gRPC-Web
+
+[gRPC-Web](https://grpc.io/docs/platforms/web/basics/) is a version of gRPC that's compatible with HTTP/1.1. It can be proxied by YARP's default configuration without any special considerations.
```
JamesNK, line 46:
```suggestion
[gRPC-Web](https://grpc.io/docs/platforms/web/basics/) is an alternative wire-format for gRPC that's compatible with HTTP/1.1.

* [`application/grpc`](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) - gRPC over HTTP/2 is how gRPC is typically used.
* [`application/grpc-web`](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) - gRPC-Web modifies the gRPC protocol to be compatible with HTTP/1.1. gRPC-Web can be used in more places. gRPC-Web can be used by browser apps and in networks without complete support for HTTP/2. Two advanced gRPC features are no longer supported: client streaming and bidirectional streaming.

gRPC-Web can be proxied by YARP's default configuration without any special considerations.
```

JamesNK: Copied and pasted some content from https://docs.microsoft.com/en-us/aspnet/core/grpc/supported-platforms?view=aspnetcore-5.0#wire-formats

I'm guessing people who are configuring YARP won't necessarily know which type of gRPC is going through the proxy, so including the content-type's and specs gives them enough info to figure out by looking at logs.

docs/docfx/articles/grpc.md
```diff
@@ -0,0 +1,46 @@
+# Proxing gRPC
+
+## Introduction
+
+[gRPC](https://grpc.io/) is a common protocol used to exchange messages between clients and servers. It's based on top of HTTP (mainly HTTP/2) and can be proxied through YARP. While YARP doesn't need to be aware of the message formats (proto files), you do need to make sure the right protocols are enabled for the incoming server and outgoing client.
+
+## Configure the Incoming Server
+
+gRPC requires HTTP/2 for most scenarios. HTTP/1.1 and HTTP/2 are enabled by default on ASP.NET Core servers but https (TLS) is required to negotiate HTTP/2. HTTP/2 over http (non-TLS) is only supported on Kestrel and requires specific settings.  For details see [here](https://docs.microsoft.com/aspnet/core/grpc/aspnetcore#server-options).
```
samsp-msft, line 9:
```suggestion
gRPC requires HTTP/2 for most scenarios. HTTP/1.1 and HTTP/2 are enabled by default in Kestrel (YARPs front end) but HTTP/2 requires https (TLS) so YARP needs to be listening on a https:// URL. 

HTTP/2 over http (non-TLS) is only supported on Kestrel and requires specific settings.  For details see [here](https://docs.microsoft.com/aspnet/core/grpc/aspnetcore#server-options).
```

docs/docfx/articles/grpc.md
samsp-msft, line 7:
```suggestion
## Configure the YARP Endpoint
```

docs/docfx/articles/grpc.md
```diff
@@ -0,0 +1,46 @@
...
+      "http": {
+        "Url": "http://localhost:5000",
+        "Protocols": "Http2"
+      }
+    }
+  }
+}
+```
+
+## Configure the Outgoing Client
```
samsp-msft, line 25:
```suggestion
## Configure the Outgoing Connection
```

samsp-msft (changes requested):
I think this doc also requires info on timeouts, and that enabling gRPC opens up the potential for attacks using long running requests.

Tratcher:
> I think this doc also requires info on timeouts, and that enabling gRPC opens up the potential for attacks using long running requests.

@samsp-msft I'm holding off on that part until we resolve #118. I expect to update the doc afterwards.

docs/docfx/articles/grpc.md
Tratcher: Why eliminate the other servers? We have people using them.

docs/docfx/articles/grpc.md
Tratcher: How about `Configure YARP's Incoming Protocols`?

docs/docfx/articles/grpc.md
```diff
@@ -0,0 +1,53 @@
+# Proxing gRPC
+
+## Introduction
+
+[gRPC](https://grpc.io/) is a language agnostic, high-performance Remote Procedure Call (RPC) framework. It's built on top of HTTP/2 and can be proxied through YARP. While YARP doesn't need to be aware of the gRPC messages, you do need to make sure the right HTTP protocol is enabled. gRPC requires HTTP/2 and gRPC calls will fail if YARP isn't correctly configured to send and receive HTTP/2 requests.
+
+## Configure YARP's Incoming Protocols
+
+gRPC requires HTTP/2 for most scenarios. HTTP/1.1 and HTTP/2 are enabled by default on ASP.NET Core servers (YARP's front end) but they require https (TLS) for HTTP/2 so YARP needs to be listening on a `https://` URL.
```
samsp-msft, line 9: do we want to point at the URLs property or the other one (i get confused as there are two)

samsp-msft (approved):

docs/docfx/articles/grpc.md
Tratcher: It's different in each server so it's better not to be that specific.

Issue #462, opened 2020-10-08
Organize files


Tratcher:
The current folder hierarchy is a side-effect of merging multiple assemblies and some organic growth. It's confusing to find things and could use some restructuring. One example is that some interfaces are adjacent to their implementations and some aren't. A shallower structure might also be easier to work with.

I don't anticipate this would affect product code, but we might choose to align namespaces with the new structure. Wait until after we complete some of the big features in flight this milestone.

alnikola:
Summary:

0. DON'T DO IT while there are major PRs open (like health check).
1. Flatten project structure.
2. Keep interfaces and implementation together, so we group things by domains not types.
3. Verify and fix namespaces after the project structure compaction
4. Top level folders ideas
   4.1 Discovery
      4.1.1 Abstractions (code config)
      4.1.2 Serializable (Configuration/Contract currently, *Data types)
   4.2 Model (currently RuntimeModel)
   4.3 Routing
   4.4 Load balancing
   4.5 Health
   4.6 Proxy
   4.7 Session affinity
   4.8 Transforms
   4.9 Management (all config-related services go here)
   4.10 Utilities (Signals, timers and other current utilities)

ManickaP:
### My questions
Mostly about things you didn't even talk about

#### Documentation
- Article file names: underscores (https://github.com/microsoft/reverse-proxy/blob/master/docs/docfx/articles/getting_started.md) vs. dashes (https://github.com/microsoft/reverse-proxy/blob/master/docs/docfx/articles/direct-proxying.md) vs. nothing (https://github.com/microsoft/reverse-proxy/blob/master/docs/docfx/articles/proxyhttpclientconfig.md)
- Markdown docs file names camel-case (https://github.com/microsoft/reverse-proxy/blob/master/docs/DailyBuilds.md) seems like all are camel-case :+1: 
- Readme files uppercase (https://github.com/microsoft/reverse-proxy/blob/master/docs/designs/README.md) vs. lowercase (https://github.com/microsoft/reverse-proxy/blob/master/docs/docfx/readme.md) vs. camel-case (https://github.com/microsoft/reverse-proxy/blob/master/src/ReverseProxy/Signals/Readme.md)
- Random readme files in sources (https://github.com/microsoft/reverse-proxy/blob/master/src/ReverseProxy/Signals/Readme.md), should this rather be in articles?
  - Can we document the new structure/namespaces with readme files? Can this be than picked up by docfx and used in API reference docs?

[...]

PR #1035, opened 2021-05-29
Move files, change namespaces

Tratcher:
Contributes to #462

- Organizes files in Yarp.ReverseProxy
  - Flatter directory structure, limit 2 deep
  - Organized as outlined in the issue.
  - Split or merged a few files to match the overall organization.
- Updated namespaces
- Moved test files to match product layout

Overall there are now fewer and shorter namespaces.

TODO:
- Other feedback from #462 like docs and tests (Future PR)

In hindsight I should have done only the file moves first, updating namespaces is painful.

I also shouldn't have started this before taking a long weekend 😁. Feel free to continue and push directly to this branch.

As for reviewing, let's focus on the directory structure first, since that drives the namespaces. Is anything horribly out of place? Do the categories still make sense?


MihaZupan:
Where does the `Discovery` namespace come from? It just sounds unfamiliar to me compared to `Abstractions`, `Configuration`, etc.

src/ReverseProxy/Health/EntityActionScheduler.cs
```diff
@@ -8,7 +8,7 @@
 using System.Threading.Tasks;
 using Yarp.ReverseProxy.Utilities;
 
-namespace Yarp.ReverseProxy.Service.Management
+namespace Yarp.ReverseProxy.Health
```
alnikola, line 11: Not sure if `Health` is the best place for `EntityActionScheduler` since it's functionality is quite generic. Would it make sense to move it into `Yarp.ReverseProxy.Management` namespace?

src/ReverseProxy/LoadBalancing/AppBuilderLoadBalancingExtensions.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Yarp.ReverseProxy.LoadBalancing;
+
+namespace Microsoft.AspNetCore.Builder
```
alnikola, line 6: Why it's inside `Microsoft....` namespace, but not in `Yarp.ReverseProxy.....` namespace? I thought we should put all YARP code into `Yarp.ReverseProxy.....` namespace.

src/ReverseProxy/Health/AppBuilderHealthExtensions.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Yarp.ReverseProxy.Health;
+
+namespace Microsoft.AspNetCore.Builder
```
alnikola, line 6: Why it's inside `Microsoft....` namespace, but not in `Yarp.ReverseProxy.....` namespace? I thought we should put all YARP code into `Yarp.ReverseProxy.....` namespace.

src/ReverseProxy/Health/IClusterDestinationsUpdater.cs
```diff
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Yarp.ReverseProxy.RuntimeModel;
+using Yarp.ReverseProxy.Model;
 
-namespace Yarp.ReverseProxy.Service.HealthChecks
+namespace Yarp.ReverseProxy.Health
```
alnikola, line 6: Would it make sense to move `IClusterDestinationsUpdater` in `Yarp.ReverseProxy.Management` namespace?

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -6,14 +6,13 @@
...
-using Yarp.ReverseProxy.Service.Proxy.Infrastructure;
+using Yarp.ReverseProxy.Discovery;
+using Yarp.ReverseProxy.Discovery.Configuration;
+using Yarp.ReverseProxy.Management;
+using Yarp.ReverseProxy.Proxy;
+using Yarp.ReverseProxy.Routing;
+using Yarp.ReverseProxy.Transforms.Builder;
 using Yarp.ReverseProxy.Utilities;
 
 namespace Microsoft.Extensions.DependencyInjection
```
alnikola, line 18: This namespace starts with `Microsoft....` too, not sure if it's correct.

src/ReverseProxy/Model/IReverseProxyApplicationBuilder.cs
```diff
@@ -1,9 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using Microsoft.AspNetCore.Builder;
-
-namespace Yarp.ReverseProxy.Middleware
+namespace Microsoft.AspNetCore.Builder
```
alnikola, line 4: This namespace starts with `Microsoft....` too, not sure if it's correct.

src/ReverseProxy/Model/ReverseProxyApplicationBuilder.cs
```diff
@@ -7,7 +7,7 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 
-namespace Yarp.ReverseProxy.Middleware
+namespace Yarp.ReverseProxy.Model
```
alnikola, line 10: `ReverseProxyApplicationBuilder` is used only for configuration. Would it make sense to move it into some special namespace? In my opinion, `Yarp.ReverseProxy.Model` is for the types we use in runtime, not configuration.

src/ReverseProxy/Routing/ReverseProxyConventionBuilder.cs
```diff
@@ -4,11 +4,10 @@
...
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.AspNetCore.Builder;
-using Yarp.ReverseProxy.Abstractions;
-using Yarp.ReverseProxy.RuntimeModel;
+using Yarp.ReverseProxy.Discovery;
+using Yarp.ReverseProxy.Model;
 
-namespace Yarp.ReverseProxy
+namespace Microsoft.AspNetCore.Builder
```
alnikola, line 10: Same question about namespace.

alnikola (commented):
Looks good overall besides a few questions.

src/ReverseProxy/Health/EntityActionScheduler.cs
Tratcher: It's only used by health checks today. It's internal so we can move it later if we want.

alnikola: OK

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
Tratcher: The convention here is to put extension methods in the same namespace as the type being extended so they're easy to discover.

src/ReverseProxy/Health/IClusterDestinationsUpdater.cs
Tratcher: Would you also recommend moving IAvailableDestinationsPolicy with it? The two should stay together.

alnikola: Yes. Let's move `IAvailableDestinationsPolicy`, too.

Tratcher: I'd put them into Health to mirror that the config for AvailableDestinationsPolicy is also part of HeathCheckConfig.

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
alnikola: OK. I see, then I will resolve all comments regarding `Microsoft` namespaces.

src/ReverseProxy/Health/IClusterDestinationsUpdater.cs
alnikola: Ah, yeah, correct. Let it be inside `Health` since we put config there.

Tratcher:
> Where does the `Discovery` namespace come from? It just sounds unfamiliar to me compared to `Abstractions`, `Configuration`, etc.

`Discovery` comes from implementers discovering the routes, clusters, and destinations for their environment. That's what the SF and K8 implementations do.

`Abstractions` is more commonly used for types people are supposed to implement, we don't have as many of those.
`Configuration` is relevant here, but overloaded since a common data source is IConfiguration.

We can cover this in today's API review.

src/ReverseProxy/Model/ReverseProxyApplicationBuilder.cs
Tratcher: It used to be in the middleware namespace but we got rid of that. All of the other components that are used by the middleware are here in model so I put it there.

Tratcher:
API review:
- Discovery -> Configuration
- Discovery.Configuration -> Configuration.ConfigProvider
- IReverseProxyBuilderExtensions namespace -> Management, internal. no longer extension methods?
- Yarp.ReverseProxy.ServiceFabric (package and namespace) -> Yarp.ServiceFabric 
- Yarp.ReverseProxy.Kubernetese.* (package and namespace) -> Yarp.Kubernetese.*

Tratcher:
Updated with API review feedback. Please prioritize this review so we can minimize merge conflicts.

src/ReverseProxy/Proxy/RequestUtilities.cs
```diff
@@ -9,15 +9,33 @@
...
     /// <summary>
     /// APIs that can be used when transforming requests.
     /// </summary>
     public static class RequestUtilities
     {
+        /// <summary>
+        /// Converts the given HTTP method (usually obtained from <see cref="HttpRequest.Method"/>)
+        /// into the corresponding <see cref="HttpMethod"/> static instance.
+        /// </summary>
+        public static HttpMethod GetHttpMethod(string method) => method switch
```
MihaZupan, line 24: Was making this public intentional?

MihaZupan (approved):
I'm very glad we're getting rid of the nesting and splitting of domains across folders.

Nit: Given that `IClusterChangeListener` is the only public thing in the `Management` namespace, could it fit under Configuration/Model instead? What kind of other functionality would we put into `Management`?

alnikola (approved):
LGTM

Tratcher:
> I'm very glad we're getting rid of the nesting and splitting of domains across folders.
> 
> Nit: Given that `IClusterChangeListener` is the only public thing in the `Management` namespace, could it fit under Configuration/Model instead? What kind of other functionality would we put into `Management`?

Moved to Model since this deals with ClusterState objects.

src/ReverseProxy/Proxy/RequestUtilities.cs
Tratcher: No, it moved from an internal class to a public one. Fixed.

PR #1036, opened 2021-05-31
Only TE:trailers is allowed when proxying from 1.1 to 2 request version

### LINKED ISSUE #969: [HTTP/2] Filter TE header
In HTTP/2 the [TE header field is only accepted](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.2) if the `trailers` value is set. We need to filter out all prohibited combinations.

alnikola:
If the incoming to proxy request has the protocol "HTTP/1.1" and contains the `TE` header whereas the outgoing to downstream request has the protocol "HTTP/2", then `TE` header is proxied only if it has the single `trailers` value, otherwise it's removed.

Fixes #969

### DIFF
--- a/src/ReverseProxy/Utilities/RequestUtilities.cs
+++ b/src/ReverseProxy/Utilities/RequestUtilities.cs
@@ -27,7 +27,7 @@ namespace Yarp.ReverseProxy.Utilities
 
             if (headerName.Equals(HeaderNames.TE, StringComparison.OrdinalIgnoreCase)
                 && ProtocolHelper.IsHttp11(httpContext.Request.Protocol)
-                && proxyRequest.Version == ProtocolHelper.Http11Version
+                && proxyRequest.Version == ProtocolHelper.Http2Version
                 && httpContext.Request.Headers.TryGetValue(HeaderNames.TE, out var teValues))
             {
                 var containsTrailers = false;
@@ -40,10 +40,10 @@ namespace Yarp.ReverseProxy.Utilities
                     else
                     {
                         // Unexpected TE header's value found.
-                        return false;
+                        return true;
                     }
                 }
-                return containsTrailers;
+                return !containsTrailers;
             }
 
             // Filter out HTTP/2 pseudo headers like ":method" and ":path", those go into other fields.
--- a/src/ReverseProxy/Utilities/RequestUtilities.cs
+++ b/src/ReverseProxy/Utilities/RequestUtilities.cs
@@ -18,7 +18,7 @@ namespace Yarp.ReverseProxy.Utilities
     /// </summary>
     public static class RequestUtilities
     {
-        internal static bool ShouldSkipRequestHeader(string headerName, StringValues headerValue, HttpContext httpContext, HttpRequestMessage proxyRequest)
+        internal static bool ShouldSkipRequestHeader(string headerName, StringValues headerValue, Version destinationProtocol)
         {
             if (_headersToExclude.Contains(headerName))
             {
@@ -26,8 +26,7 @@ namespace Yarp.ReverseProxy.Utilities
             }
 
             if (headerName.Equals(HeaderNames.TE, StringComparison.OrdinalIgnoreCase)
-                && ProtocolHelper.IsHttp11(httpContext.Request.Protocol)
-                && proxyRequest.Version == ProtocolHelper.Http2Version)
+                && destinationProtocol == ProtocolHelper.Http2Version)
             {
                 return headerValue.Count != 1 || !string.Equals(headerValue[0], "trailers", StringComparison.OrdinalIgnoreCase);
             }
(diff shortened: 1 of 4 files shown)


src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -18,13 +18,34 @@ namespace Yarp.ReverseProxy.Utilities
...
+                    {
+                        containsTrailers = true;
+                    }
+                    else
+                    {
+                        // Unexpected TE header's value found.
+                        return false;
+                    }
+                }
+                return containsTrailers;
```
MihaZupan, line 46: This check looks inverted - we are deciding if the header should be removed.

alnikola: Fixed. Thanks.

src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -18,13 +18,34 @@ namespace Yarp.ReverseProxy.Utilities
...
+                    {
+                        containsTrailers = true;
+                    }
+                    else
+                    {
+                        // Unexpected TE header's value found.
+                        return true;
+                    }
+                }
+                return !containsTrailers;
```
MihaZupan, line 46: Right now, this will allow values with multiple "trailers" (`TE: trailers trailers trailers`). Likely not problematic though.

We make the assumption that `teValues` is a `string[]` in [`HttpTransformer`](https://github.com/microsoft/reverse-proxy/blob/a0e449f42e0b662adc9f7f940588fccee2a6fa0d/src/ReverseProxy/Service/Proxy/HttpTransformer.cs#L113-L114). Perhaps this could be

```c#
Debug.Assert(teValues is string[]);
var values = teValues as string[] ?? teValues.ToArray();
return values.Length != 1 || !string.Equals(values[0], "trailers", StringComparison.OrdinalIgnoreCase);
```

MihaZupan (approved):

src/ReverseProxy/Utilities/RequestUtilities.cs
alnikola: I just realized that I don't need to get the header value here again because HttpTransformer already does it. Will change PR.

alnikola: @MihaZupan I changed this logic which now uses `headerValue` passed as an argument and `StringValue`'s indexer. Please, check it again.

MihaZupan: Looks good 👍

src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -18,13 +18,20 @@ namespace Yarp.ReverseProxy.Utilities
...
+        internal static bool ShouldSkipRequestHeader(string headerName, StringValues headerValue, HttpContext httpContext, HttpRequestMessage proxyRequest)
         {
             if (_headersToExclude.Contains(headerName))
             {
                 return true;
             }
 
+            if (headerName.Equals(HeaderNames.TE, StringComparison.OrdinalIgnoreCase)
+                && ProtocolHelper.IsHttp11(httpContext.Request.Protocol)
+                && proxyRequest.Version == ProtocolHelper.Http2Version)
```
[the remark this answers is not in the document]
alnikola: I think if somebody configured YARP to go over `HTTP/2` protocol to the destination, they would expect YARP to prepare a valid `HTTP/2` request regardless of what will be the actual protocol

src/ReverseProxy/Utilities/RequestUtilities.cs
```diff
@@ -18,13 +18,20 @@ namespace Yarp.ReverseProxy.Utilities
...
-        internal static bool ShouldSkipRequestHeader(string headerName)
+        internal static bool ShouldSkipRequestHeader(string headerName, StringValues headerValue, HttpContext httpContext, HttpRequestMessage proxyRequest)
         {
             if (_headersToExclude.Contains(headerName))
             {
                 return true;
             }
 
+            if (headerName.Equals(HeaderNames.TE, StringComparison.OrdinalIgnoreCase)
+                && ProtocolHelper.IsHttp11(httpContext.Request.Protocol)
```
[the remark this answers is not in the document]
alnikola: Removed

src/ReverseProxy/Utilities/RequestUtilities.cs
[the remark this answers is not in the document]
Tratcher: HTTP/2 is YARP's default, they usually aren't configuring that.

Tratcher:
I've re-read the TE spec and found this part:
https://datatracker.ietf.org/doc/html/rfc7230#section-4.3

>   Since the TE header field only applies to the immediate connection, a
   sender of TE MUST also send a "TE" connection option within the
   Connection header field (Section 6.1) in order to prevent the TE
   field from being forwarded by intermediaries that do not support its
   semantics.

Since we remove all per-connection headers including the Connection header then we should also remove the TE header without considering the protocol version. That makes this much simpler.

alnikola:
>Since we remove all per-connection headers including the Connection header then we should also remove the TE header without considering the protocol version. That makes this much simpler.

Agree. I will create a new PR because the change will be much simpler.

alnikola:
Opened new PR #1050

PR #1038, opened 2021-05-31
Test Expect: 100-continue when body is not sent

### LINKED ISSUE #617: Test 100-continue scenarios

alnikola:
Fixes #617

### DIFF
--- a/test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs
+++ b/test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs
@@ -1645,6 +1645,33 @@ namespace Yarp.ReverseProxy.Service.Proxy.Tests
         }
 
         [Theory]
+        [InlineData("HTTP/1.1", "1.1")]
+        [InlineData("HTTP/2", "2.0")]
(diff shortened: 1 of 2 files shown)


MihaZupan (approved):

test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs
```diff
@@ -1644,6 +1644,33 @@ public async Task ProxyAsync_WithHttpClient_Fails()
                 destinationPrefix, httpClient, requestOptions, transforms));
         }
 
+        [Theory]
+        [InlineData("HTTP/1.1", "1.1")]
+        [InlineData("HTTP/2", "2.0")]
+        public async Task ProxyAsync_Expect100ContinueWithFailedResponse_ReturnResponse(string fromProtocol, string toProtocol)
+        {
+            var httpContext = new DefaultHttpContext();
+            httpContext.Request.Method = "GET";
```
Tratcher, line 1653: This request doesn't make much sense, it's a GET with no body headers (Content-Length or Transfer-Encoding) and a data stream?

PR #1040, opened 2021-06-01
Response buffering in IIS is disabled by default

### LINKED ISSUE #533: Proxying SSE in IIS buffers
Websocket protocol support, add SignalR support

So the proxy has support for the actual websocket protocol not just the inital http handshake.

First of, I'm a bit of a noob when it comes to websockets. So please bare with me.
So I'd need a long living websocket connection.(for events send from the server)

So from what I could gather from the source code this proxy has websockets supported (or at least the handshake).
The initial handshake goes over HTTP. When the client gets an OK from the server a websocket connection is established. 
So correct me if I'm wrong, but the connection is established between the client and server directly without the proxy.
Because the proxy only handles http requests and not websocket protocol.

Looks like SignalR has some features that could come in handy.
Like managing multiple connections.

TLDR: Is there a way to proxy the actual websocket protocol? Is this even  good practice?

alnikola:
Disables write buffering of response in IIS when sending it back to the client. Having buffering enabled breaks Server Side Events scenario. Buffering can be enabled again via the new config parameter `RequestProxyConfig.AllowResponseBuffering`.

Fixes #533


alnikola:
Is `EnableResponseBuffering` good name for this parameter?

MihaZupan:
Does this have an impact on perf? If so, we should consider disabling it in the benchmark app.

alnikola:
Not sure, but it's now disabled by default so the benchmark app will see it.

alnikola:
The main purpose of this PR is to support SSE (server side events). They are now broken in some cases by this buffering.

src/ReverseProxy/Service/Proxy/RequestProxyConfig.cs
```diff
@@ -36,6 +36,12 @@ public sealed record RequestProxyConfig
         public HttpVersionPolicy? VersionPolicy { get; init; }
 #endif
 
+        /// <summary>
+        /// Enables write buffering in Kestrel when sending a response back to the client.
+        /// NOTE: enabling it can break SSE (server side event) scenarios.
+        /// </summary>
+        public bool? EnableResponseBuffering { get; init; }
```
Tratcher, line 43: This feature no-ops on Kestrel
https://github.com/dotnet/aspnetcore/blob/6427d9cc718f8093c506b62b6fd12544411b477f/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs#L303
It only does something on IIS (where this bug was reported)
https://github.com/dotnet/aspnetcore/blob/52eff90fbcfca39b7eb58baad597df6a99a542b0/src/Servers/IIS/IIS/src/Core/IISHttpContext.FeatureCollection.cs#L491

test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs
```diff
@@ -1383,6 +1383,39 @@ public async Task ProxyAsync_ResponseBodyCancelledAfterStart_Aborted()
             events.AssertContainProxyStages(hasRequestContent: false);
         }
 
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        [InlineData(null)]
+        public async Task ProxyAsync_ResponseBodyDisableBuffering_Success(bool? enableBuffering)
```
Tratcher, line 1390: Were you able to repro the original issue and verify this fix manually?

This test verifies that DisableResponseBuffering was called, but it doesn't confirm that was enough for the scenario to succeed.

src/ReverseProxy/Service/Proxy/RequestProxyConfig.cs
Tratcher: Should this be AllowResponseBuffering? Setting it to true doesn't turn on buffering everywhere, it avoids turning something off that's only available sometimes anyways.

test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs
alnikola: No, I didn't test E2E scenario because as far as I understood the root cause was reliably proved in the original issue. Do you have any doubts about this fix?

Tratcher: I wouldn't consider the issue resolved without at least a manual test confirming the E2E.

Tratcher: Put another way, we're sure this change is required, but we want to verify there are not any other issues.

alnikola: Ok. I got it.

davidfowl:
The entire description here seems off. There's no response buffering in kestrel by default.

alnikola:
>The entire description here seems off. There's no response buffering in kestrel by default.

Sorry, I misunderstood that the issue is relevant to IIS only. Will fix the description.

src/ReverseProxy/Service/Proxy/RequestProxyConfig.cs
alnikola: Renamed.

alnikola:
I manually check this PR with YARP hosted by IIS Express and it does fixes SSE issue.

Tratcher (approved):

kikaragyozov:
@Tratcher I'm sorry for the necro, but I figured this was the best place to ask this as I was tackling issues with SSE on a live IIS server. Everything worked as expected in IIS Express/.NET Aspire, but the moment it got hosted in production, SSE stopped working.

After investigating, the Dynamic Response Caching module was at fault here.

Even though I've configured the following in my `web.config`, the issue persisted:

```
	  <httpCompression>
		<dynamicTypes>
			<add mimeType="text/event-stream" enabled="false" />
		</dynamicTypes>
	  </httpCompression>
```

The only solution was to disable it via `HttpContext!.Features.Get<IHttpResponseBodyFeature>()!.DisableBuffering()`.

Could you please shine some light as to why the `web.config` approach didn't work, like, at all?
After further inspecting, it seems the dynamic compression module for IIS was not even installed on the system, so I'm not sure what's causing it to buffer responses when by default it does not do that.

image

PR #1041, opened 2021-06-01
Using options.HasFlag

danirzv:
as mentioned in https://github.com/microsoft/reverse-proxy/pull/1026#discussion_r640079973 used options.HasFlag.
also corrected `ProcessingOptions.All` to `0x7FFFFFFF`
fixes https://github.com/microsoft/reverse-proxy/pull/1026

### DIFF
--- a/src/ReverseProxy/Utilities/TlsFrameHelper.cs
+++ b/src/ReverseProxy/Utilities/TlsFrameHelper.cs
@@ -107,7 +107,7 @@ namespace Yarp.ReverseProxy.Utilities.Tls
             ApplicationProtocol = 0x2,
             Versions = 0x4,
             CipherSuites = 0x8,
-            All = 0x15,
+            All = 15,
         }
 
         [Flags]
--- a/src/ReverseProxy/Utilities/TlsFrameHelper.cs
+++ b/src/ReverseProxy/Utilities/TlsFrameHelper.cs
@@ -492,7 +492,7 @@ namespace Yarp.ReverseProxy.Utilities.Tls
 
                 ReadOnlySpan<byte> extensionData = extensions.Slice(0, extensionLength);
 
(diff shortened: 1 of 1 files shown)


src/ReverseProxy/Utilities/TlsFrameHelper.cs
```diff
@@ -103,11 +103,11 @@ public static class TlsFrameHelper
         [Flags]
         public enum ProcessingOptions
         {
-            All = 0,
             ServerName = 0x1,
             ApplicationProtocol = 0x2,
             Versions = 0x4,
             CipherSuites = 0x8,
+            All = 15,
```
Tratcher, line 110: I think this was intentional (though I agree it was unusual). @wfurt?

wfurt: If we add new option, `All` will need to change as well, right? While adding to enum is OK, changing value later may be seen as breaking change. I don't know if that matters for reverse-proxy. In he past, I also saw use of 0xff (or equivalent) for `All` to solve the same problem.
It seems like there is no place where we would depend on default value e.g. 0 so using specific value seems ok.

danirzv: I've changed it to 0x7FFFFFFF

Tratcher: It's a bit odd to have an enum without a 0 value, usually it's "None". If there is no scenario for None, then All might be the better 0 value. You moved All from 0 so you wouldn't have to check it separately in HasFlags, right?

danirzv: Yup!

Tratcher (approved):

Tratcher:
Thanks

PR #1045, opened 2021-06-02
Fix Expect 100-continue test

### LINKED ISSUE #617: Test 100-continue scenarios

alnikola:
Test now uses `POST` verb and sets `Content-Length` to make sure the body can be transferred under normal conditions.

Fixes #617


test/ReverseProxy.Tests/Service/Proxy/HttpProxyTests.cs
```diff
@@ -1650,10 +1650,12 @@ public async Task ProxyAsync_WithHttpClient_Fails()
...
         {
             var httpContext = new DefaultHttpContext();
-            httpContext.Request.Method = "GET";
+            httpContext.Request.Method = "POST";
             httpContext.Request.Protocol = fromProtocol;
             httpContext.Request.Headers[HeaderNames.Expect] = "100-continue";
-            using var contentStream = new MemoryStream(Encoding.UTF8.GetBytes(new string('a', 1024 * 1024 * 10)));
+            var content = Encoding.UTF8.GetBytes(new string('a', 1024 * 1024 * 10));
+            httpContext.Request.Headers[HeaderNames.ContentLength] = content.Length.ToString();
+            using var contentStream = new MemoryStream();
```
Tratcher, line 1658: The content is missing 😁.

Also add an assert in the mock client that HttpRequestMessage has an HttpContent instance.

alnikola: Fixed. Assert added.

Tratcher (approved):

PR #1046, opened 2021-06-02
Fix Path syntax link

alnikola:


MihaZupan (approved):

Tratcher (approved):

Kahbazi:
Should the language be removed from other links too?

Tratcher:
Using language and version neutral links is preferable. Do you see a few more that should be fixed?

Kahbazi:
Here's a PR removing languages from all links in docs. https://github.com/microsoft/reverse-proxy/pull/1047

PR #1048, opened 2021-06-02
Added summaries and cleaned up some text

samsp-msft:


docs/docfx/articles/transforms.md
```diff
@@ -1,9 +1,11 @@
-# Transforms
+# Header Transforms
```
Tratcher, line 1: Transforms are not restricted to headers, they can manipulate other fields on the request/response like protocol versions, urls, etc.

docs/docfx/articles/transforms.md
```diff
@@ -1,9 +1,11 @@
...
 
 Introduced: preview2
 
 ## Introduction
-When proxying a request it's common to modify parts of the request or response to adapt to the destination server's requirements or to flow additional data such as the client's original IP address. This process is implemented via Transforms. Types of transforms are defined globally for the application and then individual routes supply the parameters to enable and configure those transforms. The original request objects are not modified by these transforms, only the proxy requests.
+When proxying a request it's common to modify parts of the request or response headers to adapt to the destination server's requirements or to flow additional data such as the client's original IP address. This process is implemented via Transforms. Types of transforms are defined globally for the application and then individual routes supply the parameters to enable and configure those transforms. The original request objects are not modified by these transforms, only the proxy requests.
```
Tratcher, line 6: Here too, it's not just headers.

docs/docfx/articles/transforms.md
```diff
@@ -31,12 +33,12 @@ X-Forwarded-Proto: http
...
 
-Transforms fall into a few categories: request, response, and response trailers. Request and response body transforms are not supported by YARP but you can write middleware to do this. Request trailers are not supported because they are not supported by the underlying HttpClient.
-
+## Transform Categories
 
+Transforms fall into a few categories: request, response, and response trailers.  Request trailers are not supported because they are not supported by the underlying HttpClient.
 Transforms can be added to routes either through configuration or programmatically.
 
-## From Configuration
+### From Configuration
```
Tratcher, line 41: Why make these a sub-bullet of Transform Categories? The comment about adding them through config or programmatically is unrelated to categories.

docs/docfx/articles/transforms.md
```diff
@@ -149,6 +151,8 @@ The following are built in transforms identified by their primary config key. Th
 
 ### PathPrefix
 
+**Modifies to the request path adding a prefix value**
```
Tratcher, line 154:
```suggestion
**Modifies the request path adding a prefix value**
```

docs/docfx/articles/transforms.md
```diff
@@ -216,6 +224,8 @@ This will set the request path with the given value.
 
 ### PathPattern
 
+**Replaces segment of the request path using substitution patterns**
```
Tratcher, line 227:
```suggestion
**Replaces the request path using a pattern template**
```

docs/docfx/articles/transforms.md
```diff
@@ -280,6 +292,8 @@ Example:
 
 ### QueryRouteParameter
 
+**Adds or replaces a query string parameter with a value from the route configurtaion**
```
Tratcher, line 295:
```suggestion
**Adds or replaces a query string parameter with a value from the route configuration**
```

docs/docfx/articles/transforms.md
```diff
@@ -464,6 +490,8 @@ This removes the named header.
 
 ### X-Forwarded
 
+**Selects which headers should be prefixed with `X-Forwarded`**
```
Tratcher, line 493:
```suggestion
**Adds headers with information about the original client request**
```

docs/docfx/articles/transforms.md
```diff
@@ -521,6 +549,8 @@ The {Prefix}Prefix header value is taken from `HttpContext.Request.PathBase`. Th
 
 ### Forwarded
 
+**Selects which headers should be prefixed with `Forwarded`**
```
Tratcher, line 552:
```suggestion
**Adds a header with information about the original client request**
```

docs/docfx/articles/transforms.md
```diff
@@ -599,8 +631,7 @@ Example:
 ```
 X-Client-Cert: SSdtIGEgY2VydGlmaWNhdGU...
 ```
-
-This transform causes the client certificate taken from `HttpContext.Connection.ClientCertificate` to be Base64 encoded and set as the value for the given header name. This is needed because client certificates from incoming connections are not used when making connections to the destination server. The destination server may need that certificate to authenticate the client. There is no standard that defines this header and implementations vary, check your destination server for support.
+As the inbound and outbound connections are independent, and the proxy may authenticate with the destination, there needs to be a way to pass any inbound client certificate to the destination server. This transform causes the client certificate taken from `HttpContext.Connection.ClientCertificate` to be Base64 encoded and set as the value for the given header name. The destination server may need that certificate to authenticate the client. There is no standard that defines this header and implementations vary, check your destination server for support.
```
Tratcher, line 634: We couldn't re-use the client cert when communicating with the destination regardless, we're sharing destination connections across multiple clients.
```suggestion
As the inbound and outbound connections are independent, there needs to be a way to pass any inbound client certificate to the destination server. This transform causes the client certificate taken from `HttpContext.Connection.ClientCertificate` to be Base64 encoded and set as the value for the given header name. The destination server may need that certificate to authenticate the client. There is no standard that defines this header and implementations vary, check your destination server for support.
```

docs/docfx/articles/transforms.md
samsp-msft: Grr, it feels a bit vague without that.

samsp-msft: Isn't the URL and verb still technically a header?

docs/docfx/articles/transforms.md
samsp-msft: Because the last line of categories talks about config & code. It breaks it up a bit.

docs/docfx/articles/transforms.md
Tratcher: In HTTP/2 they're psudo headers, in HTTP/1.x they're not headers at all.

Tratcher: How about "Request and Response Transforms"?

docs/docfx/articles/transforms.md
Tratcher: You're missing an addition section heading then, those two lines were unrelated.
```
## Adding transforms

Transforms can be added to routes either through configuration or programmatically.

### From Configuration

...
```

samsp-msft:
Triage: @samsp-msft - rebase and cleanup. (yes, I am sending triage notes to myself)

samsp-msft:
@tratcher - this should be ready to merge now!

Tratcher (approved):

PR #1053, opened 2021-06-03
Remove unused test certificates

MihaZupan:
#994 removed certs from config and the related tests.
This PR deletes the unused certificates and updates the path of `ConfigurationConfigProviderTests.cs` in CredScan suppressions since it was moved in #1035


.config/CredScanSuppressions.json
```diff
@@ -5,22 +5,13 @@
...
-                "test/TestCertificates/https-rsa.key"
+                "test/TestCertificates/testCert.pfx"
             ]
         },
         {
             "_justification": "Unit test dummy credentials",
             "file": [
-                "test/ReverseProxy.Tests/Configuration/CertificateConfigLoaderTests.cs",
-                "test/ReverseProxy.Tests/Configuration/ConfigurationConfigProviderTests.cs"
+                "test/ReverseProxy.Tests/Configuration/ConfigProvider/ConfigurationConfigProviderTests.cs"
```
Tratcher, line 14: Does this still need any exclusions?

Tratcher (approved):

.config/CredScanSuppressions.json
MihaZupan: It's hitting against https://github.com/microsoft/reverse-proxy/blob/4bd1fb82ff1391b4df566c68407c9b87e2ca1c78/test/ReverseProxy.Tests/Configuration/ConfigProvider/ConfigurationConfigProviderTests.cs#L227

I assume this section isn't needed after #994 anymore at all? I'll remove it

Tratcher: Right, that's obsolete.

MihaZupan: Removed

PR #1059, opened 2021-06-07
Packages referenced by YARP

### LINKED ISSUE #996: Package references and their licenses should be clearly documented
YARP itself does not reference other packages, but the Kubernetes ingress and some samples do. We should make sure that those are called out in the readme for each part, and that the licenses are listed.

Close compliance item [1272006](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1272006) when complete.

alnikola:
Fixes #996

### DIFF
--- a/docs/docfx/articles/packages-refs.md
+++ b/docs/docfx/articles/packages-refs.md
@@ -7,6 +7,7 @@ Packages installed to all projects.
 Microsoft.SourceLink.AzureRepos.Git |  Apache-2.0
 Microsoft.SourceLink.GitHub |  Apache-2.0
 XliffTasks* |  MIT
+
 \* - installed to all projects excluding the ones ending with "*Tests"
 
 ## docfx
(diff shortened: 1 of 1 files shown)


Tratcher:
"We should make sure that those are called out in the readme for each part" I don't think this was supposed to go into the conceptual docs, but rather into the project source code folders.

samsp-msft:
I'm not opposed to having this as a top level document - In some ways the samples (where most of the breadth comes from) are self documenting in that they have the references listed.

Can this be converted into a table so it's easier to read, and grouped by project?

Tratcher (approved):

Kahbazi:
@alnikola Does this also need to be added to [Table of Content](https://github.com/microsoft/reverse-proxy/blob/main/docs/docfx/articles/toc.yml)?

alnikola:
@Kahbazi, good point. I will open a PR.

PR #1064, opened 2021-06-09
Proxy -> Forwarder API review

Tratcher:
The naming of the types in this namespace was highly redundant, hopefully this is better?

@MihaZupan can you take a look at the Telemetry classes to see if the naming makes sense, and if we need more changes?


src/ReverseProxy/Forwarder/IHttpForwarder.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Forward an HTTP request to a chosen destination.
+    /// </summary>
+    public interface IHttpForwarder
+    {
+        /// <summary>
+        /// Forwards the incoming request to the destination server, and the response back to the client.
+        /// </summary>
+        /// <param name="context">The HttpContent to forward.</param>
```
alnikola, line 18:
```suggestion
        /// <param name="context">The HttpContext to forward.</param>
```

test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
```diff
@@ -20,17 +20,17 @@
 using Xunit;
 using Yarp.ReverseProxy.Common.Tests;
 
-namespace Yarp.ReverseProxy.Proxy.Tests
+namespace Yarp.ReverseProxy.Forwarder.Tests
 {
-    public class HttpProxyTests
+    public class HttpForwarderTests
```
alnikola, line 25: Should we rename all test methods which are currently starting with "ProxyAsync_*"?

alnikola (approved):
LGTM

src/ReverseProxy/Forwarder/ForwarderTelemetry.cs
```diff
@@ -48,23 +48,23 @@ public void ProxyStop(int statusCode)
...
         }
 
         [Event(4, Level = EventLevel.Informational)]
-        public void ProxyStage(ProxyStage stage)
+        public void ProxyStage(ForwarderStage stage)
         {
             if (IsEnabled(EventLevel.Informational, EventKeywords.All))
             {
-                Debug.Assert(sizeof(ProxyStage) == sizeof(int), "Backing type of ProxyStage MUST NOT be changed");
+                Debug.Assert(sizeof(ForwarderStage) == sizeof(int), "Backing type of ProxyStage MUST NOT be changed");
```
MihaZupan, line 67:
```suggestion
                Debug.Assert(sizeof(ForwarderStage) == sizeof(int), "Backing type of ForwarderStage MUST NOT be changed");
```

MihaZupan (commented):
The proxy telemetry is really `IHttpProxy` telemetry, so we would then want to rename everything there too
-  [`ProxyStage`](https://github.com/microsoft/reverse-proxy/blob/f6a268782a3dc8678536c95fe9ddfbacb8f076a3/src/ReverseProxy.TelemetryConsumption/ProxyStage.cs#L9) (telemetry has a copy since the enum is internal) => `ForwarderStage`
- `IProxyTelemetryConsumer` => `IForwarderTelemetryConsumer`
- Same for the methods on `IProxyTelemetryConsumer`
  - On**Proxy** Start/Stop/Failed/Stage => On`Forwarder`
- `ProxyMetrics` => `ForwarderMetrics`
- `IProxyMetricsConsumer.OnProxyMetrics` => `IForwarderMetricsConsumer.OnForwarderMetrics`

If you want, I can put up a follow-up PR for telemetry consumption after this.

docs/docfx/articles/http-client-config.md
```diff
@@ -1,12 +1,12 @@
...
+Each [Cluster](xref:Yarp.ReverseProxy.Configuration.ClusterConfig) has a dedicated [HttpMessageInvoker](https://docs.microsoft.com/dotnet/api/system.net.http.httpmessageinvoker) instance used to proxy requests to its [Destination](xref:Yarp.ReverseProxy.Configuration.DestinationConfig)s. The configuration is defined per cluster. On YARP startup, all clusters get new `HttpMessageInvoker` instances, however if later the cluster configuration gets changed the [IForwarderHttpClientFactory](xref:Yarp.ReverseProxy.Forwarder.IForwarderHttpClientFactory) will re-run and decide if it should create a new `HttpMessageInvoker` or keep using the existing one. The default `IForwarderHttpClientFactory` implementation creates a new `HttpMessageInvoker` when there are changes to the [HttpClientConfig](xref:Yarp.ReverseProxy.Configuration.HttpClientConfig).
```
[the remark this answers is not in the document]
Tratcher: We'll want to change a lot of them, but not all. I'll do a quick pass for this file and direct-forwarding. I think we want to use `forward/er` when talking about this specific component or step, but `proxy` when talking about the general scenario and the library as a whole. E.g. YARP is the proxy with affinity, load balancing, etc., and the last step is to forward the request.

Tratcher:
Merging to avoid any more conflicts. Let me know if there's any additional follow-up needed.

Kahbazi:
@Tratcher Does this method also need to rename to `GetProxyErrorFeature`?

https://github.com/microsoft/reverse-proxy/blob/16ab855a5be980a0cf5fd21e9a7028190b6e7dda/src/ReverseProxy/Model/HttpContextFeaturesExtensions.cs#L49-L52

Tratcher:
@Kahbazi thanks, I missed that one. Do you want to handle it?

Tratcher:
Fixed in https://github.com/microsoft/reverse-proxy/pull/1087

Issue #1068, opened 2021-06-09
StreamCopyHttpContent's Started check is not thread safe


Tratcher:
See https://github.com/dotnet/runtime/issues/53914

https://github.com/microsoft/reverse-proxy/blob/961608d01adc9ab07d4cd3f5adf7d4d3dfff3f0f/src/ReverseProxy/Proxy/StreamCopyHttpContent.cs#L118-L125

Under some retry conditions in SocketsHttpHandler SerializeToStreamAsync can be called twice. There seems to be a race condition in SocketsHttpHandler where those calls can happen in parallel, and right now the Started check is not threads safe. That can lead to errors down the stack from parallel reads on SslStream.

Mitigation: Make the Started check thread safe.

Fix: Investigate why SocketsHttpHandler can call SerializeToStreamAsync multiple times in parallel (https://github.com/dotnet/runtime/issues/53914).

karelz:
Triage: Hit in production of IslandGateway - we should address it.

PR #1071, opened 2021-06-10
Make the Started check in StreamCopyHttpContent thread safe

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/1068

### DIFF
--- a/src/ReverseProxy/Proxy/StreamCopyHttpContent.cs
+++ b/src/ReverseProxy/Proxy/StreamCopyHttpContent.cs
@@ -68,7 +68,10 @@ namespace Yarp.ReverseProxy.Proxy
         /// <see cref="HttpClient.SendAsync(HttpRequestMessage, HttpCompletionOption, CancellationToken)"/>
         /// completes, even when using <see cref="HttpCompletionOption.ResponseHeadersRead"/>.
         /// </remarks>
-        public bool Started { get; private set; }
+        public bool Started => Volatile.Read(ref _started) == 1;
+
+        private int _started;
+
 
         /// <summary>
         /// Copies bytes from the stream provided in our constructor into the target <paramref name="stream"/>.
@@ -117,12 +120,10 @@ namespace Yarp.ReverseProxy.Proxy
         // would break bidirectional streaming scenarios.
         protected override async Task SerializeToStreamAsync(Stream stream, TransportContext? context)
         {
-            if (Started)
+            if (Interlocked.CompareExchange(ref _started, 1, 0) != 0)
             {
                 throw new InvalidOperationException("Stream was already consumed.");
             }
-
-            Started = true;
 
             if (_autoFlushHttpClientOutgoingStream)
             {


src/ReverseProxy/Proxy/StreamCopyHttpContent.cs
```diff
@@ -117,13 +120,11 @@ public StreamCopyHttpContent(Stream source, bool autoFlushHttpClientOutgoingStre
         // would break bidirectional streaming scenarios.
         protected override async Task SerializeToStreamAsync(Stream stream, TransportContext? context)
         {
-            if (Started)
+            if (Interlocked.CompareExchange(ref _started, 1, 0) != 0)
```
MihaZupan, line 123:
```suggestion
            if (Interlocked.Exchange(ref _started, 1) != 0)
```

src/ReverseProxy/Proxy/StreamCopyHttpContent.cs
```diff
@@ -68,7 +68,10 @@ public StreamCopyHttpContent(Stream source, bool autoFlushHttpClientOutgoingStre
         /// <see cref="HttpClient.SendAsync(HttpRequestMessage, HttpCompletionOption, CancellationToken)"/>
         /// completes, even when using <see cref="HttpCompletionOption.ResponseHeadersRead"/>.
         /// </remarks>
-        public bool Started { get; private set; }
+        public bool Started => Volatile.Read(ref _started) == 1;
```
MihaZupan, line 71: Nit: one test is using `== 1`, the other `!= 0`
```suggestion
        public bool Started => Volatile.Read(ref _started) != 0;
```

MihaZupan (approved):
Thanks!

src/ReverseProxy/Proxy/StreamCopyHttpContent.cs
Kahbazi: I wanted to be consistent with here:
https://github.com/microsoft/reverse-proxy/blob/f6a268782a3dc8678536c95fe9ddfbacb8f076a3/src/ReverseProxy/Health/EntityActionScheduler.cs#L52
https://github.com/microsoft/reverse-proxy/blob/f6a268782a3dc8678536c95fe9ddfbacb8f076a3/src/ReverseProxy/Health/EntityActionScheduler.cs#L74

Should I change both?

src/ReverseProxy/Proxy/StreamCopyHttpContent.cs
Tratcher: @MihaZupan is right, only the exchange needs to be thread safe. The comparison ends up being thread safe because Exchange returns the old value and only one thread can ever return 0.

src/ReverseProxy/Proxy/StreamCopyHttpContent.cs
Tratcher: Oh, I see where you copied EntityActionScheduler. Yes, that one can also use Exchange and == 1.

PR #1072, opened 2021-06-10
Alnikola/test nonvalid headers

alnikola:


src/ReverseProxy/Proxy/HttpTransformer.cs
```diff
@@ -68,11 +68,19 @@ public virtual ValueTask TransformRequestAsync(HttpContext httpContext, HttpRequ
         public virtual ValueTask<bool> TransformResponseAsync(HttpContext httpContext, HttpResponseMessage proxyResponse)
         {
             var responseHeaders = httpContext.Response.Headers;
+#if NET6_0_OR_GREATER
+            CopyResponseHeaders(proxyResponse.Headers.NonValidated, responseHeaders);
```
stephentoub, line 72: What is the signature of CopyResponseHeaders?  Given that it already existed and is being used in the else block, I assume it's in terms of `IEnumerable<string>`?  That's going to both box the enumerable and make interface calls for all of the MoveNext/Current calls.  It'd be better to have a loop that uses the strongly-typed struct enumerator.

stephentoub: (Nevermind, I see you updated it as well below.)

src/ReverseProxy/Proxy/HttpTransformer.cs
```diff
@@ -110,8 +125,15 @@ private static void CopyResponseHeaders(HttpHeaders source, IHeaderDictionary de
                     continue;
                 }
 
+#if NET6_0_OR_GREATER
```
Tratcher, line 128: Appending one value at a time is still inefficient. It ends up here:
https://github.com/dotnet/aspnetcore/blob/6a929c3796f048d619bdd05969972d3a84669aac/src/Http/Http.Abstractions/src/Internal/ParsingHelpers.cs#L161

You'd want append all the values in one operation.

stephentoub: The 99% case should only have one entry, even if there are multiple values.  Unless something else has caused it to parse, this will yield a single string containing the entire value portion of the header, e.g. if the header was "a: b, c, d", this will be a single string "b, c, d".

alnikola: Got it. But I believe in the GET stress test scenario I measured most headers are single-valued.

Issue #1067, opened 2021-06-09
docfx doesn't recognize Microsoft.* namespaces in xrefs


alnikola:
Build produces "invalid cross reference" warnings for the following references:
| File | Ref type |
| :--- | :--- |
middleware.md | Microsoft.AspNetCore.Builder.ReverseProxyIEndpointRouteBuilderExtensions.MapReverseProxy
dests-health-checks.md | Microsoft.AspNetCore.Builder.ProxyMiddlewareAppBuilderExtensions
transforms.md | Microsoft.Extensions.DependencyInjection.ReverseProxyServiceCollectionExtensions

Tratcher:
Similar to https://github.com/microsoft/reverse-proxy/issues/829.

karelz:
Triage:
- We should check if generate documentation for these types (API docs)
- If we do, then check with docfx if it is a known issue.

Keeping it in 1.0 to decide what is broken where and having a plan (e.g. we won't block on fix in docfx).

karelz:
Triage:
- Remaining issues only in OperatorFramework - @alnikola please paste the list here
- hen we can put it to Backlog milestone

alnikola:
Remaining `Found project reference without a matching metadata reference` warnings:
```
[21-09-16 05:25:09.791]Warning:[MetadataCommand.ExtractMetadata](C:/repos/reverse-proxy/src/OperatorFramework/examples/BasicOperator/BasicOperator.csproj)Workspace failed with: [Warning] Found project reference without a matching metadata reference: C:\repos\reverse-proxy\src\OperatorFramework\src\Core\Microsoft.Kubernetes.Core.csproj
[21-09-16 05:25:09.791]Warning:[MetadataCommand.ExtractMetadata](C:/repos/reverse-proxy/src/OperatorFramework/examples/BasicOperator/BasicOperator.csproj)Workspace failed with: [Warning] Found project reference without a matching metadata reference: C:\repos\reverse-proxy\src\OperatorFramework\src\CustomResource\Microsoft.Kubernetes.CustomResources.csproj
[21-09-16 05:25:09.791]Warning:[MetadataCommand.ExtractMetadata](C:/repos/reverse-proxy/src/OperatorFramework/examples/BasicOperator/BasicOperator.csproj)Workspace failed with: [Warning] Found project reference without a matching metadata reference: C:\repos\reverse-proxy\src\OperatorFramework\src\Controller\Microsoft.Kubernetes.Controller.csproj
[...]

PR #1075, opened 2021-06-11
Update docfx to 2.58

### LINKED ISSUE #829: xref links from Articles to ref docs are not working.
### Describe the bug
x:ref links for example the ones to Cluster and Route in https://microsoft.github.io/reverse-proxy/articles/configfiles.html are not linking correctly.

MihaZupan:
Fixes #829

### DIFF
--- a/docs/docfx/docfx.csproj
+++ b/docs/docfx/docfx.csproj
@@ -11,6 +11,6 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="docfx.console" Version="2.52.0" />
+        <PackageReference Include="docfx.console" Version="2.58.0" />
     </ItemGroup>
 </Project>


alnikola (approved):
LGTM

Tratcher:
Did you see if it helped with #1067?

Tratcher (approved):

MihaZupan:
The build log is so noisy it's hard to tell

Build output before/after (you can see the diff but it's not that useful)
https://gist.github.com/MihaZupan/c8e6019d2db138b0fd2f0a3f84f71f65

I still see a bunch of `invalid cross reference` warnings Alexander mentioned

Tratcher:
Some of those look like broken links that can be fixed. We should do a cleanup pass after the renames are done to make sure.

MihaZupan:
We can look at all the warnings as part of #1067

Issue #1079, opened 2021-06-12
YARP request queuing or latency


mgnoonan:
### Describe the bug

My use case is to setup YARP as a drop-in replacement for 3 of our API services, which are migrating to APIM/app services in Azure. This allows our internal clients to maintain existing configuration, but still get data from our services in Azure. The flow is:
`YARP -> Express Route -> APIM -> Region 1/Region 2 app services`

I have been conducting some load tests, and I see some behavior that I cannot explain. Looking at the screen cap from App Insights, I see the proxy call begin, but then nothing happens for ~22s, then the call hits Azure and completes in ~1.2s, but the overall time taken is 24s.
![image](https://user-images.githubusercontent.com/3255096/121784891-a6d48800-cb84-11eb-98aa-bdd6e31518db.png)

Looking at the test run overall, the timings on the YARP proxy servers looks good to start, but then hits a wall and suddenly the slowest calls regularly start to hit the 30s timeout (as set in the config).
![image](https://user-images.githubusercontent.com/3255096/121784936-e69b6f80-cb84-11eb-82ad-e4e026ae0638.png)

I do not see any corresponding slow responses in the downstream calls, so I don't see how this is a reflection of problems with other systems. It seems more like the requests are being queued, or there is some kind of throttling or connection latency that kicks in during the test.

Do I need to set the MaxConnectionsPerServer? Any insights you have are greatly appreciated. Love the project, I just need to get over this hurdle.

### To Reproduce

I can't provide the exact steps as this is a performance test customized for my API and built using VS Test. The target goals are 1200 req/sec, and the automation ramps up the number of users to 27k over the course of the test.

Here is a redacted version of my appSettings.json:
```json
"ReverseProxy": {
    "Routes": {
      "route-instore": {
        "ClusterId": "cluster1",
        "Match": {
          "Path": "{**catch-all}",
          "Hosts": [ "appservice1" ]
        },
        "Transforms": [
          { "PathPrefix": "/svc1" }
        ]
      },
      "route-auth": {
        "ClusterId": "cluster1",
        "Match": {
          "Path": "{**catch-all}",
          "Hosts": [ "appservice2" ]
        },
        "Transforms": [
          { "PathRemovePrefix": "/api" },
          { "PathPrefix": "/svc2" }
        ]
      },
      "route-api": {
        "ClusterId": "cluster1",
        "Match": {
          "Path": "{**catch-all}",
          "Hosts": [ "appservice3" ]
        }
      }
    },
[...]

PR #1086, opened 2021-06-14
Pass EventSource name via attribute

MihaZupan:
There is a very subtle difference between the two:
- The current version (passing the name in ctor) uses the public `EventSource` constructor, which sets the `EtwSelfDescribingEventFormat` option.
- The new version (attribute) uses the empty ctor, which is a protected one that sets `EtwManifestEventFormat` instead.

There are some subtle differences (for example bools are encoded with 1 vs 4 bytes).

This PR makes us consistent with what all other `EventSource`s do.

`dotnet trace` appears to have problems working with `EtwSelfDescribingEventFormat`, which can be observed as garbage data for events that include a bool.
image

cc: @alnikola this is what you have been seeing in the past

Related to #1079

Opened an issue in diagnostics regarding `dotnet trace`: dotnet/diagnostics#2367

### DIFF
--- a/src/ReverseProxy/Forwarder/ForwarderTelemetry.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderTelemetry.cs
@@ -8,9 +8,10 @@ using System.Threading;
 
 namespace Yarp.ReverseProxy.Forwarder
 {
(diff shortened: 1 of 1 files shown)


alnikola (approved):
LGTM
Thanks for fixing it!

PR #1106, opened 2021-06-22
K8s sample fix and cleanups

### LINKED ISSUE #1092: K8s ingress sample: Missing Cluster Id exception
I started to learn `Yarp` to see how we could use it for `OrchardCore`. And i'm a beginner on k8s, so i followed the `kubernetes-ingress.md` article.

### Describe the bug

One step of the article is `kubectl apply -f ingress-definition.yaml` but this yaml file doesn't exist, so I replaced it with `-f ingress-controller.yaml`. Then another step is `kubectl apply -f .\backend.yaml` but this yaml file doesn't exist, so I created it to define a backend deployment and service.

Then the `yarp-proxy` was failing on `The proxy config is invalid. (Missing Cluster Id.)`. In the below logs we can see that in the `Cluster` property of the message an `Id` is passed, but it was expecting a `ClusterId`. So, my workaround was to rename in `ClusterConfig` the `ClusterId` prop to `Id`, and rebuild the image, then it was working.

 https://github.com/microsoft/reverse-proxy/blob/f75df4d7ec6fb23c8f7af53e853e6fe9ed02cf6e/src/ReverseProxy/Configuration/ClusterConfig.cs#L20

### To Reproduce

Follow the `kubernetes-ingress.md` article.

Part of the logs in the `yarp-proxy` pod.

    [14:09:27 INF] Received Update for default:minimal-ingress
    [14:09:27 INF] {"MessageType":"Update","Key":"default:minimal-ingress",
      "Routes":[{"RouteId":"/","Match":{"Methods":null,"Hosts":null,"Path":"/{**catch-all}","Headers":null},"Order":null,"ClusterId":"10.1.0.108","AuthorizationPolicy":null,"CorsPolicy":null,"Metadata":null,"Transforms":null}],
      "Cluster":[{"Id":"10.1.0.108","LoadBalancingPolicy":null,"SessionAffinity":null,"HealthCheck":null,"HttpClient":null,"HttpRequest":null,"Destinations":{"http://10.1.0.108:8080":{"Address":"http://10.1.0.108:8080","Health":null,"Metadata":null}},"Metadata":null}]}
    [14:09:27 INF] Update
    [14:09:27 ERR] Failed to apply the new config.
    System.AggregateException: The proxy config is invalid. (Missing Cluster Id.)
    ---> System.ArgumentException: Missing Cluster Id.
    --- End of inner exception stack trace ---
    at Yarp.ReverseProxy.Management.ProxyConfigManager.ApplyConfigAsync(IProxyConfig config) in /src/src/ReverseProxy/Management/ProxyConfigManager.cs:line 217
    at Yarp.ReverseProxy.Management.ProxyConfigManager.ReloadConfigAsync() in /src/src/ReverseProxy/Management/ProxyConfigManager.cs:line 187

### Further technical details

- I cloned the repository 2 days ago.
- The platform: Windows

jtkech:
Fixes #1092 

- K8s ingress article updated.

- Docker files updated.

- The main fix to make the sample working was to replace in `DispatchController.cs`

      this:  return new Task<IActionResult>(() => new DispatchActionResult(...));
      with:  return Task.FromResult<IActionResult>(new DispatchActionResult(...));

    Then I made a new image that i pushed on my registry (see below), and used it.

- Some other little cleanups.

- For now the `ingress-controller.yaml` uses the `jtkech/yarp-controller:1.0.0` image (including the above fix) in place of the `jkotalik/yarp-controller:0.0.0-19c9e0d733` image, let me know if I need to replace it e.g. with.

      <REGISTRY_NAME>/yarp-controller:<TAG>


samples/KuberenetesIngress.Sample/Ingress/ingress-controller.yaml
```diff
@@ -125,7 +125,7 @@ spec:
       containers:
       - name: yarp-controller
         imagePullPolicy: Always
-        image: jkotalik/yarp-controller:0.0.0-19c9e0d733
+        image: jtkech/yarp-controller:1.0.0
```
Tratcher, line 128: Sorry, for security reasons we can't check in a reference your image here. It would be better to leave this as `[placeholder]` and add instructions for updating it.

Tratcher (commented):
Thanks for the help.

samples/KuberenetesIngress.Sample/Ingress/ingress-controller.yaml
jtkech: Okay no problem

Tratcher (approved):

PR #1119, opened 2021-06-28
Remove headers with new line characters

MihaZupan:
Added filtering of headers with new line characters to `RequestUtilities.AddHeader`.
Users can still force such headers be sent in custom transforms by calling `HttpRequestMessage.Headers.TryAddWithoutValidation`.

Fixed the prohibited headers test that was using `Name: Value` instead of `Name` when asserting that headers are present.

### DIFF
--- a/src/ReverseProxy/Forwarder/RequestUtilities.cs
+++ b/src/ReverseProxy/Forwarder/RequestUtilities.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
+using System.Runtime.CompilerServices;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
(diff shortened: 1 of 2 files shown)


alnikola (approved):
LGTM

src/ReverseProxy/Forwarder/RequestUtilities.cs
```diff
@@ -155,6 +154,26 @@ internal static void AddHeader(HttpRequestMessage request, string headerName, St
...
+                {
+                    AddHeader(request, headerName, string.Join("; ", headerValues));
+                    return;
+                }
+
+                foreach (var headerValue in headerValues)
+                {
+                    if (ContainsNewLines(headerValue))
+                    {
+                        // TODO: Log
```
Tratcher, line 172: I thought you were planning to replace the characters and add the header anyways?

Tratcher (commented):
Keep an eye on the benchmarks after merging this, scanning each header will have an impact.

src/ReverseProxy/Forwarder/RequestUtilities.cs
MihaZupan: Given that IIS/Kestrel will drop these headers/reject whole requests, I didn't think it would add much value

MihaZupan:
> Keep an eye on the benchmarks after merging this, scanning each header will have an impact.

A citrine run shows a ~0.7 % regression.

The cost is partially mitigated here by moving the cookie check out of the fast path

Tratcher (approved):

MihaZupan:
A second run shows a 0.15% improvement so it's all close to error margins

PR #1122, opened 2021-06-30
Rebase release/latest on 1.0.0-preview12

alnikola:
release/latest contains 1 commit which is not in main, so I cannot simply reset it to the head of 1.0.0-preview12. Instead, I opened this PR.


MihaZupan:
It may be easier to cherry pick that 1 commit into main and preview 12 branches, so that we don't have to do this every time.

The preview12 and release/latest should also have the updated package versions in samples.

PR #1126, opened 2021-06-30
Merge Preview12 into Release/latest

MihaZupan:


Tratcher:
Was this merge squashed? That puts release/latest in a strange state and looses commit history. Please do another PR that merges from e0c76114 (the preview12 label) into release/latest with a merge commit instead of a squash and merge.

PR #1130, opened 2021-06-30
Revert "Merge Preview12 into Release/latest"

alnikola:
Reverts microsoft/reverse-proxy#1126


Tratcher (approved):
For context, this is because the original PR was squashed.

PR #1135, opened 2021-07-01
Use RequestHeaderRemoveKey for transformation

BennyM:
Currently you are unable to use RequestHeaderRemove transformation since the wrong key is used.

### DIFF
--- a/test/ReverseProxy.Tests/Transforms/RequestHeadersTransformExtensionsTests.cs
+++ b/test/ReverseProxy.Tests/Transforms/RequestHeadersTransformExtensionsTests.cs
@@ -65,6 +65,16 @@ namespace Yarp.ReverseProxy.Transforms.Tests
             ValidateRequestHeader(append, builderContext);
         }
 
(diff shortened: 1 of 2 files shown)


test/ReverseProxy.Tests/Transforms/RequestHeadersTransformExtensionsTests.cs
```diff
@@ -65,6 +65,16 @@ public void AddRequestHeader(bool append)
...
 
+        [Fact]
+        public void RemoveRequestHeader()
+        {
+            var routeConfig = new RouteConfig();
+            routeConfig = routeConfig.WithTransformRequestHeaderRemove("MyHeader");
+
+            ValidateAndBuild(routeConfig, _factory);
+            
+        }
```
alnikola, line 76:
```suggestion
            ValidateAndBuild(routeConfig, _factory);   
        }
```

alnikola (approved):
LGTM

PR #1142, opened 2021-07-06
Fix off setting description in do X-Forwarded doc

Tornhoof:
In previous versions it was the explicit `""`string, by now it is `"Off"` as referenced multiple times above and below.


Tratcher (approved):

Tratcher:
Thanks

PR #1166, opened 2021-07-26
Pass cluster directly in Round Robin load balancing policy

Kahbazi:


src/ReverseProxy/LoadBalancing/RoundRobinLoadBalancingPolicy.cs
```diff
@@ -22,7 +22,7 @@ internal sealed class RoundRobinLoadBalancingPolicy : ILoadBalancingPolicy
                 return null;
             }
 
-            var counter = _counters.GetOrCreateValue(context.GetClusterState());
```
Tratcher, line 25: It looks like this was the last call to GetClusterState. Can you remove it?

Kahbazi: Done.

Tratcher (approved):

Tratcher:
Thanks

Kahbazi:
@Tratcher Does this PR need the breaking-change label?

Tratcher:
I guess that was a public API we removed 😁. Applied.

franklixuefei:
We were calling `context.GetClusterState()`. What should we call now?

Issue #1172, opened 2021-07-30
IHttpForwarder extensions for basic request transforms


Tratcher:
Doing any transforms with IHttpForwarder requires creating a derived HttpTransformer instance.

Here's a comparison with go:
https://gist.github.com/davidfowl/cf68de7da1f6cb4d7dcbedd6e1a9c6a4

We have some extension methods already:
https://github.com/microsoft/reverse-proxy/blob/34fe0134ecd290b5d17b7ccaba15f0715ae902d4/src/ReverseProxy/Forwarder/IHttpForwarderExtensions.cs#L23-L24

We can provide extension methods for simple request transforms:
```
public static ValueTask<ForwarderError> SendAsync(this IHttpForwarder forwarder, HttpContext context, string destinationPrefix,
            HttpMessageInvoker httpClient, Action<HttpRequestMessage> requestTransform)
```
Note TransformRequestAsync takes several parameters that aren't needed in the simple case:
https://github.com/microsoft/reverse-proxy/blob/34fe0134ecd290b5d17b7ccaba15f0715ae902d4/src/ReverseProxy/Forwarder/HttpTransformer.cs#L38

Which / how many overloads should we provide?

Requested by @davidfowl

adityamandaleeka:
Triage: Maybe we could prototype these by putting them in the sample.

Tratcher:
These were merged into the sample. Re-opening to consider moving into the product.

PR #1175, opened 2021-08-04
Add request transform extension method

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/1172

### DIFF
--- a/samples/ReverseProxy.Direct.Sample/HttpForwarderExtensions.cs
+++ b/samples/ReverseProxy.Direct.Sample/HttpForwarderExtensions.cs
@@ -1,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Yarp.ReverseProxy.Forwarder;
+
+namespace Yarp.Sample
+{
+    public static class HttpForwarderExtensions
+    {
+        /// <summary>
+        /// Forwards the incoming request to the destination server, and the response back to the client.
+        /// </summary>
+        /// <param name="context">The HttpContext to forward.</param>
+        /// <param name="destinationPrefix">The url prefix for where to forward the request to.</param>
+        /// <param name="httpClient">The HTTP client used to forward the request.</param>
+        /// <param name="requestTransform">Transform function to apply to the forwarded request.</param>
+        /// <returns>The status of a forwarding operation.</returns>
+        public static ValueTask<ForwarderError> SendAsync(this IHttpForwarder forwarder, HttpContext context, string destinationPrefix,
+            HttpMessageInvoker httpClient, Func<HttpContext, HttpRequestMessage, ValueTask> requestTransform)
+        {
+            return forwarder.SendAsync(context, destinationPrefix, httpClient, ForwarderRequestConfig.Empty, requestTransform);
+        }
+
+        /// <summary>
+        /// Forwards the incoming request to the destination server, and the response back to the client.
+        /// </summary>
+        /// <param name="context">The HttpContext to forward.</param>
+        /// <param name="destinationPrefix">The url prefix for where to forward the request to.</param>
+        /// <param name="httpClient">The HTTP client used to forward the request.</param>
+        /// <param name="requestConfig">Config for the outgoing request.</param>
+        /// <param name="requestTransform">Transform function to apply to the forwarded request.</param>
+        /// <returns>The status of a forwarding operation.</returns>
+        public static ValueTask<ForwarderError> SendAsync(this IHttpForwarder forwarder, HttpContext context, string destinationPrefix,
+            HttpMessageInvoker httpClient, ForwarderRequestConfig requestConfig, Func<HttpContext, HttpRequestMessage, ValueTask> requestTransform)
+        {
+            return forwarder.SendAsync(context, destinationPrefix, httpClient, requestConfig, new RequestTransformer(requestTransform));
+        }
+    }
+}
--- a/samples/ReverseProxy.Direct.Sample/RequestTransformer.cs
+++ b/samples/ReverseProxy.Direct.Sample/RequestTransformer.cs
@@ -1,0 +1,27 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Net.Http;
+using System.Threading.Tasks;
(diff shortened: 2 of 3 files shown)


src/ReverseProxy/Forwarder/IHttpForwarderExtensions.cs
```diff
@@ -49,5 +49,29 @@ public static class IHttpForwarderExtensions
...
+
+        /// <summary>
+        /// Forwards the incoming request to the destination server, and the response back to the client.
+        /// </summary>
+        /// <param name="context">The HttpContext to forward.</param>
+        /// <param name="destinationPrefix">The url prefix for where to forward the request to.</param>
+        /// <param name="httpClient">The HTTP client used to forward the request.</param>
+        /// <param name="requestTransform">Delegate to apply to forwarded the request.</param>
+        /// <returns>The result of forwarding the request and response.</returns>
+        public static ValueTask<ForwarderError> SendAsync(this IHttpForwarder forwarder, HttpContext context, string destinationPrefix,
```
alnikola, line 61: Please, add unit tests for this new logic.

src/ReverseProxy/Forwarder/IHttpForwarderExtensions.cs
```diff
@@ -49,5 +49,29 @@ public static class IHttpForwarderExtensions
...
             return forwarder.SendAsync(context, destinationPrefix, httpClient, requestConfig, HttpTransformer.Default);
         }
+
+        /// <summary>
+        /// Forwards the incoming request to the destination server, and the response back to the client.
+        /// </summary>
+        /// <param name="context">The HttpContext to forward.</param>
+        /// <param name="destinationPrefix">The url prefix for where to forward the request to.</param>
+        /// <param name="httpClient">The HTTP client used to forward the request.</param>
+        /// <param name="requestTransform">Delegate to apply to forwarded the request.</param>
```
alnikola, line 59:
```suggestion
        /// <param name="requestTransform">Transform function to apply to the forwarded request.</param>
```

src/ReverseProxy/Forwarder/IHttpForwarderExtensions.cs
```diff
@@ -49,5 +49,29 @@ public static class IHttpForwarderExtensions
...
+        /// <returns>The result of forwarding the request and response.</returns>
```
alnikola, line 60: Sounds a bit strange to me since the method returns `ForwarderError` indicating if the operation succeeded or not.
```suggestion
        /// <returns>The status of a forwarding operation.</returns>
```

Tratcher:
There are three overload questions we need to resolve:
1) How many parameters should the callback have? TransformRequestAsync takes up to 3 (HttpContext, HttpRequestMessage, and the destinationPrefix). If we only offered the HttpRequestMessage, the others could be captured by a closure if required. However,  I expect HttpContext will be needed most of the time, and I think we want to match the parameter ordering from TransformRequestAsync so excluding HttpContext on some overloads would be confusing. Proposal: Always have both HttpContext and HttpRequestMessage.
2) How many SendAsync parameters do we include? It takes up to 4 not counting the transforms, HttpContext, destinationPrefix, HttpClient, and the requestConfig. Only the request config is optional.
3) Should the callback be async? Many will be sync, but returning a Task makes it much more flexible and isn't a lot of overhead.

How about these two?
```
public static ValueTask<ForwarderError> SendAsync(this IHttpForwarder forwarder, HttpContext context, string destinationPrefix,
            HttpMessageInvoker httpClient, Func<HttpContext, HttpRequestMessage, ValueTask> requestTransform)
public static ValueTask<ForwarderError> SendAsync(this IHttpForwarder forwarder, HttpContext context, string destinationPrefix,
            HttpMessageInvoker httpClient, ForwarderRequestConfig requestConfig, Func<HttpContext, HttpRequestMessage, ValueTask> requestTransform)
```

Can you try these patterns out in the sample app and see how if they're easy enough to use?

src/ReverseProxy/Forwarder/IHttpForwarderExtensions.cs
Kahbazi: This is the current summary for the existing extension methods. Should I change those?

alnikola:
@Kahbazi Since it's already 2 weeks since the last commit here, could you please tell us if you are going to continue working on this PR and address [remaining questions](https://github.com/microsoft/reverse-proxy/pull/1175#issuecomment-892792441)?

Kahbazi:
@alnikola I thought the questions was directed to reviewers and I was waiting for feedback. I agree with methods that @Tratcher suggested and they are easy to use, but what bothers me is the fact that this requires creating a new `Transfomer` on each request.

https://github.com/microsoft/reverse-proxy/blob/a866d7910662125e3165836d4899a562b1693660/samples/ReverseProxy.Direct.Sample/HttpForwarderExtensions.cs#L40

and also it's very likely for users to capture `httpContext` by mistaker and cause a closure

https://github.com/microsoft/reverse-proxy/blob/a866d7910662125e3165836d4899a562b1693660/samples/ReverseProxy.Direct.Sample/Startup.cs#L53-L57

Tratcher:
> but what bothers me is the fact that this requires creating a new `Transfomer` on each request.

Correct, it's a trade-off for convenience. By the time you reach a scale where the extra allocation matters, you can create a re-usable transformer.

> and also it's very likely for users to capture httpContext by mistaker and cause a closure

True, but since this is per request that shouldn't pose any lifetime issues. Or do you mean the closure causes yet another allocation? The new static lambdas are a good mitigation for that and we could recommend it in the docs.

Kahbazi:
> > and also it's very likely for users to capture httpContext by mistaker and cause a closure
> 
> True, but since this is per request that shouldn't pose any lifetime issues. Or do you mean the closure causes yet another allocation? The new static lambdas are a good mitigation for that and we could recommend it in the docs.

Yes, I was talking about the extra allocation. I'll update the sample using static lambda.

Tratcher:
@alnikola we should review the proposed APIs tomorrow as a team.

samsp-msft:
Triage: We are happy with this pattern and the APIs. 
@alnikola - Needs code review.

alnikola (approved):
LGTM

alnikola:
Thanks!

Tratcher:
@samsp-msft @alnikola since we are happy with the APIs are we ready to move them from the sample to the product?

MihaZupan:
> are we ready to move them from the sample to the product?

That was my understanding of the team's decision

PR #1184, opened 2021-08-16
Register cluster for active health check when it's enabled on reload

### LINKED ISSUE #1134: Active health checks are not enabled after configuration reload
### Describe the bug
Active health checks are not enabled after configuration reload, only after a full restart. Strange thing is that the disabling seems to work, only the enabling doesn't

### To Reproduce

1. Create a config with a cluster with an active health check, but disabled (Enabled = false)
2. Start the proxy
3. While running the proxy, change and save the config, setting the health check to enabled=true
4. Watch the reload of the config happening, but the health checks calls don't appear in the logs

### Further technical details

- Yarp preview 12
- Windows

alnikola:
Fixes #1134

### DIFF
--- a/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
+++ b/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
@@ -36,6 +36,8 @@ namespace Yarp.ReverseProxy.Health
         }
 
         public bool InitialProbeCompleted { get; private set; }
+        
+        internal EntityActionScheduler<ClusterState> Scheduler { get; }
 
         public Task CheckHealthAsync(IEnumerable<ClusterState> clusters)
         {
@@ -98,8 +100,6 @@ namespace Yarp.ReverseProxy.Health
         {
             Scheduler.Dispose();
         }
-
-        internal EntityActionScheduler<ClusterState> Scheduler { get; }
 
         private async Task ProbeCluster(ClusterState cluster)
         {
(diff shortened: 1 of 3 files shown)


src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
```diff
@@ -83,24 +81,26 @@ public void OnClusterChanged(ClusterState cluster)
...
+            Scheduler.UnscheduleEntity(cluster);
         }
 
         public void Dispose()
         {
-            _scheduler.Dispose();
+            Scheduler.Dispose();
         }
 
+        internal EntityActionScheduler<ClusterState> Scheduler { get; private set; }
```
Kahbazi, line 102: ?
```suggestion
        internal EntityActionScheduler<ClusterState> Scheduler { get; }
```

alnikola: Thanks!

src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
```diff
@@ -83,24 +81,26 @@ public void OnClusterChanged(ClusterState cluster)
...
+            Scheduler.UnscheduleEntity(cluster);
         }
 
         public void Dispose()
         {
-            _scheduler.Dispose();
+            Scheduler.Dispose();
         }
 
+        internal EntityActionScheduler<ClusterState> Scheduler { get; }
```
Tratcher, line 102: Organization: properties go before methods, even internal ones.

Tratcher (approved):

src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
alnikola: Fixed

PR #1185, opened 2021-08-16
Expect: 100-continue functional tests w/ and w/o response body

### LINKED ISSUE #617: Test 100-continue scenarios

alnikola:
It adds functional tests for the scenarios listed below.

The first scenario set is based on [this list](https://github.com/microsoft/reverse-proxy/issues/617#issuecomment-852243785), but modified to be compatible with the actual HttpClient behavior.

- A. HTTP/1.1 POST with a Content-Length is proxied over HTTP/1.1, destination responds with a 100 Continue, the proxy responds with a 100 Continue, the data is uploaded, there's a 200 response.
- B. HTTP/1.1 POST with a Transfer-Encoding is proxied over HTTP/1.1, destination responds with a 100 Continue, the proxy responds with a 100 Continue, the data is uploaded, there's a 200 response.
- C. A & B proxied over HTTP/2
- D. HTTP/1.1 POST with a Content-Length is proxied over HTTP/1.1, destination responds with a 4xx, the data is not uploaded.
- E. D but with Transfer-Encoding.
- F. HTTP/1.1 POST with a Content-Length is proxied over HTTP/2, destination responds with a 4xx and a RESET, the data is not uploaded.
- G. HTTP/2 POST with a Content-Length is proxied over HTTP/2, destination responds with a 4xx and a RESET, the proxy app sends the 4xx and a RESET to cancel the body.

The second scenario set verifies handling of response payload in the cases of successful and terminated responses.

Fixes #617

### DIFF
--- a/test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
+++ b/test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
@@ -82,7 +82,7 @@ namespace Yarp.ReverseProxy
 
             await test.Invoke(async uri =>
             {
-                await ProcessHttpRequest(new Uri(uri), proxyProtocol, contentString, useContentLength, destResponseCode, false);
+                await ProcessHttpRequest(new Uri(uri), proxyProtocol, contentString, useContentLength, destResponseCode, false, destResponseCode == 200);
 
                 Assert.True(headerTcs.Task.IsCompleted);
                 var expectHeader = await headerTcs.Task;
@@ -103,19 +103,22 @@ namespace Yarp.ReverseProxy
         }
 
         [Theory]
-        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, false, false)]
-        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, false, true)]
-        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, true, false)]
-        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, true, true)]
-        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, false, false)]
-        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, false, true)]
-        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, true, false)]
-        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, true, true)]
-        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, false, false)]
-        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, false, true)]
-        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, true, false)]
-        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, true, true)]
-        public async Task PostExpect100_ResponseWithPayload(HttpProtocols proxyProtocol, HttpProtocols destProtocol, bool useContentLength, bool cancelResponse)
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, true, true, 200)]
+        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, true, true, 200)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, true, true, 200)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, false, false, 400)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, false, true, 400)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, true, false, 400)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, true, true, 400)]
+        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, false, false, 400)]
+        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, false, true, 400)]
+        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, true, false, 400)]
+        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, true, true, 400)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, false, false, 400)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, false, true, 400)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, true, false, 400)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, true, true, 400)]
+        public async Task PostExpect100_ResponseWithPayload(HttpProtocols proxyProtocol, HttpProtocols destProtocol, bool useContentLength, bool cancelResponse, int responseCode)
         {
             var requestBodyTcs = new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously);
 
@@ -125,7 +128,7 @@ namespace Yarp.ReverseProxy
                 {
                     await ReadContent(context, requestBodyTcs, Encoding.UTF8.GetByteCount(contentString));
 
-                    context.Response.StatusCode = 200;
+                    context.Response.StatusCode = responseCode;
 
                     var responseBody = Encoding.UTF8.GetBytes(contentString + "Response");
                     if (useContentLength)
@@ -164,10 +167,9 @@ namespace Yarp.ReverseProxy
 
             await test.Invoke(async uri =>
             {
-                var expectedStatusCode = 200;
-                await ProcessHttpRequest(new Uri(uri), proxyProtocol, contentString, useContentLength, expectedStatusCode, cancelResponse, async response =>
+                await ProcessHttpRequest(new Uri(uri), proxyProtocol, contentString, useContentLength, responseCode, cancelResponse, true, async response =>
                 {
-                    Assert.Equal(expectedStatusCode, (int)response.StatusCode);
+                    Assert.Equal(responseCode, (int)response.StatusCode);
 
                     var actualResponse = await response.Content.ReadAsStringAsync();
                     Assert.Equal(contentString + "Response", actualResponse);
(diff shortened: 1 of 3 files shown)


test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
```diff
@@ -0,0 +1,251 @@
...
+                }
+            });
+        }
+
+        [Theory]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, false, false)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, false, true)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, true, false)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, true, true)]
+        public async Task PostExpect100_ResponseWithPayload(HttpProtocols proxyProtocol, HttpProtocols destProtocol, bool useContentLength, bool cancelResponse)
```
Tratcher, line 110: Why not use the same matrix as the other test, is there something special about HTTP/1 here?

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
```diff
@@ -0,0 +1,251 @@
...
+                    context.Response.StatusCode = 200;
+
+                    var responseBody = Encoding.UTF8.GetBytes(contentString + "Response");
+                    if (useContentLength)
+                    {
+                        context.Response.Headers.ContentLength = responseBody.Length;
+                    }
+
+                    var cts = cancelResponse ? new CancellationTokenSource(1) : new CancellationTokenSource();
+                    await context.Response.Body.WriteAsync(responseBody.AsMemory(), cts.Token);
```
Tratcher, line 129:
```suggestion
                    if (cancelResponse)
                    {
                        context.Abort();
                    }
                    else
                    {
                        await context.Response.Body.WriteAsync(responseBody.AsMemory());
                    }
```

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
```diff
@@ -0,0 +1,251 @@
...
+
+                if (responseAction != null)
+                {
+                    await responseAction(response);
+                }
+            }
+            else
+            {
+                var exception = await Assert.ThrowsAsync<HttpRequestException>(() => client.SendAsync(message));
+                Assert.Equal(typeof(IOException), exception.InnerException.GetType());
```
Tratcher, line 238:
```suggestion
                Assert.IsType<IOException>(exception.InnerException);
```

Tratcher: Also check `Assert.Equal(0, contentStream.Position);` in this scenario.

alnikola: Sorry, I didn't get it. Here, it checks if the **request** content has been read by server which is true even if the **response** is cancelled. Could it be that you actually meant `Assert.Equal(content.Length, contentStream.Position);`?

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
```diff
@@ -0,0 +1,251 @@
...
+                    {
+                        // 100 response code is sent automatically on reading Body.
+                        await ReadContent(context, bodyTcs, Encoding.UTF8.GetByteCount(contentString));
+                    }
+
+                    context.Response.StatusCode = destResponseCode;
+                },
+                proxyBuilder => {
+                    proxyBuilder.Services.RemoveAll(typeof(IForwarderHttpClientFactory));
+                    proxyBuilder.Services.TryAddSingleton<IForwarderHttpClientFactory, TestForwarderHttpClientFactory>();
```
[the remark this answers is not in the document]
alnikola: Fixed

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
```diff
@@ -0,0 +1,251 @@
...
+        [Theory]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, true, 200)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, false, 200)]
+        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, true, 200)]
+        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, false, 200)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, true, 400)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, false, 400)]
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, true, 400)]
+        [InlineData(HttpProtocols.Http2, HttpProtocols.Http2, true, 400)]
+        public async Task PostExpect100_BodyAlwaysUploaded(HttpProtocols proxyProtocol, HttpProtocols destProtocol, bool useContentLength, int destResponseCode)
```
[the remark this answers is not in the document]
alnikola: I forgot to rename it to match the actual HttpClient behavior. Fixed.

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
alnikola: It was a typo. Fixed.

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
```diff
@@ -0,0 +1,266 @@
...
+        {
+            var requestBodyTcs = new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            var contentString = new string('a', 1024 * 1024 * 10);
+            var test = new TestEnvironment(
+                async context =>
+                {
+                    await ReadContent(context, requestBodyTcs, Encoding.UTF8.GetByteCount(contentString));
+
+                    context.Response.StatusCode = 200;
```
Tratcher, line 128: You're missing the 4XX scenarios from the other test where we don't expect the request body to be sent? Since we're mainly concerned about the 4XX scenario breaking, you don't need to test the full matrix for the 200 scenario.

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
Tratcher: Nevermind, I misread this test.

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
alnikola: This test verifies cancellation of a response stream by the server. It checks only 200 responses because for 400 responses the server terminates H/2 connection before the response content is sent.

alnikola: These new tests verify [this cancellation handling logic](https://github.com/dotnet/runtime/issues/53914#issuecomment-857983829) explained by @geoffkizer on dotnet/runtime#53914 discussion

Tratcher: > It checks only 200 responses because for 400 responses the server terminates H/2 connection before the response content is sent.

? 200 and 400 response bodies are treated the same by the server. Nor does a manual 400 response cause an HTTP/2 connection closure.

alnikola: Hmm, that's interesting. When I ran the [previous test version](https://github.com/microsoft/reverse-proxy/blob/c67c25b7b14165d0c3b406b71e6145238dca86fc/test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs) (first commit in the branch), I was getting exceptions on the `SendAsync` call for the 2 following configurations with 400 response
1.  proxyProtocol: Http1, destProtocol: Http2, useContentLength: True, hasResponseBody: True, destResponseCode: 400
```
System.Net.Http.HttpRequestException : Error while copying content to a stream.
    ---- System.IO.IOException : The response ended prematurely.
```
2. proxyProtocol: Http2, destProtocol: Http2, useContentLength: True, hasResponseBody: True, destResponseCode: 400
```
System.Net.Http.HttpRequestException : Error while copying content to a stream.
    ---- System.IO.IOException : The request was aborted.
    -------- System.Net.Http.Http2StreamException : The HTTP/2 server reset the stream. HTTP/2 error code 'PROTOCOL_ERROR' (0x1).
```
Both of the above use H/2 from proxy to destination, thus I assumed it's by design which seems wrong now.

alnikola: I switched the full matrix to use 400 code and kept only 3 cases for 200.

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
```diff
@@ -0,0 +1,269 @@
...
+        [InlineData(HttpProtocols.Http1, HttpProtocols.Http2, true, true, 400)]
+        public async Task PostExpect100_ResponseWithPayload(HttpProtocols proxyProtocol, HttpProtocols destProtocol, bool useContentLength, bool cancelResponse, int responseCode)
+        {
+            var requestBodyTcs = new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            var contentString = new string('a', 1024 * 1024 * 10);
+            var test = new TestEnvironment(
+                async context =>
+                {
+                    await ReadContent(context, requestBodyTcs, Encoding.UTF8.GetByteCount(contentString));
```
Tratcher, line 129: Don't you want ReadContent to be conditional on StatusCode like the other test?

Tratcher (approved):

test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
alnikola: I don't think an extra `if` is necessary here because this test verifies how the response is processed not the request.

alnikola: Actually, when I tried adding `if` here like below
```C#
if (responseCode == 200)
{
    await ReadContent(context, requestBodyTcs, Encoding.UTF8.GetByteCount(contentString));
}
```
I started getting the same exceptions as mentioned [above](https://github.com/microsoft/reverse-proxy/pull/1185#discussion_r692872324). I'm not sure if it's by design or bug. Do you know anything about this?

Tratcher: That looks like a product bug. It should be possible for the client to receive a 4XX response including body without uploading the request body.

Please re-open and investigate.

alnikola: I submitted a separate issue for this potential bug #1200

PR #1188, opened 2021-08-18
Remove HTTP2-Settings request header

### LINKED ISSUE #1112: Remove `HTTP2-settings` header
Header should not be passed through the proxy.

alnikola:
Fixes #1112

### DIFF
--- a/src/ReverseProxy/Forwarder/RequestUtilities.cs
+++ b/src/ReverseProxy/Forwarder/RequestUtilities.cs
@@ -58,7 +58,7 @@ namespace Yarp.ReverseProxy.Forwarder
             return _headersToExclude.Contains(headerName);
         }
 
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Forwarder/RequestUtilities.cs
```diff
@@ -73,6 +73,7 @@ internal static bool ShouldSkipResponseHeader(string headerName)
             "Security-Scheme",
             "ALPN",
             "Close",
+            "HTTP2-Settings",
```
MihaZupan, line 76: Please update the capacity passed to the constructor as well
https://github.com/microsoft/reverse-proxy/blob/30b731a97b6b719c33972ef52ad244f72fd56c4c/src/ReverseProxy/Forwarder/RequestUtilities.cs#L61

MihaZupan (approved):

PR #1190, opened 2021-08-18
Update enum usage

WeihanLi:
Thought there's no need to specify `int`, is there some concern?

### DIFF
--- a/src/ReverseProxy.TelemetryConsumption/Forwarder/ForwarderStage.cs
+++ b/src/ReverseProxy.TelemetryConsumption/Forwarder/ForwarderStage.cs
@@ -6,7 +6,7 @@ namespace Yarp.ReverseProxy.Telemetry.Consumption
     /// <summary>
     /// Stages of forwarding a request.
     /// </summary>
-    public enum ForwarderStage : int
+    public enum ForwarderStage
     {
         SendAsyncStart = 1,
         SendAsyncStop,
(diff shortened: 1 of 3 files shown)


Kahbazi:
@WeihanLi FYI this is the previous conversation about the issue : https://github.com/microsoft/reverse-proxy/pull/501/files#r517487666

MihaZupan:
I explicitly added the backing type here as changing it could break binary compatibility of our EventSource events between versions (enums may not always serialize as 4 bytes)

karelz:
Perhaps we could add comment there why we explicitly added the backing type?

src/ReverseProxy.TelemetryConsumption/Forwarder/ForwarderStage.cs
```diff
@@ -6,7 +6,7 @@ namespace Yarp.ReverseProxy.Telemetry.Consumption
     /// <summary>
     /// Stages of forwarding a request.
     /// </summary>
-    public enum ForwarderStage : int
+    public enum ForwarderStage
```
WeihanLi, line 9: Seemed this enum definition is the same with `src/ReverseProxy/Forwarder/ForwarderStage.cs`, maybe we could share the same one?

MihaZupan: They are almost the same, but one is public while the other is internal.
Adding conditional compilation for just this file seemed like more effort than it's worth.

PR #1193, opened 2021-08-19
Fix race conditions in EntityActionScheduler

### LINKED ISSUE #1136: Fix race conditions in EntityActionScheduler
The main issue is in `ScheduleEntity` where it will register the timer (autostart) before adding it to the dictionary.

A few scenarios:

1.
- Thread A calls ScheduleEntity and autostart is false
- Thread B calls Start, switching autostart to true
- Thread B enumerates the dictionary and enables all timers
- Thread A adds the unscheduled entry to the dictionary, where it will live forever taking up memory and never firing

2.
- Thread A calls ScheduleEntity and autostart is true
- The timer we just created fires and the callback is invoked
- Thread B executes the timer callback
- Thread A adds the entry into the dictionary
- Since the timer callback never saw the entry in the dictionary, it will never remove it from it. Any timer that isn't "runOnce" will never be re-scheduled.

alnikola:
Fixes #1136

### DIFF
--- a/src/ReverseProxy/Health/EntityActionScheduler.cs
+++ b/src/ReverseProxy/Health/EntityActionScheduler.cs
@@ -62,13 +62,12 @@ namespace Yarp.ReverseProxy.Health
 
         public void ScheduleEntity(T entity, TimeSpan period)
         {
-            var isStarted = Volatile.Read(ref _isStarted);
             var entry = new SchedulerEntry(entity, (long)period.TotalMilliseconds, _timerCallback, _timerFactory);
             var added = _entries.TryAdd(entity, entry);
 
             // Scheduler could have been started while we were adding the new entry.
             // Start timer here to ensure it's not forgotten.
-            if (isStarted == 1 || Volatile.Read(ref _isStarted) == 1)
+            if (Volatile.Read(ref _isStarted) == 1)
             {
                 entry.SetTimer();
             }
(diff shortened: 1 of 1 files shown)


src/ReverseProxy/Health/EntityActionScheduler.cs
```diff
@@ -56,14 +56,23 @@ public void Start()
...
             {
-                entry.Timer.Change(entry.Period, Timeout.Infinite);
+                entry.SetTimer();
             }
         }
 
         public void ScheduleEntity(T entity, TimeSpan period)
         {
-            var entry = new SchedulerEntry(entity, (long)period.TotalMilliseconds, _timerCallback, Volatile.Read(ref _isStarted) == 1, _timerFactory);
+            var isStarted = Volatile.Read(ref _isStarted);
```
gfoidl, line 65: I think this read here can be omitted, as `isStarted` isn't used up until L71 which could be written as
```c#
if (Volatile.Read(ref _isStarted) == 1)
{
    entry.SetTimer();
}
```
So no matter what state `_isStarted` has, there's only one volatile read. And that read is the right place to get the freshest value (due to the memory barriers from Interlocked.Exchange).

alnikola: Oh, you are right. Thanks for finding that. Fixed.

MihaZupan (approved):

PR #1194, opened 2021-08-20
Use the max of reactivation and detection periods for reactivation delay

### LINKED ISSUE #1082: Enforce Reactivation Period >= Detection Window in TransportFailureRateHealthPolicy
TransportFailureRateHealthPolicy measures the share of the failed request out of the total request made do destination happened in the time period defined as Detection Window. Once a destination's passive health is set to `Unhealthy`, such destination gets scheduled for a reactivation after the defined Reactivation Period. Reactivation means changing the passive health status from `Unhealthy` to `Unknown` which again makes destination eligible for receiving traffic.

We need to ensure that Reaction Period >= Detection Window to avoid cases when a reactivated destination gets marked as `Unhealthy` right after the first request.

alnikola:
When `TransportFailureRatePolicy` is configured with ReactivationPeriod < DetectionWindowSize, an unhealthy destination sent for cool down and then reactivated can be marked as unhealthy on the first request right after reactivation even if such a request completed successfully.

To avoid that issue, we will use the max of reactivation and detection periods for the reactivation delay.

Fixes #1082

### DIFF
--- a/src/ReverseProxy/Health/TransportFailureRateHealthPolicy.cs
+++ b/src/ReverseProxy/Health/TransportFailureRateHealthPolicy.cs
@@ -48,7 +48,8 @@ namespace Yarp.ReverseProxy.Health
         {
             var error = context.Features.Get<IForwarderErrorFeature>();
             var newHealth = EvaluateProxiedRequest(cluster, destination, error != null);
-            var reactivationPeriod = cluster.Model.Config.HealthCheck?.Passive?.ReactivationPeriod ?? _defaultReactivationPeriod;
+            var clusterReactivationPeriod = cluster.Model.Config.HealthCheck?.Passive?.ReactivationPeriod ?? _defaultReactivationPeriod;
+            var reactivationPeriod = clusterReactivationPeriod >= _policyOptions.DetectionWindowSize ? clusterReactivationPeriod : _policyOptions.DetectionWindowSize;
             _healthUpdater.SetPassive(cluster, destination, newHealth, reactivationPeriod);
         }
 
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Health/TransportFailureRateHealthPolicy.cs
```diff
@@ -64,7 +64,12 @@ private DestinationHealth EvaluateProxiedRequest(ClusterState cluster, Destinati
                     (long)_policyOptions.DetectionWindowSize.TotalMilliseconds,
                     _policyOptions.MinimalTotalCountThreshold,
                     failed);
-                return failureRate < rateLimit ? DestinationHealth.Healthy : DestinationHealth.Unhealthy;
+                var newHealthState =  failureRate < rateLimit ? DestinationHealth.Healthy : DestinationHealth.Unhealthy;
+                if (newHealthState == DestinationHealth.Unhealthy)
+                {
+                    history.Clear();
```
Tratcher, line 70: Clearing at this point is problematic since there could still be other requests in flight that would be operating on the newly cleared state. Ideally the clear would happen during the re-activation.

If that's hard to do then consider the alternative of using a Max reactivation period of the history length and the configured ReactivationPeriod.

alnikola: Thanks for catching this.
I spent some time, but couldn't come up with a reasonable solution for clearing history on a reactivation, so instead use the max of detection and reactivation periods for reactivation.

src/ReverseProxy/Health/TransportFailureRateHealthPolicy.cs
```diff
@@ -48,7 +48,8 @@ public void RequestProxied(HttpContext context, ClusterState cluster, Destinatio
         {
             var error = context.Features.Get<IForwarderErrorFeature>();
             var newHealth = EvaluateProxiedRequest(cluster, destination, error != null);
-            var reactivationPeriod = cluster.Model.Config.HealthCheck?.Passive?.ReactivationPeriod ?? _defaultReactivationPeriod;
+            var clusterReactivationPeriod = cluster.Model.Config.HealthCheck?.Passive?.ReactivationPeriod ?? _defaultReactivationPeriod;
+            var reactivationPeriod = clusterReactivationPeriod >= _policyOptions.DetectionWindowSize ? clusterReactivationPeriod : _policyOptions.DetectionWindowSize;
```
Tratcher, line 52:
```suggestion
            // Avoid reactivating until the history has expired so that it does not affect future health assessments.
            var reactivationPeriod = clusterReactivationPeriod >= _policyOptions.DetectionWindowSize ? clusterReactivationPeriod : _policyOptions.DetectionWindowSize;
```

Tratcher (approved):

PR #1195, opened 2021-08-22
Improve EntityActionScheduler reliability

MihaZupan:
The scheduler has a few issues right now:

- Changing the period or removing+adding an entry for the same key while the callback is (or is about to be) run will result in the callback firing multiple times. The callbacks will keep rescheduling themselves even if the entry has since been removed (it was only checking the key was in the dictionary)
- Updating the period of a timer on a frequency higher than that of the period would result in the callback never firing
  - Changed the entry so that updating the period does not call `Change` on the timer (which is racy). Instead, the timer is only rescheduled during the callback. This means that a period change will only take place after the timer fires - it is a slight change in behavior, but one that should not matter for YARP.
- Calls to `ScheduleEntity` or `ChangePeriod` after calling dispose (or in a race) could keep the scheduler and all its state rooted forever (the callback passed to the timer contained a captured reference to the scheduler) - this could result in random memory leaks if the app was recycling the state without restarting (likely not common with YARP)
  - Changed the implementation so that the timer callback does not capture state, and a weak reference to the scheduler is kept by entries instead
- The timers it creates will capture the user's `ExecutionContext`, keeping random `AsyncLocal`s alive
  - Added EC suppression when creating the timer
- An exception thrown by the callback would not be cought while executing on the ThreadPool (not currently an issue since none of our callbacks will throw)

### DIFF
--- a/src/ReverseProxy/Health/EntityActionScheduler.cs
+++ b/src/ReverseProxy/Health/EntityActionScheduler.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


src/ReverseProxy/Health/EntityActionScheduler.cs
```diff
@@ -24,62 +25,80 @@ namespace Yarp.ReverseProxy.Health
...
+        private readonly WeakReference<EntityActionScheduler<T>> _weakThisRef;
         private readonly Func<T, Task> _action;
         private readonly bool _runOnce;
         private readonly ITimerFactory _timerFactory;
-        private readonly TimerCallback _timerCallback;
-        private int _isStarted;
+
+        private const int Status_NotStarted = 0;
+        private const int Status_Started = 1;
+        private const int Status_Disposed = 2;
```
alnikola, line 35: Naming seems strange because I've not seen previously us using `_` in the middle on member names in production code.
```suggestion
        private const int NotStarted = 0;
        private const int Started = 1;
        private const int Disposed = 2;
```

src/ReverseProxy/Health/EntityActionScheduler.cs
```diff
@@ -24,62 +25,80 @@ namespace Yarp.ReverseProxy.Health
...
             }
         }
 
         public void ScheduleEntity(T entity, TimeSpan period)
         {
-            var entry = new SchedulerEntry(entity, (long)period.TotalMilliseconds, _timerCallback, _timerFactory);
-            var added = _entries.TryAdd(entity, entry);
+            // Ensure the Timer has a weak reference to this scheduler; otherwise,
+            // EntityActionScheduler can be rooted by the Timer implementation.
+            var entry = new SchedulerEntry(_weakThisRef, entity, (long)period.TotalMilliseconds, _timerFactory);
```
alnikola, line 74: I doubt that this is a real issue that worth introducing an increased complexity with the weak reference. `SchedulerEntry` and the timer callback implementation are nested private types thus we can rely that they correctly handle Scheduler reference.

src/ReverseProxy/Health/EntityActionScheduler.cs
```diff
@@ -24,62 +25,80 @@ namespace Yarp.ReverseProxy.Health
...
+                // Scheduler could have been started while we were adding the new entry.
+                // Start timer here to ensure it's not forgotten.
+                if (Volatile.Read(ref _status) == Status_Started)
+                {
+                    entry.EnsureStarted();
+                }
+            }
+            else
+            {
+                entry.Dispose();
```
alnikola, line 87: This code is unreachable because `entry` is a newly created instance used as the dictionary key, so `_entries.TryAdd` will always return `true`. Optionally, we can add `Debug.Assert` here to clearly indicate that invariant.

alnikola (commented):
First brief review

src/ReverseProxy/Health/EntityActionScheduler.cs
MihaZupan: That was my read in #865 as well, but the key is the entity and not the entry. The key may already exist.

src/ReverseProxy/Health/EntityActionScheduler.cs
MihaZupan: That was my read in #865 as well, but the key is the entity and not the entry. The key may already exist.

src/ReverseProxy/Health/EntityActionScheduler.cs
MihaZupan: We've used it in a few places in [runtime](https://github.com/dotnet/runtime/blob/3d5d26181cd7bc07ea6c6f87710c14ccd043b415/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnection.cs#L71-L73), but I get it's not common. I'm fine with renaming them.

src/ReverseProxy/Health/EntityActionScheduler.cs
alnikola: Ah, you're right! It's my mistake.

src/ReverseProxy/Health/EntityActionScheduler.cs
MihaZupan: > thus we can rely that they correctly handle Scheduler reference.

They don't currently, the timer implementation will root the scheduler through the callback.

It may not be 100% needed after the other changes in this PR (dispose setting the flag), but at this point the only difference is using a strong vs weak reference (not too much complexity IMO),

alnikola: > They don't currently, the timer implementation will root the scheduler through the callback.

I might be missing something, but why is it an issue? We don't recreate scheduler and aren't going to do so because there is no need.

MihaZupan: If the scheduler is disposed, it might still be rooted, resulting in a memory leak. As is, the only "safe way" to dispose of it is to shut down the process.

As you said, we don't dispose of it ourselves, but the user may recreate the DI container YARP is in at will.

alnikola: But, as far as I see after a timer has been [disposed and garbage collected](https://docs.microsoft.com/en-us/dotnet/api/system.threading.timer.dispose?view=net-5.0) it won't reference anything. The only exception is the calls which have already been scheduled on ThreadPool, but they will be eventually executed and collected, too.

MihaZupan: In the current implementation, calls to `ScheduleEntity` or `ChangePeriod` in a race with `Dispose` may keep timers alive and running forever.

Also in a situation where the type is (improperly) not disposed, we should let the GC reclaim all resources (rn it can't).
See [comment in `HttpConnectionPoolManager`](https://github.com/dotnet/runtime/blob/3edda177b07120f761a9ac6ab7f33089972d7170/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs#L102-L104) where we guard against the same thing with the timers used there.

alnikola: OK, if you think a weak reference is really necessary here let's keep it.

PR #1196, opened 2021-08-23
Update direct-forwarding.md

sharkAndshark:


alnikola:
@sharkAndshark Could you please explain why do we need this change? Is there something broken right now?

MihaZupan:
> Could you please explain why we need this change?

The `C#` is currently its own paragraph and not the language specifier for the markdown code block
image

Issue #708, opened 2021-02-02
Add telemetry to application insights


luizfbicalho:
### What should we add or change to make your life better?
Would be interesting to make integtation with application insights easier, like adding all the telemetry to azure application inghts
### Why is this important to you?
Because we can use all of the appliction insights infrastructure to monitor the proxy.

Tratcher:
YARP operates as an ASP.NET Core component. What do you need from YARP beyond Application Insights existing ASP.NET Core integration?

luizfbicalho:
The application insights log doens't  show the proxy request

For example, my proxy calls an api to get a valid token for this request, if I open the application insights log, it show the token request, but it doesn't show the proxy request to the destination.

Maybe I did something wrong in the telemetry configuration


```
            services.AddReverseProxy()
                .LoadFromConfig(Configuration.GetSection("ReverseProxy"));
            services.AddTelemetryListeners();
           
            services.AddApplicationInsightsTelemetry(Configuration["APPINSIGHTS_CONNECTIONSTRING"]);
```

Tratcher:
@MihaZupan do you think this is because of https://github.com/dotnet/runtime/issues/31261?

karelz:
Triage: We will need some guidance/docs by @MihaZupan

luizfbicalho:
I didn't understand @karelz  and @MihaZupan , is there something that I can do to log all of the information about the proxy information on the fowarded request?

MihaZupan:
Not currently, we will need to add more instrumentation as described in https://github.com/microsoft/reverse-proxy/issues/776#issuecomment-786241459 to support AppInsights out of the box.

Out-of-the-box you will currently only see the incoming requests to the proxy, and we will forward the trace information in the request to the backend, so if you have other services participating in the distributed trace, those will light up as well.

luizfbicalho:
Thanks, if it doesn't work out of the box, is there something that I can do? some extension would help? I could try to develop something if I had the direction

brentnewbury-work:
Is there any update to this? I'd also like to know if there's anything we can do now to light up dependency tracking of proxied requests in AppInsights.

MihaZupan:
It is actively being worked on in runtime. Relevant issue: https://github.com/dotnet/runtime/issues/50658

If you are blocked, see [outgoing dependencies tracking](
[...]

PR #1199, opened 2021-08-23
Set ActivityContextHeaders default to CorrelationContext only

MihaZupan:
Contributes to #708

This matches the default used by runtime
https://github.com/dotnet/runtime/blob/c83b6edc1c1f500565bc5d53cfd726463ce2f56b/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DistributedContextPropagator.cs#L123
and will make the change smaller if/when we remove the `ActivityPropagationHandler` from YARP.


alnikola (approved):
LGTM

PR #1202, opened 2021-08-24
Allow empty headers

Tratcher:
Customers concerned with proxying requests as faithfully as possible have raised concerns about not proxying headers with empty values. It looks like empty header values are supported by Kestrel and HttpClient, it was YARP that was filtering them out.

Note I think the ASP.NET Core HttpSys and IIS servers filter out empty response headers, but that's not a blocker here.


alnikola (approved):
LGTM

PR #1218, opened 2021-08-30
Improve K8S Support For Multiple Ingress

### LINKED ISSUE #1210: Kubernetes MessageConfigProvider Doesn't Support Multiple Ingress
### Describe the bug
Right now if you configure multiple ingress the kubernetes protocol only updates the config to the last message received from the controller. Fundamentally this issue seems to stem from the fact that the `MessageConfigProvider` class as it is implemented today does not handle incremental updates to the config as they are provided. Furthermore, the controller reconciler only ever sends `Update` events, there is no handling should an ingress go away.

### To Reproduce
1. Configure two separate ingresses
2. Only one of the ingress will actually end up configured on the yarp-ingress service.

archonandrewhunt:
Mostly fixes #1210

Doesn't add support for ingress being removed, only additions/updates.

I took a fairly simplest approach for fixing the issue of additive updates to the config on the receiving side. This resolves my immediate issue, although I understand it may not be the desired approach for this project. I am open to feedback and willing to adjust this code as necessary to get it merged.

### DIFF
--- a/src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
+++ b/src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
@@ -97,21 +97,21 @@ namespace Yarp.ReverseProxy.Kubernetes.Controller.Converters
                         {
                             Address = uri
                         };
-
-                        var pathMatch = FixupPathMatch(path);
-                        var host = rule.Host;
-
-                        routes.Add(new RouteConfig()
-                        {
-                            Match = new RouteMatch()
-                            {
-                                Hosts = host != null ? new[] { host } : Array.Empty<string>(),
-                                Path = pathMatch
-                            },
-                            ClusterId = cluster.ClusterId,
-                            RouteId = $"{context.Ingress.Metadata.Name}:{path.Path}"
-                        });
                     }
+
+                    var pathMatch = FixupPathMatch(path);
+                    var host = rule.Host;
+
+                    routes.Add(new RouteConfig()
+                    {
+                        Match = new RouteMatch()
+                        {
+                            Hosts = host != null ? new[] { host } : Array.Empty<string>(),
+                            Path = pathMatch
+                        },
+                        ClusterId = cluster.ClusterId,
+                        RouteId = $"{context.Ingress.Metadata.Name}:{path.Path}"
+                    });
                 }
             }
         }
--- a/src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
+++ b/src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
@@ -109,7 +109,7 @@ namespace Yarp.ReverseProxy.Kubernetes.Controller.Converters
                                 Path = pathMatch
                             },
                             ClusterId = cluster.ClusterId,
-                            RouteId = path.Path
+                            RouteId = $"{context.Ingress.Metadata.Name}:{path.Path}"
                         });
                     }
                 }
--- a/src/ReverseProxy.Kubernetes.Protocol/MessageConfigProvider.cs
+++ b/src/ReverseProxy.Kubernetes.Protocol/MessageConfigProvider.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using Microsoft.Extensions.Primitives;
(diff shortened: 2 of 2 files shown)


samsp-msft:
@karelz @adityamandaleeka - this probably needs a sponsor from the team to help shepherd it through.

Tratcher:
Several tests need to be updated.

src/ReverseProxy.Kubernetes.Protocol/MessageConfigProvider.cs
```diff
@@ -22,6 +23,17 @@ public MessageConfigProvider()
...
+
+            if (oldConfig.Routes != null)
+            {
+                routes = routes.Union(oldConfig.Routes.Where(or => !routes.Any(r => r.RouteId == or.RouteId))).ToList();
+            }
+
+            if (oldConfig.Clusters != null)
+            {
+                clusters = clusters.Union(oldConfig.Clusters?.Where(oc => !clusters.Any(r => r.ClusterId == oc.ClusterId))).ToList();
+            }
```
Tratcher, line 35: This doesn't look like the right place to unify changes with unchanged entries, with this approach there's no way to ever remove a route or cluster.

adityamandaleeka:
Triage: @archonandrewhunt Just checking, are you still interested in working on this PR?

archonandrewhunt:
@adityamandaleeka I am! I was actually off last week hence the lack of any response here. Will get back into it this week.

src/ReverseProxy.Kubernetes.Protocol/MessageConfigProvider.cs
archonandrewhunt: If the code were to live here, which I am very open to moving, I was thinking of adding a complimentary remove method that could handle those types of changes. I did note in the issue I opened relating to this PR as well that there is no support for removal in the code that I could see so far. The `Reconciler` class only sends update events at the moment. I am happy to improve that situation as well but did want to get some feedback a bit earlier to avoid having to rework a lot of stuff. 

With all that in mind, did you have a different place where you would like me to move this change unification logic to?

Tratcher: Update was supposed to be called with a complete set of routes and clusters, not just the changes. Is that not what was happening before your changes?

archonandrewhunt: No that's not the current behaviour. Right now `IngressController` and `Reconciler` are sending individual `Message` objects for each ingress that gets processed. Each of those would then get passed through to this Update method overriding the previous config.

dpbevin: I agree that the Update call should be based on all ingresses, rather than just individual ones.
It would likely need a rate limiter though.

src/ReverseProxy.Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -97,21 +97,21 @@ private static void HandleIngressRulePath(YarpIngressContext context, List<Endpo
...
-                            },
-                            ClusterId = cluster.ClusterId,
-                            RouteId = path.Path
-                        });
-                    }
+                            Hosts = host != null ? new[] { host } : Array.Empty<string>(),
+                            Path = pathMatch
+                        },
+                        ClusterId = cluster.ClusterId,
+                        RouteId = $"{context.Ingress.Metadata.Name}:{path.Path}"
```
dpbevin, line 113: This is a good step forward.
Eventually the route and cluster IDs will need to include the namespace. I meant to add an issue for that last week but I forgot. I'll add one now and I'm happy to fix as a separate PR.

zainaoo: Good job 🤨

src/ReverseProxy.Kubernetes.Protocol/MessageConfigProvider.cs
MihaZupan: > I agree that the Update call should be based on all ingresses, rather than just individual ones.
It would likely need a rate limiter though.

@dpbevin this is essentially the behavior after #1210 right?

dpbevin: @MihaZupan correct. #1253 solves this same issue as this PR set out to.

Issue #1229, opened 2021-09-03
Add Cluster parameter to IProxyConfigFilter.ConfigureRouteAsync


Kahbazi:
### What should we add or change to make your life better?
Add `ClusterConfig?` parameter to `IProxyConfigFilter.ConfigureRouteAsync`.
```diff
namespace Yarp.ReverseProxy.Configuration
{
    public interface IProxyConfigFilter
    {
-        ValueTask<RouteConfig> ConfigureRouteAsync(RouteConfig route, CancellationToken cancel);     
+        ValueTask<RouteConfig> ConfigureRouteAsync(RouteConfig route, ClusterConfig? cluster, CancellationToken cancel);
    }
}
```

### Why is this important to you?
Beside grouping for route config (#164), this could be used to apply certain rule for routes based on `cluster.Metadata`.

Tratcher:
This is logistically challenging since the filter is called before a route has been associated with a cluster. We don't even have a dictionary of clusters yet, just a flat list.
https://github.com/microsoft/reverse-proxy/blob/16ab855a5be980a0cf5fd21e9a7028190b6e7dda/src/ReverseProxy/Management/ProxyConfigManager.cs#L212-L213
We could reverse these so clusters are verified first, but you'd still need a temporary dictionary of clusters to enable lookup for the associated routes.

PR #1231, opened 2021-09-03
Add cluster to ConfigureRouteAsync

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/1229

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -231,7 +231,7 @@ csharp_using_directive_placement = outside_namespace:error
 
 # New-line options
 # https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-formatting-conventions?view=vs-2019#new-line-options
-csharp_new_line_before_open_brace = methods, properties, control_blocks, types, anonymous_methods, lambdas, object_collection_array_initializers
+csharp_new_line_before_open_brace = methods, properties, control_blocks, types, anonymous_methods, lambdas, object_collection_array_initializers, accessors
 csharp_new_line_before_else = true
 csharp_new_line_before_catch = true
 csharp_new_line_before_finally = true
--- a/src/ReverseProxy/Management/ProxyConfigManager.cs
+++ b/src/ReverseProxy/Management/ProxyConfigManager.cs
@@ -209,8 +209,8 @@ namespace Yarp.ReverseProxy.Management
         // Throws for validation failures
         private async Task<bool> ApplyConfigAsync(IProxyConfig config)
         {
-            var (configuredRoutes, routeErrors) = await VerifyRoutesAsync(config.Routes, cancellation: default);
             var (configuredClusters, clusterErrors) = await VerifyClustersAsync(config.Clusters, cancellation: default);
+            var (configuredRoutes, routeErrors) = await VerifyRoutesAsync(config.Routes, configuredClusters, cancellation: default);
 
             if (routeErrors.Count > 0 || clusterErrors.Count > 0)
             {
@@ -223,7 +223,7 @@ namespace Yarp.ReverseProxy.Management
             return routesChanged;
         }
 
-        private async Task<(IList<RouteConfig>, IList<Exception>)> VerifyRoutesAsync(IReadOnlyList<RouteConfig> routes, CancellationToken cancellation)
+        private async Task<(IList<RouteConfig>, IList<Exception>)> VerifyRoutesAsync(IReadOnlyList<RouteConfig> routes, IList<ClusterConfig> clusters, CancellationToken cancellation)
         {
             if (routes == null)
             {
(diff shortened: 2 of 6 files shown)


src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -234,6 +234,16 @@ private async Task<(IList<RouteConfig>, IList<Exception>)> VerifyRoutesAsync(IRe
...
 
+            Dictionary<string, ClusterConfig>? clusterDictionary = null;
+            if (_filters.Length != 0)
+            {
+                clusterDictionary = new Dictionary<string, ClusterConfig>();
+                foreach (var cluster in clusters)
+                {
+                    clusterDictionary.Add(cluster.ClusterId, cluster);
+                }
+            }
```
Tratcher, line 245: It looks like VerifyClustersAsync is already building a HashSet and a list. Want to combine those into a Dictionary and return it so we don't have to do the same work twice?
https://github.com/microsoft/reverse-proxy/blob/08cecb0d93e82a5efaad127a1d47fc8ae1d50a2c/src/ReverseProxy/Management/ProxyConfigManager.cs#L299-L300

.editorconfig
```diff
@@ -231,7 +231,7 @@ csharp_using_directive_placement = outside_namespace:error
 
 # New-line options
 # https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-formatting-conventions?view=vs-2019#new-line-options
-csharp_new_line_before_open_brace = methods, properties, control_blocks, types, anonymous_methods, lambdas, object_collection_array_initializers
+csharp_new_line_before_open_brace = methods, properties, control_blocks, types, anonymous_methods, lambdas, object_collection_array_initializers, accessors
```
Kahbazi, line 234: This was bugging me for a while.

test/ReverseProxy.Tests/Management/ProxyConfigManagerTests.cs
```diff
@@ -387,45 +387,72 @@ public ValueTask<ClusterConfig> ConfigureClusterAsync(ClusterConfig cluster, Can
...
-            var clusterState = routeConfig.Cluster;
-            Assert.NotNull(clusterState);
-            Assert.True(clusterState.Model.Config.HealthCheck.Active.Enabled);
-            Assert.Equal(TimeSpan.FromSeconds(12), clusterState.Model.Config.HealthCheck.Active.Interval);
-            var destination = Assert.Single(clusterState.DestinationsState.AllDestinations);
+            Assert.Equal(2, dataSource.Endpoints.Count);
+
+            var endpoint1 = dataSource.Endpoints[0];
+            var routeConfig1 = endpoint1.Metadata.GetMetadata<RouteModel>();
+            Assert.Equal(47, routeConfig1.Config.Order);
```
Kahbazi, line 445: I'm not quiet sure why this tests is failing. This is also fails randomly on my machine. 
I'm just guessing, could it be the fact that we are using `with` operator in dotnet 3.1?

alnikola: What error does it show when failed? It's strange that there is some randomness.

Kahbazi: Unfortunately I'm seeing this : `Assert.Equal() Failure\nExpected: 47\nActual:   12`
I really hope I'm missing something.

Kahbazi: FOUND IT! It looks like the order of endpoints is not always the same. I wonder why!
https://github.com/microsoft/reverse-proxy/blob/65fbfb4fdb792c1a8ba34a765ff55a6ab2f71bef/test/ReverseProxy.Tests/Management/ProxyConfigManagerTests.cs#L443

src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -246,9 +249,13 @@ private async Task<(IList<RouteConfig>, IList<Exception>)> VerifyRoutesAsync(IRe
 
                 try
                 {
-                    foreach (var filter in _filters)
+                    if (_filters.Length != 0)
                     {
-                        route = await filter.ConfigureRouteAsync(route, cancellation);
+                        clusters.TryGetValue(route.ClusterId!, out var cluster);
```
Tratcher, line 254: This nullability override is inappropriate because ClusterId really can be null, so TryGetValue will throw. Either check that the id isn't null before calling this, or add `?? string.Empty`.

Tratcher (commented):
For triage: While this is a breaking API change, it's not in the primary code path so the impact should be small.

Tratcher (approved):

Tratcher:
Thanks

Issue #1162, opened 2021-07-22
Kubernetes Ingress Controller Ignores Port Set in Kind:Ingress


michaelkacher:
### Describe the bug
The Kubernetes Ingress controller always tries to access services on port 80 even if the Ingress resource is set at a different port.

For example, when the following configuration for port 5011 is provided, the logs read that it is attempting to add the service on port 80.
```
apiVersion: v1
kind: Service
metadata:
  name: test-ui
  labels:
    app: test-ui
spec:
  type: ClusterIP
  ports:
    - port: 5011
      targetPort: 80
  selector:
    app: test-ui
---
kind: Ingress
apiVersion: networking.k8s.io/v1
metadata:
  name: test-ui-ingress
spec:
  rules:
  - http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: test-ui
            port:
              number: 5011
```

```
[11:38:32 DBG] New client created for cluster 'test-ui:80'.
[11:38:32 DBG] Cluster 'test-ui:80' has been added.
[11:38:32 DBG] Cluster 'test-api:80' has been removed.
[11:38:45 INF] Received Heartbeat for null
```

### To Reproduce

What steps can we follow to reproduce the issue?
1) Navigate to the tutorial at https://github.com/microsoft/reverse-proxy/blob/main/docs/docfx/articles/kubernetes-ingress.md
2) Find the ```backend.yaml``` file referenced in the tutorial and change the ```port: 80``` field to ```port:5011```
3) Find the  ```ingress-sample.yaml``` file referenced in the tutorial and change the ```number: 80``` field to ```number:5011```
4) Run through the tutorial, when you try to access ```localhost``` you will receive a 404 in the logs. You can revert back to 80 to get it to work

I have tested the configuration with Contour and nginx ingress controllers to confirm it is a valid configuration.

Got Exceptions? Include both the message and the stack trace
No Exceptions, found, logs included above.

### Further technical details

- Include the version of the packages you are using
Built docker images from latest branch as referenced in documentation

- The platform (Linux/macOS/Windows)
- Windows 10 
- Tested on Kind (Kubernetes in Docker) and Docker Desktop for cluster

dpbevin:
I've starting looking into this one. Looks like I need to add Service resource monitoring.

PR #1232, opened 2021-09-05
Add support for Kubernetes Services (#1162)

dpbevin:
- Added watching and caching of Kubernetes Service resources
- Skip reconciliation until Service and Ingress resources are both retrieved
- Changed cluster_id to use Service name and port name/number

For the tests:
- Added Services to Kubernetes unit tests (ingress.yaml files)
- Added test for different port/targetPort configured in Service (as per #1162)


Tratcher:
Thanks for the interest @dpbevin. We're not fully up to speed on the Kubernetes work yet so this might take us a little while to review.

Tratcher (approved):
Nice work. I don't understand all the details, but we'll merge this and let you keep iterating.

PR #1234, opened 2021-09-08
Remove Forwarded header when X-Forwarded transforms added and vice versa

### LINKED ISSUE #1070: All X-Forwarded-* transforms should replace rather than append values, strip values if x-forwarded is disabled.
Default action needs to change.

alnikola:
If X-Forwarded-* transforms are enabled by any way (default or manually), a transform removing Forwarded header is added automatically. On the opposite, if Forwarded transform is added, the transforms removing all X-Forwarded-* headers are added automatically.

Fixes #1070

### DIFF
--- a/src/ReverseProxy/Transforms/Builder/TransformHelpers.cs
+++ b/src/ReverseProxy/Transforms/Builder/TransformHelpers.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Yarp.ReverseProxy.Utilities;
 
 namespace Yarp.ReverseProxy.Transforms.Builder
 {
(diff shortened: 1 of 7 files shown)


Tratcher (approved):

Kahbazi:
One scenario that both forwarder _might_ be useful is when we want to change the forward methods in YARP, but the clusters are not ready yet. So YARP would produce both headers and don't have know when exactly cluster updates the code.

src/ReverseProxy/Transforms/Builder/TransformHelpers.cs
```diff
@@ -23,5 +24,18 @@ public static void CheckTooManyParameters(IReadOnlyDictionary<string, string> ra
...
+        {
+            context.AddXForwardedFor(prefix + ForwardedTransformFactory.ForKey, ForwardedTransformActions.Remove);
+            context.AddXForwardedPrefix(prefix + ForwardedTransformFactory.PrefixKey, ForwardedTransformActions.Remove);
+            context.AddXForwardedHost(prefix + ForwardedTransformFactory.HostKey, ForwardedTransformActions.Remove);
+            context.AddXForwardedProto(prefix + ForwardedTransformFactory.ProtoKey, ForwardedTransformActions.Remove);
+        }
+
+        internal static void RemoveForwadedHeader(TransformBuilderContext context, IRandomFactory randomFactory)
+        {
+            context.RequestTransforms.Add(new RequestHeaderForwardedTransform(randomFactory, NodeFormat.Random, NodeFormat.Random, false, false, ForwardedTransformActions.Remove));
```
Kahbazi, line 38: This might be a far too much effort but we could create a `NullRandomFactory` and then stashed an instance of `RequestHeaderForwardedTransform` which remove the header.

Tratcher: Hmm, it might be worth it if we're going to add this to every route by default.

alnikola: Make sense. I added `NullRandomFactory` and made the transform removing header a singleton.

PR #1236, opened 2021-09-08
Remove cancellation token from IProxyConfigFilter

Kahbazi:
I think we could remove the `CancellationToken` from `IProxyConfigFilter` now. This would make sense before https://github.com/microsoft/reverse-proxy/commit/8841dac2a64b90a0eb6144f229885e832110bace with the present of `IReverseProxyConfigManager` but currently is always `CancellationToken.None`.

https://github.com/microsoft/reverse-proxy/blob/04e0afabf50eccc6e75d1029b2fed6c098bf8005/src/ReverseProxy/Management/ProxyConfigManager.cs#L215-L216

### DIFF
--- a/src/ReverseProxy/Configuration/IProxyConfigFilter.cs
+++ b/src/ReverseProxy/Configuration/IProxyConfigFilter.cs
@@ -15,13 +15,13 @@ namespace Yarp.ReverseProxy.Configuration
         /// Allows modification of a cluster configuration.
         /// </summary>
         /// <param name="cluster">The <see cref="ClusterConfig"/> instance to configure.</param>
-        ValueTask<ClusterConfig> ConfigureClusterAsync(ClusterConfig cluster, CancellationToken cancel);
+        ValueTask<ClusterConfig> ConfigureClusterAsync(ClusterConfig cluster);
 
         /// <summary>
         /// Allows modification of a route configuration.
         /// </summary>
         /// <param name="route">The <see cref="RouteConfig"/> instance to configure.</param>
         /// <param name="cluster">The <see cref="ClusterConfig"/> instance related to <see cref="RouteConfig"/>.</param>
-        ValueTask<RouteConfig> ConfigureRouteAsync(RouteConfig route, ClusterConfig? cluster, CancellationToken cancel);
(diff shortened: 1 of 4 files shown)


Tratcher:
The token is passed to user defined async config filters, the cancellation token seems appropriate if we can come up with guidelines around when it would actually be cancelled. E.g. should it have a timeout?

Kahbazi:
uhmm In that case, how about if config got reloaded in the middle of another reload?

Tratcher:
Interesting idea. The risk there is that if changes were applied too frequently you'd never finish. With the current model, we'd finish what we were applying and check for new updates, possibly skipping some intermediate stale updates.

Kahbazi:
> The risk there is that if changes were applied too frequently you'd never finish. That might be OK?

I guess it's ok given that the file configuration provider have [250ms delay per reload](https://github.com/dotnet/runtime/blob/1705dacc8cd2fdf423e65f230c0f25cbb96714e5/src/libraries/Microsoft.Extensions.Configuration.FileExtensions/src/FileConfigurationProvider.cs#L36) by default, but any other config provider might behave differently.


> With the current model we'd finish what we were applying and check for new updates, possibly skipping some intermediate stale updates.

This sounds like a bug. What if it skip the last update. Maybe we should `RegisterChangeCallback` before `ApplyConfigAsync`?

Another question to answer is that should we only pass the `CancellationToken` to filters and and let the user code decide what to do or we check it to see if it is cancelled on every loop on clusters and routes and throw an exception?

Tratcher:
> This sounds like a bug. What if it skip the last update. Maybe we should `RegisterChangeCallback` before `ApplyConfigAsync`?

To clarify, since every 'update' needs to contain the full config it doesn't matter if we skip some intermediate ones, so long as we check for the latest one.

Tratcher:
> Another question to answer is that should we only pass the `CancellationToken` to filters and and let the user code decide what to do or we check it to see if it is cancelled on every loop on clusters and routes and throw an exception?

Fair. Yes, we should check between filters and one more time before UpdateRuntimeClusters.

adityamandaleeka:
Triage: Closing this PR. @Kahbazi can you please file a new issue to continue the design discussion of when the cancellation should be fired?

PR #1237, opened 2021-09-08
Add WebSockets telemetry

### LINKED ISSUE #1209: WebSockets insights
```
EstablishedTime
ClientCloseTime
ServerCloseTime
CleanupTime
ClientTcpDisconnectTime
MessagesRead
MessagesWritten
BytesRead
BytesWritten
```

Some systems like ARR give you the above stats at the end of a WebSocket request. Today with YARP you'd only be able to get the following because WebSockets are proxied in an opaque way:
```
EstablishedTime (request start time)
BytesRead
BytesWritten
CleanupTime (request end time)
```

To get others we'd have to parse out the WebSocket frames, count messages, and look for the close messages in each direction. We do not want to fully parse messages, or deal with other complexities like masking or compression, so we should not try to do anything based on the full WebSocket API.

Proposal:
- Create an opt-in middleware.
- Have it wrap and replace the IHttpUpgradeFeature kindof like what the [WebSockets middleware](https://github.com/dotnet/aspnetcore/blob/d2c48ca1a7d97ab136dcdf029e96e38ebb16ba21/src/Middleware/WebSockets/src/WebSocketMiddleware.cs#L68-L71) does, except without implementing the IHttpWebSocketFeature.
- On upgrade, verify a WebSockets were negotiated, and then return a wrapper stream around the opaque upgrade stream.
- The wrapper stream acts as a pass through observer and doesn't have its own buffers.
  - The Read code path observes the read result, tracks the frame boundaries, counts the messages, and looks for close frames.
  - The Write code path observes the write calls, tracks the frame boundaries, counts the messages, and looks for close frames.
- Emit the event data at the end of the request.

MihaZupan:
Fixes #1209

I followed the approach outlined in #1209 of wrapping the upgrade feature and MITM the upgrade transport stream.

Public API is just the `UseWebSocketsTelemetry` extension method.
I fed the data to an `EventSource` event, but it could also alternatively be something like `IWebSocketsTelemetryFeature`.

I didn't know where to put the code so I made a `WebSocketsTelemetry` folder in the main project (all types are internal though).

ToDo: I will add tests when we're happy with the overall approach

### DIFF
--- a/samples/ReverseProxy.Metrics.Sample/WebSocketsTelemetryConsumer.cs
+++ b/samples/ReverseProxy.Metrics.Sample/WebSocketsTelemetryConsumer.cs
@@ -1,0 +1,21 @@
+using System;
+using Microsoft.Extensions.Logging;
+using Yarp.Telemetry.Consumption;
+
+namespace Yarp.Sample
+{
+    public sealed class WebSocketsTelemetryConsumer : IWebSocketsTelemetryConsumer
+    {
+        private readonly ILogger<WebSocketsTelemetryConsumer> _logger;
+
+        public WebSocketsTelemetryConsumer(ILogger<WebSocketsTelemetryConsumer> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public void OnWebSocketClosed(DateTime timestamp, DateTime establishedTime, WebSocketCloseReason closeReason, long messagesRead, long messagesWritten)
+        {
+            _logger.LogInformation($"WebSocket connection closed ({closeReason}) after reading {messagesRead} and writing {messagesWritten} messages over {(timestamp - establishedTime).TotalSeconds:N2} seconds.");
+        }
+    }
+}
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -576,7 +576,7 @@ namespace Yarp.ReverseProxy.Forwarder
                 var (secondResult, secondException) = await secondTask;
                 if (secondResult != StreamCopyResult.Success)
                 {
-                    error = ReportResult(context, requestFinishedFirst, secondResult, secondException!);
+                    error = ReportResult(context, !requestFinishedFirst, secondResult, secondException!);
                 }
                 else
                 {
--- a/src/ReverseProxy/WebSocketsTelemetry/WebSocketsParser.cs
+++ b/src/ReverseProxy/WebSocketsTelemetry/WebSocketsParser.cs
@@ -1,0 +1,132 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Diagnostics;
+
+namespace Yarp.ReverseProxy.WebSocketsTelemetry
+{
+    internal unsafe struct WebSocketsParser
+    {
+        private const int MaskLength = 4;
+        private const int MinHeaderSize = 2;
+        private const int MaxHeaderSize = MinHeaderSize + MaskLength + sizeof(ulong);
+
+        private fixed byte _leftoverBuffer[MaxHeaderSize - 1];
+        private readonly byte _minHeaderSize;
+        private byte _leftover;
+        private ulong _bytesToSkip;
+
+        public long MessageCount { get; private set; }
+
+        public DateTime? CloseTime { get; private set; }
+
+        public WebSocketsParser(bool isServer)
+        {
+            _minHeaderSize = (byte)(MinHeaderSize + (isServer ? MaskLength : 0));
+            _leftover = 0;
+            _bytesToSkip = 0;
+            MessageCount = 0;
+            CloseTime = null;
+        }
+
+        // The WebSocket Protocol: https://datatracker.ietf.org/doc/html/rfc6455#section-5.2
+        //  0                   1                   2                   3
+        //  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+        // +-+-+-+-+-------+-+-------------+-------------------------------+
+        // |F|R|R|R| opcode|M| Payload len |    Extended payload length    |
+        // |I|S|S|S|  (4)  |A|     (7)     |             (16/64)           |
+        // |N|V|V|V|       |S|             |   (if payload len==126/127)   |
+        // | |1|2|3|       |K|             |                               |
+        // +-+-+-+-+-------+-+-------------+ - - - - - - - - - - - - - - - +
+        // |     Extended payload length continued, if payload len == 127  |
+        // + - - - - - - - - - - - - - - - +-------------------------------+
+        // |                               |Masking-key, if MASK set to 1  |
+        // +-------------------------------+-------------------------------+
+        // | Masking-key (continued)       |          Payload Data         |
+        // +-------------------------------- - - - - - - - - - - - - - - - +
+        // :                     Payload Data continued ...                :
+        // +---------------------------------------------------------------+
+        //
+        // The header can be 2-10 bytes long, followed by a 4 byte mask if the message was sent by the client.
+        // We have to read the first 2 bytes to know how long the frame header will be.
+        // Since the buffer may not contain the full frame, we make use of a leftoverBuffer
+        // where we store leftover bytes that don't represent a complete frame header.
+        // On the next call to Consume, we interpret the leftover bytes as the beginning of the frame.
+        // As we are not interested in the actual payload data, we skip over (payload length + mask length) bytes after each header.
+        public void Consume(ReadOnlySpan<byte> buffer)
+        {
+            int leftover = _leftover;
+            var bytesToSkip = _bytesToSkip;
+
+            while (true)
+            {
+                var toSkip = Math.Min(bytesToSkip, (ulong)buffer.Length);
+                buffer = buffer.Slice((int)toSkip);
+                bytesToSkip -= toSkip;
+
+                var available = leftover + buffer.Length;
+                int headerSize = _minHeaderSize;
+
+                if (available < headerSize)
+                {
+                    break;
+                }
+
+                var length = (leftover > 1 ? _leftoverBuffer[1] : buffer[1 - leftover]) & 0x7FUL;
+
+                if (length > 125)
+                {
+                    // The actual length will be encoded in 2 or 8 bytes, based on whether the length was 126 or 127
+                    var lengthBytes = 2 << (((int)length & 1) << 1);
+                    headerSize += lengthBytes;
+                    Debug.Assert(leftover < headerSize);
+
+                    if (available < headerSize)
+                    {
+                        break;
+                    }
+
+                    lengthBytes += MinHeaderSize;
+
+                    length = 0;
+                    for (var i = MinHeaderSize; i < lengthBytes; i++)
+                    {
+                        length <<= 8;
+                        length |= i < leftover ? _leftoverBuffer[i] : buffer[i - leftover];
+                    }
+                }
+
+                Debug.Assert(leftover < headerSize);
+                bytesToSkip = length;
+
+                int header = leftover > 0 ? _leftoverBuffer[0] : buffer[0];
+
+                if ((header & 0xF) == 0x8) // CLOSE
+                {
+                    CloseTime ??= DateTime.UtcNow;
+                }
+                else if ((header & 0x80) != 0) // FIN
+                {
+                    MessageCount++;
+                }
+
+                // Advance the buffer by the number of bytes read for the header,
+                // accounting for any bytes we may have read from the leftoverBuffer
+                buffer = buffer.Slice(headerSize - leftover);
+                leftover = 0;
+            }
+
+            Debug.Assert(bytesToSkip == 0 || buffer.Length == 0);
+            _bytesToSkip = bytesToSkip;
+
+            Debug.Assert(leftover + buffer.Length < MaxHeaderSize);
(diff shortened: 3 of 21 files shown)


src/ReverseProxy/WebSocketsTelemetry/WebSocketsParser.cs
```diff
@@ -0,0 +1,105 @@
...
+        public WebSocketsParser(bool isServer)
+        {
+            _minHeaderSize = (byte)(MinHeaderSize + (isServer ? MaskLength : 0));
+            _leftover = 0;
+            _bytesToSkip = 0;
+            MessageCount = 0;
+            CloseTime = null;
+        }
+
+        public void Consume(ReadOnlySpan<byte> buffer)
```
Tratcher, line 33: Can you copy a description or some ascii art here that shows the websocket frame layout for reference?
Also, a link to that RFC section.

src/ReverseProxy/WebSocketsTelemetry/WebSocketsTelemetryMiddleware.cs
```diff
@@ -0,0 +1,80 @@
...
+        public WebSocketsTelemetryMiddleware(RequestDelegate next)
+        {
+            _next = next ?? throw new ArgumentNullException(nameof(next));
+        }
+
+        public Task Invoke(HttpContext context)
+        {
+            if (WebSocketsTelemetry.Log.IsEnabled())
+            {
+                if (context.Features.Get<IHttpUpgradeFeature>() is { } upgradeFeature)
```
Tratcher, line 26:
```suggestion
                if (context.Features.Get<IHttpUpgradeFeature>() is { } upgradeFeature
                    && upgradeFeature.IsUpgradableRequest)
```

src/ReverseProxy/WebSocketsTelemetry/WebSocketsTelemetryMiddleware.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                if (upgradeWrapper.TelemetryStream is { } telemetryStream)
+                {
+                    WebSocketsTelemetry.Log.WebSocketClosed(
+                        telemetryStream.EstablishedTime.Ticks,
+                        telemetryStream.ClientCloseTime?.Ticks ?? -1,
+                        telemetryStream.ServerCloseTime?.Ticks ?? -1,
+                        telemetryStream.MessagesRead,
+                        telemetryStream.MessagesWritten);
+                }
```
Tratcher, line 54: Best practice: Unwrap/cleanup on exit
```suggestion
                }
                context.Features.Set<IHttpUpgradeFeature>(upgradeFeature);
```

src/ReverseProxy/WebSocketsTelemetry/WebSocketsTelemetryMiddleware.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            public async Task<Stream> UpgradeAsync()
+            {
+                Debug.Assert(TelemetryStream is null);
+                var opaqueTransport = await _upgradeFeature.UpgradeAsync();
+                TelemetryStream = new WebSocketsTelemetryStream(opaqueTransport);
+                return TelemetryStream;
+            }
+
+            public bool IsUpgradableRequest => _upgradeFeature.IsUpgradableRequest;
```
Tratcher, line 77: Style: Properties before the constructor.

samples/ReverseProxy.Metrics.Sample/WebSocketsTelemetryConsumer.cs
```diff
@@ -0,0 +1,49 @@
...
+                if (error == ForwarderError.UpgradeRequestClient || error == ForwarderError.UpgradeRequestCanceled ||
+                    error == ForwarderError.UpgradeResponseClient || error == ForwarderError.UpgradeResponseCanceled)
+                {
+                    closedBy = "Client (disconnect)";
+                }
+                else if (error == ForwarderError.UpgradeRequestDestination || error == ForwarderError.UpgradeResponseDestination)
+                {
+                    closedBy = "Server (disconnect)";
+                }
+            }
```
Tratcher, line 44: IForwarderErrorFeature wouldn't be available for out-of-proc event consumers. The ForwarderError should be read by the WebSocketsTelemetryMiddleware and either passed into WebSocketClosed, or else the closedBy should be computed in the middleware and passed to the event as a enum.

Bonus: Then you don't need IHttpContextAccessor.

Tratcher (commented):
It looks like the right structure. Testing is going to be interesting.

Work with the customer to decide the right field types for the telemetry output.

src/ReverseProxy.TelemetryConsumption/TelemetryConsumptionExtensions.cs
```diff
@@ -38,6 +39,12 @@ public static IServiceCollection AddTelemetryConsumer(this IServiceCollection se
         {
             var implementsAny = false;
 
+            if (consumer is IWebSocketsTelemetryConsumer)
+            {
+                services.TryAddEnumerable(new ServiceDescriptor(typeof(IWebSocketsTelemetryConsumer), consumer));
```
Kahbazi, line 44: This is a matter of style. How about this one?
```suggestion
                services.TryAddEnumerable(ServiceDescriptor.Singleton<IWebSocketsTelemetryConsumer>(consumer);
```

samples/ReverseProxy.Metrics.Sample/WebSocketsTelemetryConsumer.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            _logger.LogInformation($"WebSocket connection closed by {closedBy} after reading {messagesRead} and writing {messagesWritten} messages over {(timestamp - establishedTime).TotalSeconds:N2} seconds.");
```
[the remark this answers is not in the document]
MihaZupan: This is in sample code showcasing the use of the event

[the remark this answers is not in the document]
Kahbazi: Oh I missed the sample path.

src/ReverseProxy.TelemetryConsumption/TelemetryConsumptionExtensions.cs
MihaZupan: Since `consumer` in this case is `object`, `.Singleton<T>(T foo)` can't be used.

But it could be changed to
```c#
if (consumer is IWebSocketsTelemetryConsumer webSocketsTelemetryConsumer)
{
    services.TryAddEnumerable(ServiceDescriptor.Singleton(webSocketsTelemetryConsumer));
}
```

which does still seem nicer since we're avoiding the `typeof` 👍.

src/ReverseProxy/WebSocketsTelemetry/WebSocketsTelemetryStream.cs
```diff
@@ -0,0 +1,85 @@
...
+            _readParser = new WebSocketsParser(isServer: true);
+            _writeParser = new WebSocketsParser(isServer: false);
+        }
+
+        public WebSocketCloseReason GetCloseReason(HttpContext context)
+        {
+            var clientCloseTime = _readParser.CloseTime;
+            var serverCloseTime = _writeParser.CloseTime;
+
+            if (clientCloseTime.HasValue && (!serverCloseTime.HasValue || serverCloseTime.Value > clientCloseTime.Value))
```
Tratcher, line 35: The close is only graceful if both sides acknowledge it. Otherwise you can attribute the reason as a disconnect by the unacknowledged side.
```suggestion
            if (clientCloseTime.HasValue && serverCloseTime.HasValue && serverCloseTime.Value > clientCloseTime.Value)
```

MihaZupan: So something like
```c#
if (clientSentClose && serverSentClose)
{
    return clientCloseTime < serverCloseTime ? ClientGraceful : ServerGraceful;
}
else
{
    Disconnect
}
```
?

Tratcher:
```c#
if (clientSentClose && serverSentClose)
{
    return clientCloseTime < serverCloseTime ? ClientGraceful : ServerGraceful;
}
else if (clientSentClose)
{
    ServerDisconnect
}
else if (serverSentClose)
{
   ClientDisconnect
}
else // check forwarder errors
```

MihaZupan: If the client sends a Close message and immediately closes the connection, we would be reporting that as a ServerDisconnect (and vice-versa).

Would that be the desired behavior?

Tratcher: Interesting... how would you detect that? Would the read return 0? Writes might fail, depending on the timing.

MihaZupan: I changed the logic so that it is now
```c#
if (clientSentClose && serverSentClose)
{
    return clientCloseTime < serverCloseTime ? ClientGraceful : ServerGraceful;
}

// One side sent a WebSocket close, but we never saw a response from the other side
// It is possible an error occurred, but we saw a graceful close first, so that is the intiator
if (clientSentClose)
{
    return ClientGraceful;
}
if (serverSentClose)
{
    return ServerGraceful;
}

return forwarderError ...
```

While "graceful" doesn't necessarily mean "both sides sent close", I believe reporting what likely initiated the close is more useful as the `WebSocketCloseReason`.
Either side may still perform protocol violations, but it's not our aim to catch that.

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -576,7 +576,7 @@ private static void RestoreUpgradeHeaders(HttpContext context, HttpResponseMessa
                 var (secondResult, secondException) = await secondTask;
                 if (secondResult != StreamCopyResult.Success)
                 {
-                    error = ReportResult(context, requestFinishedFirst, secondResult, secondException!);
+                    error = ReportResult(context, !requestFinishedFirst, secondResult, secondException!);
```
MihaZupan, line 579: This appears to be a bug where if the request copy finishes first successfully, we will attribute potential response errors to the request side.

src/ReverseProxy/WebSocketsTelemetry/WebSocketsParser.cs
```diff
@@ -0,0 +1,132 @@
...
+                }
+
+                Debug.Assert(leftover < headerSize);
+                bytesToSkip = length;
+
+                int header = leftover > 0 ? _leftoverBuffer[0] : buffer[0];
+
+                if ((header & 0xF) == 0x8) // CLOSE
+                {
+                    CloseTime ??= DateTime.UtcNow;
```
Tratcher, line 107: Consider using IClock for testability.

src/ReverseProxy/WebSocketsTelemetry/WebSocketsParser.cs
```diff
@@ -0,0 +1,132 @@
...
+                }
+                else if ((header & 0x80) != 0) // FIN
```
Tratcher, line 109: Does this mean control frames like Pings are not counted? Should they be?

Tratcher (commented):
I was expecting some unit tests for the WebSocketsParser?

src/ReverseProxy/WebSocketsTelemetry/WebSocketsParser.cs
MihaZupan: As written, it would count control frames (other than Close).
I don't think it should - I'm changing it to ignore control messages.

MihaZupan:
Added more tests

src/ReverseProxy/WebSocketsTelemetry/WebSocketsTelemetryMiddleware.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                _clock = clock ?? throw new ArgumentNullException(nameof(clock));
+                InnerUpgradeFeature = upgradeFeature ?? throw new ArgumentNullException(nameof(upgradeFeature));
+            }
+
+            public async Task<Stream> UpgradeAsync()
+            {
+                Debug.Assert(TelemetryStream is null);
+                var opaqueTransport = await InnerUpgradeFeature.UpgradeAsync();
+                TelemetryStream = new WebSocketsTelemetryStream(_clock, opaqueTransport);
```
Tratcher, line 82: This assumes the upgrade is a websocket. We also support upgrades for SPDY. Check that the response headers include `Upgrade: websocket` before wrapping the stream.

test/ReverseProxy.Tests/WebSocketsTelemetry/WebSocketsParserTests.cs
```diff
@@ -0,0 +1,230 @@
...
+            }
+
+            return header;
+        }
+
+        private ReadOnlySpan<byte> GetCloseFrame() => GetHeader(opcode: 8, length: 0);
+
+        private ReadOnlySpan<byte> GetPingFrame() => GetHeader(opcode: 9, length: 0);
+
+        private ReadOnlySpan<byte> GetPongFrame() => GetHeader(opcode: 10, length: 0);
```
Tratcher, line 60: Pings and pongs may contain data, test with and without.
Close can also contain a message.

Tratcher (approved):
Have you tried running this in the sample with real websocket traffic?

MihaZupan:
Yes, I tried it with a simple demo app and the default Blazor template, and it's working as expected.

PR #1239, opened 2021-09-10
Tests veryfing skip body request behavior in Expect 100 scenarios

### LINKED ISSUE #1200: [HTTP/2] Exception in POST with 100-Continue when server returns 4XX before reading the request body
While implementing, the functional tests for Except: 100-Continue scenarios, it was discovered that in some cases the server can resets the HTTP/2 stream with the error code 'PROTOCOL_ERROR' that is propagated as `IOException` on the original client. This happens when the test returns a 4XX error without reading the request body first. The issue can be reproduced by [this test](https://github.com/microsoft/reverse-proxy/blob/c67c25b7b14165d0c3b406b71e6145238dca86fc/test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs) using the following test arguments:

1. proxyProtocol: Http1, destProtocol: Http2, useContentLength: True, hasResponseBody: True, destResponseCode: 400
```
System.Net.Http.HttpRequestException : Error while copying content to a stream.
    ---- System.IO.IOException : The response ended prematurely.
```
2. proxyProtocol: Http2, destProtocol: Http2, useContentLength: True, hasResponseBody: True, destResponseCode: 400
```
System.Net.Http.HttpRequestException : Error while copying content to a stream.
    ---- System.IO.IOException : The request was aborted.
    -------- System.Net.Http.Http2StreamException : The HTTP/2 server reset the stream. HTTP/2 error code 'PROTOCOL_ERROR' (0x1).
```

alnikola:
Fixes #1200

### DIFF
--- a/test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
+++ b/test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
@@ -177,6 +177,7 @@ namespace Yarp.ReverseProxy
             });
         }
 
+        // Fix was implemented in https://github.com/dotnet/runtime/pull/58548
 #if NET7_0_OR_GREATER
         [Theory]
         [InlineData(HttpProtocols.Http1, HttpProtocols.Http1, false)]
--- a/test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
+++ b/test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
@@ -177,6 +177,66 @@ namespace Yarp.ReverseProxy
             });
         }
 
+#if NET7_0_OR_GREATER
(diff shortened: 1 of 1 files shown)


test/ReverseProxy.FunctionalTests/Expect100ContinueTests.cs
```diff
@@ -177,6 +177,66 @@ public async Task PostExpect100_ResponseWithPayload(HttpProtocols proxyProtocol,
             });
         }
 
+#if NET7_0_OR_GREATER
```
Tratcher, line 180: Include a comment linking to the 7.0 HttpClient fix PR.

Tratcher (approved):

PR #1246, opened 2021-09-15
Use the same default transformer for HttpForwarder and the whole YARP

### LINKED ISSUE #1093: Inconsistent header transforms defaults between YARP and Forwarder
YARP Direct Proxy for one of the endpoint screams with the following error:

> The SSL connection could not be established, see inner exception

Inner exception says:

> Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host.

In the current, non-YARP proxy solution, the problem is solved (or rather, by-passed) by setting the `ServerCertificateCustomValidationCallback` flag:

```c#
public class IgnoreSslHandler : HttpClientHandler
{
    public IgnoreSslHandler()
    {
        if (Environment.IsDevelopment())
        {
            ServerCertificateCustomValidationCallback = DangerousAcceptAnyServerCertificateValidator;
        }
    }
}
```

But, `HttpMessageInvoker` with the `HttpClientHandler` that has the `ServerCertificateCustomValidationCallback ` flag set to either true or to `DangerousAcceptAnyServerCertificateValidator` doesn't seem to be used in the YARP.

Code snippet that is used for proxy context:

```c#
await httpProxy.ProxyAsync(httpContext, "https://url", httpClient, requestOptions, transformer);
```

Any idea how to solve that?

alnikola:
Fixes #1093

### DIFF
--- a/samples/ReverseProxy.Direct.Sample/README.md
+++ b/samples/ReverseProxy.Direct.Sample/README.md
@@ -37,7 +37,7 @@ Normal proxying comprises the following steps:
 
 To enable control over mapping request and response fields and headers between the client and destination (steps 4 and 7 above), the HttpForwarder.ProxyAsync method takes a HttpTransformer. Your implementation can modify the request url, method, protocol version, response status code, or decide which headers are copied, modify them, or insert additional headers as required.
 
-**Note:** When using the HttpForwarder class directly there are some [header transforms](https://microsoft.github.io/reverse-proxy/articles/transforms.html) included by default, such as adding ```X-Forwarded-For``` and removing the original Host header. The same set of transforms is by default included into the alternate YARP pipeline model (see BasicYarpSample).
+**Note:** When using the HttpForwarder class directly there are some [header transforms](https://microsoft.github.io/reverse-proxy/articles/transforms.html) included by default, such as adding ```X-Forwarded-For``` and removing the original Host header. This is the same set of transforms included by default in the YARP pipeline model (see BasicYarpSample).
 
 ## Files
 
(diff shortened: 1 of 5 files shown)


src/ReverseProxy/Forwarder/HttpTransformer.cs
```diff
@@ -20,7 +21,12 @@ public class HttpTransformer
         /// A default set of transforms that copies all request and response fields and headers, except for some
```
Tratcher, line 21: Update the comment too.

Tratcher (approved):

Kahbazi:
Does this also need to be updated?
https://github.com/microsoft/reverse-proxy/blob/6a7b4fe43a0d1136897aeba0f6bcf532657522e1/samples/ReverseProxy.Direct.Sample/README.md#L40

src/ReverseProxy/Forwarder/HttpTransformer.cs
alnikola: Updated

samples/ReverseProxy.Direct.Sample/README.md
```diff
@@ -35,9 +35,9 @@ Normal proxying comprises the following steps:
...
+To enable control over mapping request and response fields and headers between the client and destination (steps 4 and 7 above), the HttpForwarder.ProxyAsync method takes a HttpTransformer. Your implementation can modify the request url, method, protocol version, response status code, or decide which headers are copied, modify them, or insert additional headers as required.
 
-**Note:** When using the HttpProxy class directly there are no transforms included by default, you have full control of the transforms in your HttpTransformer implementation. The alternate YARP pipeline model (see BasicYarpSample), has some [default header transforms](https://microsoft.github.io/reverse-proxy/articles/transforms.html), such as adding ```X-Forwarded-For``` and removing the original Host header.
+**Note:** When using the HttpForwarder class directly there are some [header transforms](https://microsoft.github.io/reverse-proxy/articles/transforms.html) included by default, such as adding ```X-Forwarded-For``` and removing the original Host header. The same set of transforms is by default included into the alternate YARP pipeline model (see BasicYarpSample).
```
alnikola, line 40: @Tratcher @Kahbazi Does it look good now?

Tratcher:
```suggestion
**Note:** When using the HttpForwarder class directly there are some [header transforms](https://microsoft.github.io/reverse-proxy/articles/transforms.html) included by default, such as adding ```X-Forwarded-For``` and removing the original Host header. This is the same set of transforms included by default in the YARP pipeline model (see BasicYarpSample).
```

Kahbazi (approved):

PR #1247, opened 2021-09-15
Fix many docfx warnings

### LINKED ISSUE #1067: docfx doesn't recognize Microsoft.* namespaces in xrefs

alnikola:
Fixes most of the docfx warnings. It leaves only 5 active warnings which complain about missing metadata references in `src/OperatorFramework/examples/BasicOperator/BasicOperator.csproj`. They can be ignored for now.

Contributes to #1067

### DIFF
--- a/src/OperatorFramework/src/Core/GroupKindNamespacedName.cs
+++ b/src/OperatorFramework/src/Core/GroupKindNamespacedName.cs
@@ -87,13 +87,11 @@ namespace Microsoft.Kubernetes
                 NamespacedName.From(resource));
         }
 
-        /// <inheritdoc/>
         public override bool Equals(object obj)
         {
             return obj is GroupKindNamespacedName name && Equals(name);
         }
 
-        /// <inheritdoc/>
         public bool Equals([AllowNull] GroupKindNamespacedName other)
         {
             return Group == other.Group &&
@@ -101,7 +99,6 @@ namespace Microsoft.Kubernetes
                    NamespacedName.Equals(other.NamespacedName);
         }
 
-        /// <inheritdoc/>
         public override int GetHashCode()
         {
             return HashCode.Combine(Group, Kind, NamespacedName);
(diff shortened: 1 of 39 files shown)


src/OperatorFramework/src/Core/GroupKindNamespacedName.cs
```diff
@@ -87,21 +87,18 @@ public static GroupKindNamespacedName From(IKubernetesObject<V1ObjectMeta> resou
                 NamespacedName.From(resource));
         }
 
-        /// <inheritdoc/>
```
Tratcher, line 90: What's wrong with inheritdoc?

Tratcher (approved):
Should that be "contributes to #"? If we're not going to fix all of the warnings yet then the issue should stay open, even if moved to the backlog.

src/OperatorFramework/src/Core/GroupKindNamespacedName.cs
alnikola: It doesn't find all of them. I removed the ones docfx complained about.

alnikola:
>If we're not going to fix all of the warnings yet then the issue should stay open, even if moved to the backlog.

@Tratcher Are you sure we are going to fix them for 1.0? It's 1 project inside Kubernetes integration package.

Tratcher:
We don't have to fix them for 1.0, but the tracking issue should stay open and indicate what work remains. We can re-triage the issue to decide when to do it.

alnikola:
> We don't have to fix them for 1.0, but the tracking issue should stay open and indicate what work remains.
Got it. I changed it to Contributes to.

PR #1257, opened 2021-09-21
Transforms in "Always" mode are run even if there is no response

### LINKED ISSUE #1056: No response transforms if the destination fails to respond
https://github.com/microsoft/reverse-proxy/issues/1051#issuecomment-854811689 (@brentnewbury-work)

### Describe the bug
Response transforms are skipped if the destination fails to respond due to a connection issue or timeout. A fixed status code is set and processing exits:
https://github.com/microsoft/reverse-proxy/blob/2fd46596c7cc370660f024e442ac3c8c00b88acf/src/ReverseProxy/Proxy/HttpProxy.cs#L134-L143
https://github.com/microsoft/reverse-proxy/blob/2fd46596c7cc370660f024e442ac3c8c00b88acf/src/ReverseProxy/Proxy/HttpProxy.cs#L460-L477

Some response transforms like ResponseHeader have the setting to run on "Success" or "Always", but always does not currently include the failure-to-respond scenario. It might even make sense to add a "Failure" mode to only run transforms for 5xx and connection failures. 

I'm mainly concerned about the "Always" case since people expect it run and generate consistent output. Failure cases can already be handled by the caller, but you don't want to duplicate logic for the Always scenario.

We would want to make the ProxyError value available to these transforms.

alnikola:
Fixes #1056

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -144,7 +144,7 @@ namespace Yarp.ReverseProxy.Forwarder
                 }
                 catch (Exception requestException)
                 {
-                    return await HandleRequestFailureAsync(context, requestContent, requestException);
+                    return await HandleRequestFailureAsync(context, requestContent, requestException, transformer);
                 }
                 finally
                 {
@@ -457,7 +457,7 @@ namespace Yarp.ReverseProxy.Forwarder
             return requestBodyError;
         }
 
-        private async ValueTask<ForwarderError> HandleRequestFailureAsync(HttpContext context, StreamCopyHttpContent? requestContent, Exception requestException)
+        private async ValueTask<ForwarderError> HandleRequestFailureAsync(HttpContext context, StreamCopyHttpContent? requestContent, Exception requestException, HttpTransformer transformer)
         {
             // Check for request body errors, these may have triggered the response error.
             if (requestContent?.ConsumptionTask.IsCompleted == true)
@@ -473,6 +473,8 @@ namespace Yarp.ReverseProxy.Forwarder
             // We couldn't communicate with the destination.
             ReportProxyError(context, ForwarderError.Request, requestException);
             context.Response.StatusCode = StatusCodes.Status502BadGateway;
+
+            await transformer.TransformResponseAsync(context, null);
 
             return ForwarderError.Request;
         }
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -131,16 +131,23 @@ namespace Yarp.ReverseProxy.Forwarder
                 }
                 catch (OperationCanceledException canceledException)
                 {
+                    ForwarderError error;
                     if (!requestAborted.IsCancellationRequested && requestTimeoutToken.IsCancellationRequested)
                     {
                         ReportProxyError(context, ForwarderError.RequestTimedOut, canceledException);
                         context.Response.StatusCode = StatusCodes.Status504GatewayTimeout;
-                        return ForwarderError.RequestTimedOut;
+                        error = ForwarderError.RequestTimedOut;
+                    }
+                    else
+                    {
+                        ReportProxyError(context, ForwarderError.RequestCanceled, canceledException);
+                        context.Response.StatusCode = StatusCodes.Status502BadGateway;
+                        error = ForwarderError.RequestCanceled;
                     }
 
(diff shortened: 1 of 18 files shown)


src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -144,7 +144,7 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
                 }
                 catch (Exception requestException)
```
Tratcher, line 145: What about the OperationCanceledException code path?

src/ReverseProxy/Transforms/ResponseTrailersTransform.cs
```diff
@@ -41,6 +41,12 @@ public static StringValues TakeHeader(ResponseTrailersTransformContext context,
             }
 
             var existingValues = StringValues.Empty;
+
+            if (context.ProxyResponse == null)
```
Tratcher, line 45: Debug.Assert.

src/ReverseProxy/Transforms/ResponseTransform.cs
```diff
@@ -39,6 +39,12 @@ public static StringValues TakeHeader(ResponseTransformContext context, string h
             }
 
             var existingValues = StringValues.Empty;
+
+            if (context.ProxyResponse == null)
```
Tratcher, line 43: This check could be delayed until the `else if` block. We want to avoid making assumptions about the HttpContext content.

src/ReverseProxy/Transforms/ResponseHeaderRemoveTransform.cs
```diff
@@ -35,7 +35,7 @@ public override ValueTask ApplyAsync(ResponseTransformContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            if (Always || Success(context))
+            if (context.ProxyResponse != null && (Always || Success(context)))
```
Tratcher, line 38: This check isn't necessary, ProxyResponse isn't accessed on this code path.

src/ReverseProxy/Transforms/ResponseTrailersAllowedTransform.cs
```diff
@@ -41,6 +41,11 @@ public override ValueTask ApplyAsync(ResponseTrailersTransformContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
+            if (context.ProxyResponse == null)
```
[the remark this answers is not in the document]
alnikola: Replaced all null checks with Debug.Assert in the response trailers transforms. The relevant tests are also removed since we don't expect this code to be called when context.ProxyResponse is null.

src/ReverseProxy/Transforms/ResponseHeaderRemoveTransform.cs
alnikola: Fixed

src/ReverseProxy/Transforms/ResponseTransform.cs
alnikola: Fixed

src/ReverseProxy/Forwarder/HttpForwarder.cs
alnikola: Added call to the response transformer there too.

test/ReverseProxy.Tests/Transforms/ResponseHeaderValueTransformTests.cs
```diff
@@ -12,35 +12,38 @@ public class ResponseHeaderValueTransformTests
...
+        [InlineData("start", 200, "new", true, true, "start;new", false)]
+        [InlineData("start,value", 400, "new", true, false, "start,value", false)]
+        [InlineData("start,value", 200, "new", true, false, "start,value;new", false)]
+        [InlineData("start,value", 400, "new", true, true, "start,value;new", false)]
+        [InlineData("start,value", 200, "new", true, true, "start,value;new", false)]
+        [InlineData("start;value", 400, "new", true, false, "start;value", false)]
+        [InlineData("start;value", 200, "new", true, false, "start;value;new", false)]
+        [InlineData("start;value", 400, "new", true, true, "start;value;new", false)]
+        [InlineData("start;value", 200, "new", true, true, "start;value;new", false)]
+        public async Task AddResponseHeader_Success(string startValue, int status, string value, bool append, bool always, string expected, bool responseNull)
```
MihaZupan, line 38:
```suggestion
        public async Task AddResponseHeader_Success(string startValue, int status, string value, bool append, bool always, string expected, bool responseNull = false)
```
Nit: May be more readable since only a handful of cases set it.

MihaZupan (approved):

PR #1263, opened 2021-09-22
Add support for extended YARP config in Kubernetes via annotations

macsux:
Adds support for YARP extended configuration in Kubernetes Ingress resource via annotations. Since some config is complex, it is defined as JSON blocks.

Also since some config is inherently scoped to a route or cluster, use of such annotations would implicitly mean that it is applicable to all paths defined in the resource. If other settings need to be applied to different routes, they need to be isolated as separate Kubernetes resources. Doesn't seem to be any way around these limitations as Ingress schema does not have any extension points that would work for this in the body of the "spec" and the use of Kubernetes structured schemas would not allow us to extend the schema in place.

Some areas of concern is System.Text.Json struggles with serialization / deserialization of `Flags` enums such as for `HttpClient:SslProtocols`  (workaround is to use numeric value).


src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -149,6 +160,34 @@ private static YarpIngressOptions HandleAnnotations(YarpIngressContext context,
...
+            {
+                options.AuthorizationPolicy = authorizationPolicy;
+            }
+            if (annotations.TryGetValue("yarp.ingress.kubernetes.io/cors-policy", out var corsPolicy))
+            {
+                options.CorsPolicy = corsPolicy;
+            }
+            if (annotations.TryGetValue("yarp.ingress.kubernetes.io/session-affinity", out var sessionAffinity))
+            {
+                options.SessionAffinity = JsonSerializer.Deserialize<SessionAffinityConfig>(sessionAffinity, jsonOptions);
```
Tratcher, line 177: What happens if Deserialize fails for any of these? Where is the exception caught and how gracefully can it recover?

src/Kubernetes.Controller/Yarp.Kubernetes.Controller.csproj
```diff
@@ -18,6 +19,8 @@
         <PackageReference Include="Serilog.Formatting.Compact" Version="1.1.0" />
         <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
         <PackageReference Include="Nerdbank.GitVersioning" Version="3.3.37" />
+        <PackageReference Condition=" '$(TargetFramework)' != 'net6.0' " Include="System.Text.Json" Version="6.0.0-*" />
```
Tratcher, line 22: What does using the 6.0 json package get you on 3.1 or 5.0? System.Text.Json should already be available in the runtime. Are there issues in the old implementations you're working around?

Keep in mind that these projects don't target net6.0 yet so the conditional isn't doing anything.

src/Kubernetes.Controller/Dispatching/DispatchActionResult.cs
```diff
@@ -25,6 +26,7 @@ public class DispatchActionResult : IActionResult, IDispatchTarget
         private Task _task = Task.CompletedTask;
         private readonly object _taskSync = new();
         private HttpContext _httpContext;
+        private static readonly JsonSerializerOptions _jsonOptions = new() {Converters = {new JsonStringEnumConverter()}};
```
Tratcher, line 29: Include a comment here on why the converter is needed.

src/Kubernetes.Controller/Yarp.Kubernetes.Controller.csproj
macsux: `JsonSerializer` didn't handle `TimeSpan` serialization before net6.0. Conditional is future-proofing.

src/Kubernetes.Controller/Dispatching/DispatchActionResult.cs
macsux: Some options use enums, and we want to be able to use strings in annotation json instead of numeric values. Did you want me to put comment in the code itself?

src/Kubernetes.Controller/Converters/YarpParser.cs
macsux: The error would bubble up to `IngressController` and cause it to get requeued for processing. I guess this can be considered a non-transient error, so no point in retrying it any further. Propose adding a second catch to `Reconciler` [here](https://github.com/microsoft/reverse-proxy/blob/dc4f55243b66f929367aa6871870558f61add707/src/Kubernetes.Controller/Services/Reconciler.cs#L64) to log warning and stop processing
```csharp
            catch (JsonException ex)
            {
                _logger.LogWarning(ex.Message);
            }
```

src/Kubernetes.Controller/Dispatching/DispatchActionResult.cs
Tratcher: Yes please.

test/Kubernetes.Tests/IngressConversionTests.cs
```diff
@@ -4,35 +4,51 @@
...
 using System.Collections.Generic;
 using System.IO;
+using System.Text;
 using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Threading.Tasks;
+using FluentAssertions.Json;
 using k8s;
 using k8s.Models;
+using Newtonsoft.Json;
```
nitinsapru, line 14: Is there a reason why both Newtonsoft and System.Text.Json are used in this class ? Seems like additional overhead during the packaging to add two different types of serializers.

macsux:
It's used for assertions that exist in Newtonsoft.Json to allow deep
comparisons of json trees. This is just for tests.

src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -149,6 +160,34 @@ private static YarpIngressOptions HandleAnnotations(YarpIngressContext context,
             {
                 options.Https = http.Equals("https", StringComparison.OrdinalIgnoreCase);
             }
+            if (annotations.TryGetValue("yarp.ingress.kubernetes.io/transforms", out var transforms))
```
dpbevin, line 163: Wouldn't the options vary wildly for transforms?

Would it be more consumable if the annotation was more specific, e.g. `yarp.ingress.kubernetes.io/transforms/pathRemovePrefix`?

macsux: Well, transforms are one of the extension points, so you can't even anticipate all the config that can appear in that block. Overall we're trying to pigeon-hole a WHOLE bunch of configurations into unstructured metadata as the default Ingress in Kubernetes is so basic. But if people want to do it this way, we give them that option via this pull request. 

IMO the whole approach around Ingress resource is flawed as it's meant to act as a very rudimentary primitive for Layer 7 HTTP ingress that other components can use without knowledge of the actual ingress controller implementation. It somehow devolved into stuffing implementation-specific configuration that fundamentally affects routing behavior in a way that implementations are no longer interchangeable. Personally, I think only what is supported by the `Ingress` resources should be configurable and nothing more. If you want extra features of individual implementation, a YARP specific CRD should be used to specifically exposes YARP configuration features as structured schema instead of this metaprogramming we're doing now...  /rant off

dpbevin: @macsux I understand the limitations of the Kubernetes Ingress resource, so I understand your frustration around having to "stuff things into annotations".

To give you some more context to my comment, I currently use Nginx as my Kube Ingress Controller. About 90% of my ingress resources have to include a regex to remove the start of the original path. This requires two separate annotations: one to enable regex, the other to specify the rewrite path. And then there's the regex capture in the path:

```
metadata:
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /$2
    nginx.ingress.kubernetes.io/use-regex: "true"
spec:
  rules:
  - http:
      paths:
      - path: /myappone(/|$)(.*)
```

So having a simple (one, non-JSON) annotation to remove a path prefix is very appealing to me because it's easy to read and I'm less likely to get a string incorrect than getting JSON formatting inside YAML incorrect.

What's good about this PR is that is it would give access to every possible transform we could ever imagine. But does that mean we shouldn't make life simple for the 80% of the use cases as well? Could there be an DI-injected interface that provides a fallback for the non-trivial annotations?

test/Kubernetes.Tests/IngressConversionTests.cs
```diff
@@ -45,11 +61,39 @@ private static void VerifyRoutes(string routesJson, string name)
...
         private static void VerifyJson(string json, string name, string fileName)
         {
             var other = File.ReadAllText(Path.Combine("testassets", name, fileName));
+            json = StripNullProperties(json);
+            other = StripNullProperties(other);
 
-            Assert.True(JToken.DeepEquals(JToken.Parse(json), JToken.Parse(other)));
+            var actual = JToken.Parse(json);
+            var jOther = JToken.Parse(other);
+            actual.Should().BeEquivalentTo(jOther);
```
MihaZupan, line 96: Is this just a style difference over `Assert.Equals(other, actual)` / `JToken.DeepEquals`?
If so, we shouldn't add a reference to `FluentAssertions` for it.

MihaZupan (commented):
Sorry for the long delay. We made style changes across the repo (file scoped namespaces), so this PR will need to be rebased.

test/Kubernetes.Tests/IngressConversionTests.cs
macsux: It goes beyond that as when the assertion fails it tells you exactly which section of JSON doesn't match like this:
```
JSON document has a different value at $.a.
Actual document
{
  "a": "a"
}
was expected to be equivalent to
{
  "a": "b"
}
.
```

MihaZupan: (gentle ping) @macsux  would you be able to resolve the merge conflicts here?

macsux: Rebased

macsux:
One thing that may be worth improving is that hierarchical properties are currently expressed as JSON blocks in the annotations via multiline YAML, which is suboptimal as we're mixing multiple serialization formats. This could be rewritten to take a multiline YAML block instead, but it would require taking on dependency on YamlDotNet nuget package. Potentially use another annotation to determine nested serialization format - maybe some like copy / posting json.

test/Kubernetes.Tests/IngressConversionTests.cs
```diff
@@ -58,11 +74,39 @@ private static void VerifyRoutes(string routesJson, string name)
         VerifyJson(routesJson, name, "routes.json");
     }
 
+        private static string StripNullProperties(string json)
```
MihaZupan, line 77: Nit: Indentation is off on this method

src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -126,6 +137,7 @@ private static string FixupPathMatch(V1HTTPIngressPath path)
     private static YarpIngressOptions HandleAnnotations(YarpIngressContext context, V1ObjectMeta metadata)
     {
         var options = context.Options;
+        var jsonOptions = new JsonSerializerOptions { Converters = { new JsonStringEnumConverter() } };
```
MihaZupan, line 140: Can we store this into a static field aswell? `System.Text.Json` uses the options as a cache.

src/Kubernetes.Controller/Dispatching/DispatchActionResult.cs
```diff
@@ -57,7 +60,7 @@ public async Task ExecuteResultAsync(ActionContext context)
             var utf8Bytes = JsonSerializer.SerializeToUtf8Bytes(new Message
             {
                 MessageType = MessageType.Heartbeat
-            });
+                }, _jsonOptions);
```
MihaZupan, line 63:
```suggestion
            }, _jsonOptions);
```

MihaZupan (approved):
> This could be rewritten to take a multiline YAML block instead

I can see the appeal in staying with just Yaml in the config file instead of mixing formats.

What would be your preference? @dpbevin any thoughts on this?

src/Kubernetes.Controller/Converters/YarpParser.cs
MihaZupan: My understanding is that configuration like the generic `yarp.ingress.kubernetes.io/transforms` can coexist with more specific transforms, e.g. `yarp.ingress.kubernetes.io/transforms/pathRemovePrefix`.

As in, support for easily configuring common transforms can be added on top of this PR.

MihaZupan:
> This could be rewritten to take a multiline YAML block instead

@macsux We discussed this in triage as well and everyone feels like staying with Yaml instead of mixing formats is the cleaner option here.
`YamlDotNet` is a well-established package so I have no pushback against using it for deserializing here. My understanding is that the difference from code perspective is swapping `Json.Deserialize<T>` to `Yaml.Deserialize<T>`, does that sound right?

MihaZupan:
Gentle ping @macsux

macsux:
Bah, fell off my radar with other things. I've made the change to YAML, hopefully this is good to merge now.

dpbevin:
@MihaZupan good suggestion on not mixing formats. Thanks for sticking with this one @macsux 👍

src/Kubernetes.Controller/Dispatching/DispatchActionResult.cs
```diff
@@ -25,6 +26,7 @@ public class DispatchActionResult : IActionResult, IDispatchTarget
     private Task _task = Task.CompletedTask;
     private readonly object _taskSync = new();
     private HttpContext _httpContext;
+    // some config options use enums, we need to enable conversion from string based representations
```
MihaZupan, line 29:
```suggestion
```

src/Kubernetes.Controller/Dispatching/DispatchActionResult.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Text;
 using System.Text.Json;
+using System.Text.Json.Serialization;
```
MihaZupan, line 7:
```suggestion
```

src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -134,7 +147,36 @@ private static YarpIngressOptions HandleAnnotations(YarpIngressContext context,
 
         if (annotations.TryGetValue("yarp.ingress.kubernetes.io/backend-protocol", out var http))
         {
-            options.Https = http.Equals("https", StringComparison.OrdinalIgnoreCase);
+        	options.Https = http.Equals("https", StringComparison.OrdinalIgnoreCase);
+        }
+        if (annotations.TryGetValue("yarp.ingress.kubernetes.io/transforms", out var transforms))
+        {
+
```
MihaZupan, line 154:
```suggestion
```

src/Kubernetes.Controller/Yarp.Kubernetes.Controller.csproj
```diff
@@ -10,13 +10,17 @@
...
       <ProjectReference Include="..\Kubernetes.Protocol\Yarp.Kubernetes.Protocol.csproj" />
+
     </ItemGroup>
 
     <ItemGroup>
         <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.10.12" />
         <PackageReference Include="Serilog.Extensions.Logging" Version="3.0.1" />
         <PackageReference Include="Serilog.Formatting.Compact" Version="1.1.0" />
         <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
+        <PackageReference Condition=" '$(TargetFramework)' != 'net6.0' " Include="System.Text.Json" Version="6.0.0-*" />
```
MihaZupan, line 21:
```suggestion
```

src/Kubernetes.Controller/Yarp.Kubernetes.Controller.csproj
```diff
@@ -10,13 +10,17 @@
     <ItemGroup>
       <ProjectReference Include="..\OperatorFramework\src\Controller\Microsoft.Kubernetes.Controller.csproj" />
       <ProjectReference Include="..\Kubernetes.Protocol\Yarp.Kubernetes.Protocol.csproj" />
+
```
MihaZupan, line 13:
```suggestion
```

src/Kubernetes.Protocol/Receiver.cs
```diff
@@ -5,6 +5,8 @@
 using System.IO;
 using System.Net.Http;
 using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
```
MihaZupan, line 9:
```suggestion
```

src/Kubernetes.Protocol/Yarp.Kubernetes.Protocol.csproj
```diff
@@ -8,6 +8,10 @@
   <ItemGroup>
     <ProjectReference Include="..\OperatorFramework\src\Controller\Microsoft.Kubernetes.Controller.csproj" />
     <ProjectReference Include="..\ReverseProxy\Yarp.ReverseProxy.csproj" />
+
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Condition=" '$(TargetFramework)' != 'net6.0' " Include="System.Text.Json" Version="6.0.0-*" />
+  </ItemGroup>
```
MihaZupan, line 16:
```suggestion
```

src/Kubernetes.Protocol/Yarp.Kubernetes.Protocol.csproj
```diff
@@ -8,6 +8,10 @@
   <ItemGroup>
     <ProjectReference Include="..\OperatorFramework\src\Controller\Microsoft.Kubernetes.Controller.csproj" />
     <ProjectReference Include="..\ReverseProxy\Yarp.ReverseProxy.csproj" />
+
```
MihaZupan, line 11:
```suggestion
```

src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -29,6 +34,7 @@ internal static void ConvertFromKubernetesIngress(YarpIngressContext ingressCont
         {
             HandleIngressRule(ingressContext, ingressContext.Endpoints, defaultSubsets, rule, configContext);
         }
+
```
MihaZupan, line 37:
```suggestion
```

src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -1,15 +1,20 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.Json;
+using System.Text.Json.Serialization;
```
MihaZupan, line 5:
```suggestion
```

MihaZupan (approved):
Thanks @macsux! This LGTM.

I'm just removing any references to `System.Text.Json` from product code that shouldn't be needed anymore.

src/Kubernetes.Controller/Converters/YarpParser.cs
macsux: Why do you want to remove the transforms block?

MihaZupan: Oh I just meant the blank line

test/Kubernetes.Tests/testassets/annotations/clusters.json
```diff
@@ -0,0 +1,47 @@
+[
+  {
+    "ClusterId": "frontend:80",
```
MihaZupan, line 3:
```suggestion
    "ClusterId": "frontend.default:80",
```

test/Kubernetes.Tests/testassets/annotations/routes.json
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "RouteId": "minimal-ingress:/foo",
```
MihaZupan, line 3:
```suggestion
    "RouteId": "minimal-ingress.default:/foo",
```

test/Kubernetes.Tests/testassets/annotations/routes.json
```diff
@@ -0,0 +1,23 @@
...
+    "Match": {
+      "Methods": null,
+      "Hosts": [],
+      "Path": "/foo/{**catch-all}",
+      "Headers": null
+    },
+    "Order": null,
+    "ClusterId": "frontend:80",
```
MihaZupan, line 11:
```suggestion
    "ClusterId": "frontend.default:80",
```

test/Kubernetes.Tests/testassets/annotations/ingress.yaml
```diff
@@ -0,0 +1,81 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: minimal-ingress
+  namespace: default
+  annotations:
+    yarp.ingress.kubernetes.io/authorization-policy: authzpolicy
+    yarp.ingress.kubernetes.io/cors-policy: corspolicy
+    yarp.ingress.kubernetes.io/load-balancing: Random
+    yarp.ingress.kubernetes.io/health-check: |
```
MihaZupan, line 10: Is the `|` here still needed?

macsux: yes, because the content is just string. `|` denotes start of multiline string, and it just so happens that our string is yaml block that gets it's own parsing. the inner block will have no validation applied to it when `ingress` resource is submitted though, so if they made a syntax mistake in it they won't know till code runs.

MihaZupan:
Thanks!

PR #1265, opened 2021-09-22
Move to 1ES servicing pools in release/latest

jonfortescue:
We're required to move all servicing branches to the new 1ES servicing pools by September 30. This accomplishes that for release/latest.

Tracking issue: https://github.com/dotnet/core-eng/issues/14276

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -47,11 +47,11 @@ stages:
       - job: Windows
         pool:
           ${{ if eq(variables['System.TeamProject'], 'public') }}:
-            name: NetCorePublic-Pool
-            queue: BuildPool.Windows.10.Amd64.VS2019.Pre.Open
+            name: NetCore1ESPool-Svc-Public
+            demands: ImageOverride -equals Build.Windows.10.Amd64.VS2019.Pre.Open
           ${{ if ne(variables['System.TeamProject'], 'public') }}:
-            name: NetCoreInternal-Pool
-            queue: BuildPool.Windows.10.Amd64.VS2019.Pre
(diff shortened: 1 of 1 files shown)


MihaZupan:
We're not actively producing builds from `release*` branches, perhaps this should go into `main` instead?

jonfortescue:
@MihaZupan I show a build from this branch on Sep 2 &ndash; is that a one-off? We'll be shutting down these pools soon so if there are any future builds from `release*` branches, they will fail without a change.

If you're sure that there won't be any builds, though, then yes; this can be closed and we'll make the switch to the R&D pools in main.

MihaZupan:
There would only be a new build on that branch if we do a servicing release of an existing preview, which is rare. If we do, we can cherry-pick the pool update there.
We will merge `main` into `release/latest` within the next few weeks.

jonfortescue:
To be clear, the pools for `main` and the `release*` branches are different. We would use the R&D pools in `main` and the servicing pools in the `release*` branches, so a pure cherry-pick from main will be inadequate.

MihaZupan:
Oh okay, in that case doing it just in release makes sense, thanks for the explanation

jonfortescue:
@MihaZupan any chance w e can get this merged in?

MihaZupan (approved):

MihaZupan:
@jonfortescue Is the timeline known for when all branches (including non-servicing) will be moved onto the new pools?

jonfortescue:
@MihaZupan best I can give you is "as soon as possible." We're working every day on moving all active servicing branches over first but we have ~80 repos we're working in, so it's going to take some time.

Issue #1235, opened 2021-09-08
Header Match - Support notContains and other modes


ksauce:
### What should we add or change to make your life better?

There are scenarios to route based on a Header match modes:

NotContains
notContains: this is a modified NOT version of contains. This is not a strict NOT logic of contains rule because we want to maintains the AND relationships between items in the dictionary.

It will match only if the following are satisfied:

Header key does not exist.
For all header keys that exist, the value substrings are not found in header value.
In the example below, if request contains fooHeader: fooValue, barHeader: badValue, it would not be considered as match since foo is a substring of fooValue.

{
  "headers": {
    "notContains": {
      "fooHeader": "foo",
      "barHeader": "bar",
    }
  }
}
Psedocode would be

bool match = (!(headers?["fooHeader"].hasSubString("foo") ?? true) &&
              !(headers?["barHeader"].hasSubString("bar") ?? true))

AnyContains
anyContains: ALL header keys must be present. ANY one of the values for each of the header keys must be present as substring. All keys must have a substring matched.

{
  "headers": {
    "anyContains": {
      "fooHeader": [
        "foo1",
        "foo2"
        ],
      "barHeader": [
        "bar1",
        "bar2"
        ]
    }
  }
}

AnyValues
anyValues: ALL header keys must be present. ANY one of the values for each of the header keys must be present. All keys must have a value matched.

{
  "headers": {
    "anyValues": {
      "fooHeader": [
        "foo1Value",
        "foo2Values"
        ],
      "barHeader": [
        "bar1Value",
        "bar2Value"
        ]
    }
  }
}

### Why is this important to you?

fairly common rules that teams would use when making routing decisions

karelz:
Triage:
- **NotContains** - this seems highly specialized. It brings OR into the routing logic, which does not exist there. It feels like very specific case, not much general. Unless we see more interest, we will likely not support it from config. It should be achievable via RouteConstraint. We should add small doc/example how to do these type of things.
- **AnyContains** - we can extend our [HeaderMatchMode](https://github.com/microsoft/reverse-proxy/blob/main/src/ReverseProxy/Configuration/HeaderMatchMode.cs) with 'contains'. Adding 2 rules, one for each header should then solve this scenario.
- **AnyValues** - we already have it with [HeaderMatchMode.ExactHeader](https://github.com/microsoft/reverse-proxy/blob/main/src/ReverseProxy/Configuration/HeaderMatchMode.cs#L15) - this should be achievable today.

[...]

PR #1277, opened 2021-10-06
Add Query Parameter MatchPolicy

ksauce:
Add query parameter match policy which is similar to the header match policy
Add NotContains to header match policy


Tratcher:
This fixes includes https://github.com/microsoft/reverse-proxy/issues/1235

ksauce:
> Can you check on the failures in these tests? https://github.com/microsoft/reverse-proxy/blob/main/test/Kubernetes.Tests/IngressConversionTests.cs I think you just need to update their baseline files like this one:
> 
> https://github.com/microsoft/reverse-proxy/blob/95c0b15c95252330d481184a9a99ff998acff2d9/test/Kubernetes.Tests/testassets/basic-ingress/routes.json#L8

docs/docfx/articles/queryparameter-routing.md
```diff
@@ -0,0 +1,351 @@
...
+- Exact - The query parameter must match in its entirety, subject to the value of `IsCaseSensitive`. Only single query parameters are supported. If there are multiple query parameters with the same name then the match fails.
+- Prefix - The query parameter must match by prefix, subject to the value of `IsCaseSensitive`. Only single query parameters are supported. If there are multiple query parameters with the same name then the match fails.
+- Exists - The query parameter must exist and contain any non-empty value.
+- Contains - The query parameter must exist and contains the non-empty value.
+- NotContains - The query parameter must exist and not contain the non-empty value.
+
+### IsCaseSensitive
+
+Indicates if the value match should be performed as case sensitive or insensitive. The default is `false`, insensitive.
+
```
Tratcher, line 258: Add a section about encoding.

```
### Encoding

The request query string will be parsed and decoded before matching against the route rules.

/````
   "route8" : {
    "ClusterId": "cluster1",
    "Match": {
      "Path": "{**catch-all}",
      "QueryParameters": [
        {
          "Name": "queryparam8",
          "Values": [ "another value" ],
          "Mode": "Exact"
        }
      ]
    }
/````

Matches to
/```
?queryparam8=another%20value
/```
or
/```
?queryparam8=another+value
/```

```

(Awkwardly, I can't specify a code block in a suggestion, remove the /'s.)

test/ReverseProxy.Tests/Routing/HeaderMatcherPolicyTests.cs
```diff
@@ -186,17 +187,17 @@ public async Task ApplyAsync_MultipleHeaderValues_NotSupported()
 
         [Theory]
         [InlineData("abc", HeaderMatchMode.ExactHeader, false, null, false)]
-        [InlineData("abc", HeaderMatchMode.ExactHeader, false, "", false)]
+        [InlineData("abc", HeaderMatchMode.ExactHeader, false, " ", false)]
```
Tratcher, line 190: ?

test/ReverseProxy.Tests/Routing/QueryMatcherPolicyTests.cs
```diff
@@ -0,0 +1,408 @@
...
+        [Theory]
+        [InlineData(null, false)]
+        [InlineData("", false)]
+        [InlineData("abc", true)]
+        public async Task ApplyAsync_MatchingScenarios_AnyQueryParamValue(string incomingQueryParamValue, bool shouldMatch)
+        {
+            var context = new DefaultHttpContext();
+            if (incomingQueryParamValue != null)
+            {
+                var queryStr = "?org-id=" + incomingQueryParamValue;
```
Tratcher, line 166: Add one test with escaped keys and values like `?org+id=val%20ue`.

src/ReverseProxy/Configuration/RouteQueryParameter.cs
```diff
@@ -0,0 +1,63 @@
...
+                && Mode == other.Mode
+                && IsCaseSensitive == other.IsCaseSensitive
+                && (IsCaseSensitive
+                    ? CaseSensitiveEqualHelper.Equals(Values, other.Values)
+                    : CaseInsensitiveEqualHelper.Equals(Values, other.Values));
+        }
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(Mode, IsCaseSensitive,
```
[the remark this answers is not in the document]
ksauce: What do you mean here?

test/ReverseProxy.Tests/Routing/HeaderMatcherPolicyTests.cs
ksauce: looking into that .  passing in "" fails the tests
﻿ Yarp.ReverseProxy.Routing.Tests.HeaderMatcherPolicyTests.ApplyAsync_MatchingScenarios_TwoHeaderValues(header1Value: "abc", header2Value: "def", headerValueMatchMode: NotContains, isCaseSensitive: False, incomingHeaderValue: "", shouldMatch: True)

src/ReverseProxy/Configuration/RouteQueryParameter.cs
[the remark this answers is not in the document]
MihaZupan: There are a lot of `GetHashCode` issues across the configuration types (hash codes are using collection instances instead of contents).
 I will put up a PR after this one is merged.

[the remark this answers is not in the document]
Tratcher: Yeah, it's unclear if GetHashCode is even being used in practice. We mainly implement it because there's a C# rule that says you have to if you implemented Equals.

PR #1278, opened 2021-10-07
Clean up proxy url generation 

### LINKED ISSUE #1219: path percent encoding - wierdness
### Describe the bug
currently if a path is (double) percent encoded, it will be decoded by yarp, but unfortunatly yarp will try to sent the unencoded path to the backend server

### To Reproduce
send the following path:

```
GET /encoded%2520path
```

yarp should now send the following to the backend:

```
GET /encoded%2520path
```

what it does send:
```
GET /encoded%20path
```

whoever we actually used double percent encoding in some places of our code and this completly breaks it

it would be good if there was a option to keep the path as-is.

btw. we can fix it by using (or by using a HttpTransformer):
```c#
    public class PercentEncodingTransform : RequestTransform
    {
        public override ValueTask ApplyAsync(RequestTransformContext context)
        {
            context.ProxyRequest.RequestUri = new Uri(
                new Uri(context.DestinationPrefix),
                Uri.EscapeUriString(context.HttpContext.Request.GetEncodedPathAndQuery()));
            return ValueTask.CompletedTask;
        }
    }
```

but it's probably strange that a proxy rewrite's the path. (btw. we use double percent encoding because the framework's and routers which we use is a little bit wierd and this fixed it)

not sure if that is even fixable since it depends on `PathString` which always change the behavior. maybe a good idea would be to expose a RawRequestPath in dotnet and allow yarp to have a config to use RawRequestPath instead of `context.Path` for the backend

### Further technical details

- 1.0.0-preview12
- platform: macOS

Tratcher:
Fixes #1219

The default PathString.ToUriComponent() doesn't round trip all the encoded values we want so we now encode it ourselves.

### DIFF
--- a/src/ReverseProxy/Forwarder/RequestUtilities.cs
+++ b/src/ReverseProxy/Forwarder/RequestUtilities.cs
@@ -134,18 +134,12 @@ namespace Yarp.ReverseProxy.Forwarder
 
             // Check if any escaping is required.
             var value = path.Value!;
-            var i = 0;
-            for (; i < value.Length; i++)
+            for (var i = 0; i < value.Length; i++)
             {
                 if (!IsValidPathChar(value[i]))
                 {
-                    break;
+                    return EncodePath(value, i);
                 }
-            }
-
-            if (i < value.Length)
-            {
-                return EncodePath(value, i);
             }
 
             return value;
--- a/src/ReverseProxy/Forwarder/RequestUtilities.cs
+++ b/src/ReverseProxy/Forwarder/RequestUtilities.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
 using System.Runtime.CompilerServices;
+using System.Text;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Forwarder/RequestUtilities.cs
```diff
@@ -118,11 +119,141 @@ public static Uri MakeDestinationAddress(string destinationPrefix, PathString pa
...
+                if (!IsValidPathChar(value[i]))
+                {
+                    break;
+                }
+            }
+
+            if (i < value.Length)
+            {
+                return EncodePath(value, i);
+            }
```
MihaZupan, line 149:
```suggestion
            for (var i = 0; i < value.Length; i++)
            {
                if (!IsValidPathChar(value[i]))
                {
                    return EncodePath(value, i);
                }
            }
```

src/ReverseProxy/Forwarder/RequestUtilities.cs
```diff
@@ -118,11 +119,141 @@ public static Uri MakeDestinationAddress(string destinationPrefix, PathString pa
...
+
+            return value;
+        }
+
+        private static string EncodePath(string value, int i)
```
MihaZupan, line 154: This makes for the strongest use-case for an API like dotnet/runtime#40603 I've seen so far. Ideally `ValueStringBuilder` would be public.

MihaZupan (approved):

PR #1284, opened 2021-10-11
Correct config model GetHashCode calculations

MihaZupan:
> If you override the GetHashCode method, you should also override Equals, and vice versa. **If your overridden Equals method returns true when two objects are tested for equality, your overridden GetHashCode method must return the same value for the two objects.**

The configuration model currently breaks this contract. When checking equality it will compare the contents of collections, but it will use the collection instance instead of contents for `GetHashCode`. This means that two "equal" instances may return different hash codes unless all properties share the same instances.

I went through the configuration types and fixed up a few details (e.g. `Path` being case-insensitive for active health checks).

`Transforms` and `Metadata` were also compared without case sensitivity, but I don't believe they should be.
Transforms may behave differently depending on case, and metadata is user input that should be opaque to us.

### DIFF
--- a/src/ReverseProxy/Utilities/CollectionEqualityHelper.cs
+++ b/src/ReverseProxy/Utilities/CollectionEqualityHelper.cs
@@ -126,18 +126,38 @@ namespace Yarp.ReverseProxy.Utilities
             return hashCode.ToHashCode();
         }
 
-        public static int GetHashCode<T>(IReadOnlyDictionary<string, T>? dictionary)
+        public static int GetHashCode<T>(IReadOnlyDictionary<string, T>? dictionary, IEqualityComparer<T>? valueComparer = null)
         {
             if (dictionary is null)
             {
                 return 0;
             }
 
-            // We don't know what comparer the dictionary was created with, so we can't reliably create correct hash codes based on keys
-            return dictionary.Count;
+            if (dictionary.Count == 0)
+            {
+                return 42;
+            }
+
+            // We don't know what comparer the dictionary was created with, so we assume it's Ordinal/OrdinalIgnoreCase
+            // If a culture-sensitive comparer was used, this may result in GetHashCode returning different values for "equal" strings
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Utilities/CollectionEqualityHelper.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            var hashCode = new HashCode();
+            foreach (var value in values)
+            {
+                hashCode.Add(value, valueComparer);
+            }
+            return hashCode.ToHashCode();
+        }
+
+        public static int GetHashCode<T>(IReadOnlyDictionary<string, T>? dictionary)
```
Tratcher, line 129: Then require the IEqualityComparer to be passed in.

MihaZupan: We don't have it for keys, we just have the (potentially user-provided) `IReadOnlyDictionary` instances

Tratcher: We can pass in what we expect it to be. E.g. You're now treating transforms as ordinal.

MihaZupan: > You're now treating transforms as ordinal.

I believe they are. Different casing can result in different behavior (e.g. the casing of an appended header value).

> We can pass in what we expect it to be

Since these can be user-controlled, we risk breaking the `Equals`/`GetHashCode` contract if we choose an arbitrary comparer. E.g. the comparer may return true for strings that are not "ordinal equal".

Realistically it would only make sense for the user to use `Ordinal(IgnoreCase)` when constructing the dictionary, so I'd be okay if we choose to always use `OrdinalIgnoreCase` here. It would risk breaking the contract for a better (but not perfect) hash code.

MihaZupan: As far as I can tell, we don't use these as keys internally.

PR #1289, opened 2021-10-14
Change to an activity based timeout

### LINKED ISSUE #1052: Security: disable web socket and have mechanism to enable
way to control per route / cluster

Data rates are disabled for web sockets.

Tratcher:
Fixes #1052 
Fixes https://github.com/microsoft/reverse-proxy/issues/118#issuecomment-896417517

The request Timeout used to cover the time until the response headers were received, which may also include sending the request body. This timeout could be problematic for large requests or gRPC client streaming. There was no timeout on the response body or on WebSockets which risked a resource leak.

Now a single activity token and timeout is used to track 'liveness', where that timeout is reset any time there's observable progress such as:
- Response headers received
- Request body data received
- Request body data sent
- Response body data received
- Response body data sent
- WebSocket data sent or received, including pings.

What doesn't count:
- HTTP/2 pings
- TCP keep-alives

This way we're not limiting the size of a request nor the direction of the traffic, so long as there continues to be any traffic.

Server request and response body data rates may still apply (though we opt out of those for gRPC & Kestrel).

(Thanks to @MihaZupan's last change, this was pretty easy :grin:, with many of the changes being aesthetic.)

### DIFF
--- a/test/ServiceFabric.Tests/ServiceDiscovery/Util/LabelsParserTests.cs
+++ b/test/ServiceFabric.Tests/ServiceDiscovery/Util/LabelsParserTests.cs
@@ -37,7 +37,7 @@ namespace Yarp.ReverseProxy.ServiceFabric.Tests
                 { "YARP.Backend.SessionAffinity.Cookie.Path", "mypath" },
                 { "YARP.Backend.SessionAffinity.Cookie.SameSite", "Strict" },
                 { "YARP.Backend.SessionAffinity.Cookie.SecurePolicy", "SameAsRequest" },
-                { "YARP.Backend.HttpRequest.Timeout", "00:00:17" },
+                { "YARP.Backend.HttpRequest.ActivityTimeout", "00:00:17" },
                 { "YARP.Backend.HttpRequest.AllowResponseBuffering", "true" },
                 { "YARP.Backend.HttpRequest.Version", "1.1" },
 #if NET


src/ReverseProxy/Forwarder/StreamCopyHttpContent.cs
```diff
@@ -138,19 +139,23 @@ async Task SerializeToStreamAsync(Stream stream, TransportContext? context, Canc
             // On HTTP/1.1: Linked HttpContext.RequestAborted + Request Timeout
             // On HTTP/2.0: SocketsHttpHandler error / the server wants us to stop sending content / H2 connection closed
             // _cancellation will be the same as cancellationToken for HTTP/1.1, so we can avoid the overhead of linking them
-            CancellationTokenSource? linkedCts = null;
+            IDisposable? registration = null;
```
MihaZupan, line 142:
```suggestion
            CancellationTokenRegistration registration = null;
```

Avoids boxing

docs/docfx/articles/http-client-config.md
```diff
@@ -84,15 +84,15 @@ At the moment, there is no solution for changing encoding for response headers i
 HTTP request configuration is based on [ForwarderRequestConfig](xref:Yarp.ReverseProxy.Forwarder.ForwarderRequestConfig) and represented by the following configuration schema.
 ```JSON
 "HttpRequest": {
-    "Timeout": "<timespan>",
+    "ActivityTimeout": "<timespan>",
```
Kahbazi, line 87: This also needs to be changed.
https://github.com/microsoft/reverse-proxy/blob/e74dade21281b546f8d9bbf9838ef63aae543405/docs/docfx/articles/config-files.md?plain=1#L199

Kahbazi:
@Tratcher breaking-change label?

PR #1291, opened 2021-10-15
Adding http version and streaming info to logging output

samsp-msft:
Final part of #600

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -756,10 +756,10 @@ namespace Yarp.ReverseProxy.Forwarder
                 EventIds.HttpDowngradeDetected,
                 "The request was downgraded from HTTP/2.");
 
-            private static readonly Action<ILogger, string, string, string, Exception?> _proxying = LoggerMessage.Define<string, string, string>(
+            private static readonly Action<ILogger, string, string, string, string, Exception?> _proxying = LoggerMessage.Define<string, string, string, string>(
                 LogLevel.Information,
                 EventIds.Forwarding,
-                "Proxying to {targetUrl} - {verson} {isStreaming}");
+                "Proxying to {targetUrl} {version} {versionPolicy} {isStreaming}");
 
             private static readonly Action<ILogger, ForwarderError, string, Exception> _proxyError = LoggerMessage.Define<ForwarderError, string>(
                 LogLevel.Information,
@@ -776,12 +776,14 @@ namespace Yarp.ReverseProxy.Forwarder
                 // Avoid computing the AbsoluteUri unless logging is enabled
                 if (logger.IsEnabled(LogLevel.Information))
                 {
+                    var streaming = isStreamingRequest ? "streaming" : "no-streaming";
+                    var version = ProtocolHelper.GetHttpProtocol(msg.Version);
 #if NET
-                    var versionInfo = $"http/{msg.Version} {msg.VersionPolicy}";
+                    var versionPolicy = ProtocolHelper.GetVersionPolicy(msg.VersionPolicy);
 #else
-                    var versionInfo = $"http/{msg.Version}";
+                    var versionPolicy = "RequestVersionOrLower";
 #endif
-                    _proxying(logger, msg.RequestUri!.AbsoluteUri, versionInfo, (isStreamingRequest) ? "streaming" : "no-streaming", null);
+                    _proxying(logger, msg.RequestUri!.AbsoluteUri, version, versionPolicy, streaming, null);
                 }
             }
 
--- a/src/ReverseProxy/Forwarder/ProtocolHelper.cs
+++ b/src/ReverseProxy/Forwarder/ProtocolHelper.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Net;
+using System.Net.Http;
+using Microsoft.AspNetCore.Http;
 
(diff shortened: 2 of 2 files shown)


src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -771,12 +771,17 @@ public static void HttpDowngradeDetected(ILogger logger)
...
+            public static void Proxying(ILogger logger, HttpRequestMessage msg, bool isStreamingRequest)
             {
                 // Avoid computing the AbsoluteUri unless logging is enabled
                 if (logger.IsEnabled(LogLevel.Information))
                 {
-                    _proxying(logger, targetUrl.AbsoluteUri, null);
+#if NET
+                    var versionInfo = $"http{msg.Version} {msg.VersionPolicy}";
+#else
+                    var versionInfo = $"http {msg.Version}";
```
Tratcher, line 782:
```suggestion
                    var versionInfo = $"http/{msg.Version} {msg.VersionPolicy}";
#else
                    var versionInfo = $"http/{msg.Version}";
```

Tratcher (approved):
Can you show what the new log output looks like now? I don't think Version.ToString formats the way you want by default.
You can borrow GetHttpProtocol:
https://github.com/dotnet/aspnetcore/blob/393a86c0e7218964e557adeb833010e1482bb96b/src/Http/Http.Abstractions/src/HttpProtocol.cs#L112
And put it here:
https://github.com/microsoft/reverse-proxy/blob/2015d992e0061c6849fbe11c62311161b3399210/src/ReverseProxy/Forwarder/ProtocolHelper.cs#L9

samsp-msft:
> Can you show what the new log output looks like now? I don't think Version.ToString formats the way you want by default. You can borrow GetHttpProtocol: https://github.com/dotnet/aspnetcore/blob/393a86c0e7218964e557adeb833010e1482bb96b/src/Http/Http.Abstractions/src/HttpProtocol.cs#L112 And put it here:
> 
> https://github.com/microsoft/reverse-proxy/blob/2015d992e0061c6849fbe11c62311161b3399210/src/ReverseProxy/Forwarder/ProtocolHelper.cs#L9

Current output:

```console
info: Yarp.ReverseProxy.Forwarder.HttpForwarder[9]
      Proxying to http://www.example.com/ - http2.0 RequestVersionOrLower no-streaming
```

How exactly do we want it to read? Options:

*      Proxying to http://www.example.com/ - HTTP/2 RequestVersionOrLower no-streaming
*      Proxying to http://www.example.com/ - http2 ≤  false
*      Proxying to http://www.example.com/ - http≤2  no-stream

Tratcher:
> ` Proxying to http://www.example.com/ - HTTP/2 RequestVersionOrLower no-streaming`

Looks good.

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -756,10 +756,10 @@ private static class Log
                 EventIds.HttpDowngradeDetected,
                 "The request was downgraded from HTTP/2.");
 
-            private static readonly Action<ILogger, string, Exception?> _proxying = LoggerMessage.Define<string>(
+            private static readonly Action<ILogger, string, string, string, Exception?> _proxying = LoggerMessage.Define<string, string, string>(
                 LogLevel.Information,
                 EventIds.Forwarding,
-                "Proxying to {targetUrl}");
+                "Proxying to {targetUrl} - {verson} {isStreaming}");
```
Kahbazi, line 762:
```suggestion
                "Proxying to {targetUrl} - {version} {isStreaming}");
```

Kahbazi: Shouldn't `isStreaming` be `bool` so it's easier to find it when using structured logs?

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -771,12 +771,17 @@ public static void HttpDowngradeDetected(ILogger logger)
...
 
-            public static void Proxying(ILogger logger, Uri targetUrl)
+            public static void Proxying(ILogger logger, HttpRequestMessage msg, bool isStreamingRequest)
             {
                 // Avoid computing the AbsoluteUri unless logging is enabled
                 if (logger.IsEnabled(LogLevel.Information))
                 {
-                    _proxying(logger, targetUrl.AbsoluteUri, null);
+#if NET
+                    var versionInfo = $"http/{msg.Version} {msg.VersionPolicy}";
```
Kahbazi, line 780: Why not create another property for `VersionPolicy` in logs? When `Version` and `VersionPolicy` are separate properties it's easier to have different queries based on `Version` alone.

Tratcher (commented):
I pushed some minor changes to this branch. The output now looks like this:
`Proxying to https://example.com/ HTTP/2 RequestVersionOrLower no-streaming`

src/ReverseProxy/Forwarder/HttpForwarder.cs
Kahbazi: In the log message "streaming" and "non-streaming" is a better choice but I thought in a structured log view like Elasticsearch, storing bool needs less space than string also writing a query like `isStreaming = true` is better than `isStreaming = streaming`.

Tratcher: Can you show an example of that?

PR #1295, opened 2021-10-18
Change I*MetricsConsumer to a single interface

MihaZupan:
Doing this now since it is a breaking API change.
Allows us to remove a bunch of boilerplate from the implementation.

```c#
public interface IMetricsConsumer<TMetrics>
{
    void OnMetrics(TMetrics previous, TMetrics current);
}
```

Any thoughts on `previous`/`current` vs `oldMetrics`/`newMetrics`


Tratcher (approved):
Show it to the rest of the team tomorrow before merging.

PR #1302, opened 2021-10-19
Fix ActiveHealthCheckMonitor's ProbeCluster logic

### LINKED ISSUE #1299: ActiveHealthCheckMonitor may report probe results for wrong destinations
`ActiveHealthCheckMonitor` will try to initialize all probe requests for a given cluster at once, storing tasks into [`probeTasks`](https://github.com/microsoft/reverse-proxy/blob/762b1a43216cb2161a0dcb422036bb6fb9e84351/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs#L117).
If [`SendAsync` fails synchronously](https://github.com/microsoft/reverse-proxy/blob/762b1a43216cb2161a0dcb422036bb6fb9e84351/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs#L132-L138), the task will not be stored in the list.

Later on, it will try to[ await tasks and turn the responses into `DestinationProbingResult`s](https://github.com/microsoft/reverse-proxy/blob/762b1a43216cb2161a0dcb422036bb6fb9e84351/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs#L141-L157).
It will [take the `DestinationState` from `allDestinations` at index `i`](https://github.com/microsoft/reverse-proxy/blob/762b1a43216cb2161a0dcb422036bb6fb9e84351/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs#L149), but that index will not correspond to the relevant destination if any request failed synchronously (as there is no more 1:1 mapping between `probingResults` and `allDestinations`).

As a result, `DestinationProbingResult`s passed to `IActiveHealthCheckPolicy.ProbingCompleted` will contain mismatched destination/response/exception information.

MihaZupan:
Fixes #1299

Opening this as a draft since I'm not clear on some details.

### DIFF
--- a/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
+++ b/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
@@ -3,8 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Net.Http;
-using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
```diff
@@ -114,87 +112,67 @@ private async Task ProbeCluster(ClusterState cluster)
...
+            }
+            catch (Exception ex)
+            {
+                if (waitingForSendAsync)
+                {
+                    Log.DestinationProbingFailed(_logger, destination.DestinationId, cluster.ClusterId, ex);
+                }
+                else
+                {
+                    Log.ActiveHealthProbeConstructionFailedOnCluster(_logger, destination.DestinationId, cluster.ClusterId, ex);
```
MihaZupan, line 168: Previously if `SendAsync` failed synchronously, the `DestinationProbingResult` for that destination would not be reported to `ProbingCompleted`.
This was done intentionally, but I don't understand why. @tratcher do you remember?

I think it should be the responsibility of `IActiveHealthCheckPolicy` to decide what to do with a failed request, and the monitor shouldn't be hiding failures.

MihaZupan: Secondly, is there a benefit to reporting different logs for sync vs async probe failure?

MihaZupan (commented):
Some logs are now unreachable, I assume the `EventId` should just be deleted and left as a gap?

src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
Tratcher: I think ActiveHealthProbeConstructionFailedOnCluster is trying to report a failure from _probingRequestFactory.CreateRequest, not cluster.Model.HttpClient.SendAsync. That would make sense to report separately. Sync and async HttpClient exceptions should be reported the same.

src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
```diff
@@ -114,87 +112,67 @@ private async Task ProbeCluster(ClusterState cluster)
...
+        {
+            var cts = new CancellationTokenSource(timeout);
+            var waitingForSendAsync = false;
+            try
+            {
+                var request = _probingRequestFactory.CreateRequest(cluster.Model, destination.Model);
+                Log.SendingHealthProbeToEndpointOfDestination(_logger, request.RequestUri, destination.DestinationId, cluster.ClusterId);
+
+                var responseTask = cluster.Model.HttpClient.SendAsync(request, cts.Token);
+                waitingForSendAsync = true;
```
Tratcher, line 153: Swapping these so Log.ActiveHealthProbeConstructionFailedOnCluster only covers exceptions from _probingRequestFactory.CreateRequest.
```suggestion
                waitingForSendAsync = true;
                var responseTask = cluster.Model.HttpClient.SendAsync(request, cts.Token);
```

Tratcher (approved):

src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
MihaZupan: > If _probingRequestFactory.CreateRequest throws should that probe be reported to IActiveHealthCheckPolicy?

Yes, I think so. It indicates something is seriously wrong with health check logic if we couldn't even create the request.

Issue #1055, opened 2021-06-04
Changes dependant on targetting 6.0


MihaZupan:
List of known changes we should do in Yarp once adding a 6.0 target (and/or more convenient after dropping 3.1):
- [ ] Use non-validated header enumeration https://github.com/dotnet/runtime/issues/35126#issuecomment-854040319
- [ ] Call Cancellable/Sync overloads of `HttpContent.ReadAsStream*` #770 
- [ ] Doc https://github.com/dotnet/aspnetcore/pull/33776 here: https://github.com/microsoft/reverse-proxy/blob/16ab855a5be980a0cf5fd21e9a7028190b6e7dda/docs/docfx/articles/http-client-config.md#httpclient
- [x] `CancellationTokenSource` pooling #902
- [x] Remove `ActivityPropagationHandler` in favour of the [`DistributedContextPropagator`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.distributedcontextpropagator) that can be customized on a per-handler basis https://github.com/dotnet/runtime/issues/55556
- [x] Use [Random.Shared](https://github.com/dotnet/runtime/issues/43887) in `RandomFactory`
- [x] Update `NameResolutionMetrics` to include `current-dns-lookups` https://github.com/dotnet/runtime/pull/49146
- [X] Implement the cancellable overload of HttpContent.SerializeToStreamAsync (5.0) (#1271)
- [x] Use [`WaitAsync`](https://github.com/dotnet/runtime/blob/v6.0.0-preview.7.21377.19/src/libraries/Common/tests/System/Threading/Tasks/TaskTimeoutExtensions.cs#L15) task timeout extensions instead of `TimeoutAfter`

Feel free to add anything I may be missing

Kahbazi:
How about using [Random.Shared](https://github.com/dotnet/runtime/issues/43887) in `RandomFactory`?

Tratcher:
Note adding 6.0 support is independent from dropping 3.1 support. Are there any of these you think are prohibitive to #if light-up on 5.0 or 6.0 while still supporting 3.1?

MihaZupan:
For #770 we should just wait to drop 3.1. For others it's actually that we need 6.0.

davidfowl:
Just move the ifdefs to extension methods that noop, it'll keep the code clean. It's the technique we use in Pipelines and Channels in the BCL.

[...]

PR #1306, opened 2021-10-20
Use WaitAsync

WeihanLi:
Contributes to #1055 

Use WaitAsync task timeout extensions instead of TimeoutAfter for .NET 6

### DIFF
--- a/test/ReverseProxy.Tests/Common/TaskExtensions.cs
+++ b/test/ReverseProxy.Tests/Common/TaskExtensions.cs
@@ -49,7 +49,16 @@ namespace Yarp.Tests.Common
             {
                 return await task;
             }
-
+#if NET6_0_OR_GREATER
+            try
+            {
+                return await task.WaitAsync(timeout);
+            }
+            catch (TimeoutException ex) when (ex.Source == typeof(TaskExtensions).Namespace)
+            {
+                throw new TimeoutException(CreateMessage(timeout, filePath, lineNumber));
+            }
+#else
             var cts = new CancellationTokenSource();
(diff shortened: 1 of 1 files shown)


test/ReverseProxy.Tests/Common/TaskExtensions.cs
```diff
@@ -21,24 +21,41 @@ internal static class TaskExtensions
...
+            return task.AsTask().WaitAsync(DefaultTimeoutTimeSpan);
+#else
             return task.AsTask().TimeoutAfter(DefaultTimeoutTimeSpan);
+#endif
         }
 
         public static Task<T> DefaultTimeout<T>(this Task<T> task)
         {
+#if NET6_0_OR_GREATER
+            return task.WaitAsync(DefaultTimeoutTimeSpan);
```
Tratcher, line 43: What does this throw when it expires?  Before it was throwing a specific exception with details:
`throw new TimeoutException(CreateMessage(timeout, filePath, lineNumber));`

MihaZupan: Might be better to copy what was done in AspNetCore:
https://github.com/dotnet/aspnetcore/blob/dfcf1adee2c0f6c89ecf66a8791e550bce124698/src/Shared/TaskExtensions.cs#L77-L108

WeihanLi: It would throw `TimeoutException` also for a timeout

WeihanLi: Got it, thanks @Tratcher

Tratcher:
Thanks

Issue #1351, opened 2021-11-05
File Scoped Namespaces


Kahbazi:
Is there plan to use C# 10 File Scoped Namespaces?

Tratcher:
FYI we decided to do this for aspnetcore. https://github.com/dotnet/aspnetcore/pull/38076 

This requires the .NET 6 SDK, which we are now using.

MihaZupan:
I assume this change is only visible internally - consumers of our packages would still be able to use the 5.0 SDK and not be forced to 6.0?

Tratcher:
Right, this change is only local/style. We'll have to be careful about what other C# 10 features we use though, especially ones that affect public API.

karelz:
Triage: Yes for code. We should leave samples as they are -- see comment  in PR #1352

Tratcher:
- [x] ReverseProxy
- [x] Tests
- [ ] Kubernetes
- [ ] OperatorFramework
- [ ] TestAssets
- [ ] ~Samples & Docs~ Not while we support 3.1/5.0.
- [ ]

PR #1352, opened 2021-11-05
File Scoped Namespaces

Kahbazi:
I have only change the main project for now. Should I change the other projects too in this PR or wait for this one to be merged first?

Contributes to https://github.com/microsoft/reverse-proxy/issues/1351

### DIFF
--- a/src/ReverseProxy/Yarp.ReverseProxy.csproj
+++ b/src/ReverseProxy/Yarp.ReverseProxy.csproj
@@ -7,7 +7,6 @@
     <RootNamespace>Yarp.ReverseProxy</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Nullable>enable</Nullable>
-    <LangVersion>10.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/ReverseProxy/Yarp.ReverseProxy.csproj
+++ b/src/ReverseProxy/Yarp.ReverseProxy.csproj
@@ -7,6 +7,7 @@
     <RootNamespace>Yarp.ReverseProxy</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Nullable>enable</Nullable>
+    <LangVersion>10.0</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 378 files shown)


Tratcher:
@Kahbazi we're going to look at #1351 in triage before we decide to go forward with this. We'll get back to you next week.

WeihanLi:
Would like to suggest adding namespace style config into the `.editorconfig` for better experience <<https://github.com/WeihanLi/SparkTodo/blob/01aeada4fa7107fe864696b8e06db480b73465a7/.editorconfig#L32>>

```
csharp_style_namespace_declarations=file_scoped:suggestion
```

karelz:
Triage: We are ok with it. We want to keep samples as they are for now to not force sample users to install .NET 6 SDK when they copy-paste specific sample.

src/ReverseProxy/Yarp.ReverseProxy.csproj
```diff
@@ -7,6 +7,7 @@
     <RootNamespace>Yarp.ReverseProxy</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Nullable>enable</Nullable>
+    <LangVersion>10.0</LangVersion>
```
Tratcher, line 10: This should be global:
https://github.com/microsoft/reverse-proxy/blob/b2932df6c37044f5a72a5ef9228cf43e6b64b858/Directory.Build.props#L9

test/ReverseProxy.Tests/Configuration/ConfigProvider/ConfigurationReadingExtensionsTests.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using Xunit;
+using Yarp.ReverseProxy.Configuration.ConfigProvider;
```
Tratcher, line 6: ? There shouldn't be any changes that require this.

Tratcher (commented):
Please update the other projects and tests as well. Only avoid projects in the samples and docs folders that target 3.1/5.0.

Tratcher (approved):

Tratcher:
Thanks @Kahbazi. I'll leave #1351 open for now and add notes about converting other parts of the repo.

Issue #1353, opened 2021-11-05
1.0 release checklist


MihaZupan:
See [Release.md](https://github.com/microsoft/reverse-proxy/blob/main/docs/operations/Release.md) for detailed instructions.

- [x] Ensure there's a release branch created (see [Branching](https://github.com/microsoft/reverse-proxy/blob/main/docs/operations/Branching.md))
  - https://github.com/microsoft/reverse-proxy/tree/release/1.0
  - Branched on 08f6dd95bf1a940959a41f4c4a5364f7c5486060
- [x] Ensure the `Version.props` has the `PreReleaseVersionLabel` updated to the next preview
  - #1358
- [x] Identify and validate the build on the `microsoft-reverse-proxy-official` pipeline
  - https://dev.azure.com/dnceng/internal/_build/results?buildId=1457057
- [x] Release the build (unlisted)
  - https://www.nuget.org/packages/Yarp.ReverseProxy/1.0.0
  - https://www.nuget.org/packages/Yarp.Telemetry.Consumption/1.0.0
- [x] Tag the commit
  - https://github.com/microsoft/reverse-proxy/releases/tag/v1.0.0
- [x] Draft release notes
- [x] relist the packages on nuget
- [x] Update samples to use updated package versions
  - ~~#1362~~ replaced by #1364
- [x] Publish the docs
  - #1364
- [x] Publish release notes
  - https://github.com/microsoft/reverse-proxy/releases/tag/v1.0.0
- [x] Close the [old milestone](https://github.com/microsoft/reverse-proxy/milestones)
- [x] Announce on social media
  - https://twitter.com/_MihaZupan/status/1458110656650035201
- [x] Set the preview branch to protected
- [x] Delete the [previous preview branch](https://github.com/microsoft/reverse-proxy/branches)
- [x] Request source code archival
- [ ] Update docs (for changes like #1359)

WeihanLi:
The [roadmap](https://github.com/microsoft/reverse-proxy/blob/main/docs/roadmap.md) should be updated also

karelz:
@WeihanLi good point, it is now tackled in PR #1374

PR #1356, opened 2021-11-05
Remove PreReleaseVersionLabel

### LINKED ISSUE #1317: RTM package branding
We need to figure out how to properly build the RTM build. It might just be clearing this field, but we'll need to test it.
https://github.com/microsoft/reverse-proxy/blob/4e96dde49a4d600f81ba0b232c8335dc3bf6fd06/eng/Versions.props#L10

It seems to work in a local build, but we'll still need to verify in an official build.

MihaZupan:
Fixes #1317
Contributes to #1353

Should we be switching `main` to `1.1.0.preview.1` or drop the preview altogether `1.1.0`?

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -7,6 +7,8 @@
       That ensures that if we get up to preview 10, it doesn't sort as between preview 1 and preview 2!
       See https://semver.org/spec/v2.0.0.html for details.
     -->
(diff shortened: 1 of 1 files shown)


Tratcher:
Let's stick with `1.1.0.preview.1` until we decide if there will be previews or not.

Tratcher (approved):

PR #1358, opened 2021-11-05
Bump version to 1.1.0.preview.1

MihaZupan:
Contributes to #1353


Tratcher (approved):
...if the 1.0 main build produces the versions we expect.

PR #1396, opened 2021-11-17
#1389 Added K8S namespace to cluster and route IDs

### LINKED ISSUE #1389: Same ingress or service name in multiple Kubernetes namespaces
When defining Ingress or Service resources in Kubernetes, it is legal to have two ingresses with the same name but in different *namespaces*. It is the combination of name and namespace that make the resource unique.

Yarp currently does *not* include the namespace when creating the keys for the Cluster and Routes configuration.

(Example ingresses shown here with the spec section omitted for clarity)
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: the-ingress
  namespace: ns-one
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: the-ingress
  namespace: ns-two
```

dpbevin:
Fixes #1389

### DIFF
--- a/src/Kubernetes.Controller/Converters/YarpParser.cs
+++ b/src/Kubernetes.Controller/Converters/YarpParser.cs
@@ -110,12 +110,12 @@ internal static class YarpParser
         {
             if (ingressServiceBackend.Port.Number.HasValue && ingressServiceBackend.Port.Number.Value > 0)
             {
-                return $"{ingressServiceBackend?.Name}.{namespaceName}:{ingressServiceBackend?.Port.Number}";
+                return $"{ingressServiceBackend.Name}.{namespaceName}:{ingressServiceBackend.Port.Number}";
             }
 
-            if (!string.IsNullOrWhiteSpace(ingressServiceBackend?.Name))
+            if (!string.IsNullOrWhiteSpace(ingressServiceBackend.Name))
             {
-                return $"{ingressServiceBackend?.Name}.{namespaceName}:{ingressServiceBackend?.Port.Name}";
+                return $"{ingressServiceBackend.Name}.{namespaceName}:{ingressServiceBackend.Port.Name}";
             }
         }
 
--- a/src/Kubernetes.Controller/Converters/YarpParser.cs
+++ b/src/Kubernetes.Controller/Converters/YarpParser.cs
@@ -113,7 +113,7 @@ internal static class YarpParser
                 return $"{ingressServiceBackend.Name}.{namespaceName}:{ingressServiceBackend.Port.Number}";
             }
 
-            if (!string.IsNullOrWhiteSpace(ingressServiceBackend.Name))
+            if (!string.IsNullOrWhiteSpace(ingressServiceBackend.Port.Name))
             {
                 return $"{ingressServiceBackend.Name}.{namespaceName}:{ingressServiceBackend.Port.Name}";
             }
--- a/src/Kubernetes.Controller/Converters/YarpParser.cs
+++ b/src/Kubernetes.Controller/Converters/YarpParser.cs
@@ -57,7 +57,7 @@ internal static class YarpParser
         }
 
         // Each ingress rule path can only be for one service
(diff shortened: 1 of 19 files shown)


src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -104,6 +104,24 @@ private static void HandleIngressRulePath(YarpIngressContext ingressContext, V1S
...
+        {
+            if (ingressServiceBackend.Port.Number.HasValue && ingressServiceBackend.Port.Number.Value > 0)
+            {
+                return $"{ingressServiceBackend?.Name}.{namespaceName}:{ingressServiceBackend?.Port.Number}";
+            }
+
+            if (!string.IsNullOrWhiteSpace(ingressServiceBackend?.Name))
+            {
+                return $"{ingressServiceBackend?.Name}.{namespaceName}:{ingressServiceBackend?.Port.Name}";
+            }
```
MihaZupan, line 119:
```suggestion
            if (ingressServiceBackend.Port.Number.HasValue && ingressServiceBackend.Port.Number.Value > 0)
            {
                return $"{ingressServiceBackend.Name}.{namespaceName}:{ingressServiceBackend.Port.Number}";
            }

            if (!string.IsNullOrWhiteSpace(ingressServiceBackend.Name))
            {
                return $"{ingressServiceBackend.Name}.{namespaceName}:{ingressServiceBackend.Port.Name}";
            }
```

`ingressServiceBackend` was already null checked above

src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -104,6 +104,24 @@ private static void HandleIngressRulePath(YarpIngressContext ingressContext, V1S
         }
     }
 
+    private static string UpstreamName(string namespaceName, V1IngressServiceBackend ingressServiceBackend)
+    {
+        if (ingressServiceBackend != null)
+        {
+            if (ingressServiceBackend.Port.Number.HasValue && ingressServiceBackend.Port.Number.Value > 0)
+            {
+                return $"{ingressServiceBackend?.Name}.{namespaceName}:{ingressServiceBackend?.Port.Number}";
```
MihaZupan, line 113: Does the existence of a non-default port imply that the name won't be null?

Or should this method generate something like
- `backend.Name` if it exists +
- `namespaceName` +
- `port.Number` if it exists or `port.Name` otherwise

MihaZupan (approved):

src/Kubernetes.Controller/Converters/YarpParser.cs
dpbevin: The port and name values are mutually exclusive, so one or the other _should_ be set.
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#servicebackendport-v1-networking-k8s-io

MihaZupan: Oh, I was more thinking of the `ingressServiceBackend.Name` - it looks like we will treat the name not being set differently depending on whether the port was set via name vs number.

The first `if` could be changed to `if (!string.IsNullOrWhiteSpace(ingressServiceBackend?.Name))` since it's always an invalid case regardless of port.

dpbevin:
@MihaZupan is this able to be merged now?

src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -104,6 +104,24 @@ private static void HandleIngressRulePath(YarpIngressContext ingressContext, V1S
...
+    private static string UpstreamName(string namespaceName, V1IngressServiceBackend ingressServiceBackend)
+    {
+        if (ingressServiceBackend != null)
+        {
+            if (ingressServiceBackend.Port.Number.HasValue && ingressServiceBackend.Port.Number.Value > 0)
+            {
+                return $"{ingressServiceBackend.Name}.{namespaceName}:{ingressServiceBackend.Port.Number}";
+            }
+
+            if (!string.IsNullOrWhiteSpace(ingressServiceBackend.Name))
```
MihaZupan, line 116: Re-reading it, I think this was actually been the indented behavior?
```suggestion
            if (!string.IsNullOrWhiteSpace(ingressServiceBackend.Port.Name))
```

MihaZupan: @dpbevin Are you okay with me merging this with the above suggestion?

PR #1398, opened 2021-11-17
#1397 Add Kubernetes IngressClass support

### LINKED ISSUE #1397: Kubernetes Ingress Class Support
### Background

[Kubernetes 1.18 introduced](https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/) a standard mechanism for assigning individual Ingress resources to specific Ingress Controllers, using the `ingressClassName` field on Ingresses and the *IngressClass* resource. Up to that point, Ingress Controllers used a custom annotation `kubernetes.io/ingress.class` on Ingress resources but in subtly different ways.

Kubernetes 1.22 (released Oct 2021) removed the beta Ingress resource API which means that all Ingress resources support the `ingressClassName` field.

### Why is this important?

Support for `IngressClass` means that groups of Ingress resources can be explicitly allocated to specific instances of an Ingress Controller, thereby segregating your traffic.

For example, you may run two Ingress Controllers in a cluster: one for external access, the other for internal access. You could set the `ingressClassName` field for your public APIs to `yarp-public` , and your monitoring Ingresses to `yarp-internal`.

Another example would be to run multiple Ingress Controller deployments/daemonsets to help distribute the load more effectively.

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: the-ingress
  namespace: default
spec:
  ingressClassName: yarp
  rules:
    - http:
        paths:
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: the-service
                port: 
                  name: http
```

dpbevin:
Fixes #1397

* Added informer for IngressClass resources
* Updated RBAC in sample manifest to allow list/read of IngressClass resources
* Delayed start of Ingress resource watching until IngressClass, Service, and Endpoints are synchronized
* Modified ingress cache to only store Ingresses that match the IngressClass rules
* Bound the existing ConfigMap to Application Settings and included sensible default for Ingress Class name (this allows the value to be changed when multiple, independent instances of the Yarp Ingress Controller are deployed into the same cluster)


dpbevin:
@MihaZupan bump (please)

lejsekt:
Is there any update on this? I'd really like to configure yarp using ingress controller in gcp and start using it instead of nginx. Thank you

MihaZupan:
Closing in favour of #1539

PR #1417, opened 2021-11-29
Skip checking whether a header is content-specific in RequestUtilities.AddHeader

MihaZupan:
[`TryAddWithoutValidation`](https://github.com/dotnet/runtime/blob/d9a578934827ec7792407b37852b1eb8fbf2b19b/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpHeaders.cs#L112-L114) will only return `false` if the header name does not fit into the header collection (e.g. trying to add content header into request headers). We can avoid checking whether the header is content-specific in our logic and always fallback to trying on the content.

### DIFF
--- a/src/ReverseProxy/Forwarder/RequestUtilities.cs
+++ b/src/ReverseProxy/Forwarder/RequestUtilities.cs
@@ -276,14 +276,10 @@ public static class RequestUtilities
 
             if (!request.Headers.TryAddWithoutValidation(headerName, headerValue))
             {
-                if (request.Content is null && _contentHeaders.Contains(headerName))
-                {
-                    request.Content = new EmptyHttpContent();
-                }
+                var content = request.Content ??= new EmptyHttpContent();
 
-                var added = request.Content?.Headers.TryAddWithoutValidation(headerName, headerValue);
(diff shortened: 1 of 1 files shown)


Tratcher:
> [`TryAddWithoutValidation`](https://github.com/dotnet/runtime/blob/d9a578934827ec7792407b37852b1eb8fbf2b19b/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpHeaders.cs#L112-L114) will only return `false` if the header name does not fit into the header collection (e.g. trying to add content header into request headers).

It will also fail if the header name is invalid. We've seen some cases where AspNetCore's header name validation is less strict than HttpClient's, so I don't think the assumption is safe.
https://github.com/dotnet/runtime/blob/d9a578934827ec7792407b37852b1eb8fbf2b19b/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HeaderDescriptor.cs#L58

MihaZupan:
> It will also return false if the header name is invalid.

Right, and getting an extra `Content-Length: 0` injected if you attempted to add an invalid header would be odd indeed.

PR #1432, opened 2021-12-06
Put generated `InvalidClientHelloDataTruncatedBytes` data into one test

### LINKED ISSUE #473: Deal with SniHelper_TruncatedData_Fails test warnings
Every official build produces 6 copies of the following warning:
```
##[warning]No of subresults in test case: Microsoft.ReverseProxy.Utilities.Tls.TlsFrameHelperTests.SniHelper_TruncatedData_Fails are more than supported limit 1000, truncating remaning ones.
```

The test uses programmatically generated member data to exhaustively cover the scenario (3980 variations!).
https://github.com/microsoft/reverse-proxy/blob/a3b5e1b9b03fcec8bc764b033b5c066838a8ffe9/test/ReverseProxy.Tests/Utilities/TlsFrameHelperTests.cs#L99-L122

The test should either be broken up or be made less exhaustive so it can properly report results.

greenEkatherine:
Fixes #473 

Moving 3980 generated cases into one test to not break XUnit limits and get valid results.

### DIFF
--- a/test/ReverseProxy.Tests/Utilities/TlsFrameHelperTests.cs
+++ b/test/ReverseProxy.Tests/Utilities/TlsFrameHelperTests.cs
@@ -26,11 +26,14 @@ public class TlsFrameHelperTests
         InvalidClientHello(clientHello, id, shouldPass: false);
     }
 
-    [Theory]
-    [MemberData(nameof(InvalidClientHelloDataTruncatedBytes))]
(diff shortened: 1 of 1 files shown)


Tratcher (approved):

wfurt (approved):
LGTM.

It would be nice to make same change in runtime.

PR #1436, opened 2021-12-09
Fixes build of docker image

specialforest:
Fixes the following docker image build error:
When using COPY with more than one source file, the destination must be
a directory and end with a /


specialforest:
@dpbevin please take a look.

dpbevin:
LGTM but there's another instance of the same issue in another Dockerfile: https://github.com/microsoft/reverse-proxy/blob/main/samples/KuberenetesIngress.Sample/Ingress/Dockerfile
Are you happy to also change that?

MihaZupan (approved):
Thanks!

Issue #1409, opened 2021-11-24
Behavior of HttpTransformer.Default should match the behavior of new HttpTransformer() or be moved/renamed


garrettm-msft:
### Some details

**How many backends are in your application?**

- [ ] 1-2
- [ ] 3-5
- [ ] 6-10
- [X] 10+

**How do you host your application?**

- [ ] Kubernetes
- [ ] Azure App Service
- [ ] Azure VMs
- [ ] Other Cloud Provider (please include details below)
- [X] Other Hosting model (please include details below)
Azure Service Fabric

### What did you think of YARP?

We recently root caused a direct proxying regression on our upgrade to YARP 1.0.0 caused by #1246 , which changed the behavior of HttpTransformer.Default to match the default behavior of TransformBuilder.CreateTransformer rather than the behavior of new HttpTransformer(). While it seems fair for the default behavior of IHttpForwarder.SendAsync to match the default forwarding behavior of non-direct proxy methods, it is quite sneaky for the HttpTransformer.Default property to additionally include the behavior of StructuredTransformer and default transforms added by TransformBuilder.CreateTransformer.

We suggest:

- Moving/renaming HttpTransformer.Default to make it clearer what logic it includes
- Still providing a static value for new HttpTransformer() and/or making the parameterless HttpTransformer constructor public to make a default HttpTransformer less awkward to obtain

@davidni as FYI and in case he has more to add on this

Tratcher:
> * Moving/renaming HttpTransformer.Default to make it clearer what logic it includes

That's a breaking change we're unlikely to make now that 1.0 has shipped.

> * Still providing a static value for new HttpTransformer() and/or making the parameterless HttpTransformer constructor public to make a default HttpTransformer less awkward to obtain

Maybe `HttpTransformer.Empty`?

karelz:
Triage: @garrettm-msft can you please confirm if the above suggestion of adding `HttpTransformer.Empty` would address your key concerns?

garrettm-msft:
Thanks both for the replies. While we believe the semantics are a bit unfortunate it's understandable that making such a breaking change is non-viable currently, and adding HttpTransformer.Empty should adequately address the concerns here.

Tratcher:
Re-opening to add a test

PR #1444, opened 2021-12-13
Add Empty HttpTransformer

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/1409

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpTransformer.cs
+++ b/src/ReverseProxy/Forwarder/HttpTransformer.cs
@@ -21,19 +21,13 @@ public class HttpTransformer
     /// A default set of transforms that adds X-Forwarded-* headers, removes the original Host value and
     /// copies all other request and response fields and headers, except for some protocol specific values.
     /// </summary>
-    public static readonly HttpTransformer Default;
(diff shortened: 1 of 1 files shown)


greenEkatherine:
@Kahbazi thanks for your PR. I was actually working on it as well, that's why I self-assigned, but I forgot to remove the "help wanted" label. I will add a test for this change after it is merged.

src/ReverseProxy/Forwarder/HttpTransformer.cs
```diff
@@ -23,9 +23,16 @@ public class HttpTransformer
     /// </summary>
     public static readonly HttpTransformer Default;
 
+    /// <summary>
+    /// A default set of transforms that copies all request and response fields and headers, except for some
```
Tratcher, line 27:
```suggestion
    /// An empty transformer that copies all request and response fields and headers, except for some
```

Kahbazi:
> @Kahbazi thanks for your PR. I was actually working on it as well, that's why I self-assigned, but I forgot to remove the "help wanted" label. I will add a test for this change after it is merged.

Sorry, I didn't notice the Assignees of the issue.

Tratcher (approved):

PR #1450, opened 2021-12-14
Add test for `HttpTransformer.Empty`

### LINKED ISSUE #1409: Behavior of HttpTransformer.Default should match the behavior of new HttpTransformer() or be moved/renamed

greenEkatherine:
Fixes #1409

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpTransformer.cs
+++ b/src/ReverseProxy/Forwarder/HttpTransformer.cs
@@ -23,6 +23,12 @@ public class HttpTransformer
     /// </summary>
     public static readonly HttpTransformer Default;
 
+    /// <summary>
+    /// A minimum set of transforms that copies all other request and response fields and headers,
+    /// except for some protocol specific values.
+    /// </summary>
+    public static readonly HttpTransformer Empty = new HttpTransformer();
+
     static HttpTransformer()
     {
         Default = TransformBuilder.CreateTransformer(new TransformBuilderContext());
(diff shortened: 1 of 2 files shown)


karelz:
@greenEkatherine FYI I changed the keyword above to close the issue automatically - see https://docs.devart.com/studio-for-sql-server/source-controlling-databases/associating-commits-with-github-issues.html

karelz (approved):

greenEkatherine:
> @greenEkatherine FYI I changed the keyword above to close the issue automatically - see https://docs.devart.com/studio-for-sql-server/source-controlling-databases/associating-commits-with-github-issues.html

Thank you! I always forget which words work 😃

PR #1452, opened 2021-12-14
Update HealthChecks docs

karelz:
- Add Active health checks example explanation
- Link existing API docs
- Update ActiveHealthCheckOptions to new name ActiveHealthCheckConfig
- Unify destination formats without trailing /
- Unify list formatting (period at the end of each list item)

### DIFF
--- a/docs/docfx/articles/dests-health-checks.md
+++ b/docs/docfx/articles/dests-health-checks.md
@@ -14,7 +14,7 @@ There are several cluster-wide configuration settings controlling active health 
           "Active": {
             "Enabled": "true",
             "Interval": "00:00:10",
-            "Timeout": "00:00:10",
+            "Timeout": "00:00:05",
             "Policy": "ConsecutiveFailures",
             "Path": "/api/health"
           }
@@ -24,11 +24,11 @@ There are several cluster-wide configuration settings controlling active health 
         },
         "Destinations": {
           "cluster1/destination1": {
-            "Address": "https://localhost:10000/"
+            "Address": "https://localhost:10000"
           },
           "cluster1/destination2": {
-            "Address": "http://localhost:10010/",
-            "Health": "http://localhost:10020/"
+            "Address": "http://localhost:10010/MyService",
+            "Health": "http://localhost:10020/MyHealth"
           }
         }
       }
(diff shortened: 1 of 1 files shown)


docs/docfx/articles/dests-health-checks.md
```diff
@@ -47,35 +47,41 @@ var clusters = new[]
...
+
+Per chosen policy (set in config via `Policy`), each destination will keep track of returned unhealthy HTTP response codes (2xx is considered healthy). When a destination returns unhealthy HTTP response code 3 times in a row (set in config via `Threshold`), it will be marked `Unhealthy` and it will not receive additional incoming traffic. All requests in flight to the destination will be left intact to finish on their own.
+
+YARP will keep probing the unhealthy destination at the same rate, i.e. every 10 seconds with timeout of 5 seconds. When there is first healthy response, the destination will be marked `Healthy` and will be used for routing of incoming traffic again.
```
MihaZupan, line 69:
```suggestion
In the examples above, YARP will proactively probe each destination for health every 10 seconds with a timeout of 5 seconds (set in config via `Interval` and `Timeout`). The active health probes will happen regardless of whether there is any load on the destinations from dispatching requests. The probing URLs will be `https://localhost:10000/api/health` for `destination1` (based on config settings `Address` and `Path`) and `https://localhost:10020/MyHealth/api/health` for `destination2` (based on config settings `Health` and `Path`).

Per chosen policy (set in config via `Policy`), each destination will keep track of returned unhealthy HTTP response codes (2xx is considered healthy). When a destination returns an unhealthy HTTP response code or fails to respond 3 times in a row (set in config via `Threshold`), it will be marked `Unhealthy` and it will not receive additional incoming traffic. All requests in flight to the destination will be left intact to finish on their own.

YARP will keep probing the unhealthy destination at the same rate, i.e. every 10 seconds with a timeout of 5 seconds. When receiving the first healthy response, the destination will be marked `Healthy` and will be used for routing of incoming traffic again.
```

MihaZupan (approved):
Just some wording nits

docs/docfx/articles/dests-health-checks.md
```diff
@@ -112,9 +118,9 @@ There are default built-in implementations for all of the aforementioned compone
...
+[IActiveHealthCheckPolicy](xref:Yarp.ReverseProxy.Health.IActiveHealthCheckPolicy) analyzes how destinations respond to active health probes sent by [IActiveHealthCheckMonitor](xref:Yarp.ReverseProxy.Health.IActiveHealthCheckMonitor), evaluates new active health states for all the probed destinations, and then call [IDestinationHealthUpdater.SetActive](xref:Yarp.ReverseProxy.Health.IDestinationHealthUpdater.SetActive) to set new active health states and rebuild the healthy destination collection based on the updated values.
 
-The below is a simple example of a custom `IActiveHealthCheckPolicy` marking destination as `Healthy`, if a successful response code was returned for a probe, and as `Unhealthy` otherwise.
+The below is a simple example of a custom [IActiveHealthCheckPolicy](xref:Yarp.ReverseProxy.Health.IActiveHealthCheckPolicy) marking destination as `Healthy`, if a successful response code was returned for a probe, and as `Unhealthy` otherwise.
```
Tratcher, line 123: When an API is referenced multiple times we only need to link the first reference in a given section. Linking every instance unnecessarily clutters the doc.

Tratcher (approved):

docs/docfx/articles/dests-health-checks.md
karelz: I would argue that the visual presentation is more consistent when all the references are the same. Perhaps we can discuss on Thu?

samsp-msft:
@karelz - we decided when looking at this last time that we don't need the duplication of definition links. Can you update and push this change please.

karelz:
Closing for now

PR #1461, opened 2021-12-16
Fixed invalid heading markdown for Metrics.Sample Readme

tomaustin700:
Previously the heading wasn't rendered correctly as there was no space after the #. Just added the #


MihaZupan (approved):
Thanks!

PR #1462, opened 2021-12-16
Throw on wrong Content-Length

### LINKED ISSUE #1412: Timeout when request body is consumed before proxying
### Describe the bug
The simple fact of accessing the Request.Form in a Middleware, while Posting a form, will freeze the Core App.

### To Reproduce
Add this line to a middleware 
```c#
if (context.Request.HasFormContentType && context.Request.Form != null)
{ 
     //...
}
```

This will let GETs go through fine, but will lock the callint core application when posting a form. 
Tested while making a postback into a WebFormsApplication

Thank you

greenEkatherine:
Fixes #1412 

Throwing an exception when `Content-Length` is set but doesn't match with actual length in request. This allows to avoid unclear hangings as described in the issue above.

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -644,6 +644,7 @@ internal sealed class HttpForwarder : IHttpForwarder
         if (destinationResponseContent != null)
         {
             using var destinationResponseStream = await destinationResponseContent.ReadAsStreamAsync();
+            // The response content-length is enforced by the server.
             return await StreamCopier.CopyAsync(isRequest: false, destinationResponseStream, clientResponseStream, -1, _clock, activityCancellationSource, activityCancellationSource.Token);
         }
 
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -406,7 +406,6 @@ internal sealed class HttpForwarder : IHttpForwarder
             // but for now, out of an abundance of caution, we only do it for requests that look like gRPC.
             return new StreamCopyHttpContent(
                 source: request.Body,
-                contentLength: request.Headers.ContentLength,
                 autoFlushHttpClientOutgoingStream: isStreamingRequest,
                 clock: _clock,
                 activityToken);
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -406,6 +406,7 @@ internal sealed class HttpForwarder : IHttpForwarder
             // but for now, out of an abundance of caution, we only do it for requests that look like gRPC.
             return new StreamCopyHttpContent(
                 source: request.Body,
+                contentLength: request.Headers.ContentLength,
                 autoFlushHttpClientOutgoingStream: isStreamingRequest,
                 clock: _clock,
                 activityToken);
@@ -585,8 +586,8 @@ internal sealed class HttpForwarder : IHttpForwarder
         // :: Step 7-A-2: Copy duplex streams
         using var destinationStream = await destinationResponse.Content.ReadAsStreamAsync();
 
-        var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
-        var responseTask = StreamCopier.CopyAsync(isRequest: false, destinationStream, clientStream, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
+        var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, context.Request.ContentLength, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
+        var responseTask = StreamCopier.CopyAsync(isRequest: false, destinationStream, clientStream, null, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
 
         // Make sure we report the first failure.
         var firstTask = await Task.WhenAny(requestTask, responseTask);
@@ -643,7 +644,7 @@ internal sealed class HttpForwarder : IHttpForwarder
         if (destinationResponseContent != null)
         {
             using var destinationResponseStream = await destinationResponseContent.ReadAsStreamAsync();
-            return await StreamCopier.CopyAsync(isRequest: false, destinationResponseStream, clientResponseStream, _clock, activityCancellationSource, activityCancellationSource.Token);
+            return await StreamCopier.CopyAsync(isRequest: false, destinationResponseStream, clientResponseStream, null, _clock, activityCancellationSource, activityCancellationSource.Token);
         }
 
         return (StreamCopyResult.Success, null);
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -586,8 +586,8 @@ internal sealed class HttpForwarder : IHttpForwarder
         // :: Step 7-A-2: Copy duplex streams
         using var destinationStream = await destinationResponse.Content.ReadAsStreamAsync();
 
-        var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, context.Request.ContentLength ?? -1, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
-        var responseTask = StreamCopier.CopyAsync(isRequest: false, destinationStream, clientStream, -1, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
+        var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, StreamCopier.UnknownLength, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
+        var responseTask = StreamCopier.CopyAsync(isRequest: false, destinationStream, clientStream, StreamCopier.UnknownLength, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
 
         // Make sure we report the first failure.
         var firstTask = await Task.WhenAny(requestTask, responseTask);
@@ -645,7 +645,7 @@ internal sealed class HttpForwarder : IHttpForwarder
         {
             using var destinationResponseStream = await destinationResponseContent.ReadAsStreamAsync();
             // The response content-length is enforced by the server.
-            return await StreamCopier.CopyAsync(isRequest: false, destinationResponseStream, clientResponseStream, -1, _clock, activityCancellationSource, activityCancellationSource.Token);
+            return await StreamCopier.CopyAsync(isRequest: false, destinationResponseStream, clientResponseStream, StreamCopier.UnknownLength, _clock, activityCancellationSource, activityCancellationSource.Token);
         }
 
         return (StreamCopyResult.Success, null);
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -586,8 +586,8 @@ internal sealed class HttpForwarder : IHttpForwarder
         // :: Step 7-A-2: Copy duplex streams
         using var destinationStream = await destinationResponse.Content.ReadAsStreamAsync();
 
-        var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, context.Request.ContentLength, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
-        var responseTask = StreamCopier.CopyAsync(isRequest: false, destinationStream, clientStream, null, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
+        var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, context.Request.ContentLength ?? -1, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
+        var responseTask = StreamCopier.CopyAsync(isRequest: false, destinationStream, clientStream, -1, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
 
         // Make sure we report the first failure.
         var firstTask = await Task.WhenAny(requestTask, responseTask);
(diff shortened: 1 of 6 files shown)


test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
```diff
@@ -678,6 +681,86 @@ public async Task RequestWithBodies_HasHttpContent(string method, string protoco
...
+        httpContext.Request.Method = "POST";
+        httpContext.Request.ContentLength = contentLength;
+        httpContext.Request.Body = StringToStream(body);
+
+        var destinationPrefix = "https://localhost/";
+        var sut = CreateProxy();
+        var client = MockHttpHandler.CreateClient(
+            async (HttpRequestMessage request, CancellationToken cancellationToken) =>
+            {
+                Assert.Equal(new Version(2, 0), request.Version);
```
MihaZupan, line 701: We should make these tests run against HTTP/1.1 as well

MihaZupan (approved):
Looks good overall, just left some minor comments

src/ReverseProxy/Forwarder/StreamCopier.cs
```diff
@@ -32,13 +32,14 @@ public static ValueTask<(StreamCopyResult, Exception?)> CopyAsync(bool isRequest
             ? new StreamCopierTelemetry(isRequest, clock)
             : null;
 
-        return CopyAsync(input, output, telemetry, activityToken, cancellation);
+        return CopyAsync(input, output, promisedContentLength, telemetry, activityToken, cancellation);
     }
 
-    private static async ValueTask<(StreamCopyResult, Exception?)> CopyAsync(Stream input, Stream output, StreamCopierTelemetry? telemetry, ActivityCancellationTokenSource activityToken, CancellationToken cancellation)
+    private static async ValueTask<(StreamCopyResult, Exception?)> CopyAsync(Stream input, Stream output, long? promisedContentLength, StreamCopierTelemetry? telemetry, ActivityCancellationTokenSource activityToken, CancellationToken cancellation)
```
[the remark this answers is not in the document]
greenEkatherine: Only here or back in call stack?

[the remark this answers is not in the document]
MihaZupan: I would change it here and for the StreamCopyHttpContent field. At which point, it is likely easiest to just always use `long`.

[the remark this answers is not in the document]
greenEkatherine: Done

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -406,6 +406,7 @@ private static void RestoreUpgradeHeaders(HttpContext context, HttpRequestMessag
             // but for now, out of an abundance of caution, we only do it for requests that look like gRPC.
             return new StreamCopyHttpContent(
                 source: request.Body,
+                contentLength: request.Headers.ContentLength,
```
Tratcher, line 409: There's a scenario where we don't copy the Content-Length:
https://github.com/microsoft/reverse-proxy/blob/6eba401c82affa71475a0a17e1a0f452013e06fb/src/ReverseProxy/Forwarder/HttpTransformer.cs#L61-L73

Can StreamCopyHttpContent capture the ContentLength from its own headers on start?

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -585,8 +586,8 @@ private static void RestoreUpgradeHeaders(HttpContext context, HttpResponseMessa
         // :: Step 7-A-2: Copy duplex streams
         using var destinationStream = await destinationResponse.Content.ReadAsStreamAsync();
 
-        var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
-        var responseTask = StreamCopier.CopyAsync(isRequest: false, destinationStream, clientStream, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
+        var requestTask = StreamCopier.CopyAsync(isRequest: true, clientStream, destinationStream, context.Request.ContentLength ?? -1, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
+        var responseTask = StreamCopier.CopyAsync(isRequest: false, destinationStream, clientStream, -1, _clock, activityCancellationSource, activityCancellationSource.Token).AsTask();
```
Tratcher, line 590: Upgrades must not have a content-length.

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -643,7 +644,7 @@ ForwarderError ReportResult(HttpContext context, bool reqeuest, StreamCopyResult
         if (destinationResponseContent != null)
         {
             using var destinationResponseStream = await destinationResponseContent.ReadAsStreamAsync();
-            return await StreamCopier.CopyAsync(isRequest: false, destinationResponseStream, clientResponseStream, _clock, activityCancellationSource, activityCancellationSource.Token);
+            return await StreamCopier.CopyAsync(isRequest: false, destinationResponseStream, clientResponseStream, -1, _clock, activityCancellationSource, activityCancellationSource.Token);
```
Tratcher, line 647: Make a constant for this instead of passing in -1. `UnknownLength = -1`

src/ReverseProxy/Forwarder/StreamCopier.cs
```diff
@@ -67,16 +68,26 @@ private static async ValueTask<(StreamCopyResult, Exception?)> CopyAsync(Stream
...
+                telemetry?.AfterRead(contentLength);
 
                 // Success, reset the activity monitor.
                 activityToken.ResetTimeout();
 
                 // End of the source stream.
                 if (read == 0)
                 {
-                    return (StreamCopyResult.Success, null);
+                    if (promisedContentLength < 0 || contentLength == promisedContentLength)
```
Tratcher, line 81:
```suggestion
                    if (promisedContentLength == UnknownLength || contentLength == promisedContentLength)
```

src/ReverseProxy/Forwarder/StreamCopier.cs
```diff
@@ -67,16 +68,26 @@ private static async ValueTask<(StreamCopyResult, Exception?)> CopyAsync(Stream
...
+                    {
+                        return (StreamCopyResult.Success, null);
+                    }
+                    else
+                    {
+                        return (StreamCopyResult.InputError,
+                            new Exception($"Sent {contentLength} request content bytes, but Content-Length promised {promisedContentLength}."));
+                    }
+
```
Tratcher, line 90:
```suggestion
```

src/ReverseProxy/Forwarder/StreamCopier.cs
```diff
@@ -67,16 +68,26 @@ private static async ValueTask<(StreamCopyResult, Exception?)> CopyAsync(Stream
...
                 if (read == 0)
                 {
-                    return (StreamCopyResult.Success, null);
+                    if (promisedContentLength < 0 || contentLength == promisedContentLength)
+                    {
+                        return (StreamCopyResult.Success, null);
+                    }
+                    else
+                    {
+                        return (StreamCopyResult.InputError,
```
Tratcher, line 87:
```suggestion
                        // This can happen if something in the proxy consumes or modifies part or all of the request body before proxying.
                        return (StreamCopyResult.InputError,
```

src/ReverseProxy/Forwarder/StreamCopier.cs
```diff
@@ -67,16 +68,26 @@ private static async ValueTask<(StreamCopyResult, Exception?)> CopyAsync(Stream
                 }
 
                 read = await input.ReadAsync(buffer.AsMemory(), cancellation);
+                contentLength += read;
```
Tratcher, line 71: You've handled the case where too few bytes were read, but not the case where too many are read. Normally the server would handle that, but it would also be possible for people to cause that by modifying the request in the proxy.
```
                    // Normally this is enforced by the server, but it could get out of sync if something in the proxy modified the body.
                    if (promisedContentLength != UnknownLength || contentLength >= promisedContentLength)
                    {
                        return (StreamCopyResult.InputError, new InvalidOperationException("More bytes received than the specified content-length.");
                    }
```

src/ReverseProxy/Forwarder/StreamCopier.cs
```diff
@@ -67,16 +68,26 @@ private static async ValueTask<(StreamCopyResult, Exception?)> CopyAsync(Stream
...
                 {
-                    return (StreamCopyResult.Success, null);
+                    if (promisedContentLength < 0 || contentLength == promisedContentLength)
+                    {
+                        return (StreamCopyResult.Success, null);
+                    }
+                    else
+                    {
+                        return (StreamCopyResult.InputError,
+                            new Exception($"Sent {contentLength} request content bytes, but Content-Length promised {promisedContentLength}."));
```
Tratcher, line 88:
```suggestion
                            new InvalidOperationException($"Received {contentLength} request content bytes, but Content-Length promised {promisedContentLength}."));
```

src/ReverseProxy/Forwarder/HttpForwarder.cs
Tratcher: Or go back to nullable, its intent is clearer 😁.

Tratcher:
Looks good overall. My comments are mostly style and flow.

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -643,7 +644,7 @@ ForwarderError ReportResult(HttpContext context, bool reqeuest, StreamCopyResult
         if (destinationResponseContent != null)
         {
             using var destinationResponseStream = await destinationResponseContent.ReadAsStreamAsync();
```
Tratcher, line 646:
```suggestion
            using var destinationResponseStream = await destinationResponseContent.ReadAsStreamAsync();
            // The response content-length is enforced by the server.
```

src/ReverseProxy/Forwarder/StreamCopyHttpContent.cs
```diff
@@ -181,7 +183,7 @@ async Task SerializeToStreamAsync(Stream stream, TransportContext? context, Canc
                 return;
             }
```
Tratcher, line 185:
```suggestion

            // Check that the content-length matches the request body size. This can be removed in .NET 7 now that SocketsHttpHandler enforces this: https://github.com/dotnet/runtime/issues/6225.
```

src/ReverseProxy/Forwarder/StreamCopier.cs
MihaZupan: The `contentLength == promisedContentLength` check will catch the "too many" case as well.

src/ReverseProxy/Forwarder/StreamCopier.cs
MihaZupan: I would expect that in most cases a user would hit this because they incorrectly modified the header/body, in which case "Sent" should be clearer. In 7.0, HttpClient will also report this as "sent".

src/ReverseProxy/Forwarder/StreamCopier.cs
Tratcher: Not soon enough to prevent the protocol violation though, only after the fact. We need to check before writing too many bytes to HttpClient.

MihaZupan: Fair point, `SocketsHttpHandler` in 7.0 would prevent this.

src/ReverseProxy/Forwarder/StreamCopier.cs
greenEkatherine: Replaced by `InvalidOperationException`

src/ReverseProxy/Forwarder/HttpForwarder.cs
greenEkatherine: Added a const

src/ReverseProxy/Forwarder/HttpForwarder.cs
greenEkatherine: Replaced with unknown

src/ReverseProxy/Forwarder/HttpForwarder.cs
greenEkatherine: I considered it but ``

src/ReverseProxy/Forwarder/StreamCopier.cs
greenEkatherine: Added

src/ReverseProxy/Forwarder/HttpForwarder.cs
Tratcher: You're correct on the ordering, but you can capture the header at the start of SerializeToStreamAsync.

src/ReverseProxy/Forwarder/StreamCopier.cs
```diff
@@ -19,8 +19,9 @@ internal static class StreamCopier
 {
     // Based on performance investigations, see https://github.com/microsoft/reverse-proxy/pull/330#issuecomment-758851852.
     private const int DefaultBufferSize = 65536;
+    public static readonly long UnknownLength = -1;
```
Tratcher, line 22: Not const?

test/ReverseProxy.Tests/Forwarder/StreamCopierTests.cs
```diff
@@ -51,7 +51,7 @@ public async Task SourceThrows_Reported(bool isRequest)
         var destination = new MemoryStream();
 
         using var cts = ActivityCancellationTokenSource.Rent(TimeSpan.FromSeconds(10), CancellationToken.None);
-        var (result, error) = await StreamCopier.CopyAsync(isRequest, source, destination, clock, cts, cts.Token);
+        var (result, error) = await StreamCopier.CopyAsync(isRequest, source, destination, -1, clock, cts, cts.Token);
```
Tratcher, line 54:
```suggestion
        var (result, error) = await StreamCopier.CopyAsync(isRequest, source, destination, UnknownLength, clock, cts, cts.Token);
```

src/ReverseProxy/Forwarder/HttpForwarder.cs
greenEkatherine: Got it, I moved capturing out of the constructor.

src/ReverseProxy/Forwarder/StreamCopier.cs
greenEkatherine: Replaced

test/ReverseProxy.Tests/Forwarder/StreamCopierTests.cs
```diff
@@ -104,7 +104,7 @@ public async Task Cancelled_Reported(bool isRequest)
 
         using var cts = ActivityCancellationTokenSource.Rent(TimeSpan.FromSeconds(10), CancellationToken.None);
         cts.Cancel();
-        var (result, error) = await StreamCopier.CopyAsync(isRequest, source, destination, new ManualClock(), cts, cts.Token);
+        var (result, error) = await StreamCopier.CopyAsync(isRequest, source, destination, -1, new ManualClock(), cts, cts.Token);
```
Tratcher, line 107:
```suggestion
        var (result, error) = await StreamCopier.CopyAsync(isRequest, source, destination, StreamCopier.UnknownLength, new ManualClock(), cts, cts.Token);
```

Tratcher (approved):

Issue #1465, opened 2021-12-17
Service Fabric AuthorizationPolicy support for routes


francescocristallo:
### What should we add or change to make your life better?

Would be great to have an AutorizationPolicy Label for Service Fabric services in ServiceManifest.xml, like:

`<Label Key="YARP.Routes.route2.AuthorizationPolicy">requireAuth</Label>`

### Why is this important to you?

Having this would remove the burden of Authorization from each single service for simple case scenarios

Tratcher:
SF work has moved to https://github.com/microsoft/service-fabric-yarp

PR #1469, opened 2021-12-18
Add authorization policy to Service Fabric

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/1465


Tratcher:
This SF code is no longer under development and we expect to remove it soon now that https://github.com/microsoft/service-fabric-yarp is available.

Issue #155, opened 2020-05-12
Test end-to-end multi-value header support


Tratcher:
In theory the HTTP spec allows any multi-value headers to be represented as a single comma separated list or as multiple `key: value` headers. In practice there are at least two headers that break clients if merged to a single line, Set-Cookie and WWW-Authenticate.

Test the end-to-end experience for multi-value request and response headers, especially Set-Cookie and WWW-Authenticate. A header's format should be maintained as much as possible in either direction. E.g. don't split up a comma separated list into multiple headers, and don't combine multiple headers into a comma separated list.

There are known issues with Http.Sys and IIS data structures combining multi-value request headers.

Tratcher:
https://github.com/microsoft/reverse-proxy/issues/437 is a good example of how this can break.

Tratcher:
Another example:
https://github.com/microsoft/reverse-proxy/issues/467#issuecomment-734093762

samsp-msft:
Tagging as its the kind of thing that will cause problems for one of our services.

PR #1498, opened 2022-01-06
Test multi-value headers

greenEkatherine:
I'm raising draft PR to discuss results for #155 

I suspect that now there are could be few false-positive and false-negative tests, so I have questions:

1. Is it ok to send empty string if no value was passed?
2.  It seems that `WWW-Authenticate` and `Set-Cookie` should use semicolon delimiter as well as `Cookie` header but now it is not working
3. Should I add tests for multi-value headers into https://github.com/microsoft/reverse-proxy/blob/main/test/ReverseProxy.FunctionalTests/HeaderTests.cs as well?

- [ ] Add tests for HTTP/1.1

### DIFF
--- a/test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
+++ b/test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
@@ -879,6 +879,80 @@ public class HttpForwarderTests
         events.AssertContainProxyStages(hasRequestContent: false);
     }
 
+    [Theory]
+    [MemberData(nameof(MultiHeadersData))]
+    public async Task RequestWithMultiHeaders(string headerName, string[] headers, string[] expectedHeader)
+    {
+        var events = TestEventListener.Collect();
+
+        var httpContext = new DefaultHttpContext();
+        httpContext.Request.Method = "GET";
+        httpContext.Request.Headers.Add(headerName, headers);
+
+        var destinationPrefix = "https://localhost/";
+        var sut = CreateProxy();
+        var client = MockHttpHandler.CreateClient(
+            (HttpRequestMessage request, CancellationToken cancellationToken) =>
+            {
+
+                Assert.Equal(new Version(2, 0), request.Version);
+                Assert.Equal("GET", request.Method.Method, StringComparer.OrdinalIgnoreCase);
+                Assert.True(request.Headers.TryGetValues(headerName, out var sentHeaders));
+                Assert.NotNull(sentHeaders);
+                Assert.Equal(expectedHeader, sentHeaders);
+
+                var response = new HttpResponseMessage(HttpStatusCode.OK) { Content = new ByteArrayContent(Array.Empty<byte>()) };
+                return Task.FromResult(response);
+            });
+
+        await sut.SendAsync(httpContext, destinationPrefix, client);
+
+        Assert.Null(httpContext.Features.Get<IForwarderErrorFeature>());
+        Assert.Equal(StatusCodes.Status200OK, httpContext.Response.StatusCode);
+
+        AssertProxyStartStop(events, destinationPrefix, httpContext.Response.StatusCode);
+        events.AssertContainProxyStages(hasRequestContent: false);
+    }
+
+    public static IEnumerable<string> MultiHeaderNames()
+    {
+        var headers = new[]
+        {
+            HeaderNames.Accept,
+            HeaderNames.AcceptCharset,
+            HeaderNames.AcceptEncoding,
+            HeaderNames.AcceptLanguage,
+            HeaderNames.SetCookie,
+            HeaderNames.Via,
+            HeaderNames.Warning,
+            HeaderNames.WWWAuthenticate
+        };
+
+        foreach (var header in headers)
+        {
+            yield return header;
+        }
+    }
+
+    public static IEnumerable<object[]> MultiHeadersData()
+    {
+        var values = new string[][] {
+            new[] { "testA=A_Value", "testB=B_Value", "testC=C_Value" },
+            new[] { "testA=A_Value, testB=B_Value", "testC=C_Value" },
+            new[] { "testA=A_Value", "",  "testB=B_Value, testC=C_Value" },
+            new[] { "", "" },
+            new[] { "testA=A_Value, testB=B_Value, testC=C_Value" }
+        };
+
+        foreach (var header in MultiHeaderNames())
+        {
+            foreach (var value in values)
+            {
+                yield return new object[] { header, value, value.Where(s => !string.IsNullOrEmpty(s)).ToArray() };
+            }
+        }
+    }
+
     [Fact]
     public async Task OptionsWithVersion()
     {
@@ -2131,6 +2205,33 @@ public class HttpForwarderTests
         await sut.SendAsync(httpContext, destinationPrefix, client, new ForwarderRequestConfig { Version = Version.Parse(protocol) });
 
         Assert.Equal((int)HttpStatusCode.OK, httpContext.Response.StatusCode);
+    }
+
+    [Theory]
+    [MemberData(nameof(MultiHeadersData))]
+    public async Task ResponseWithMultiHeaders(string headerName, string[] headers, string[] expectedHeader)
+    {
+        var httpContext = new DefaultHttpContext();
+        httpContext.Request.Method = "GET";
+
+        var destinationPrefix = "https://localhost:123/a/b/";
+        var sut = CreateProxy();
+        var client = MockHttpHandler.CreateClient(
+            async (HttpRequestMessage request, CancellationToken cancellationToken) =>
+            {
+                await Task.Yield();
+
+                var response = new HttpResponseMessage(HttpStatusCode.OK);
+
+                response.Headers.TryAddWithoutValidation(headerName, headers);
+
+                return response;
+            });
+
+        await sut.SendAsync(httpContext, destinationPrefix, client, new ForwarderRequestConfig { Version = new Version(2, 0) });
+
+        Assert.Equal((int)HttpStatusCode.OK, httpContext.Response.StatusCode);
+        Assert.Equal(expectedHeader, headers);
     }
 
     [Theory]


Tratcher:
> 3\. Should I add tests for multi-value headers into https://github.com/microsoft/reverse-proxy/blob/main/test/ReverseProxy.FunctionalTests/HeaderTests.cs as well?

Yes, we need to see what makes it out on the wire. SocketsHttpHandler and Kestrel ultimately decide the format to write.

Tratcher:
> 2\. It seems that `WWW-Authenticate` and `Set-Cookie` should use semicolon delimiter as well as `Cookie` header but now it is not working

The specs for these are weird:
- Cookie is always supposed to be a single header (except maybe in HTTP/2 https://github.com/dotnet/aspnetcore/issues/26461), separated by ';'.
- Set-Cookie is always supposed to be separate headers, one per cookie.
- Most clients expect WWW-Authenticate to be separate headers.

Tratcher:
> 1. Is it ok to send empty string if no value was passed?

You mean `Header: `? Yes. It's not supposed to matter in most cases, but some applications rely on the presence of a header even if there's no value. YARP should flow these headers just like any other.

test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
```diff
@@ -879,6 +879,80 @@ public async Task RequestWithCookieHeaders()
...
+        await sut.SendAsync(httpContext, destinationPrefix, client);
+
+        Assert.Null(httpContext.Features.Get<IForwarderErrorFeature>());
+        Assert.Equal(StatusCodes.Status200OK, httpContext.Response.StatusCode);
+
+        AssertProxyStartStop(events, destinationPrefix, httpContext.Response.StatusCode);
+        events.AssertContainProxyStages(hasRequestContent: false);
+    }
+
+    public static IEnumerable<string> MultiHeaderNames()
```
Tratcher, line 917: Make a separate list for request and response headers. E.g. Cookie is only for requests and Set-Cookie is only for responses.

Tratcher: You may need a separate test for Cookie anyways given its unique format.

greenEkatherine:
I updated what we discussed but few more issues left. I need help to understand if it is a correct behavior:

1. HttpHeader removes empty values, so both `RequestWithEmptyMultiHeaders` and `ResponseWithEmptyMultiHeaders` failed for all headers.
2. For request headers I got empty value untouched, if it was not the only one, but for response headers behavior is the opposite (except `Set-Cookie` header).
3. There are several headers that seems to be required but were removed: `Allow`, `Upgrade`, `TE`, `Tranfer-Encoding` and `Content-Length/Type/Encoding`. Do we need to test them as well?

Tratcher:
> 1. HttpHeader removes empty values, so both `RequestWithEmptyMultiHeaders` and `ResponseWithEmptyMultiHeaders` failed for all headers.

There's a test over here for empty single-value headers.

Tratcher:
> 3\. There are several headers that seems to be required but were removed: `Allow`, `Upgrade`, `TE`, `Tranfer-Encoding` and `Content-Length/Type/Encoding`. Do we need to test them as well?

Most of these require special processing since they affect the request format on the wire. See:
https://github.com/microsoft/reverse-proxy/blob/07aa1683fb1b697c81f5fb47492dcbf52f203e0c/src/ReverseProxy/Forwarder/RequestUtilities.cs#L63-L85

Allow, Content-Type, and Content-Encoding shouldn't have been removed. Content-Length is rarely removed.

greenEkatherine:
> There's a test over here for empty single-value headers. How is that working but multi-value fails? https://github.com/microsoft/reverse-proxy/pull/1202/files#diff-4ffa10100339146ae9048f52f94656c0aff1b984dabdbb5200d87284069b1379R23

It's probably because of different scenarios, in your example empty values are checked by `IDictionary.TryGetValue`, my tests in HeaderTests.cs also passed. I was talking about HttpForwarderTests.cs https://github.com/microsoft/reverse-proxy/pull/1498/files#:~:text=public%20async%20Task%20RequestWithEmptyMultiHeaders(string%20version%2C%20string%20headerName%2C%20string%5B%5D%20headers), there I'm checking values in `HttpHeaders.Contains`.
> 
> > 2. For request headers I got empty value untouched, if it was not the only one, but for response headers behavior is the opposite (except Set-Cookie header).
> 
> I don't understand, can you give examples?

I mean case `{ "testA=A_Value", "",  "testB=B_Value, testC=C_Value" }`. In request headers I got as a result `{ "testA=A_Value", "",  "testB=B_Value, testC=C_Value" },` but in response it is mostly without empty value `{ "testA=A_Value", "testB=B_Value, testC=C_Value" },` except for `Set-Cookie`.

greenEkatherine:
> Most of these require special processing since they affect the request format on the wire. See:
> 
> https://github.com/microsoft/reverse-proxy/blob/07aa1683fb1b697c81f5fb47492dcbf52f203e0c/src/ReverseProxy/Forwarder/RequestUtilities.cs#L63-L85
> 
> Allow, Content-Type, and Content-Encoding shouldn't have been removed. Content-Length is rarely removed. Do you mean they were removed if they contained empty values? Maybe HttpHeaders if marking them as invalid.

Thank you, I see about special headers, I won't add them then. But I don't see `Content-Language` (instead of Content-Length, sorry, because length cannot be multi-value), `Allow`, `Content-Type`, and `Content-Encoding` even if they had values:
- for request https://github.com/microsoft/reverse-proxy/pull/1498/files#diff-e2060282ce1ef75079b6dbfc7c67f8769f2231509ee7d98e0284d766060136baR905
- for response https://github.com/microsoft/reverse-proxy/pull/1498/files#diff-e2060282ce1ef75079b6dbfc7c67f8769f2231509ee7d98e0284d766060136baR2333

Maybe it is not a proper place/method to check them?

Tratcher:
> > There's a test over here for empty single-value headers. How is that working but multi-value fails? https://github.com/microsoft/reverse-proxy/pull/1202/files#diff-4ffa10100339146ae9048f52f94656c0aff1b984dabdbb5200d87284069b1379R23
> 
> It's probably because of different scenarios, in your example empty values are checked by `IDictionary.TryGetValue`, my tests in HeaderTests.cs also passed. I was talking about HttpForwarderTests.cs https://github.com/microsoft/reverse-proxy/pull/1498/files#:~:text=public%20async%20Task%20RequestWithEmptyMultiHeaders(string%20version%2C%20string%20headerName%2C%20string%5B%5D%20headers), there I'm checking values in `HttpHeaders.Contains`.
> 
> > > 2. For request headers I got empty value untouched, if it was not the only one, but for response headers behavior is the opposite (except Set-Cookie header).
> > 
> > 
> > I don't understand, can you give examples?
> 
> I mean case `{ "testA=A_Value", "", "testB=B_Value, testC=C_Value" }`. In request headers I got as a result `{ "testA=A_Value", "", "testB=B_Value, testC=C_Value" },` but in response it is mostly without empty value `{ "testA=A_Value", "testB=B_Value, testC=C_Value" },` except for `Set-Cookie`.
> 
> UPD for p.2: I made a mistake here, the behavior for request/response is consistent - it always deletes empty values except for `Set-Cookie`

Dropping an empty value from a multi-value list is acceptable, so long as the header gets sent. Keep the test for this condition and tell it to expect only non-empty results. The example customers were concerned about was a header with a single empty value being removed completely.

Tratcher:
> > Most of these require special processing since they affect the request format on the wire. See:
> > https://github.com/microsoft/reverse-proxy/blob/07aa1683fb1b697c81f5fb47492dcbf52f203e0c/src/ReverseProxy/Forwarder/RequestUtilities.cs#L63-L85
> > 
> > Allow, Content-Type, and Content-Encoding shouldn't have been removed. Content-Length is rarely removed. Do you mean they were removed if they contained empty values? Maybe HttpHeaders if marking them as invalid.
> 
> Thank you, I see about special headers, I won't add them then. But I don't see `Content-Language` (instead of Content-Length, sorry, because length cannot be multi-value), `Allow`, `Content-Type`, and `Content-Encoding` even if they had values:
> 
> * for request https://github.com/microsoft/reverse-proxy/pull/1498/files#diff-e2060282ce1ef75079b6dbfc7c67f8769f2231509ee7d98e0284d766060136baR905
> * for response https://github.com/microsoft/reverse-proxy/pull/1498/files#diff-e2060282ce1ef75079b6dbfc7c67f8769f2231509ee7d98e0284d766060136baR2333
> 
> Maybe it is not a proper place/method to check them?

Content headers must be in HttpRequestMessage.Content.Headers, not HttpRequestMessage.Headers.

greenEkatherine:
> Dropping an empty value from a multi-value list is acceptable, so long as the header gets sent. Keep the test for this condition and tell it to expect only non-empty results. The example customers were concerned about was a header with a single empty value being removed completely.

In that case total failures of `RequestWithEmptyMultiHeaders` and `ResponseWithEmptyMultiHeaders` indicate a bug. At least if the scenario is correct.

greenEkatherine:
Thanks for the tip, I found it! So, the only question left with missed headers is about `Allow`. 
> 
> Content headers must be in HttpRequestMessage.Content.Headers, not HttpRequestMessage.Headers. https://docs.microsoft.com/en-us/dotnet/api/system.net.http.headers.httpcontentheaders?view=net-6.0

Tratcher:
Allow is also an HttpContent header.

Tratcher:
> > Dropping an empty value from a multi-value list is acceptable, so long as the header gets sent. Keep the test for this condition and tell it to expect only non-empty results. The example customers were concerned about was a header with a single empty value being removed completely.
> 
> In that case total failures of `RequestWithEmptyMultiHeaders` and `ResponseWithEmptyMultiHeaders` indicate a bug. At least if the scenario is correct.

Contains might be over-normalizing. Try enumerating under the debugger, or TryGetValue.

MihaZupan:
If a header is considered "known", we may have restrictions on the format of the value.

Add/TryGetValues/Enumeration all force validation of the value, potentially removing the header.

In order to preserve "invalid" values in the collection (and therefore on the wire), they should be added via `TryAddWithoutValidation` and their existence checked via `Headers.NonValidated`.

greenEkatherine:
> If a header is considered "known", we may have restrictions on the format of the value.
> 
> Add/TryGetValues/Enumeration all force validation of the value, potentially removing the header.
> 
> In order to preserve "invalid" values in the collection (and therefore on the wire), they should be added via `TryAddWithoutValidation` and their existence checked via `Headers.NonValidated`. Note that `NonValidated` only exists in 6.0+.

It works, thank you! So I will leave unit test with empty values for 6.0+.

greenEkatherine:
What's left to do to make this PR ready to merge:

1. Make check for content headers differently.
2. Add function test into HeaderTests.cs for response.

test/ReverseProxy.FunctionalTests/HeaderTests.cs
```diff
@@ -466,4 +468,248 @@ public async Task ContentLengthAndTransferEncoding_ContentLengthRemoved()
...
+        var proxyTcs = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
+        var appTcs = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+        var test = new TestEnvironment(
+            context =>
+            {
+                try
+                {
+                    Assert.True(context.Request.Headers.TryGetValue(headerName, out var headerValues));
+                    Assert.Single(headerValues);
```
Tratcher, line 485: So Kestrel preserves mutli-value request headers, but HttpClient doesn't? Please file an issue for that in runtime.

test/ReverseProxy.FunctionalTests/HeaderTests.cs
```diff
@@ -466,4 +468,248 @@ public async Task ContentLengthAndTransferEncoding_ContentLengthRemoved()
...
+            var response = responseBuilder.ToString();
+
+            Assert.Null(proxyError);
+            Assert.Null(unhandledError);
+
+            var lines = response.Split("\r\n");
+            Assert.Equal("HTTP/1.1 200 OK", lines[0]);
+            foreach (var expected in expectedValues)
+            {
+                var test = lines.Contains($"{headerName}: {expected}");
```
Tratcher, line 710:
```suggestion
```

test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
```diff
@@ -2133,6 +2327,37 @@ public async Task Request_RemoveProhibitedHeaders(string protocol, string prohib
...
+                    Assert.True(response.Content.Headers.TryAddWithoutValidation(headerName, headers));
+                }
+
+                return response;
+            });
+
+        await sut.SendAsync(httpContext, destinationPrefix, client, new ForwarderRequestConfig { Version = new Version(version) });
+
+        Assert.Equal((int)HttpStatusCode.OK, httpContext.Response.StatusCode);
+        Assert.Contains(headerName, httpContext.Response.Headers);
```
Tratcher, line 2357: How does this line work? Headers is an `IEnumerable<string, StringValues>`, but you're checking if it contains a `string`? Shouldn't this be `Assert.True(httpContext.Response.Headers.TryGetValue(headerName, var _))`?

Tratcher (approved):

test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
greenEkatherine: It doesn't look so, I see some failures, still working on it

test/ReverseProxy.FunctionalTests/HeaderTests.cs
MihaZupan: You mean the fact that HttpClient will send multiple values on the same line? That is by design.

Tratcher: That design will be a problem for proxy scenarios. Being as transparent as possible includes not changing the header format.

MihaZupan: dotnet/runtime#62981 (before the last commit) would allow you to get headers serialized in multiple lines, if you called `TryAddWithoutValidation` multiple times. We decided against that as a potentially breaking change for little/no perf benefit.
Even then, it was done on a best-effort basis, as accessing a value / enumerating the collection would force us to group the entries back into a single line.

Have we heard any requests to support this? I don't believe I've seen anything in runtime yet.

test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
greenEkatherine: Ah, got it, I didn't put all tests that check non-validated empty headers under .NET 6.0 condition. I will replace that string as well

test/ReverseProxy.FunctionalTests/HeaderTests.cs
Tratcher: We have had specific complaints about response headers, so we've been very careful about this on the server side. I'm not aware of specific complaints about request headers yet. These tests were done as a precaution to understand and document the current behavior.

The inability to preserve header formats in one direction is notable, but we won't know if it's blocking until a customer notices. I'm surprised our recent partner deployments didn't notice. Now is a good time to check if there are easy workarounds or fixes so we're prepared when it does come up.

Tratcher: https://github.com/dotnet/runtime/issues/63808

test/ReverseProxy.FunctionalTests/HeaderTests.cs
```diff
@@ -466,4 +470,246 @@ public async Task ContentLengthAndTransferEncoding_ContentLengthRemoved()
...
+            while (true)
+            {
+                var count = await stream.ReadAsync(buffer);
+                if (count == 0)
+                {
+                    break;
+                }
+                responseBuilder.Append(Encoding.ASCII.GetString(buffer, 0, count));
+            }
+            var response = responseBuilder.ToString();
```
MihaZupan, line 550:
```suggestion
            var response = await new StreamReader(stream).ReadToEndAsync();
```

test/ReverseProxy.FunctionalTests/HeaderTests.cs
```diff
@@ -466,4 +470,246 @@ public async Task ContentLengthAndTransferEncoding_ContentLengthRemoved()
...
```
MihaZupan, line 702:
```suggestion
            var response = await new StreamReader(stream).ReadToEndAsync();
```

test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
```diff
@@ -879,6 +887,203 @@ public async Task RequestWithCookieHeaders()
...
+        Assert.Null(httpContext.Features.Get<IForwarderErrorFeature>());
+        Assert.Equal(StatusCodes.Status200OK, httpContext.Response.StatusCode);
+
+        AssertProxyStartStop(events, destinationPrefix, httpContext.Response.StatusCode);
+        events.AssertContainProxyStages(hasRequestContent: false);
+    }
+#endif
+
+    internal static bool AreEqualIgnoringEmptyStrings(IEnumerable<string> left, IEnumerable<string> right)
+    => Enumerable.SequenceEqual(left.Where(s => !string.IsNullOrEmpty(s)).ToArray(), right.Where(s => !string.IsNullOrEmpty(s)).ToArray());
```
MihaZupan, line 977:
```suggestion
    => Assert.Equal(left.Where(s => !string.IsNullOrEmpty(s)), right.Where(s => !string.IsNullOrEmpty(s)));
```

MihaZupan (approved):

Issue #1479, opened 2021-12-22
Kubernetes Ingress Controller configures Route metadata


specialforest:
### What should we add or change to make your life better?
We need a way to specify metadata in Ingress object and make Kubernetes Ingress Controller populate Metadata field of RouteConfig.

### Why is this important to you?
I'd like to use Route Metadata to configure custom middlewares in Reverse Proxy.

specialforest:
To specify metadata we could use annotations. Define a prefix, e.g. "metadata.yarp.ingress.kubernetes.io":
```yaml
annotations:
  metadata.yarp.ingress.kubernetes.io/foo: bar
  metadata.yarp.ingress.kubernetes.io/another-key: value
```
A single Ingress object can define multiple paths and they are converted into multiple routes in Reverse Proxy. Here I propose to copy metadata to each of these routes.

karelz:
It is being added in PR #1263

specialforest:
Thanks for linking this PR! It doesn't cover routes metadata currently though. I made a change previously and re-wrote it to follow the same approach as in the PR. It's currently in my branch https://github.com/specialforest/reverse-proxy/tree/k8-ingress-metadata

MihaZupan:
Added in #1518

PR #1518, opened 2022-01-14
Add support for Route Metadata in Kubernetes via annotations

specialforest:
Add support for Route Metadata in Kubernetes via annotations #1479.

A change similar to #1263.

### DIFF
--- a/src/Kubernetes.Controller/Converters/YarpParser.cs
+++ b/src/Kubernetes.Controller/Converters/YarpParser.cs
@@ -98,6 +98,7 @@ internal static class YarpParser
                     Transforms = ingressContext.Options.Transforms,
                     AuthorizationPolicy = ingressContext.Options.AuthorizationPolicy,
                     CorsPolicy = ingressContext.Options.CorsPolicy,
+                    Metadata = ingressContext.Options.Metadata,
                 });
 
                 // Add destination for every endpoint address
@@ -192,6 +193,10 @@ internal static class YarpParser
         if (annotations.TryGetValue("yarp.ingress.kubernetes.io/health-check", out var healthCheck))
         {
             options.HealthCheck = YamlDeserializer.Deserialize<HealthCheckConfig>(healthCheck);
+        }
+        if (annotations.TryGetValue("yarp.ingress.kubernetes.io/metadata", out var routeMetadata))
+        {
+            options.Metadata = YamlDeserializer.Deserialize<Dictionary<string, string>>(routeMetadata);
         }
 
         // metadata to support:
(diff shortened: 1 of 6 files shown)


src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -137,6 +140,14 @@ private static YarpIngressOptions HandleAnnotations(YarpIngressContext context,
             options.Https = http.Equals("https", StringComparison.OrdinalIgnoreCase);
         }
 
+        var routeMetadata = annotations
```
specialforest, line 143: I think can follow the same pattern as #1263 and put the whole metadata as JSON dictionary under single "yarp.ingress.kubernetes.io/metadata" or "yarp.ingress.kubernetes.io/route-metadata" annotation. I'll wait before the other change is merge and then add this part.

MihaZupan: Hey @specialforest, #1263 has just been merged so this should now be unblocked.

src/Kubernetes.Controller/Converters/YarpParser.cs
```diff
@@ -193,6 +194,10 @@ private static YarpIngressOptions HandleAnnotations(YarpIngressContext context,
         {
             options.HealthCheck = YamlDeserializer.Deserialize<HealthCheckConfig>(healthCheck);
         }
+        if (annotations.TryGetValue("yarp.ingress.kubernetes.io/metadata", out var routeMetadata))
```
specialforest, line 197: It may worth to use "route-metadata" here, since ClusterConfig also has Metdata field.

src/Kubernetes.Controller/Converters/YarpParser.cs
specialforest: Rebased.

src/Kubernetes.Controller/Converters/YarpParser.cs
MihaZupan: K8s annotations already represent a mix of settings for both clusters and routes.
E.g. `LoadBalancingPolicy`, `HttpClientConfig` for a cluster or `Transforms`, `CorsPolicy` for a route is part of the same "dictionary".

Metadata is something that appears on both routes & clusters in YARP's config model, but as separate collections.
I'm wondering if setting both `cluster.Metadata` and `route.Metadata` to the same thing would be fine in this case. Is there any value in making them separate given that the config only comes from one place for an ingress?

@macsux @dpbevin @Tratcher any thoughts here?

samsp-msft:
Triage: We'd recommend using separate properties for cluster and route metadata as combining them may have weird implications. So have `route-metdata` as the k8s property.

MihaZupan (approved):
Thanks!

PR #1533, opened 2022-01-26
Handle distributed context headers in custom propagator

### LINKED ISSUE #1497: YARP is filtering headers (with common names) because of distributed tracing
Long story short, change in net6/yarp filtering out a set of headers breaking our customers  consuming these headers in their business logic.  We did have a workaround deployed to unblock customers.   The ask is whether net6/yarp could expose an simpler option to opt out of such behavior (eg. one line fix).  This would allow us to write less codes especially avoiding manually propagate a pre-hardcoded set of headers.   It would sustain longer term and benefit other using net6/yarp as well.

[Article about headers removed](https://github.com/microsoft/reverse-proxy/blob/main/docs/docfx/articles/header-guidelines.md#traceparent-request-id-tracestate-baggage-correlation-context)
> TraceParent, Request-Id, TraceState, Baggage, Correlation-Context
> These headers relate to distributed tracing. They are automatically removed on .NET 6 or later so that the forwarding HttpClient can replace them with updated values.

greenEkatherine:
Resolves #1497 

Added `ReverseProxyPropagator` that handles headers removal. Default inner implementation removes distributed context headers. It can be changed by overriding `Propagator` property in `ForwarderHttpClientFactory` which is used to set up `handler.ActivityHeadersPropagator`.

### DIFF
--- a/docs/docfx/articles/direct-forwarding.md
+++ b/docs/docfx/articles/direct-forwarding.md
@@ -53,7 +53,7 @@ public void Configure(IApplicationBuilder app, IHttpForwarder forwarder)
         UseProxy = false,
         AllowAutoRedirect = false,
         AutomaticDecompression = DecompressionMethods.None,
-        UseCookies = false
+        UseCookies = false,
         ActivityHeadersPropagator = new ReverseProxyPropagator(DistributedContextPropagator.Current)
     });
     var transformer = new CustomTransformer(); // or HttpTransformer.Default;
--- a/docs/docfx/articles/direct-forwarding.md
+++ b/docs/docfx/articles/direct-forwarding.md
@@ -60,17 +60,16 @@ public void Configure(IApplicationBuilder app, IHttpForwarder forwarder)
     var requestConfig = new ForwarderRequestConfig { ActivityTimeout = TimeSpan.FromSeconds(100) };
 
     app.UseRouting();
-    app.UseAuthorization();
     app.UseEndpoints(endpoints =>
     {
         endpoints.Map("/{**catch-all}", async httpContext =>
         {
             var error = await forwarder.SendAsync(httpContext, "https://localhost:10000/",
-                httpClient, requestConfig, transformer);
+                httpClient, requestOptions, transformer);
             // Check if the operation was successful
             if (error != ForwarderError.None)
             {
-                var errorFeature = httpContext.GetForwarderErrorFeature();
+                var errorFeature = httpContext.Features.Get<IForwarderErrorFeature>();
                 var exception = errorFeature.Exception;
             }
         });
@@ -84,9 +83,18 @@ private class CustomTransformer : HttpTransformer
     public override async ValueTask TransformRequestAsync(HttpContext httpContext,
         HttpRequestMessage proxyRequest, string destinationPrefix)
     {
-        // Copy headers normally and then remove the original host.
-        // Use the destination host from proxyRequest.RequestUri instead.
+        // Copy all request headers
         await base.TransformRequestAsync(httpContext, proxyRequest, destinationPrefix);
+
+        // Customize the query string:
+        var queryContext = new QueryTransformContext(httpContext.Request);
+        queryContext.Collection.Remove("param1");
+        queryContext.Collection["area"] = "xx2";
+
+        // Assign the custom uri. Be careful about extra slashes when concatenating here. RequestUtilities.MakeDestinationAddress is a safe default.
+        proxyRequest.RequestUri = RequestUtilities.MakeDestinationAddress("https://example.com", httpContext.Request.Path, queryContext.QueryString);
+
+        // Suppress the original request header, use the one from the destination Uri.
         proxyRequest.Headers.Host = null;
     }
 }
--- a/docs/docfx/articles/direct-forwarding.md
+++ b/docs/docfx/articles/direct-forwarding.md
@@ -54,6 +54,7 @@ public void Configure(IApplicationBuilder app, IHttpForwarder forwarder)
         AllowAutoRedirect = false,
         AutomaticDecompression = DecompressionMethods.None,
         UseCookies = false
+        ActivityHeadersPropagator = new ReverseProxyPropagator(DistributedContextPropagator.Current)
     });
     var transformer = new CustomTransformer(); // or HttpTransformer.Default;
     var requestConfig = new ForwarderRequestConfig { ActivityTimeout = TimeSpan.FromSeconds(100) };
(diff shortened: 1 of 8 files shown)


src/ReverseProxy/Forwarder/ForwarderHttpClientFactory.cs
```diff
@@ -45,7 +46,10 @@ public HttpMessageInvoker CreateClient(ForwarderHttpClientContext context)
             UseProxy = false,
             AllowAutoRedirect = false,
             AutomaticDecompression = DecompressionMethods.None,
-            UseCookies = false
+            UseCookies = false,
+#if NET6_0_OR_GREATER
+            ActivityHeadersPropagator = Propagator
```
MihaZupan, line 51:
```suggestion
            ActivityHeadersPropagator = new ReverseProxyPropagator(DistributedContextPropagator.Current)
```

We should still pick up the `DistributedContextPropagator.Current` over `CreateDefaultPropagator`.

I also don't think we need a property on `ForwarderHttpClientFactory` for this as the user can already change the propagator inside `ConfigureHandler`.

MihaZupan (approved):
We should update the direct-forwarding docs to include changing this property on the `SocketsHttpHandler` as well:
https://github.com/microsoft/reverse-proxy/blob/96ed54fd0286638838edb1543333a00dbb3c595b/docs/docfx/articles/direct-forwarding.md?plain=1#L51-L57

greenEkatherine:
> We should update the direct-forwarding docs to include changing this property on the `SocketsHttpHandler` as well: https://github.com/microsoft/reverse-proxy/blob/96ed54fd0286638838edb1543333a00dbb3c595b/docs/docfx/articles/direct-forwarding.md?plain=1#L51-L57

I noticed that example there doesn't match with `Direct.Sample` and not only with this new property. Should I update `Direct.Sample` as well?

MihaZupan:
Yes please.

We should also consider updating https://microsoft.github.io/reverse-proxy/articles/header-guidelines.html#traceparent-request-id-tracestate-baggage-correlation-context to mention you can turn header removal off by calling
```c#
.ConfigureHttpHandler(handler => handler.Propagator = null)
```
in Startup

src/ReverseProxy/Forwarder/ReverseProxyPropagator.cs
```diff
@@ -0,0 +1,47 @@
...
+#if NET6_0_OR_GREATER
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Net.Http;
+
+namespace Yarp.ReverseProxy.Forwarder;
+
+public sealed class ReverseProxyPropagator : DistributedContextPropagator
```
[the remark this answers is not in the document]
greenEkatherine: Added more info into docs

docs/docfx/articles/header-guidelines.md
```diff
@@ -37,7 +37,10 @@ This response header is used with HTTP/3 upgrades and only applies to the immedi
 
 ### TraceParent, Request-Id, TraceState, Baggage, Correlation-Context
 
-These headers relate to distributed tracing. They are automatically removed on .NET 6 or later so that the forwarding HttpClient can replace them with updated values.
+These headers relate to distributed tracing. They are automatically removed on .NET 6 or later so that the forwarding HttpClient can replace them with updated values. You can turn header removal off by calling:
+```C#
+.ConfigureHttpHandler((_, handler) => handler.ActivityHeadersPropagator = null)
```
MihaZupan, line 42:
```suggestion
These headers relate to distributed tracing. They are automatically removed on .NET 6 or later so that the forwarding HttpClient can replace them with updated values.
You can opt out of modifying these headers by setting `SocketsHttpHandler.ActivityHeadersPropagator ` to `null`:
```C#
services.AddReverseProxy()
    .ConfigureHttpClient((_, handler) => handler.ActivityHeadersPropagator = null);
```

docs/docfx/articles/direct-forwarding.md
```diff
@@ -54,22 +54,22 @@ public void Configure(IApplicationBuilder app, IHttpForwarder forwarder)
...
 
     app.UseRouting();
-    app.UseAuthorization();
     app.UseEndpoints(endpoints =>
     {
         endpoints.Map("/{**catch-all}", async httpContext =>
         {
             var error = await forwarder.SendAsync(httpContext, "https://localhost:10000/",
-                httpClient, requestConfig, transformer);
+                httpClient, requestOptions, transformer);
```
MihaZupan, line 68:
```suggestion
                httpClient, requestConfig, transformer);
```

This are still named `requestConfig` in this sample.

docs/docfx/articles/direct-forwarding.md
```diff
@@ -54,22 +54,22 @@ public void Configure(IApplicationBuilder app, IHttpForwarder forwarder)
...
             // Check if the operation was successful
             if (error != ForwarderError.None)
             {
-                var errorFeature = httpContext.GetForwarderErrorFeature();
+                var errorFeature = httpContext.Features.Get<IForwarderErrorFeature>();
```
MihaZupan, line 72:
```suggestion
                var errorFeature = httpContext.GetForwarderErrorFeature();
```

I would update the `Direct.Sample` project to use the helper method instead of reverting that here.

Overall I don't think the samples have to be exactly the same except for how you construct the handler itself.

docs/docfx/articles/direct-forwarding.md
```diff
@@ -54,12 +54,12 @@ public void Configure(IApplicationBuilder app, IHttpForwarder forwarder)
         AllowAutoRedirect = false,
         AutomaticDecompression = DecompressionMethods.None,
         UseCookies = false
```
MihaZupan, line 56:
```suggestion
        UseCookies = false,
```

src/ReverseProxy/Forwarder/ReverseProxyPropagator.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Net.Http;
+
+namespace Yarp.ReverseProxy.Forwarder;
+
+/// <summary>
+/// A default propagator that removes headers.
```
MihaZupan, line 14:
```suggestion
/// Removes existing headers and then delegates to the inner propagator.
```

Tratcher (approved):

greenEkatherine:
> https://github.com/microsoft/reverse-proxy/blob/96ed54fd0286638838edb1543333a00dbb3c595b/samples/ReverseProxy.Direct.Sample/Startup.cs#L36-L42
> 
> should include the `ActivityHeadersPropagator`.

Oh, forgot this one, adding...

Issue #368, opened 2020-08-18
Handling severe errors


Tratcher:
https://github.com/microsoft/reverse-proxy/pull/357#discussion_r466628404

There are some severe error conditions that we need to come up with a consistent, extensible policy for addressing. Some examples:

1. Fail to load config data at startup. Current behavior: throw with intent to crash the app.
1. Fail to reload config data. Current behavior: Log and keep running with the last known good config. Some of these errors will prevent the proxy from listening for future config changes because IChangeToken is single use.

An application developer may want to FailFast the app or trigger a graceful shutdown in the case of a config reload failure, but app lifetime should be something controlled by the developer, not by the ReverseProxy library. We may need to provide extensibility points where the developer can handle these errors in their own way.

karelz:
Triage: We need to decide what is the right API/behavior. FailFast is likely not a good behavior.

karelz:
Triage: This does not seem to be blocking @davidni (right?), moving to Backlog ...

PR #1534, opened 2022-01-27
Support multiple configuration sources

### LINKED ISSUE #979: Support multiple configuration sources
The current implementation of IProxyConfigProvider is registered as a singleton service, which prevents the use of multiple configuration sources.

Registering a custom programatic source prevents the use of the configuration file.

Having multiple configuration sources could allow us to have a set of default Routes / Clusters on the appsettings.json while we perform programatic configuration on a second implementation of IProxyConfigProvider.

It may also allow for overriding some Routes/Clusters with less destinations for Debugging/Developer workflow.

Tratcher:
Fixes #979
Contributes to #368

Customers want to be able to do more advanced things with config like:
- Load it from multiple places: JSON, in memory, k8, etc.
- Be able to mix and match across those sources. E.g. routes are defined in memory and clusters in json.

Design:
- Multiple LoadFromX calls can now be made. No new APIs are needed, but we will want to ensure the existing ones behave as expected if called multiple times. I've only verified LoadFromConfig so far. Others might not make sense to call multiple times but could still be mixed and matched.
- The config manager will resolve all IProxyConfigProvider's from the DI container, concatenate their contents, run validation, and update the internal state.
- RouteConfig and ClusterConfig are the smallest units supported. Duplicates are an error and will not be merged.
- Changes in any provider trigger a reload. I tried only reloading routes/clusters from the modified provider but it was much harder to support mixing clusters and routes across sources that way. We still diff each route and cluster and only update them if modified.
- Errors in any config source will prevent all from loading/being applied. If this happens at startup the proxy will crash (as before). If it happens on reload then the proxy will continue to function using the last know good config as before. I've added a little extra logic to retry reload errors after 5 minutes.

### DIFF
--- a/src/ReverseProxy/Management/ProxyConfigManager.cs
+++ b/src/ReverseProxy/Management/ProxyConfigManager.cs
@@ -8,7 +8,6 @@ using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
@@ -154,10 +153,7 @@ internal sealed class ProxyConfigManager : EndpointDataSource, IDisposable
             for (var i = 0; i < _providers.Length; i++)
             {
                 var config = _providers[i].GetConfig();
-                _configs[i] = new ConfigInstance
-                {
-                    LatestConfig = config,
-                };
+                _configs[i] = new ConfigInstance(config);
                 routes.AddRange(config.Routes ?? Array.Empty<RouteConfig>());
                 clusters.AddRange(config.Clusters ?? Array.Empty<ClusterConfig>());
             }
@@ -217,11 +213,16 @@ internal sealed class ProxyConfigManager : EndpointDataSource, IDisposable
 
     private class ConfigInstance
     {
+        public ConfigInstance(IProxyConfig config)
+        {
+            LatestConfig = config;
+        }
+
         public IProxyConfig LatestConfig { get; set; }
 
         public bool GetConfigFailed { get; set; }
 
-        public IDisposable CallbackCleanup { get; set; }
+        public IDisposable? CallbackCleanup { get; set; }
     }
 
     private static void ReloadConfig(object? state)
@@ -279,8 +280,6 @@ internal sealed class ProxyConfigManager : EndpointDataSource, IDisposable
     // Throws for validation failures
     private async Task<bool> ApplyConfigAsync(IReadOnlyList<RouteConfig> routes, IReadOnlyList<ClusterConfig> clusters)
     {
-        var routesChanged = false;
-
         var (configuredClusters, clusterErrors) = await VerifyClustersAsync(clusters, cancellation: default);
         var (configuredRoutes, routeErrors) = await VerifyRoutesAsync(routes, configuredClusters, cancellation: default);
 
@@ -291,8 +290,7 @@ internal sealed class ProxyConfigManager : EndpointDataSource, IDisposable
 
         // Update clusters first because routes need to reference them.
         UpdateRuntimeClusters(configuredClusters.Values);
-        routesChanged |= UpdateRuntimeRoutes(configuredRoutes);
-
+        var routesChanged = UpdateRuntimeRoutes(configuredRoutes);
         return routesChanged;
     }
 
@@ -498,7 +496,7 @@ internal sealed class ProxyConfigManager : EndpointDataSource, IDisposable
                 Log.ClusterRemoved(_logger, existingCluster.ClusterId);
                 var removed = _clusters.TryRemove(existingCluster.ClusterId, out var _);
                 Debug.Assert(removed);
-                    
+
                 foreach (var listener in _clusterChangeListeners)
                 {
                     listener.OnClusterRemoved(existingCluster);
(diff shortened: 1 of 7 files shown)


samsp-msft:
We should make it clear in the existing sample that the Routes and Cluster properties could be called more than once, and so caching their entries would be a good idea. I could see there being issues where one provider changes, which causes others to be read, and they give different data from before without a change token being fired. Would that be a problem?

src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -48,15 +49,16 @@ internal sealed class ProxyConfigManager : EndpointDataSource, IDisposable
...
     private readonly IClusterDestinationsUpdater _clusterDestinationsUpdater;
-    private IDisposable? _changeSubscription;
 
     private List<Endpoint>? _endpoints;
-    private CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
-    private IChangeToken _changeToken;
+    private CancellationTokenSource _endpointsChangeSource = new();
+    private IChangeToken _endpointsChangeToken;
+
+    private CancellationTokenSource _configChangeSource = new();
```
Kahbazi, line 57: If I followed the code correctly, this `new()` instance that is created here would never be used and it's being replaced in here
https://github.com/microsoft/reverse-proxy/blob/a33d36b54c08606a3f064a18c427410beef60b10/src/ReverseProxy/Management/ProxyConfigManager.cs#L252-L253

src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -200,17 +231,75 @@ private async Task ReloadConfigAsync()
             Log.ErrorApplyingConfig(_logger, ex);
         }
 
-        if (newConfig.ChangeToken.ActiveChangeCallbacks)
+        ListenForConfigChanges();
+    }
+
+    private static void ValidateConfigProperties(IProxyConfig config)
+    {
+        if (config == null)
```
Kahbazi, line 239: In order to follow https://github.com/dotnet/aspnetcore/issues/39323?
```suggestion
        if (config is null)
```

src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -200,17 +231,75 @@ private async Task ReloadConfigAsync()
...
+        {
+            throw new InvalidOperationException($"{nameof(IProxyConfigProvider.GetConfig)} returned a null value.");
+        }
+        if (config.ChangeToken == null)
```
Kahbazi, line 243:
```suggestion
        if (config.ChangeToken is null)
```

src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -200,17 +231,75 @@ private async Task ReloadConfigAsync()
...
+            configState.CallbackCleanup?.Dispose();
+            var token = configState.LatestConfig.ChangeToken;
+            if (token.ActiveChangeCallbacks)
+            {
+                configState.CallbackCleanup = token.RegisterChangeCallback(SignalChange, source);
+            }
+            else
+            {
+                poll = true;
+            }
```
Kahbazi, line 274: Should probably mentioned in the docs that in case of a static config the implementation of the `ChangeToken` must return `true` for `ActiveChangeCallbacks`, so there won't be polling every 5 minutes.

Kahbazi, line 274: Should probably mentioned in the docs that in case of a static config the implementation of the `ChangeToken` must return `true` for `ActiveChangeCallbacks`, so there won't be polling every 5 minutes.

src/ReverseProxy/Management/ProxyConfigManager.cs
Tratcher: You're right, it was initialized mainly for the nullability requirements, it's easier to work with it it's never null.

test/ReverseProxy.Tests/Management/ProxyConfigManagerTests.cs
```diff
@@ -146,6 +167,180 @@ public async Task BuildConfig_OneClusterOneDestinationOneRoute_Works()
...
+        actualDestinations = clusterState.Destinations.Values;
+        destination = Assert.Single(actualDestinations);
+        Assert.Equal("d2", destination.DestinationId);
+        Assert.NotNull(destination.Model);
+        Assert.Equal(TestAddress, destination.Model.Config.Address);
+    }
+
+    [Fact]
+    public async Task BuildConfig_TwoOverlappingConfigs_Works()
+    {
```
greenEkatherine, line 259: Could you please add negative scenarios as well: partial configs from both resources and load failure?

src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -200,17 +231,75 @@ private async Task ReloadConfigAsync()
...
+        }
+
+        // Don't register until we're done hooking everything up to avoid cancellation races.
+        source.Token.Register(ReloadConfig, this);
+
+        if (poll)
         {
-            _changeSubscription = newConfig.ChangeToken.RegisterChangeCallback(ReloadConfig, this);
+            source.CancelAfter(TimeSpan.FromMinutes(5));
```
MihaZupan, line 282: Similar to the below comment about `ObjectDisposedException` during `Cancel()`, the source may already be disposed of by the time we get here.

The order `Register()` and `CancelAfter` should be swapped.

MihaZupan (approved):
Just some comments about race conditions around signaling, otherwise LGTM.

> I've added a little extra logic to retry reload errors after 5 minutes.

If there was a need, we could be more fancy with this in the future (configurable / start with 1 second and have exponential backoff ...).
Given this is already an error condition I assume we don't care (yet)?
I like the approach of using the same source for polling and how little it complicates the logic.
Might be good having a test covering it.

src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -160,31 +174,48 @@ public async Task<EndpointDataSource> InitialLoadAsync()
...
+            routes.AddRange(instance.LatestConfig.Routes ?? Array.Empty<RouteConfig>());
+            clusters.AddRange(instance.LatestConfig.Clusters ?? Array.Empty<ClusterConfig>());
         }
-        catch (Exception ex)
+
+        if (!sourcesChanged)
         {
-            Log.ErrorReloadingConfig(_logger, ex);
-            // If we can't load the config then we can't listen for changes anymore.
+            // Polling mode, only reload if at least one provider changed.
```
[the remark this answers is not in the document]
MihaZupan: And the log message should probably also be updated
https://github.com/microsoft/reverse-proxy/blob/a33d36b54c08606a3f064a18c427410beef60b10/src/ReverseProxy/Management/ProxyConfigManager.cs#L754

src/ReverseProxy/Management/ProxyConfigManager.cs
Tratcher: Fair, but that's not a style we've adopted yet in this repo and I don't want to be inconsistent. Can you file an issue for recommended style changes?

Issue #126, opened 2020-04-28
Design A/B testing strategies


Tratcher:
A/B testing is a common scenario for proxies, letting services incrementally upgrade/experiment using live traffic.

How does a service decide which requests to route to which test group?
- Assuming all requests are stateless you could decide this randomly on a per-request basis.
- For a stateful service you'd want to do a random assignment and then use affinization (https://github.com/microsoft/reverse-proxy/issues/45).
- A service may also decide to statically affinitize based on known groups like tenants (e.g. move small customers before big ones).

Initial theory:
- This is a separate feature from load balancing. Within a load balancing set we assume all servers are equally able to handle a request aside from considerations of health and load. You'd select a test group and then load balance within that group.
- This should take advantage of the routing layer. For something like tenant based affinity that information may already be part of the route definition.
- Alternatively we could put it into the route config (ProxyRoute), allowing it to specify multiple backends and an assignment strategy.
- This should be compatible with retry mechanisms to allow failing over to the other test group.

Tratcher:
@samsp-msft in your surveys have customers explained how they handle this today? And/or how they want to handle it?

karelz:
Triage: Design should be done in 1.0 - customers will ask for it.

Tratcher:
This could also come into play for rolling upgrades: https://github.com/microsoft/reverse-proxy/issues/190#issuecomment-632798853

1. Remove some nodes from pool1
2. Shut down those nodes
3. Upgrade and restart them
4. Add those nodes to pool2
5. Gradually transition a percentage of traffic from pool1 to pool2 and check for errors. (A/B testing)

Repeat until all nodes have been upgraded and moved.

Tratcher:
Note: Backed.PartitioningOptions seems to be related to this scenario, but it's not actually implemented anywhere. Implement or remove this option section. The same for quota options.

Tratcher:
Notes from a partner discussion:
```
Routing - One Catch-All
 - Partitioning function to select cluster
Clusters, each with one destination
	- Next
	- Current
        - Previous
```

Tratcher:
@johnazariah to your question from: https://github.com/microsoft/reverse-proxy/issues/405#issuecomment-689092153

You could still use the headers to decide where requests get assigned, but it would be applied at a stage after routing and would be much more customizable. Affinity is a definite question, the best option is to have a stable selection algorithm like the headers.

johnazariah:
[...]

PR #1538, opened 2022-01-28
Add cluster access API, reassignment

Tratcher:
Contributes to https://github.com/microsoft/reverse-proxy/issues/126#issuecomment-1024494147

This provides the minimal infrastructure necessary to build your own A/B testing or rolling upgrade system.
A) You can now directly look up routes and clusters
B) You can re-assign a request to a new cluster

That should allow a developer to provide their own custom middleware that examines requests, chooses a cluster, and assigns the request accordingly.

Note: Changing the cluster can cause some issues with Session Affinity if the two clusters have different affinity configurations, especially if it's enabled in one and disabled in the other. I've left this as a warning for now, not sure if anyone will need different settings across A/B clusters.

### DIFF
--- a/docs/docfx/articles/ab-testing.md
+++ b/docs/docfx/articles/ab-testing.md
@@ -1,0 +1,51 @@
+# A/B Testing and Rolling Upgrades
+
+## Introduction
+
+A/B testing and rolling upgrades require procedures for dynamically assigning incoming traffic to evaluate changes in the destination application. YARP does not have a built in model for this, but it does expose some infrastructure useful for building such a system. See [issue #126](https://github.com/microsoft/reverse-proxy/issues/126) for additional details about this scenario.
+
+## Example
+
+```
+    public void Configure(IApplicationBuilder app, IProxyStateLookup lookup)
+    {
+        app.UseRouting();
+        app.UseEndpoints(endpoints =>
+        {
+            endpoints.MapReverseProxy(proxyPipeline =>
+            {
+                // Custom cluster selection
+                proxyPipeline.Use((context, next) =>
+                {
+                    if (lookup.TryGetCluster(ChooseCluster(context), out var cluster))
+                    {
+                        context.ReassignProxyRequest(cluster);
+                    }
+
+                    return next();
+                });
+                proxyPipeline.UseSessionAffinity();
+                proxyPipeline.UseLoadBalancing();
+            });
+        });
+    }
+
+    private string ChooseCluster(HttpContext context)
+    {
+        // Decide which cluster to use. This could be random, weighted, based on headers, etc.
+        return Random.Shared.Next(2) == 1 ? "cluster1" : "cluster2";
+    }
+```
+
+## Usage
+
+This scenario makes use of two APIs, [IProxyStateLookup](xref:Yarp.ReverseProxy.IProxyStateLookup) and [ReassignProxyRequest](xref:Microsoft.AspNetCore.Http.HttpContextFeaturesExtensions.ReassignProxyRequest).
+
+`IProxyStateLookup` is a service available in the Dependency Injection container that can be used to look up or enumerate the current routes and clusters. Note this data may change if the configuration changes. An A/B orchestration algorithm can examine the request, decide which cluster to send it to, and then retrieve that cluster from `IProxyStateLookup.TryGetCluster`.
+
+Once the cluster is selected, `ReassignProxyRequest` can be called to assign the request to that cluster. This updates the [IReverseProxyFeature](xref:Yarp.ReverseProxy.Model.IReverseProxyFeature) with the new cluster and destination information needed for the rest of the proxy middleware pipeline to handle the request.
+
+## Session affinity
+
+Note that session affinity functionality is split between middleware, which reads it settings from the current cluster, and transforms, which are part of the original route. Clusters used for A/B testing should use the same session affinity configuration to avoid conflicts.
+
--- a/docs/docfx/articles/ab-testing.md
+++ b/docs/docfx/articles/ab-testing.md
@@ -2,7 +2,7 @@
 
 ## Introduction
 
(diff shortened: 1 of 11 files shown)


src/ReverseProxy/Management/IProxyStateLookup.cs
```diff
@@ -0,0 +1,19 @@
...
+
+public interface IProxyStateLookup
+{
+    bool TryGetRoute(string id, [NotNullWhen(true)] out RouteModel? route);
+
+    IEnumerable<RouteModel> GetRoutes();
+
+    bool TryGetCluster(string id, [NotNullWhen(true)] out ClusterState? cluster);
+
+    IEnumerable<ClusterState> GetClusters();
```
Tratcher, line 18: Yes, it's a little weird that the types here are asymmetric, but these are the types that are actually useful to people. RouteState isn't public and is only used to track endpoint routing items.

src/ReverseProxy/Management/IProxyStateLookup.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Yarp.ReverseProxy.Model;
+
+namespace Yarp.ReverseProxy;
+
+public interface IProxyStateLookup
+{
+    bool TryGetRoute(string id, [NotNullWhen(true)] out RouteModel? route);
```
samsp-msft, line 12: can this just return null or the route? And then doesn't need the try.

src/ReverseProxy/Management/IProxyStateLookup.cs
samsp-msft: That's the right choice then.

src/ReverseProxy/Model/HttpContextFeaturesExtensions.cs
```diff
@@ -40,4 +40,20 @@ public static IReverseProxyFeature GetReverseProxyFeature(this HttpContext conte
     {
         return context.Features.Get<IForwarderErrorFeature>();
     }
+
+    // Compare to ProxyPipelineInitializerMiddleware
+    public static void ReassignProxyRequest(this HttpContext context, ClusterState cluster)
```
samsp-msft, line 45: Is this what the user will use to redirect as part of middleware?

samsp-msft (approved):
LGTM

src/ReverseProxy/Management/IProxyStateLookup.cs
Tratcher: If it returned null then you'd have to check for null before you could use it. The try is a bit cleaner and less error prone.
```
var route = lookup.GetRoute(id);
if (route != null)
{
  ///
}
```
```
if (lookup.TryGetRoute(id, out var route))
{
  ///
}
```

src/ReverseProxy/Model/HttpContextFeaturesExtensions.cs
Tratcher: Right, see the sample.

docs/docfx/articles/ab-testing.md
```diff
@@ -0,0 +1,51 @@
...
+    }
+```
+
+## Usage
+
+This scenario makes use of two APIs, [IProxyStateLookup](xref:Yarp.ReverseProxy.IProxyStateLookup) and [ReassignProxyRequest](xref:Microsoft.AspNetCore.Http.HttpContextFeaturesExtensions.ReassignProxyRequest).
+
+`IProxyStateLookup` is a service available in the Dependency Injection container that can be used to look up or enumerate the current routes and clusters. Note this data may change if the configuration changes. An A/B orchestration algorithm can examine the request, decide which cluster to send it to, and then retrieve that cluster from `IProxyStateLookup.TryGetCluster`.
+
+Once the cluster is selected, `ReassignProxyRequest` can be called to assign the request to that cluster. This updates the [IReverseProxyFeature](xref:Yarp.ReverseProxy.Model.IReverseProxyFeature) with the new cluster and destination information needed for the rest of the proxy middleware pipeline to handle the request.
```
Kahbazi, line 46: It's good to mention that `ReassignProxyRequest` must be called in proxyPipeline and not in a middleware before `UseEndpoints`.

docs/docfx/articles/toc.yml
```diff
@@ -40,3 +40,5 @@
   href: packages-refs.md
 - name: Diagnosing proxy issues
   href: diagnosing-yarp-issues.md
+- name: AB Testing
```
Kahbazi, line 43: Does it support `/ ` in the name?
```suggestion
- name: A/B Testing
```

docs/docfx/articles/ab-testing.md
```diff
@@ -0,0 +1,51 @@
+# A/B Testing and Rolling Upgrades
+
+## Introduction
+
+A/B testing and rolling upgrades require procedures for dynamically assigning incoming traffic to evaluate changes in the destination application. YARP does not have a built in model for this, but it does expose some infrastructure useful for building such a system. See [issue #126](https://github.com/microsoft/reverse-proxy/issues/126) for additional details about this scenario.
```
Kahbazi, line 5:
```suggestion
A/B testing and rolling upgrades require procedures for dynamically assigning incoming traffic to evaluate changes in the destination application. YARP does not have a built-in model for this, but it does expose some infrastructure useful for building such a system. See [issue #126](https://github.com/microsoft/reverse-proxy/issues/126) for additional details about this scenario.
```

testassets/ReverseProxy.Code/Startup.cs
```diff
@@ -112,6 +121,11 @@ public void Configure(IApplicationBuilder app)
                 // Custom endpoint selection
```
MihaZupan, line 121:
```suggestion
                // Custom endpoint and cluster selection
```

Using `ReassignProxyRequest` makes more sense in the context of the `ab-testing.md` sample.
Perhaps add a comment on when one would want to do this / link to the a/b doc?

MihaZupan (approved):

Tratcher:
I was really hoping one of you would call me out on the terrible name of `IProxyStateLookup` 😆. Still hoping to find a better name.

PR #1556, opened 2022-02-07
Add http.sys delegation support

### LINKED ISSUE #1503: Expose Http.Sys fast delegation proxying
Http.Sys server exposes a feature that lets you examine a request and then quickly 'proxy' it to another service on the same machine by transferring it to a different queue.

We could expose this as optional middleware that reads cluster/destination data and forwards requests using this mechanic.

Pros:
- Very fast/efficient
- Interops with IIS

Cons:
- You can't modify the request or observe the response.

Supported on .NET 6 and Win11.

Example:
https://github.com/dotnet/aspnetcore/commit/f72785fbf776cef17bf2500ad5a03e5cd09ac5a9#diff-66e5582beb1ea1594e35e4e0457f4676e3453bb34efecd2a9e69005ff17f06bbR55-R61

NGloreous:
This PR fixes #1503 and adds support for delegating requests to other processes via the http.sys request delegation feature. 

There are two main components added:
- HttpSysDelegator- Delegates request and manages DelegationRule lifetime based on updates to cluster config. Destinations opt-in to delegation by adding a 'HttpSysQueueName' metadata entry defining the name of the http.sys queue to attach to.
- HttpSysDelegatorMiddleware - Determines if the selected destination should use queue delegation or not. If so, invokes HttpSysDelegator to delegate the request. 

Also included are UTs, a functional regression test, docs, and samples.


src/ReverseProxy/Delegation/HttpSysDelegationMiddleware.cs
```diff
@@ -0,0 +1,138 @@
...
+                {
+                    throw new InvalidOperationException(
+                        "Current request can't be delegated. Either the request body has started to be read or the response has started to be sent.");
+                }
+
+                if (!_delegationRuleManager.TryGetDelegationRule(destination, out var delegationRule))
+                {
+                    Log.DelegationRuleNotFound(_logger, cluster.Config.ClusterId, destination.DestinationId);
+                    context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
+                    context.Features.Set<IForwarderErrorFeature>(new ForwarderErrorFeature(ForwarderError.HttpSysDelegationRuleNotFound, ex: null));
```
NGloreous, line 72: Thoughts on using IForwarderErrorFeature here? It's a bit odd since this isn't the Forwarder but it's related.

The up sides of re-using it is that the existing passive health check logic just works. Although we may want add custom logic over there at some point since typically if things go bad (because the receiving processes crashed/re-started), the only way to fix is to re-create the rule.

src/ReverseProxy/Delegation/HttpSysDelegationRuleManager.cs
```diff
@@ -0,0 +1,221 @@
...
+                var ruleKey = new RuleKey(incomingDestination.DestinationId, queueName);
+
+                var added = currentRules.Contains(ruleKey);
+                if (!added)
+                {
+                    try
+                    {
+                        var rule = _delegationFeature.CreateDelegationRule(queueName, incomingDestination.Model.Config.Address);
+                        added = _rules.TryAdd(ruleKey, rule);
+                        if (!added)
```
NGloreous, line 113: Right now, I'm only allowing a destination queue to be used once in all YARP config. I could allow the same queue to be used by multiple destinations, but I couldn't think of a reason why that would be needed. Thoughts? It's easy enough to change later if needed.

test/ReverseProxy.FunctionalTests/Common/HttpSysFactAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public partial class HttpSysDelegationTests
+{
+    public class HttpSysFactAttribute : FactAttribute
+    {
+        public HttpSysFactAttribute()
+        {
+            if (!OperatingSystem.IsWindows())
+            {
+                Skip = "Http.sys tests are only supported on Windows";
```
NGloreous, line 15: What Windows OS are YARP functional tests run? If any older versions are used, I'll need to also check if the delegation feature is available.

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -149,4 +152,21 @@ public static IReverseProxyBuilder ConfigureHttpClient(this IReverseProxyBuilder
...
+#if NET6_0_OR_GREATER
+    /// <summary>
+    /// Adds a singleton <see cref="Yarp.ReverseProxy.Delegation.IHttpSysDelegationRuleManager"/> implementation to manage Http.sys delegation rules created from cluster config.
+    /// </summary>
+    public static IReverseProxyBuilder AddHttpSysDelegation(this IReverseProxyBuilder builder)
+    {
+        builder.Services.AddSingleton(p =>
+            p.GetRequiredService<IServer>().Features?.Get<IServerDelegationFeature>()
+            ?? throw new InvalidOperationException($"{typeof(IHttpSysRequestDelegationFeature).FullName} is missing. Http.sys delegation is only supported when using the Http.sys server"));
+        builder.Services.AddSingleton<Yarp.ReverseProxy.Delegation.HttpSysDelegationRuleManager>();
```
Tratcher, line 165: Trim namespaces:
```suggestion
        builder.Services.AddSingleton<HttpSysDelegationRuleManager>();
```

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -13,6 +13,9 @@
 using Yarp.ReverseProxy.Routing;
 using Yarp.ReverseProxy.Transforms.Builder;
 using Yarp.ReverseProxy.Utilities;
+using Yarp.ReverseProxy.Model;
+using Microsoft.AspNetCore.Hosting.Server;
+using Microsoft.AspNetCore.Server.HttpSys;
```
Tratcher, line 18: Sort

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -149,4 +152,21 @@ public static IReverseProxyBuilder ConfigureHttpClient(this IReverseProxyBuilder
         });
         return builder;
     }
+
+#if NET6_0_OR_GREATER
+    /// <summary>
+    /// Adds a singleton <see cref="Yarp.ReverseProxy.Delegation.IHttpSysDelegationRuleManager"/> implementation to manage Http.sys delegation rules created from cluster config.
+    /// </summary>
+    public static IReverseProxyBuilder AddHttpSysDelegation(this IReverseProxyBuilder builder)
```
Tratcher, line 160: Can we setup the services by default and have them no-op if you don't add the middleware? If we don't need parameters then it's easier to use if we can avoid the extra API.

test/ReverseProxy.FunctionalTests/Common/Helpers.cs
```diff
@@ -15,4 +18,25 @@ public static string GetAddress(this IHost server)
...
+        int port;
+        do
+        {
+            port = ThreadStaticRandom.Instance.Next(5000, 65535);
+        } while (openPorts.Contains(port));
+
+        return port;
+    }
+
+    private static HashSet<int> GetOpenPorts()
```
Tratcher, line 35: Guessing ports isn't very reliable. The best approach we've found is to actually start the server on 0 and let it figure it out.
https://github.com/dotnet/aspnetcore/blob/7c554b7941fc8002141b6a91e2f0c5dc4905213b/src/Servers/HttpSys/test/FunctionalTests/Utilities.cs#L85

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -149,4 +152,21 @@ public static IReverseProxyBuilder ConfigureHttpClient(this IReverseProxyBuilder
...
+    {
+        builder.Services.AddSingleton(p =>
+            p.GetRequiredService<IServer>().Features?.Get<IServerDelegationFeature>()
```
Tratcher, line 163:
```suggestion
        // https://github.com/dotnet/aspnetcore/issues/40043
        builder.Services.TryAddSingleton(p =>
            p.GetRequiredService<IServer>().Features?.Get<IServerDelegationFeature>()
```

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -149,4 +152,21 @@ public static IReverseProxyBuilder ConfigureHttpClient(this IReverseProxyBuilder
...
+            ?? throw new InvalidOperationException($"{typeof(IHttpSysRequestDelegationFeature).FullName} is missing. Http.sys delegation is only supported when using the Http.sys server"));
+        builder.Services.AddSingleton<Yarp.ReverseProxy.Delegation.HttpSysDelegationRuleManager>();
+        builder.Services.AddSingleton<Yarp.ReverseProxy.Delegation.IHttpSysDelegationRuleManager>(p => p.GetRequiredService<Yarp.ReverseProxy.Delegation.HttpSysDelegationRuleManager>());
```
Tratcher, line 166:
```suggestion
        builder.Services.TryAddSingleton<Yarp.ReverseProxy.Delegation.IHttpSysDelegationRuleManager>(p => p.GetRequiredService<Yarp.ReverseProxy.Delegation.HttpSysDelegationRuleManager>());
```

test/ReverseProxy.FunctionalTests/HttpSysDelegationTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+public partial class HttpSysDelegationTests
+{
+    [HttpSysFact]
+    public async Task RequestDelegated()
+    {
+        IForwarderErrorFeature proxyError = null;
+        Exception unhandledError = null;
+        var expectedResone = "Hello World!";
+        var queueName = nameof(HttpSysDelegationTests) + ThreadStaticRandom.Instance.Next().ToString("x8");
```
Tratcher, line 27:
```suggestion
        var queueName = nameof(HttpSysDelegationTests) + Random.Shared.Next().ToString("x8");
```

test/ReverseProxy.FunctionalTests/HttpSysDelegationTests.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace Yarp.ReverseProxy;
+
+public partial class HttpSysDelegationTests
+{
+    [HttpSysFact]
+    public async Task RequestDelegated()
+    {
+        IForwarderErrorFeature proxyError = null;
+        Exception unhandledError = null;
+        var expectedResone = "Hello World!";
```
Tratcher, line 26:
```suggestion
        var expectedResponse = "Hello World!";
```

src/ReverseProxy/Delegation/HttpSysDelegationMiddleware.cs
Tratcher: The new enum values are out of place. Could you use the existing value Request or NoAvailableDestinations and add an InvalidOperationException with details?

It's interesting you're trying to make it work with passive health checks. How much sense does that make given it can't normally view the response? The only health you can measure is if the delegation rule is present and works.

Wouldn't the rule be present in all cases except race conditions during config changes?

Tratcher (commented):
Partial review.

src/ReverseProxy/Delegation/HttpSysDelegationMiddleware.cs
NGloreous: Currently, the only case where the rule wouldn't exist is if creating the rule failed for some reason or because the config wasn't valid (i.e. the queue was duplicated). If we eventually add some of the mitigations I wanted to for dealing with the receiving process re-starting, then I was thinking the rule would be removed while we are trying to re-create it in the background. So re-using NoAvailableDestinations feels a bit misleading, but I guess is accurate enough. I wouldn't want to add a InvalidOperationException since it may not always be invalid.

I think re-using Request could make sense. The exception would be an HttpSysException so consumers who care could filter on that.

samples/ReverseProxy.HttpSysDelegation.Sample/README.md
```diff
@@ -0,0 +1,57 @@
+# Http.sys Delegation Sample
```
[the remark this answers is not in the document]
NGloreous: Is copy the only way to have it in both?

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
NGloreous: VS failed me :/

src/ReverseProxy/Delegation/AppBuilderDelegationExtensions.cs
```diff
@@ -0,0 +1,26 @@
+#if NET6_0_OR_GREATER
```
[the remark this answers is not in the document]
NGloreous: It seems samples don't need the license?

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
NGloreous: It may be a bit odd with the need for the IServerDelegationFeature. I could change my work around to return null if the feature doesn't exist and then in HttpSysDelegationRuleManager no bother doing anything if I get a null IServerDelegationFeature.  

But if you ever want to remove the work around, then when you try to Build the app you'd hit an exception. There would also be the overhead of calling the IClusterChangeListener methods even if they no-op.

Thoughts?

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
NGloreous: Hmm, yes, this allows others to register this before us but in that case isn't it odd that we would also have HttpSysDelegationRuleManager hanging around processing cluster changes but not doing anything useful? Is there any way to work around that here with DI magic? If not, maybe a reason to keep AddHttpSysDelegation.

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
NGloreous: Sure, I did it this way to avoid having more \#if logic in the usings but I'm fine with either way.

test/ReverseProxy.FunctionalTests/Common/Helpers.cs
NGloreous: Nice, didn't know that worked 😄

test/ReverseProxy.FunctionalTests/HttpSysDelegationTests.cs
NGloreous: I was bound to have one of these 😄

src/ReverseProxy/Delegation/HttpSysDelegationMiddleware.cs
NGloreous: Unless I'm misunderstanding something, it looks the passive health check doesn't currently look at the response and everything is calculated based on pass/fail of proxying the request. This should work for delegation but not sure if anyone would use it like this. For us, we only have one destination and if it's not working, then requests will just fail. That is why I'd like to add some support to try and "fix" delegation. The only thing in the proxy's control is to try and re-create the delegation rule.

test/ReverseProxy.FunctionalTests/Common/HttpSysFactAttribute.cs
NGloreous: Thoughts on a way to do this check other than copying the logic from here: https://github.com/dotnet/aspnetcore/blob/main/src/Servers/HttpSys/src/NativeInterop/HttpApi.cs? I guess I could start up a server that that seems heavy weight or reflection :(

src/ReverseProxy/Delegation/HttpSysDelegationRuleManager.cs
Tratcher: Allow it if it's easy. People may duplicate the same destination in different clusters. This would work today with normal http proxying.

test/ReverseProxy.FunctionalTests/Common/HttpSysFactAttribute.cs
Tratcher: An OS version check is usually enough for tests.

src/ReverseProxy/Delegation/AppBuilderDelegationExtensions.cs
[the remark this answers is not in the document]
Tratcher: They don't matter, but it doesn't hurt.

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
NGloreous: Is there a way to run custom logic at service creation time to conditionally add a service?

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
Tratcher: > It may be a bit odd with the need for the IServerDelegationFeature. I could change my work around to return null if the feature doesn't exist and then in HttpSysDelegationRuleManager no bother doing anything if I get a null IServerDelegationFeature.

Try it and see if it works.

> But if you ever want to remove the work around, then when you try to Build the app you'd hit an exception. There would also be the overhead of calling the IClusterChangeListener methods even if they no-op.

This is ok because it only happens on config changes which are infrequent. Active health checks work the same way.

samples/ReverseProxy.HttpSysDelegation.Sample/README.md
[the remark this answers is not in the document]
Tratcher: Yes, though I don't expect the content to be exactly the same. E.g. the doc would inline as much of the sample as needed.

samples/ReverseProxy.HttpSysDelegation.Sample/README.md
```diff
@@ -0,0 +1,57 @@
...
+This sample shows how to use YARP to delegate requests to other Http.sys request queues instead of or in addition to proxying requests. Using Http.sys delegation requires hosting YARP on [ASP.NET Core's Http.sys server](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/httpsys) and requests can only be delegated to other processes which use Http.sys for request processing (e.g. ASP.NET Core using Http.sys server or IIS).
```
Tratcher, line 2: Leave language/culture out of the links so auto-detection works.
```suggestion
This sample shows how to use YARP to delegate requests to other Http.sys request queues instead of or in addition to proxying requests. Using Http.sys delegation requires hosting YARP on [ASP.NET Core's Http.sys server](https://docs.microsoft.com/aspnet/core/fundamentals/servers/httpsys) and requests can only be delegated to other processes which use Http.sys for request processing (e.g. ASP.NET Core using Http.sys server or IIS).
```

Tratcher: This feature requires ASP.NET Core 6

samples/ReverseProxy.HttpSysDelegation.Sample/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+**Note: delegation only works for ASP.NET Core 6+ running on new versions of Windows**
+
+## Http.sys Delegation
+Http.sys delegation is a kernel level feature added into newer versions of Windows which allows a request to be transferred from the receiving process's queue to another process's queue with very little overhead or added latency. For this delegation to work, the receiving process is only allowed to read the request headers. If the body has started to be read or a response has started, trying to delegate the request will fail. Internally, YARP leverage's ASP.NET Core's [IHttpSysRequestDelegationFeature](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.httpsys.ihttpsysrequestdelegationfeature).
```
Tratcher, line 7: Which OS versions are needed?

Tratcher: The response will not be visible to the proxy after delegation, which limits the functionality of the session affinity and passive health checks components.

src/ReverseProxy/Delegation/AppBuilderDelegationExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+/// </summary>
+public static class AppBuilderDelegationExtensions
+{
+    /// <summary>
+    /// Adds middleware to check if the selected destination should use Http.sys delegation. If so the request is delegated to the destination queue.
+    /// This should be placed after load balancing and passive health checks.
+    /// </summary>
+    /// <remarks>
+    /// This middleware only works with the ASP.NET Core Http.sys server implementation.
+    /// A <see cref="IHttpSysDelegationRuleManager"/> must be registered with DI. This can be done by calling <see cref="ReverseProxyServiceCollectionExtensions.AddHttpSysDelegation(IReverseProxyBuilder)"/>.
```
Tratcher, line 21: IHttpSysDelegationRuleManager is an implementation detail.
```suggestion
    /// This requires calling <see cref="ReverseProxyServiceCollectionExtensions.AddHttpSysDelegation(IReverseProxyBuilder)"/> to register the needed services in DI.
```

samples/ReverseProxy.HttpSysDelegation.Sample/ReverseProxy/appsettings.json
```diff
@@ -0,0 +1,45 @@
...
+        }
+      }
+    },
+    "Clusters": {
+      "delegatecluster": {
+        "Destinations": {
+          "SampleHttpSysServer": {
+            "Address": "http://localhost:5600/",
+            "Metadata": {
+              "HttpSysQueueName": "SampleHttpSysServerQueue"
```
Tratcher, line 31: `HttpSysDelegation` is used elsewhere to identify the same feature.
```suggestion
              "HttpSysDelegationQueue": "SampleHttpSysServerQueue"
```

src/ReverseProxy/Delegation/HttpSysDelegationMiddleware.cs
Tratcher: The default passive health check looks at the proxy error, but the design is intended to allow checking status codes and other response fields as well.

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
Tratcher: Not that I know of.

src/ReverseProxy/Delegation/IHttpSysDelegationRuleManager.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.AspNetCore.Server.HttpSys;
+using Yarp.ReverseProxy.Model;
+
+namespace Yarp.ReverseProxy.Delegation;
+
+/// <summary>
+/// Manages Http.sys delegation rules for the configured clusters/destination pairs.
+/// </summary>
+public interface IHttpSysDelegationRuleManager
```
Tratcher, line 14: Can you give an example of when someone would customize/replace this service? It seems like the interface isn't required to be public (or even needed) at the moment.

src/ReverseProxy/Delegation/AppBuilderDelegationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.AspNetCore.Builder;
+
+/// <summary>
+/// Extensions for adding delegation middleware to the pipeline.
+/// </summary>
+public static class AppBuilderDelegationExtensions
+{
+    /// <summary>
+    /// Adds middleware to check if the selected destination should use Http.sys delegation. If so the request is delegated to the destination queue.
+    /// This should be placed after load balancing and passive health checks.
```
Tratcher, line 20:
```suggestion
    /// Adds middleware to check if the selected destination should use Http.sys delegation. If so, the request is delegated to the destination queue
    /// instead of being proxied over HTTP. This should be placed after load balancing and passive health checks.
```

src/ReverseProxy/Management/IReverseProxyBuilderExtensions.cs
```diff
@@ -112,4 +118,24 @@ public static IReverseProxyBuilder AddPassiveHealthCheck(this IReverseProxyBuild
...
+        {
+            // IServerDelegationFeature isn't added to DI  https://github.com/dotnet/aspnetcore/issues/40043
+            // IServerDelegationFeature may not be set if not http.sys server or the OS doesn't support delegation
+            var delegationFeature = p.GetRequiredService<IServer>().Features?.Get<IServerDelegationFeature>();
+            return new HttpSysDelegationRuleManager(
+                delegationFeature,
+                p.GetRequiredService<ILogger<HttpSysDelegationRuleManager>>());
+        });
+
+        builder.Services.AddSingleton<IHttpSysDelegationRuleManager>(p => p.GetRequiredService<HttpSysDelegationRuleManager>());
```
Tratcher, line 135: Avoid ordering restrictions if this is intended to be replaced.
```suggestion
        builder.Services.TryAddSingleton<IHttpSysDelegationRuleManager>(p => p.GetRequiredService<HttpSysDelegationRuleManager>());
```

src/ReverseProxy/Delegation/IHttpSysDelegationRuleManager.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Attempts to get the <see cref="DelegationRule"/> for the given cluster/destination pair.
+    /// </summary>
+    /// <param name="destination">The destination to get the <see cref="DelegationRule"/> for.</param>
+    /// <param name="delegationRule">
+    /// <paramref name="delegationRule"/> contains the <see cref="DelegationRule"/> for the given cluster/destination pair if found or null if not found.
+    /// </param>
+    /// <returns>true if the <see cref="DelegationRule"/> was found; otherwise, false.</returns>
+    bool TryGetDelegationRule(DestinationState destination, [MaybeNullWhen(false)] out DelegationRule delegationRule);
```
Tratcher, line 24:
```suggestion
    bool TryGetDelegationRule(DestinationState destination, [NotNullWhen(true)] out DelegationRule delegationRule);
```

src/ReverseProxy/Delegation/HttpSysDelegationMiddleware.cs
```diff
@@ -0,0 +1,141 @@
...
+                    Log.DelegationRuleNotFound(_logger, cluster.Config.ClusterId, destination.DestinationId);
+                    context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
+                    context.Features.Set<IForwarderErrorFeature>(new ForwarderErrorFeature(ForwarderError.NoAvailableDestinations, ex: null));
+                    return Task.CompletedTask;
+                }
+
+                try
+                {
+                    Log.DelegatingRequest(_logger, delegationRule?.QueueName);
+                    delegationFeature.DelegateRequest(delegationRule!);
```
Tratcher, line 82: Nullability checks shouldn't be needed here, TryGetDelegationRule shouldn't return null.
```suggestion
                    Log.DelegatingRequest(_logger, delegationRule.QueueName);
                    delegationFeature.DelegateRequest(delegationRule);
```

src/ReverseProxy/Delegation/HttpSysDelegationRuleManager.cs
```diff
@@ -0,0 +1,226 @@
...
+                }
+            }
+        }
+    }
+
+    private void RemoveRule(RuleKey ruleKey)
+    {
+        if (_rules.TryRemove(ruleKey, out var rule))
+        {
+            rule?.Dispose();
```
Tratcher, line 176: Immediate removal &/ disposal of the rule can affect requests that are in flight in the pipeline. How do we avoid disrupting active requests? We've used ConditionalWeakTable to maintain state like this elsewhere.
https://github.com/microsoft/reverse-proxy/blob/701781241c60eb796794dc24ecf110a5c222c2b6/src/ReverseProxy/LoadBalancing/RoundRobinLoadBalancingPolicy.cs#L14

src/ReverseProxy/Delegation/HttpSysDelegationMiddleware.cs
NGloreous: You are correct except for UTs they are needed because I can't create a DelegationRule since the constructor in internal 😞 Any thoughts on a better way to work around this?

Tratcher: I thought NotNullWhen would have fixed this.

src/ReverseProxy/Management/IReverseProxyBuilderExtensions.cs
NGloreous: Not needed anymore.

src/ReverseProxy/Delegation/HttpSysDelegationRuleManager.cs
NGloreous: I was thinking about this some more. #2 could be accomplished without a timer by also using a weak reference to the DelegationRule from the DelegationQueue class. I'm still a bit concerned about leaving these around for so long though.

NGloreous: Per our chat, I'm going with \#2 using weak references and we'll let GC take care of the cleanup.

src/ReverseProxy/Delegation/HttpSysDelegator.cs
```diff
@@ -0,0 +1,318 @@
...
+                if (!_queuesPerDestination.TryGetValue(destination, out var queue))
+                {
+                    var queueKey = new QueueKey(queueName, urlPrefix);
+                    if (!_queues.TryGetValue(queueKey, out var queueWeakRef) || !queueWeakRef.TryGetTarget(out queue))
+                    {
+                        // Either the queue hasn't been created or it has been cleaned up.
+                        // Create a new one, and try to add it if someone else didn't beat us to it.
+                        queue = new DelegationQueue(queueName, urlPrefix);
+                        queueWeakRef = new WeakReference<DelegationQueue>(queue);
+                        _queues.AddOrUpdate(
```
NGloreous, line 131: I need to use the value from here in case someone else wins :)

src/ReverseProxy/Delegation/HttpSysDelegator.cs
```diff
@@ -0,0 +1,318 @@
...
+        {
+            context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
+            context.Features.Set<IForwarderErrorFeature>(new ForwarderErrorFeature(ForwarderError.NoAvailableDestinations, queue?.InitializationException));
+            return;
+        }
+
+        try
+        {
+            Log.DelegatingRequest(_logger, destination);
+            delegationFeature.DelegateRequest(queue.DelegationRule);
```
MihaZupan, line 74: I take it this just does the hand-off and doesn't block to wait on its completion?

src/ReverseProxy/Delegation/HttpSysDelegator.cs
```diff
@@ -0,0 +1,318 @@
...
+        // Using weak references means we ensure the queue exist as long as the referencing destinations exist.
+        // Once all the destinations are gone, GC will eventually finalize the underlying SafeHandle to the http.sys
+        // queue, which will clean up references in http.sys, allowing us to re-create it again later if needed.
+        foreach (var destination in cluster.Destinations.Select(kvp => kvp.Value))
+        {
+            var queueName = destination.GetHttpSysDelegationQueue();
+            var urlPrefix = destination.Model?.Config?.Address;
+            if (queueName != null && urlPrefix != null)
+            {
+                if (!_queuesPerDestination.TryGetValue(destination, out var queue))
```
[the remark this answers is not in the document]
NGloreous: I'm thinking yes so I'll handle this case.

src/ReverseProxy/Delegation/HttpSysDelegator.cs
```diff
@@ -0,0 +1,336 @@
...
+            // Opportunistically retry initialization if it failed previously.
+            // This helps when the target queue wasn't yet created because
+            // the target process hadn't yet started up.
+            queue.TryInitialize(_delegationFeature);
+        }
+
+        if (queue == null || !queue.IsInitialized)
+        {
+            context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
+            context.Features.Set<IForwarderErrorFeature>(new ForwarderErrorFeature(ForwarderError.NoAvailableDestinations, queue?.InitializationException));
```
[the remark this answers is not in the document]
Tratcher: You're missing a log here for the case when the request is rejected because TryInitialize failed.

Should this if block be consolidated with the one above?

src/ReverseProxy/Delegation/HttpSysDelegator.cs
```diff
@@ -0,0 +1,336 @@
...
+                            new Exception("Delegation queue is null after adding a new one. This shouldn't happen."));
+                    }
+                }
+            }
+        }
+
+        RemoveDeadQueueReferences();
+    }
+
+    private void RemoveDeadQueueReferences()
```
Tratcher, line 177: Hmm, this is called per cluster which may be a lot of redundant scanning. It's probably OK for now but it might not scale well.

src/ReverseProxy/Delegation/HttpSysDelegator.cs
```diff
@@ -0,0 +1,336 @@
...
+        //    This allows us to find the already created queue if more than one destination points to the same queue.
+        //
+        // Using weak references means we ensure the queue exist as long as the referencing destinations exist.
+        // Once all the destinations are gone, GC will eventually finalize the underlying SafeHandle to the http.sys
+        // queue, which will clean up references in http.sys, allowing us to re-create it again later if needed.
+        foreach (var destination in cluster.Destinations.Select(kvp => kvp.Value))
+        {
+            var queueName = destination.GetHttpSysDelegationQueue();
+            var urlPrefix = destination.Model?.Config?.Address;
+            if (queueName != null && urlPrefix != null)
```
Tratcher, line 120: How do you handle the scenario where the delegation que name is removed from the destination metadata? It looks like that would leak. Should there be an else `_queuesPerDestination.TryRemove(destination)`? Or does that put us right back into the original race condition?

Tratcher (approved):

src/ReverseProxy/Delegation/HttpSysDelegator.cs
NGloreous: Yes, open to suggestions how to better handle this. Maybe only do for Change/Remove instead of Add would help a bit but still not great.

src/ReverseProxy/Delegation/HttpSysDelegator.cs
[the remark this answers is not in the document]
NGloreous: Good call on the lock. As for logging, I didn't see the need to log the initialization error for every request since we also report the error on the context and if initialization consistently fails, we'd be logging quite a lot.

src/ReverseProxy/Delegation/HttpSysDelegator.cs
NGloreous: So, the case where a destination goes from being used for delegation to being used to proxy? I don't handle that here but yes I could just call Remove for every destination that doesn't have the queue metadata.

src/ReverseProxy/Management/IReverseProxyBuilderExtensions.cs
```diff
@@ -2,18 +2,24 @@
...
 using Yarp.ReverseProxy.LoadBalancing;
 using Yarp.ReverseProxy.Model;
-using Yarp.ReverseProxy.Forwarder;
 using Yarp.ReverseProxy.Routing;
 using Yarp.ReverseProxy.SessionAffinity;
 using Yarp.ReverseProxy.Transforms;
 using Yarp.ReverseProxy.Utilities;
+#if NET6_0_OR_GREATER
+using Yarp.ReverseProxy.Delegation;
+#endif
```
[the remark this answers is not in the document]
NGloreous: I wish VS would handle this but it seems to fail managing using when you add a preprocessor directive 😢 better yet is if it would just ignore invalid usings as this looks ugly either way :/

src/ReverseProxy/Delegation/HttpSysDelegator.cs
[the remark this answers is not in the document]
Tratcher: Any refused request should have at least one info level log to that effect.

PR #1560, opened 2022-02-08
Remove Service Fabric integration code & docs

MihaZupan:
🔥

There are no instances of "fabric" in the repo after this ... unless we have typos hidden somewhere :)


docs/docfx/articles/service-fabric-int.md
```diff
@@ -1,216 +0,0 @@
-# Service Fabric Integration
-
-Note: This functionality has been replaced by https://github.com/microsoft/service-fabric-yarp
```
Tratcher, line 3: Can we keep the doc and toc but remove everything after this line? We can fill this back in later with the new instructions.

Tratcher (approved):

PR #1564, opened 2022-02-11
Improvement some conditional statement

mokarchi:


MihaZupan (commented):
Thank you for looking to contribute to the project, but can you elaborate on what the motivation behind this change is? I'm not sure it helps with readability.

mokarchi:
I think so yes there is no behind motivation because if/else and switch is a code smell and specific part for refactoring code with strategy pattern or chain of responsibilty pattern but also use switch is simple way for extend source code and temporary treatment

Tratcher:
FYI we don't normally accept style-only PRs as it is very subjective and doesn't improve the functionality of the product.

mokarchi:
@Tratcher Should I Close PR?

Issue #1552, opened 2022-02-06
Consider enforcing is null instead of == null


Kahbazi:
In order to follow the style of ASP.NET Core repo, `is null` should be used instead of `== null`.

I'll send a PR, if this design is approved.

samsp-msft:
Not high priority, but seems like a good thing to do. It should probably also include turning on an analyzer so we get build time feedback if it gets broken in future.

MihaZupan:
We should wait at least until #1560 is merged to avoid conflicts

PR #1565, opened 2022-02-11
Use is null instead of == null

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/1552

### DIFF
--- a/docs/docfx/articles/dests-health-checks.md
+++ b/docs/docfx/articles/dests-health-checks.md
@@ -139,7 +139,7 @@ public class FirstUnsuccessfulResponseHealthPolicy : IActiveHealthCheckPolicy
         for (var i = 0; i < probingResults.Count; i++)
         {
             var response = probingResults[i].Response;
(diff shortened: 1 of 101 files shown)


MihaZupan (approved):
Is there an editorconfig / analyzer rule we can enforce to maintain this style in the future?

redcinelli:
I might completely be out of line, but [ide0041](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0041) is the rule you are looking for.

I believe this should look like smth like that in the `.editorconfig`: 

```ini
#IDE0041
dotnet_style_prefer_is_null_check_over_reference_equality_method = true
```

MihaZupan:
Sadly I believe that rule only applies to code like

```c#
if (ReferenceEquals(foo, null))
```
and not
```c#
if (foo == null)
```

karelz:
Triage: There is not analyzer to track this, which is sad.
We will take this. Sorry for the delay @Kahbazi. Would you mind resolving the conflicts, so that we can merge?

Thanks!

PR #1573, opened 2022-02-16
Point users of custom distributed tracing mechanisms at DistributedContextPropagator

### LINKED ISSUE #1480: Track proxied calls as dependencies in telemetry

Track proxied calls as a dependency in telemetry. I would expect this level of tracking to come as a default by YARP the same way as when you use HttpClient.

An appInsights example of how telemetry looks for request proxied by YARP:

image

... compared to a request made to a controller action which has injected a HttpClient instance and makes arbitrary http call.

image

I am experiencing this behavior with .NET 6 & YARP v1.0.

As YARP is designed to do various transformations, I would be super useful to see the actual HTTP request call made by the proxy.

I was able to find various discussions along this theme, but none of them neither solves them/nor states that this feature should not be part ot the Yarp. See: https://github.com/microsoft/reverse-proxy/issues/708, https://github.com/microsoft/reverse-proxy/pull/456, https://github.com/microsoft/reverse-proxy/discussions/1316

MihaZupan:
Fixes #1480

### DIFF
--- a/docs/docfx/articles/distributed-tracing.md
+++ b/docs/docfx/articles/distributed-tracing.md
@@ -8,6 +8,12 @@ See detailed guides for setting up your application with:
 
 .NET 6.0 has built-in configurable support for distributed tracing that YARP takes advantage of to enable such scenarios out-of-the-box.
 
+## Using custom tracing headers
+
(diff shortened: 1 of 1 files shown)


Tratcher (approved):

MihaZupan:
License checks being stuck is apparently a known issue: https://github.com/dotnet/runtime/issues/65437

PR #1574, opened 2022-02-16
Only add API documentation for released projects

MihaZupan:
Turns the [API documentation list](https://microsoft.github.io/reverse-proxy/api/index.html) into a much more manageable and relevant data source (avoids all the k8s and SF code from showing up there).

image


docs/docfx/docfx.json
```diff
@@ -5,7 +5,8 @@
         {
           "src": "../../src",
           "files": [
-            "**.csproj"
```
Tratcher, line 8: Can you leave this line here commented out? I see this as a temporary change that we'll revert after 1.1.

Tratcher (approved):

PR #1580, opened 2022-02-19
PoC for Certificate Store and TLS Support

dpbevin:
@MihaZupan @Tratcher this is early draft for discussion as it'll end up being quite big and I'd like to gauge opinions before going too much further. This change is split into two commits, which might make it easier to understand. In reality, I'd definitely want to turn this into several PRs.

### Commit 1 - Initial removal of separate ingress

This commit removes the `DispatchController` from the the controller as well as the need for a separate "Ingress" deployable. Note that I didn't delete the Ingress or update the docs because I want to make sure we're agreed here. The motivation for this change was for two reasons.

The first reason is simplicity - having only a single deployment means fewer moving parts. When I had early discussion with @samsp-msft there seemed to be a desire to spread the load of the multiple ingress from a single k8s configuration management instance. With the recent addition of the "Ingress Class" capability, having segregated ingresses is still possible without the `DispatchController` because ingresses can be associated with specific instances. So you get the best of both worlds with a single, combined deployment.

The second reason is for security, which will be come apparent in the next commit. If an ingress needs to perform TLS termination, it requires access to a set of public/private keys. The only mechanism in place previously would be for the ingress to retrieve the public/private keys via the `DispatchController` API. This would expose an unauthenticated API with access to private keys - something that is bad in anyone's book.

### Commit 2 - Initial PoC for certificate store and TLS support

The main components of this change are:
- Utilize the TLS specification of Kubernetes ingress resources: https://kubernetes.io/docs/concepts/services-networking/ingress/#tls
- Utilize Kestrel's SNI `ServerCertificateSelector` to provide TLS termination
- Maintain a cache of `X509Certificate2` objects for ingresses that are being proxied.
- Extend the `ResourceInformer` to filter out specific secrets so that we only pull in secrets we're interested in. When using `helm` with a cluster, a lot of additional information is stored in secrets that would make secret retrieval slower without a filtering mechanism.

The PoC has left me with some questions though:

1. Is using `ServerCertificateSelector` the right way to go? (I feel it is but would appreciate input from you guys)
2. Certificate conversion is more difficult before .NET 5 due to API limitations (you'll see from the `#if` statements). Given the support lifecycle of ASP.NET Core 3.1, do we even need to consider supporting 3.1?
3. Following on from 2, can we add support for .NET 6 yet?
4. The `ServerCertificateCache` is caching more than it needs to for the various scenarios of K8s ingress being updated before or after the K8s secret. Addressing this would require an ad-hoc mechanism for reading K8s secrets - likely in the `ResourceInformer` (or similar). This leads me to questions around the intentions of the "Operator Framework" and 1) is this used by MS groups outside of this project and, 2) whether there's an opportunity to simplify/streamline. Very keen to get your thoughts there.

Let me know if we need something more than responses here to discuss.


MihaZupan:
> Given the support lifecycle of ASP.NET Core 3.1, do we even need to consider supporting 3.1?

Given we haven't shipped anything here with 3.1 yet and 6.0 is LTS, I'm fine with dropping it.

> Following on from 2, can we add support for .NET 6 yet?

Yes, even YARP itself already targets 6.0.

specialforest:
> ### Commit 1 - Initial removal of separate ingress

If go this path it'd great to have in future services and other setup steps from src/Kubernetes.Controller/Startup.cs exposed as an extension method, e.g., something like services.AddIngressController(). For my use case I want to use custom YARP as ingress controller, so in order to use Kubernetes.Controller as a library need to remove Program.cs from it, right? Perhaps make KuberenetesIngress.Sample/Ingress the controller.

dpbevin:
I'm curious to know more about your use case @specialforest. DM?

specialforest:
> I'm curious to know more about your use case @specialforest. DM?

a long story short - custom middlewares

> Having the core functionality baked into a library seems reasonable to me. I still think there's a case for an out of the box image too.

agree on both

dpbevin:
Closing this since it will need to be reworked significantly after the library simplification change

PR #1584, opened 2022-02-22
Update K8S client to v5 and K8S images to .NET 6.0

dpbevin:
- Updated [KubernetesClient NuGet](https://www.nuget.org/packages/KubernetesClient/) to 5.x branch (issue identified when using 6.x branch, so will investigate separately)
- Updated Kubernetes images (ingress, controller, backend) to .NET 6.0 only
  - Selectively disabled CS8002 warning since targeting .NET 6 (see https://docs.microsoft.com/en-us/dotnet/standard/assembly/strong-named)
- Removed support for .NET 5.0 and ASP.NET Core 3.1 for Kubernetes images **only** (ingress, controller, backend).
  - **NOTE: All other libraries retain same multi-target version compatibilities.**


samples/KuberenetesIngress.Sample/Ingress/Dockerfile
```diff
@@ -5,7 +5,7 @@ WORKDIR /app
 EXPOSE 80
 EXPOSE 443
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0 AS publish
+FROM mcr.microsoft.com/dotnet/sdk:6.0.100 AS publish
```
MihaZupan, line 8: Why is this change needed?

src/Kubernetes.Controller/Yarp.Kubernetes.Controller.csproj
```diff
@@ -1,23 +1,24 @@
...
-      <UserSecretsId>8436591a-094d-47de-973a-d41a5d91b895</UserSecretsId>
-      <DockerDefaultTargetOS>Linux</DockerDefaultTargetOS>
-      <IsPackable>$('System.TeamProject') != 'internal'</IsPackable>
-    </PropertyGroup>
+  <PropertyGroup>
+    <TargetFrameworks>net6.0</TargetFrameworks>
+    <UserSecretsId>8436591a-094d-47de-973a-d41a5d91b895</UserSecretsId>
+    <DockerDefaultTargetOS>Linux</DockerDefaultTargetOS>
+    <IsPackable>$('System.TeamProject') != 'internal'</IsPackable>
+    <NoWarn>$(NoWarn);CS8002</NoWarn>
```
MihaZupan, line 8: Why would this error be reported after upgrading the framework? Is it a result of updating the referenced package versions?

MihaZupan (approved):

samples/KuberenetesIngress.Sample/Ingress/Dockerfile
specialforest: I guess because .NET SDK got updated to 6.0.200 recently, and the latest mcr sdk image uses it, but global.json still references the older version 6.0.100. Docker image build fails at the moment:

```
#20 0.572       A compatible installed .NET SDK for global.json version [6.0.100] from [/src/global.json] was not found.
#20 0.572       Install the [6.0.100] .NET SDK or update [/src/global.json] with an installed .NET SDK:
#20 0.573         6.0.200 [/usr/share/dotnet/sdk]
```

src/Kubernetes.Controller/Yarp.Kubernetes.Controller.csproj
MihaZupan: Tried it out locally - it's due to the `KubernetesClient` being updated.

samples/KuberenetesIngress.Sample/Ingress/Dockerfile
MihaZupan: Opened #1592 to track updating these

PR #1589, opened 2022-02-23
Fix Kubernetes Controller Dockerfile Build

archonandrewhunt:
global.json specifies 6.0.100 explicitly, however the dockerfile uses the 6.0 sdk tag, which has moved on to 6.0.200. This fixes the build.


MihaZupan:
Hi, thanks for looking to contribute. This specific issue will be addressed as part of #1584, so I am going to close this PR in favor of that one.

PR #1593, opened 2022-02-25
Adding support to reset http.sys delegation queues

### LINKED ISSUE #1587: Add support to Http.sys delegation to handle receiver restarts
This expands on #1503. 

When using http.sys delegation, there are two processes involved, the delegator (the process creating the DelegationRule to delegate requests it receives) and the receiver (the process the delegator transfers requests to). It's possible that the receiver may restart independently from the delegator and ideally, delegation continues to work once the receiver has finished restarting. 

The problem is the delegator has a handle open to the receiver's queue which prevents it from getting removed when the receiver shuts down. There are two possible issues due to this:

1. The receiver may fail to properly start because trying to create the queue will fail. Instead, the receiver needs to be smart enough to attach instead and properly re-configure the queue. The current ASP.NET core http.sys server does not support this (https://github.com/dotnet/aspnetcore/issues/40359), and I wouldn't expect most others to either since this is a newer concept. 
2. If the receiver does attach and re-configure the queue, delegation will fail due to this bug: https://github.com/dotnet/aspnetcore/issues/40358.

To work around \# 2, we can check if the error received while delegating is `ERROR_INVALID_PARAMETER`, and if so dispose and re-create the DelegationRule.

Solving for \# 1 is less straightforward. There is no way to get notified when the receiver detaches from the queue. We could poll but that doesn't seem ideal. Another option would be to monitor for when the receiver process exits and when it does to dispose the rule so it can properly start back up. 

I suggest we add support for consumers to "reset" a rule. This allows consumers to setup any type of monitoring they would like and then gives them the ability to force the DelegationRule to be rest.

I suggest adding the below public interface to allow for this:

```C#
public interface IHttpSysDelegator
{
    void ResetDestination(DestinationState destination);
}
```

NGloreous:
Fixes #1587.

### DIFF
--- a/src/ReverseProxy/Delegation/HttpSysDelegator.cs
+++ b/src/ReverseProxy/Delegation/HttpSysDelegator.cs
@@ -235,7 +235,7 @@ internal sealed class HttpSysDelegator : IHttpSysDelegator, IClusterChangeListen
             if (state.IsInitialized)
             {
                 delegated = TryDelegateInternal(requestDelegationFeature, state.Rule, _ruleResetEvent, _delegatingCount, out delegationException);
-                if (!delegated && ShouldTryToReset(delegationException, destination))
+                if (!delegated && ShouldTryToReset(delegationException))
                 {
                     state = Reset(serverDelegationFeature, state);
 
@@ -386,15 +386,12 @@ internal sealed class HttpSysDelegator : IHttpSysDelegator, IClusterChangeListen
             };
         }
 
-        private static bool ShouldTryToReset(Exception? delegateException, DestinationState destination)
+        private static bool ShouldTryToReset(Exception? delegateException)
         {
             return delegateException switch
             {
                 // Work around for https://github.com/dotnet/aspnetcore/issues/40358
                 HttpSysException ex when ex.ErrorCode == ERROR_INVALID_PARAMETER => true,
-
-                // Receiver has shut down or detached from the queue
-                HttpSysException ex when ex.ErrorCode == ERROR_FILE_NOT_FOUND && destination.ShouldResetDelegationQueueOnErrorNotFound() => true,
 
                 _ => false,
             };
--- a/src/ReverseProxy/Delegation/HttpSysDelegator.cs
+++ b/src/ReverseProxy/Delegation/HttpSysDelegator.cs
@@ -8,14 +8,12 @@ using System.Data;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.CompilerServices;
-using System.Threading;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.HttpSys;
 using Microsoft.Extensions.Logging;
 using Yarp.ReverseProxy.Forwarder;
 using Yarp.ReverseProxy.Model;
-using Yarp.ReverseProxy.Utilities;
 
 namespace Yarp.ReverseProxy.Delegation;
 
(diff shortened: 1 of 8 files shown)


src/ReverseProxy/Delegation/HttpSysDelegator.cs
```diff
@@ -190,69 +202,230 @@ private void RemoveDeadQueueReferences()
...
+
+        private static bool ShouldTryToReset(Exception? delegateException, DestinationState destination)
+        {
+            return delegateException switch
+            {
+                // Work around for https://github.com/dotnet/aspnetcore/issues/40358
+                HttpSysException ex when ex.ErrorCode == ERROR_INVALID_PARAMETER => true,
+
+                // Receiver has shut down or detached from the queue
+                HttpSysException ex when ex.ErrorCode == ERROR_FILE_NOT_FOUND && destination.ShouldResetDelegationQueueOnErrorNotFound() => true,
```
NGloreous, line 397: @Tratcher this was the other case I was talking about. Let me know what you think. If you are fine with this approach in general then I'll update the docs.

src/ReverseProxy/Delegation/HttpSysDelegator.cs
```diff
@@ -190,69 +202,230 @@ private void RemoveDeadQueueReferences()
...
+            static bool TryDelegateInternal(
+                IHttpSysRequestDelegationFeature requestDelegationFeature,
+                DelegationRule rule,
+                ManualResetEventSlim resetEvent,
+                AtomicCounter delegatingCount,
+                [NotNullWhen(false)] out Exception? delegateException)
+            {
+                try
+                {
+                    // This ensures we don't dispose/reset the while a request is trying to delegate
```
NGloreous, line 262: I could have used a ReaderWriterLockSlim here instead. Since the most common case is there are lots of readers and very seldom a writer (i.e. a reset), this approach seemed like less overhead because in the no-writer case, it's just an if check inside of ManualResetEventSlim. ReaderWriterLockSlim is much more complex and involves a SpinLock every time.

I'm open to other ideas/suggestions.

Tratcher: The delegation uses SafeHandles that should prevent disposal while in use. Worse case you have a race where it's disposed just before use and you get an ODE you can catch here. You shouldn't need the MRES or ref counting AtomicCounter here, let SafeHandle take care of it.

NGloreous: I was originally thinking of just catching the ODE but due to the linked line below, it's possible that we unset the delegation property then before we dispose the safehandle another thread tries to delegate. At that point I think http.sys will error out but not sure what error code it will return.

https://github.com/dotnet/aspnetcore/blob/de3019b7dab5b5995942db1fbec6aa466c4af34c/src/Servers/HttpSys/src/DelegationRule.cs#L50

Assuming I figure out the error code http.sys would return in this case, what would you want to happen if an ODE is throws or the error code is returned? Retry or just let it fail?

NGloreous: Per our chat, removing all retries here now that we back ported the fix to ASP.NET Core 6. https://github.com/dotnet/aspnetcore/pull/40967

test/ReverseProxy.FunctionalTests/HttpSysDelegationTests.cs
```diff
@@ -36,6 +38,7 @@ public async Task RequestDelegated()
             proxyBuilder => { },
             proxyApp =>
             {
+                delegator = proxyApp.ApplicationServices.GetRequiredService<HttpSysDelegator>();
```
Tratcher, line 41:
```suggestion
                delegator = proxyApp.ApplicationServices.GetRequiredService<IHttpSysDelegator>();
```

Tratcher (approved):

PR #1595, opened 2022-03-01
Consistent casing in Destination Health logs

rwkarg:
`ActiveDestinationHealthStateIsSet` has leading uppercase when logging the enum value. This makes the literal string states consistent with that.


MihaZupan (approved):
Thanks!

Issue #1596, opened 2022-03-02
1.1.0-RC.1 release checklist


MihaZupan:
See [Release.md](https://github.com/microsoft/reverse-proxy/blob/main/docs/operations/Release.md) for detailed instructions.

- [x] Create an [RC milestone](https://github.com/microsoft/reverse-proxy/milestone/18) and move all resolved 1.1 issues/PRs to it
- [x] Ensure there's a release branch created (see [Branching](https://github.com/microsoft/reverse-proxy/blob/main/docs/operations/Branching.md))
  - https://github.com/microsoft/reverse-proxy/tree/release/1.1.0-rc1 branched on 7ca6acb0594489cb8bedce12662f0c7f58d797a5
- [x] Ensure the `Version.props` has the `PreReleaseVersionLabel` updated to the next preview
- [x] Identify and validate the build on the `microsoft-reverse-proxy-official` pipeline
  - https://dev.azure.com/dnceng/internal/_build/results?buildId=1641870&view=results for commit cf3b6e1
- [x] Tag the commit
  - https://github.com/microsoft/reverse-proxy/releases/tag/v1.1.0-rc.1
- [x] Draft release notes
- [x] Release the build
  - https://www.nuget.org/packages/Yarp.ReverseProxy/1.1.0-rc.1.22152.1
- [x] Update samples to use updated package versions
  - f685a7dfb1937af85a6441278617a600c9a66bfd, 4c32d925bfe285917cbab950a780c761fa577d9d
- [x] Publish the docs
  - #1605
- [x] Publish release notes
  - https://github.com/microsoft/reverse-proxy/releases/tag/v1.1.0-rc.1
- [x] Close the [old milestone](https://github.com/microsoft/reverse-proxy/milestones)
- [x] Announce on social media
  - https://twitter.com/_MihaZupan/status/1499887488567070721
- [ ] Set the preview branch to protected
- [x] ~~Delete the [previous preview branch](https://github.com/microsoft/reverse-proxy/branches)~~
  - N/A for this release since we are keeping the 1.0 branch
- [x] Request source code archival
  - https://portal.microsofticm.com/imp/v3/incidents/details/292957616/home

PR #1603, opened 2022-03-04
Fix Prometheus sample route definitions + typos

MihaZupan:
I will backport this change onto `release/latest` once #1596 is finished.


samsp-msft:
Oopse.

adityamandaleeka (approved):

PR #1629, opened 2022-04-01
Add HTTP/3 support

### LINKED ISSUE #1208: Support for HTTP/3
Now that HTTP/3 support is included in Kestrel and HttpClient in Preview 7 of .NET 6, we should make sure that YARP will support HTTP/3.

> **Note: HTTP/3 support is dependent on support from the underlying .NET**
> YARP is going to require at least .NET 6, possibly .NET 7 depending on what issues are found and where in the stack they need to be fixed. The HTTP/3 RFC is not yet final, and so may also require additional changes. The support for HTTP/3 in .NET 6 is preview, so is not considered complete, and there will likely be issues as it gets stressed. HTTP/3 support may be updated by patches to .NET 6 but that is not guaranteed.  

In conversations amongst the team, we think there is a possibility it might just work if http/3 is [enabled](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/http3?view=aspnetcore-6.0) for Kestrel and HttpClient, but the devil is going to be in the detail. 

Things that need to be checked:
- [ ] Does header filtering account correctly for HTTP/3
- [x] Connection count management accounts for the fact that you *should* only need 1 http/3 connection, as multiple requests should be multiplexable
- [x] What happens if an upgrade fails due to firewalls preventing QUIC connections

greenEkatherine:
Resolves #1208 

Mostly docs update and adding new version into config validation.

### DIFF
--- a/docs/docfx/articles/http3.md
+++ b/docs/docfx/articles/http3.md
@@ -35,11 +35,13 @@ To use HTTP/3 with .NET 6, preview features need to be enabled via a setting in 
 
 ## HttpClient
 
-There is the switch which enables HTTP/3 either programmatically or in project features:
+There is the switch which enables HTTP/3 either programmatically:
 ```C#
-// Set this switch programmatically or in csproj:
-// <RuntimeHostConfigurationOption Include="System.Net.SocketsHttpHandler.Http3Support" Value="true" />
 AppContext.SetSwitch("System.Net.SocketsHttpHandler.Http3Support", true);
+```
+or in project feature:
+```csproj
+<RuntimeHostConfigurationOption Include="System.Net.SocketsHttpHandler.Http3Support" Value="true" />
 ```
 
 In addition to this default version of HttpRequest should be replaced by "3", find more details about [HttpRequest configuration](http-client-config.md#httprequest).
--- a/docs/docfx/articles/http3.md
+++ b/docs/docfx/articles/http3.md
@@ -17,14 +17,12 @@ myHostBuilder.ConfigureWebHostDefaults(webHostBuilder =>
         {
             kestrel.ListenAnyIP(443, portOptions =>
             {
-#if NET6_0_OR_GREATER
                 portOptions.Protocols = HttpProtocols.Http1AndHttp2AndHttp3;
-#endif
-            portOptions.UseHttps();
+                portOptions.UseHttps();
+            });
         });
-    });
     webHostBuilder.UseStartup<Startup>();
-});
+    });
 ```
 To use HTTP/3 with .NET 6, preview features need to be enabled via a setting in the project file:
 ```proj
--- a/docs/docfx/articles/http3.md
+++ b/docs/docfx/articles/http3.md
@@ -1,7 +1,10 @@
 # HTTP/3
 
 ## Introduction
-To enable HTTP/3 protocol on YARP you should set up HttpClient and Kestrel.
+YARP supports HTTP/3 for inbound and outbound connections using the HTTP/3 preview support in .NET 6. To enable the HTTP/3 protocol in YARP you need to:
+- Configure inbound connections in Kestrel
+- Configure outbound connections in HttpClient 
+- Enable preview features
 
 ## Set up HTTP/3 on Kestrel
 
@@ -23,7 +26,7 @@ myHostBuilder.ConfigureWebHostDefaults(webHostBuilder =>
     webHostBuilder.UseStartup<Startup>();
 });
 ```
-For correct project build with `HttpProtocols.Http1AndHttp2AndHttp3` preview features need to be enabled:
+To use HTTP/3 with .NET 6, preview features need to be enabled via a setting in the project file:
 ```proj
 <PropertyGroup>
   <EnablePreviewFeatures>True</EnablePreviewFeatures>
--- a/docs/docfx/articles/http3.md
+++ b/docs/docfx/articles/http3.md
@@ -25,7 +25,7 @@ myHostBuilder.ConfigureWebHostDefaults(webHostBuilder =>
     });
 ```
 To use HTTP/3 with .NET 6, preview features need to be enabled via a setting in the project file:
-```proj
+```csproj
 <PropertyGroup>
   <EnablePreviewFeatures>True</EnablePreviewFeatures>
 </PropertyGroup>
@@ -37,11 +37,11 @@ There is the switch which enables HTTP/3 either programmatically:
 ```C#
 AppContext.SetSwitch("System.Net.SocketsHttpHandler.Http3Support", true);
 ```
-or in project feature:
+or in project file:
 ```csproj
 <RuntimeHostConfigurationOption Include="System.Net.SocketsHttpHandler.Http3Support" Value="true" />
 ```
 
-In addition to this default version of HttpRequest should be replaced by "3", find more details about [HttpRequest configuration](http-client-config.md#httprequest).
+In addition to this, the default version of HttpRequest should be replaced by "3", find more details about [HttpRequest configuration](http-client-config.md#httprequest).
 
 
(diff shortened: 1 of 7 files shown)


MihaZupan:
We have some logic that's specific to HTTP/2 in `HttpForwarder`:
https://github.com/microsoft/reverse-proxy/blob/b6477b184191613db69c2826611c392cde07ba62/src/ReverseProxy/Forwarder/HttpForwarder.cs#L112

Should this be updated to handle `Http2OrGreater`?

It looks like it's also used in our downgrade log:
https://github.com/microsoft/reverse-proxy/blob/b6477b184191613db69c2826611c392cde07ba62/src/ReverseProxy/Forwarder/HttpForwarder.cs#L138-L143

This should likely be updated to notify on downgrade from HTTP/3 as well.

greenEkatherine:
> We have some logic that's specific to HTTP/2 in `HttpForwarder`:
> 
> https://github.com/microsoft/reverse-proxy/blob/b6477b184191613db69c2826611c392cde07ba62/src/ReverseProxy/Forwarder/HttpForwarder.cs#L112

It is also used for gRPC so I don't touch it. In your example I probably should replace condition

Tratcher:
> It is also used for gRPC so I don't touch it. In your example I probably should replace condition

gRPC is considered forward compatible with HTTP/3, so this check should be expanded to HTTP/2 or greater.

Tratcher:
https://github.com/microsoft/reverse-proxy/blob/b6477b184191613db69c2826611c392cde07ba62/src/ReverseProxy/Forwarder/HttpForwarder.cs#L139-L143
Consider removing this downgrade check always logging something like: "Received HTTP/(version) response (status code)."

docs/docfx/articles/http3.md
```diff
@@ -0,0 +1,44 @@
+# HTTP/3
+
+## Introduction
+To enable HTTP/3 protocol on YARP you should set up HttpClient and Kestrel.
```
samsp-msft, line 4:
```suggestion
YARP supports HTTP/3 for inbound and outbound connections using the HTTP/3 preview support in .NET 6. To enable the HTTP/3 protocol in YARP you need to:
- Configure inbound connections in Kestrel
- Configure outbound connections in HttpClient 
- Enable preview features
```

docs/docfx/articles/http3.md
```diff
@@ -0,0 +1,44 @@
...
+#if NET6_0_OR_GREATER
+                portOptions.Protocols = HttpProtocols.Http1AndHttp2AndHttp3;
+#endif
+            portOptions.UseHttps();
+        });
+    });
+    webHostBuilder.UseStartup<Startup>();
+});
+```
+For correct project build with `HttpProtocols.Http1AndHttp2AndHttp3` preview features need to be enabled:
```
samsp-msft, line 26:
```suggestion
To use HTTP/3 with .NET 6, preview features need to be enabled via a setting in the project file:
```

docs/docfx/articles/http3.md
```diff
@@ -0,0 +1,44 @@
...
+```proj
+<PropertyGroup>
+  <EnablePreviewFeatures>True</EnablePreviewFeatures>
+</PropertyGroup>
+```
+
+## HttpClient
+
+There is the switch which enables HTTP/3 either programmatically or in project features:
+```C#
```
samsp-msft, line 36: Rework to show 2 snippets, one for project, the other in code.

samsp-msft (approved):

greenEkatherine:
> Consider removing this downgrade check and always logging something like: "Received HTTP/(version) response (status code)."

Done, it required event renaming as well. Should I mark the PR as "breaking-change" after that?

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -135,12 +135,8 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
...
             }
 
-            // Detect connection downgrade, which may be problematic for e.g. gRPC.
-            if (isClientHttp2 && destinationResponse.Version.Major != 2)
-            {
-                // TODO: Do something on connection downgrade...
-                Log.HttpDowngradeDetected(_logger);
-            }
+            // TODO: Do something on connection downgrade...
+            Log.ResponseDetected(_logger, destinationResponse);
```
Tratcher, line 139:
```suggestion
            Log.ResponseReceived(_logger, destinationResponse);
```

src/ReverseProxy/Utilities/EventIds.cs
```diff
@@ -33,7 +33,7 @@ internal static class EventIds
     public static readonly EventId RouteAdded = new EventId(27, "RouteAdded");
     public static readonly EventId RouteChanged = new EventId(28, "RouteChanged");
     public static readonly EventId RouteRemoved = new EventId(29, "RouteRemoved");
-    public static readonly EventId HttpDowngradeDetected = new EventId(30, "HttpDowngradeDetected");
+    public static readonly EventId ResponseDetected = new EventId(30, "ResponseDetected");
```
Tratcher, line 36: To avoid the breaking change, leave the old EventId unused and unchanged and create a new one.

docs/docfx/articles/http3.md
```diff
@@ -0,0 +1,47 @@
...
+            {
+                portOptions.Protocols = HttpProtocols.Http1AndHttp2AndHttp3;
+                portOptions.UseHttps();
+            });
+        });
+    webHostBuilder.UseStartup<Startup>();
+    });
+```
+To use HTTP/3 with .NET 6, preview features need to be enabled via a setting in the project file:
+```proj
```
MihaZupan, line 28:
```suggestion
```csproj
```

docs/docfx/articles/http3.md
```diff
@@ -0,0 +1,47 @@
...
+</PropertyGroup>
+```
+
+## HttpClient
+
+There is the switch which enables HTTP/3 either programmatically:
+```C#
+AppContext.SetSwitch("System.Net.SocketsHttpHandler.Http3Support", true);
+```
+or in project feature:
```
MihaZupan, line 40:
```suggestion
or in project file:
```

docs/docfx/articles/http3.md
```diff
@@ -0,0 +1,47 @@
...
+```csproj
+<RuntimeHostConfigurationOption Include="System.Net.SocketsHttpHandler.Http3Support" Value="true" />
+```
+
+In addition to this default version of HttpRequest should be replaced by "3", find more details about [HttpRequest configuration](http-client-config.md#httprequest).
```
MihaZupan, line 45:
```suggestion
In addition to this, the default version of HttpRequest should be replaced by "3", find more details about [HttpRequest configuration](http-client-config.md#httprequest).
```

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -771,9 +766,13 @@ private static class Log
...
 
-        public static void HttpDowngradeDetected(ILogger logger)
+        public static void ResponseReceived(ILogger logger, HttpResponseMessage msg)
         {
-            _httpDowngradeDetected(logger, null);
+            if (logger.IsEnabled(LogLevel.Debug))
+            {
+                var version = ProtocolHelper.GetHttpProtocol(msg.Version);
+                _responseReceived(logger, version, (int)msg.StatusCode, null);
+            }
```
MihaZupan, line 775: Can we change `_responseReceived` to take the `Version` instead of `string`, and skip the extra logic here?

MihaZupan (approved):

src/ReverseProxy/Forwarder/HttpForwarder.cs
greenEkatherine: Sure, I just did the similar way like `Proxying` method

PR #1630, opened 2022-04-01
Docs for custom request and response body transforms 

### LINKED ISSUE #1204: How to transform a response using an ITransformProvider
I want to completely rewrite the body of a response - how can this be done?

The [AddResponseTransform](https://microsoft.github.io/reverse-proxy/articles/transforms.html#addresponsetransform) section of the docs mentions the `AddResponseTransform` method which:
> allows creating a custom response transform without implementing a ResponseTransform derived class

The [ResponseTransform](https://microsoft.github.io/reverse-proxy/articles/transforms.html#responsetransform) section of the docs mentions:
> Avoid reading or modifying the response body as this may disrupt the proxying flow

So how are we meant to transform the response body? It seems like the docs are saying *"you can mody the response but not the response body"* - is that the case?

I've tried the following, which results in the modified body being included in the response *followed by* the original response;

```csharp
public class MyTransformProvider : ITransformProvider
{
    public void Apply(TransformBuilderContext context)
    {
        var transform = new ResponseFuncTransform((context) =>
        {
            var responseObject = new 
            {
                Value = "Transformed Response"
            };

            var json = Newtonsoft.Json.JsonConvert.SerializeObject(responseObject);
            var bytes = System.Text.Encoding.UTF8.GetBytes(json);
            context.HttpContext.Response.Body.WriteAsync(bytes, 0, bytes.Length);
            
            return new ValueTask(Task.FromResult(responseObject));
        });

        context.ResponseTransforms.Add(transform);
    }
    
    // ...
}
```

Returns a response body of:
```json
{
    "Value":"Transformed Response"
}{
    "Value":"Original response returned by server"
}
```

Any help understanding how to completely transform a response, including the response content/body, would be greatly appreciated. I'm finding the online documentation and inline code documentation a little light when it comes to transformations.

Tratcher:
Fixes #1204

### DIFF
--- a/docs/docfx/articles/transforms.md
+++ b/docs/docfx/articles/transforms.md
@@ -971,7 +971,7 @@ All response trailers transforms must derive from the abstract base class [Respo
 
 YARP does not provide any built in transforms for modifying the request body. However, the body can be modified in custom transforms.
 
-Be careful about which kids of request are modified, how much data gets buffer, enforcing timeouts, parsing untrusted input, and updating the body related headers like `Content-Length`.
+Be careful about which kinds of requests are modified, how much data gets buffered, enforcing timeouts, parsing untrusted input, and updating the body-related headers like `Content-Length`.
 
 The below example uses simple, inefficient buffering to transform requests. A more efficient implementation would wrap and replace `HttpContext.Request.Body` with a stream that performed the needed modifications as data was proxied from client to server. That would also require removing the Content-Length header since the final length would not be known in advance.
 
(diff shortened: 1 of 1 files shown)


docs/docfx/articles/transforms.md
```diff
@@ -967,6 +967,70 @@ All response transforms must derive from the abstract base class [ResponseTransf
 
 All response trailers transforms must derive from the abstract base class [ResponseTrailersTransform](xref:Yarp.ReverseProxy.Transforms.ResponseTrailersTransform). These can freely modify the client HttpResponse trailers. These run after the response body and should not attempt to modify the response headers or body. Consider also adding a parametrized extension method on `TransformBuilderContext` for discoverability and easy of use.
 
+### Request body transforms
+
+YARP does not provide any built in transforms for modifying the request body. However, the body can be modified in custom transforms.
+
+Be careful about which kids of request are modified, how much data gets buffer, enforcing timeouts, parsing untrusted input, and updating the body related headers like `Content-Length`.
```
MihaZupan, line 974:
```suggestion
Be careful about which kinds of requests are modified, how much data gets buffered, enforcing timeouts, parsing untrusted input, and updating the body-related headers like `Content-Length`.
```

docs/docfx/articles/transforms.md
```diff
@@ -967,6 +967,70 @@ All response transforms must derive from the abstract base class [ResponseTransf
...
+        }
+    });
+});
+```
+
+### Response body transforms
+
+YARP does not provide any built in transforms for modifying the response body. However, the body can be modified in custom transforms.
+
+Be careful about which kids of responses are modified, how much data gets buffer, enforcing timeouts, parsing untrusted input, and updating the body related headers like `Content-Length`. You may need to decompress content before modifying it, as indicated by the Content-Encoding header, and afterwards re-compress it or remove the header.
```
MihaZupan, line 1005:
```suggestion
Be careful about which kinds of responses are modified, how much data gets buffered, enforcing timeouts, parsing untrusted input, and updating the body-related headers like `Content-Length`. You may need to decompress content before modifying it, as indicated by the Content-Encoding header, and afterwards re-compress it or remove the header.
```

MihaZupan (approved):

greenEkatherine (approved):

davidfowl:
@Tratcher Why not recommend calling `Response.Clear()` before writing the response? That's what I ended up doing here:

https://github.com/davidfowl/SmartLoadBalancer/blob/c317d10540863ab3ecdae7384fd6f5765667b43a/Microsoft.AspNetCore.SignalR.LoadBalancing/SignalRSessionAffinity.cs#L86-L90

PR #1635, opened 2022-04-04
Docs for IProxyConfigFilter 

### LINKED ISSUE #1492: Docs for AddConfigFilter/IProxyConfigFilter
These should be covered under / referenced from both config-files and config-providers.

There is a sample here: https://github.com/microsoft/reverse-proxy/tree/main/samples/ReverseProxy.ConfigFilter.Sample

Tratcher:
Fixes #1492

### DIFF
--- a/docs/docfx/articles/config-filters.md
+++ b/docs/docfx/articles/config-filters.md
@@ -1,0 +1,92 @@
+# Configuration Filters
+
+## Introduction
+YARP configuration for routes, clusters, and destinations can be loaded from [configuration files](config-files.md) or from [configuration providers](config-providers.md). Configuration filters can be used to modify that raw input before it's validated and applied.
+
+Filters can be used for a variety of purposes such as:
+- Supplementing config fields with data from other sources like the deployment environment
+- Applying system defaults
+- Applying global shared/global settings and enforce policies
+- Substituting placeholder values
+- Normalization and error correction
+
+## AddConfigFilter
+Configuration filters are registered in the Dependency Injection system using the [AddConfigFilter](xref:Microsoft.Extensions.DependencyInjection.ReverseProxyServiceCollectionExtensions) API. Any number of unique filters can be added and will be applied in the order added.
+
+```C#
+// Load the configuration and register a config filter
+services.AddReverseProxy()
+    .LoadFromConfig(_configuration.GetSection("ReverseProxy"))
+    .AddConfigFilter<CustomConfigFilter>();
+```
+
+## IProxyConfigFilter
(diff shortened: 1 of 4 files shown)


greenEkatherine (approved):

docs/docfx/articles/config-filters.md
```diff
@@ -0,0 +1,92 @@
+# Configuration Filters
+
+## Introduction
+YARP configuration for routes, clusters, and destinations can be loaded from [configuration files](config-files.md) or from [configuration providers](config-providers.md). Configuration filters can be used to modify that raw input before it's validated and applied.
+
+Filters can be used for a variety of purposes such as:
+- Supplementing config fields with data from other sources like the deployment environment
+- Applying system defaults
+- Applying global shared/global settings and enforce policies
```
greenEkatherine, line 9:
```suggestion
- Applying shared/global settings and enforce policies
```

MihaZupan: What did you mean here by "shared/global"? To me, these sound like the same thing.

docs/docfx/articles/config-filters.md
```diff
@@ -0,0 +1,92 @@
...
+// Load the configuration and register a config filter
+services.AddReverseProxy()
+    .LoadFromConfig(_configuration.GetSection("ReverseProxy"))
+    .AddConfigFilter<CustomConfigFilter>();
+```
+
+## IProxyConfigFilter
+Configuration filters are written using the [IProxyConfigFilter](xref:Yarp.ReverseProxy.Configuration.IProxyConfigFilter) interface. Because filters are registered in DI they may define a constructor to inject any dependencies.
+
+Filters are called for each route and cluster each time configuration is loaded or re-loaded. They can choose to return the original input unmodified or a modified copy. The C# 9 [records syntax](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#nondestructive-mutation) is convenient for making modified copies.
```
MihaZupan, line 26:
```suggestion
Filters are called for each route and cluster each time configuration is loaded or re-loaded. They can choose to return the original input unmodified or a modified copy. The C# 9 [records syntax](https://docs.microsoft.com/dotnet/csharp/language-reference/builtin-types/record#nondestructive-mutation) is convenient for making modified copies.
```

MihaZupan (approved):

docs/docfx/articles/config-filters.md
Tratcher: The only difference is scope. You may have several groups that share settings, or you may have one global setting. I'll change both to 'common' settings.
```suggestion
- Applying common settings and enforce policies
```

Issue #1640, opened 2022-04-07
Kubernetes NugGet Packaging


karelz:
### Discussed in https://github.com/microsoft/reverse-proxy/discussions/1631

<div type='discussions-op-text'>

<sup>Originally posted by @dpbevin April  3, 2022</sup>
Several people have expressed the desire to consume the Kubernetes YARP via NuGet packages. As of 3-April-2022, there are **seven** different C# projects (in this repo) that comprise the Kubernetes support for YARP.

## Proposal
I'm proposing that we collapse these seven projects into a single project before considering NuGet packaging for the Kubernetes support.

There may also be an opportunity to simplify the generalized "OperatorFramework" code as part of the process and make the code more specific to the job at hand.

@specialforest @MihaZupan </div>

karelz:
Triage: We believe that collapsing makes sense (incl. @MihaZupan's opinion as our top expert on k8s :))

dpbevin:
First PR is ready to roll - this just removes deps and moves the necessary pieces of code around.

I have a follow-up commit ready for a separate PR that cleans up namespaces. I imagine those changes will need a bit more discussion.

danielkatz:
Is there any progress done in this area?

acortelyou:
@MihaZupan Can we consider official nuget packaging now that the project structure has been updated? There are many duplicate requests for this as well as considerable confusion identifying the correct daily builds to use. At least lets get it on a roadmap.

msschl:
Any news here?

simongullberg:
Any update?

PR #1642, opened 2022-04-08
Simplify Library Structure for Kubernetes

dpbevin:
Addresses the simplicity aspect of #1640 with two commits:
* First commit combines the ingress controller into a single deployable
* Second commit moves the relevant pieces of the "Operator Framework" into the Kubernetes library

I'll update the namespaces in the `OperatorFramework` folder of the `Yarp.Kubernetes.Controller` module as a subsequent PR.


samples/KuberenetesIngress.Sample/Ingress/README.md
```diff
@@ -1,3 +1,21 @@
-# Yarp Ingress and Ingress Controller
+# Yarp Ingress Controller
 
-This directory contains a sample ingress as well as the definition for the Kubernetes manifests for the ingress and ingress controller.
+This directory contains a sample ingress as well as the definition for the Kubernetes manifests for the ingress controller.
+
+The sample ingress controller is a single deployable (previously this was two separate deployables).
```
MihaZupan, line 5: Can we keep both modes?
- Your suggested approach of having the ingress & controller live in a single process
- The existing model where you have a controller and possibly multiple ingress instances talking to it instead of the API server

We're all in favour of collapsing unnecessary projects down to simplify the code/structure, but we don't want to throw out the option of hosting the controller standalone.

How complicated do you think it would be to have a single deployable (e.g. single NuGet) as you suggest, but it having the option to act as a standalone ingress, standalone controller or 2-in-1? That is, have an option to specify 1) where to get the config and 2) should it act as a service to serve that config to other ingresses.

archonandrewhunt:
We are currently running the single ingress-controller with multiple ingress instances scenario in production. I would need to know more to understand how combining them would impact this approach. Is the idea that each ingress would monitor the k8s cluster and update their own configs accordingly?

samples/KuberenetesIngress.Sample/Ingress/README.md
dpbevin: It's possible but I still believe we will run into problems when we introduce TLS termination, specifically when it comes to providing access to Kubernetes Secrets. With the "split" design, this would involve sending TLS private/public keys over the API that currently exists (between the ingress and the controller) and given there is no authentication mechanism, this would open up a significant security hole.

My other concern with the "split" approach is that there is no redundancy in the "controller instance" as I believe an ingress instance can only connect to a single "controller instance".

dpbevin:
@archonandrewhunt just to clarify your scenario, you are running multiple replicas of ingress instances for "load balancing", correct? Could you also give me a ballpark idea of the number of ingress instances and the number of ingress/service resources please?

So the immediate positive impact of this change is that you wouldn't have a single point of failure anymore (i.e. the "controller instance").

It is true that each ingress would then listen for k8s ingress/service events independently and maintain their own in-memory cache. So from this respect, there would be a slightly higher load on the kubeapi server - so understanding how many instances you would run would be good to know (perhaps I could simulate that myself).

The combined approach is the same approach to many other ingress controllers available and while there have been some issues with really big systems, other solutions have been developed over the years depending on the needs.
One approach is to utilize [IngressClass](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class) annotations to Ingress sources so that subsets of ingresses are managed by multiple deployments (not just multiple replicas). This is already supported by Yarp.
Another approach is to apply [Field Selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/) when subscribing for k8s event updates from the kubeapi server. For example, one deployment of the ingress monitors namespaces A-K, while another monitors namespaces L-M. While Yarp doesn't support this today, adding it would be relatively straightforward and it's something that I'd already considering doing myself.

dpbevin:
Fixed merge conflict from #1664

archonandrewhunt:
@dpbevin Thanks for getting back to me. We're pretty running a fairly small deployment right now with 4 ingress replicas and ~30 services/ingresses, so based on what you are describing it sounds like combining the controller into the ingress would be fine for us, and probably even preferable as you mentioned to avoid that single point of failure.

Appreciate the info

samples/KuberenetesIngress.Sample/Ingress/README.md
dpbevin: @MihaZupan I've just pushed a change that:
* Still provides a single project with all K8S logic
* Reinstates the split deployable
* Adds a new sample with a combined (single) deployable

Hopefully this covers both bases but I still think a single deployable is the right way to go long term for the reasons I've mentioned.

samples/KuberenetesIngress.Sample/Ingress/README.md
```diff
@@ -1,3 +1,33 @@
-# Yarp Ingress and Ingress Controller
+# Yarp Ingress Controller
 
-This directory contains a sample ingress as well as the definition for the Kubernetes manifests for the ingress and ingress controller.
+This directory contains a sample ingress as well as the definition for the Kubernetes manifests for the ingress controller.
+
+This sample requires two applications to be deployed:
+* An Ingress (this application)
+* A Kubernetes Ingress Monitor ()
```
MihaZupan, line 7:
```suggestion
* A Kubernetes Ingress Monitor (a process listening for changes in k8s and dispatching the Yarp configuration to ingress instances)
```

MihaZupan (approved):

samples/KuberenetesIngress.Sample/Ingress/README.md
MihaZupan: Thanks David, really appreciate the great answers about the benefits of a combined approach.

I'd say we should get this change in as-is, still enabling a split deployment. We can revisit whether that still has merit as a separate discussion. FWIW I'm not so sure anymore after your arguments :)

samples/KuberenetesIngress.Sample/Combined/README.md
```diff
@@ -0,0 +1,21 @@
...
+From the base directory for this repo (where the .sln file is), run the command:
+
+```
+docker build -t yarp-combined:latest -f .\samples\KuberenetesIngress.Sample\Combined\Dockerfile .
+```
+
+## Deploying the Sample Ingress Controller
+
+1. Open the [ingress-controller.yaml](./ingress-controller.yaml) file
+2. Modify the container image to match the name used when building the image, e.g. change `<REGISTRY_NAME>/yarp:<TAG>` to `yarp:latest`
```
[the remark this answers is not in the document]
dpbevin: Fixed

dpbevin:
Rebased to pull in latest changes from `main` branch and addressed feedback.
Let me know if you want me to squash my commits.

PR #1653, opened 2022-04-13
Classify a ConnectTimeout as RequestTimedOut

### LINKED ISSUE #1628: ConnectTimeout bubbles out imprecisely as RequestCanceled
### Describe the bug
When `HttpForwarder.SendAsync` fails due to the SocketsHttpHandler hitting its configured `ConnectTimeout`, YARP produces result  `RequestCanceled`. This isn't necessarily wrong, but it is imprecise and misleading (since **in most other cases, RequestCanceled indicates the client deliberately aborted the request**). In this case, clearly it isn't the client's fault, but rather the destination's (or something on the way to it). This caused as escalation for us due to custom code on our end producing the wrong status code as a result of the wrong expectations on the meaning of `ForwarderError.RequestCanceled` in this case after we started setting `ConnectTimeout` to a reasonable value.

Perhaps a new member in `ForwarderError` would be useful (sugg `RequestConnectionTimedOut`).

### To Reproduce
Create a SocketsHttpHandler specifying `ConnectTimeout = TimeSpan.FromSeconds(15)`, then pass it to `HttpForwarder.SendAsync` and attempt connecting to a *really slow* destination that will not accept the connection in time.

### Further technical details

Notice that when `SocketsHttpHandler.SendAsync` fails due to a connection timeout, the resulting exception chain looks like this (as of .NET 6.0.3):

```
--> System.Threading.Tasks.TaskCanceledException: The operation was canceled.
--> inner exception: System.TimeoutException: A connection could not be established within the configured ConnectTimeout.
```

This condition is then interpreted here as `ForwarderError.RequestCanceled`, which isn't quite precise:

https://github.com/microsoft/reverse-proxy/blob/b6477b184191613db69c2826611c392cde07ba62/src/ReverseProxy/Forwarder/HttpForwarder.cs#L464-L471

- Include the version of the packages you are using: `Yarp.ReverseProxy 1.0.0` (by code inspection, latest 1.1 RC1 is also impacted)
- The platform: Windows / .NET 6.0.3

MihaZupan:
Fixes #1628

`RequestTimedOut` is described as

> Timed out when trying to connect, send the request headers, or receive the response headers.

So a `ConnectTimeout` fits this existing description.

cc: @davidni

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -456,13 +456,18 @@ internal sealed class HttpForwarder : IHttpForwarder
     {
         if (requestException is OperationCanceledException)
         {
-            if (!context.RequestAborted.IsCancellationRequested && requestCancellationSource.IsCancellationRequested)
+            if (context.RequestAborted.IsCancellationRequested)
             {
-                return await ReportErrorAsync(ForwarderError.RequestTimedOut, StatusCodes.Status504GatewayTimeout);
+                return await ReportErrorAsync(ForwarderError.RequestCanceled, StatusCodes.Status502BadGateway);
             }
             else
             {
-                return await ReportErrorAsync(ForwarderError.RequestCanceled, StatusCodes.Status502BadGateway);
+#if NET6_0_OR_GREATER
+                Debug.Assert(requestCancellationSource.IsCancellationRequested || requestException.ToString().Contains("ConnectTimeout"), requestException.ToString());
+#else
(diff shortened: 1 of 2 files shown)


Tratcher (approved):

davidni:
Thanks @MihaZupan, works for us!

nlyu:
@MihaZupan Hi thx for the fix. When would this be released?

MihaZupan:
It's available in [daily builds](https://github.com/microsoft/reverse-proxy/blob/main/docs/DailyBuilds.md).
Otherwise, it will be included in the `1.1` release, which is likely going to be released within a week.

PR #1661, opened 2022-04-20
Fix Component Governance alerts

wtgodbe:


src/OperatorFramework/src/Core/Microsoft.Kubernetes.Core.csproj
```diff
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <PackageReference Include="KubernetesClient" Version="5.0.20" />
+    <PackageReference Include="Portable.BouncyCastle" Version="1.8.8" />
```
Tratcher, line 16: @MihaZupan didn't you want to update the KubernetesClient version instead?

Tratcher (changes requested):

src/OperatorFramework/src/Core/Microsoft.Kubernetes.Core.csproj
MihaZupan: Right. I will close this in favour of #1664.
`Portable.BouncyCastle` won't be a dependency at all given the TFMs these projects target (6.0).

Issue #1665, opened 2022-04-21
HttpForwarder throw exception when trying to forward a POST request.


DeepWaterWhale:
### Describe the bug
HttpForwarder throw exception when trying to forward a POST request.

### To Reproduce
Forward any POST request with body can repro.

Got Exceptions? Include both the message and the stack trace
Error Message: (Parameter 'size') Actual value was 0.

CallStack:
   at Microsoft.AspNetCore.Server.HttpSys.RequestStream.ValidateReadBuffer(Byte[] buffer, Int32 offset, Int32 size)
   at Microsoft.AspNetCore.Server.HttpSys.RequestStream.ReadAsync(Byte[] buffer, Int32 offset, Int32 size, CancellationToken cancellationToken)
   at System.IO.Stream.ReadAsync(Memory`1 buffer, CancellationToken cancellationToken)
   at Yarp.ReverseProxy.Forwarder.StreamCopier.<CopyAsync>d__3.MoveNext() in D:\__GItRepo__\reverse-proxy\src\ReverseProxy\Forwarder\StreamCopier.cs:line 53

-->

### Further technical details

- Include the version of the packages you are using
Use the latest source code to build the package and testing.
commit id is: b8a58a6a7aec30d7d1bdda8e41f02bd1034d20ea

- The platform (Linux/macOS/Windows): Windows

I took a look at the code, in my understanding, yarp create a StreamCopyHttpContent in the proxy request to copy the original request's body.

And in StreamCopyHttpContent, it uses StreamCopier to do the real copy work, code is like below. But it throws exception when use an empty buffer directly, then return RequestBodyClient error back.

```C#
                // Issue a zero-byte read to the input stream to defer buffer allocation until data is available.
                // Note that if the underlying stream does not supporting blocking on zero byte reads, then this will
                // complete immediately and won't save any memory, but will still function correctly.
                var zeroByteReadTask = input.ReadAsync(Memory<byte>.Empty, cancellation);
                if (zeroByteReadTask.IsCompletedSuccessfully)
                {
                    // Consume the ValueTask's result in case it is backed by an IValueTaskSource
                    _ = zeroByteReadTask.Result;
                }
                else
                {
                    // Take care not to return the same buffer to the pool twice in case zeroByteReadTask throws
                    var bufferToReturn = buffer;
                    buffer = null;
                    ArrayPool<byte>.Shared.Return(bufferToReturn);

                    await zeroByteReadTask;

[...]

Issue #1657, opened 2022-04-14
HttpForwarder failed to send body after update to version 1.1.0-rc.1.22152.1 from 1.0.1


illay1994:
### Describe the bug
We have simple yarp proxy with configuration:
```
        services.AddReverseProxy().LoadFromConfig(Configuration.GetSection("ReverseProxy"));
```
 and 
```

        app.UseEndpoints(endpoints =>
        {
            endpoints.MapReverseProxy();
        });
```

config file:

```
 "ReverseProxy": {
    "Routes": {
      "route": {
        "ClusterId": "cluster",
        "AuthorizationPolicy": "policy",
        "Match": {
          "Path": "/v1/{**remainder}"
        }
      }
    },
    "Clusters": {
      "cluster": {
        "Destinations": {
          "api": {
            "Address": "http://localhost:60825/subpath1/subpath2"
          }
        }
      }
    }
  }
```

### To Reproduce
in version 1.0.1 all working fine

after update to version 1.1.0-rc.1.22152.1 get error in log and return 400 HttpError code

logs:

```
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 POST http://192.168.0.2:5000/v1/api application/json;+charset=UTF-8 36
----
internal Authentication logs 
----
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
      Executing endpoint 'route'
info: Yarp.ReverseProxy.Forwarder.HttpForwarder[9]
      Proxying to http://localhost:60825/subpath1/subpath2/v1/api HTTP/2 RequestVersionOrLower no-streaming
info: Yarp.ReverseProxy.Forwarder.HttpForwarder[48]
      RequestBodyClient: The client reported an error when copying the request body.
      System.AggregateException: One or more errors occurred. (Sent 0 request content bytes, but Content-Length promised 36.) (Error while copying content to a stream.)
       ---> System.InvalidOperationException: Sent 0 request content bytes, but Content-Length promised 36.
         --- End of inner exception stack trace ---
       ---> (Inner Exception #1) System.Net.Http.HttpRequestException: Error while copying content to a stream.
       ---> System.IO.IOException: An error occurred when reading the request body from the client.
       ---> System.InvalidOperationException: Sent 0 request content bytes, but Content-Length promised 36.
         --- End of inner exception stack trace ---
[...]

PR #1666, opened 2022-04-21
Disable zero-byte-reads on 3.1 and 5.0

MihaZupan:
Contributes to #1665 and #1657.

@Tratcher feel free to close this one if you think we need any more changes here.


Tratcher (approved):
That was quick 😁

PR #1668, opened 2022-04-24
#1667 Fix watcher for Kubernetes resources

### LINKED ISSUE #1667: Ingress Events Not Detected After Upgrade of KubernetesClient to v6
### Describe the bug
Kubernetes Ingress events don't get picked up after upgrade the KubernetesClient library as part of #1664 . 
When the ingress controller starts, it does load the ingresses correctly, however subsequent updates to ingresses do not.

### To Reproduce
1. `kubectl apply -f .\samples\KuberenetesIngress.Sample\backend\backend.yaml`
1. `kubectl apply -f .\samples\KuberenetesIngress.Sample\backend\ingress-sample.yaml`
1. `docker build -t yarp-controller:latest -f .\src\Kubernetes.Controller\Dockerfile .`
1. `kubectl apply -f .\samples\KuberenetesIngress.Sample\Ingress\ingress-controller.yaml`
1. `docker build -t backend:latest -f .\samples\KuberenetesIngress.Sample\backend\Dockerfile .`
1. `Invoke-WebRequest http://localhost` fails to be forwarded to the backend resource

Revert the KubernetesClient library to `5.0.20` and see that the `iwr` request works as expected

dpbevin:
Fixes #1667

The v6 `KubernetesClient` library changed the implementation of the watcher - see https://github.com/kubernetes-client/csharp/pull/681.

This change updates the `AnyResourceKind.ListClusterAnyResourceKindWithHttpMessagesAsync()` method to match the change in the `KubernetesClient` library.

Unfortunately, the `LineSeparatedHttpContent` class in the `KubernetesClient` library is inaccessible to us (internal), so has been duplicated here. This change then required the addition of the `HttpOperationResponseExtensions.CustomWatch()` extension method to ensure the correct stream reader was created.

### DIFF
--- a/src/OperatorFramework/src/Controller/Informers/HttpOperationResponseExtensions.cs
+++ b/src/OperatorFramework/src/Controller/Informers/HttpOperationResponseExtensions.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using k8s;
+using k8s.Exceptions;
+using Microsoft.Rest;
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Kubernetes.Controller.Informers
+{
+    internal static class HttpOperationResponseExtensions
+    {
+        public static Watcher<T> CustomWatch<T, L>(
+            this Task<HttpOperationResponse<L>> responseTask,
+            Action<WatchEventType, T> onEvent,
+            Action<Exception> onError = null,
+            Action onClosed = null)
+        {
+            return new Watcher<T>(MakeStreamReaderCreator<T, L>(responseTask), onEvent, onError, onClosed);
+        }
+
+        private static Func<Task<TextReader>> MakeStreamReaderCreator<T, L>(Task<HttpOperationResponse<L>> responseTask)
+        {
+            return async () =>
+            {
+                var response = await responseTask.ConfigureAwait(false);
+
+                if (!(response.Response.Content is Microsoft.Kubernetes.Client.LineSeparatedHttpContent content))
+                {
+                    throw new KubernetesClientException("not a watchable request or failed response");
+                }
+
+                return content.StreamReader;
+            };
+        }
+    }
+}
--- a/src/OperatorFramework/src/Controller/Informers/ResourceInformer.cs
+++ b/src/OperatorFramework/src/Controller/Informers/ResourceInformer.cs
@@ -284,7 +284,7 @@ namespace Microsoft.Kubernetes.Controller.Informers
                 cancellationToken: cancellationToken);
 
             DateTime lastEventUtc = DateTime.UtcNow;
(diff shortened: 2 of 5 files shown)


dpbevin:
@MihaZupan The `KubernetesClient` library doesn't really make it easy to have a generics-based solution, which probably explains why `AnyResourceKind` was created in the first place.

Here's an alternative potential fix that removes the need for `AnyResourceKind`, which would make library upgrades easier going forward. Keen to get your thoughts on which approach you'd prefer.

https://github.com/dpbevin/reverse-proxy/commit/c6641096663eea2f539481009ef27f1016cebcb5

MihaZupan:
Trying to understand the difference here:
Previously we had our own logic for dealing with arbitrary resources (ResourceKinds).
Since `KubernetesClient` provides out-of-the-box helpers for all the resources we care about, we are simplifying the logic to just use the helpers instead (e.g. `ListIngressClassWithHttpMessagesAsync`, `ListPodForAllNamespacesWithHttpMessagesAsync`).

I'm not sure I understand what the downsides are. Is it that the OperatorFramework can't be used with different resource kinds in this case (which is fine if we're removing it as a dedicated library anyway)?

Especially if you believe it will make future upgrades easier, I prefer the alternative approach.

dpbevin:
@MihaZupan I think you've captured it well.

The only downside is probably not a valid scenario anymore considering the removal of the OperatorFramework. Introducing concrete classes for resource specific informers (e.g. `V1IngressResourceInformer`) would not be practical for a generic library. However, because we are moving to being a specific purpose, we only need a small, known set of informers. So I agree that the removal of genetics is fine in this case.

The only question now is which order to do the two PRs I have open in. Going with the type-specific resource informers will be easier after the library is collapsed into a single project.

MihaZupan:
> Going with the type-specific resource informers will be easier after the library is collapsed into a single project.

If that means less work for you, that's fine by me.

dpbevin:
> > Going with the type-specific resource informers will be easier after the library is collapsed into a single project.
> 
> If that means less work for you, that's fine by me.

It certainly does. My preferred sequence would be:
1. Complete https://github.com/microsoft/reverse-proxy/pull/1642
2. Abandon this PR https://github.com/microsoft/reverse-proxy/pull/1668
3. Fix #1667 with a new PR based off https://github.com/dpbevin/reverse-proxy/commit/c6641096663eea2f539481009ef27f1016cebcb5

If 1. is going to take a while that I'd prefer not to leave the library in a broken state, in which case I would suggest:
1. Complete this PR
2. Complete https://github.com/microsoft/reverse-proxy/pull/1642 (after fixing merge conflict that this PR will cause)
3. Improve this fix by reworking `main` based off https://github.com/dpbevin/reverse-proxy/commit/c6641096663eea2f539481009ef27f1016cebcb5

dpbevin:
Closing in favour of https://github.com/microsoft/reverse-proxy/pull/1706

PR #1639, opened 2022-04-07
Add support for multi-value header routing

MihaZupan:
Fixes #1494

Matching now treats multiple values as *OR*s. I.e. `Exact`/`Prefix` must match *any* of the request values.
`NotContains` means the value must *not match any* request header.

Values are further split based on the separator `,` (or `;` for `Cookie`). A single outer pair of quotes is also removed before matching.

This PR is generally slightly faster on simple (single-value) inputs, and (as expected) slower on multi-value ones.
Modes that may observe a perf degradation are `Exact` and `Prefix`.
I don't think the difference is prohibitive to the point that having an opt-out switch is needed.

Benchmark scenarios are defined [here](https://gist.github.com/MihaZupan/ccac2200ff7ba2f509060ee1a061e230#file-headerroutingbenchmark-cs-L30-L64).

| Method |         Scenario |      Mean |    Error |   StdDev | Ratio |
|------- |----------------- |----------:|---------:|---------:|------:|
|    Old |   SimpleContains | 292.78 ns | 0.957 ns | 0.848 ns |  1.00 |
|    New |   SimpleContains | 243.82 ns | 0.823 ns | 0.687 ns |  0.83 |
|        |                  |           |          |          |       |
|    Old |      SimpleExact | 158.98 ns | 0.529 ns | 0.469 ns |  1.00 |
|    New |      SimpleExact | 117.40 ns | 0.236 ns | 0.184 ns |  0.74 |
|        |                  |           |          |          |       |
|    Old |     SimplePrefix | 104.09 ns | 0.284 ns | 0.222 ns |  1.00 |
|    New |     SimplePrefix |  97.69 ns | 0.512 ns | 0.454 ns |  0.94 |
|        |                  |           |          |          |       |
|    Old | MultipleContains | 294.57 ns | 1.679 ns | 1.402 ns |  1.00 |
|    New | MultipleContains | 291.51 ns | 3.079 ns | 2.880 ns |  0.99 |
|        |                  |           |          |          |       |
|    Old |    MultipleExact | 124.28 ns | 0.237 ns | 0.185 ns |  1.00 |
|    New |    MultipleExact | 175.34 ns | 0.626 ns | 0.523 ns |  1.41 |
|        |                  |           |          |          |       |
|    Old |   MultiplePrefix | 130.89 ns | 1.529 ns | 1.431 ns |  1.00 |
|    New |   MultiplePrefix | 176.16 ns | 0.513 ns | 0.428 ns |  1.35 |

[...]

PR #1674, opened 2022-04-28
Update header-routing docs after #1639

### LINKED ISSUE #1671: Update docs for multi-value header routing
I realized I forgot to update [the docs](https://microsoft.github.io/reverse-proxy/articles/header-routing.html) before merging #1639.

MihaZupan:
Closes #1671

### DIFF
--- a/docs/docfx/articles/header-routing.md
+++ b/docs/docfx/articles/header-routing.md
@@ -186,7 +186,7 @@ var routes = new[]
             Path = "{**catch-all}",
             Headers = new[]
             {
-            new RouteHeader()
+                new RouteHeader()
                 {
                     Name = "Header4",
                     Values = new[] { "value1", "value2" },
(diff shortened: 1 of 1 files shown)


docs/docfx/articles/header-routing.md
```diff
@@ -254,11 +254,11 @@ A list of possible values to search for. The header must match at least one of t
...
-- Contains - The header must contain the value for a match, subject to the value of `IsCaseSensitive`. Only single headers are supported. If there are multiple headers with the same name then the match fails.
-- NotContains - The header must not contain any of the match values, subject to the value of `IsCaseSensitive`. Only single headers are supported. If there are multiple headers with the same name then the match fails.
+- ExactHeader - Any of the headers with the given name must match in its entirety, subject to the value of `IsCaseSensitive`. If a header contains multiple values (separated by `,` or `;`), they are split before matching. A single pair of quotes will also be stripped from the value before matching.
+- HeaderPrefix - Any of the headers with the given name must match by prefix, subject to the value of `IsCaseSensitive`. If a header contains multiple values (separated by `,` or `;`), they are split before matching. A single pair of quotes will also be stripped from the value before matching.
+- Exists - The header must exist and contain any non-empty value. If there are multiple headers with the same name, the rule will also match.
```
greenEkatherine, line 259: What if there are multiple headers with different names? Any of them should exist? I would add it too if it is a possible scenario

MihaZupan: You mean if the user defined something like the following (multiple header rules)?
```json
"routeFoo" : {
  "ClusterId": "clusterBar",
  "Match": {
    "Path": "{**catch-all}",
    "Headers": [
      {
        "Name": "foo",
        "Mode": "Exists"
      },
      {
        "Name": "bar",
        "Mode": "Exists"
      }
    ]
  }
}
```

This case is called out as
> [If multiple headers rules are specified on a route then all must match for a route to be taken. OR logic must be implemented either within a header rule or as separate routes.](https://github.com/microsoft/reverse-proxy/blob/4bcff01820a3870de07cfb7950524a117b31277f/docs/docfx/articles/header-routing.md#:~:text=If%20multiple%20headers%20rules%20are%20specified%20on%20a%20route%20then%20all%20must%20match%20for%20a%20route%20to%20be%20taken.%20OR%20logic%20must%20be%20implemented%20either%20within%20a%20header%20rule%20or%20as%20separate%20routes.)

It holds for any header rule and wasn't affected by recent changes. Do you think we should have an explicit example for it?

greenEkatherine: Ah, right, this is not about values. In that case I think no need to add example

greenEkatherine (approved):

Tratcher (approved):

PR #1705, opened 2022-05-10
Add catch for Http Exception for target refused

### LINKED ISSUE #1704: Health Check return "Unhealthy" if Destination refuses to connect
In ReverseProxy.Health.ActiveHealthCheckMonitor when calling ProbeDestinationAsync when trying to get a response from the destination if the destination port is not available, the method fails to the catch with HttpRequestException "No connection could be made because the target machine actively refused it." causing the entire probe to fail and status to be updated. By catching this particular exception an "Unhealthy" status can be returned for the destination endpoint.

I am looking at using this as a micro-service api endpoint with some of the services potentially being unavailable.  It would help to know which endpoints are not reachable by having an "Unhealthy" state.

TJPoorman:
Fixes #1704 by adding a catch HttpException for a specific message when probing destinations

### DIFF
--- a/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
+++ b/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
@@ -176,6 +176,12 @@ internal partial class ActiveHealthCheckMonitor : IActiveHealthCheckMonitor, ICl
 
             return new DestinationProbingResult(destination, response, null);
         }
+        catch (HttpRequestException httpEx) when (httpEx.Message.StartsWith("No connection could be made because the target machine actively refused it."))
+        {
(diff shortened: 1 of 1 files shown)


MihaZupan:
I don't think we should make such changes. A connection failing to establish may have different semantics than receiving a 404.

If you know that in your scenario you want to treat them the same, you can choose to do so in your `IActiveHealthCheckPolicy` implementation.

TJPoorman:
> I don't think we should make such changes. A connection failing to establish may have different semantics than receiving a 404.
> 
> If you know that in your scenario you want to treat them the same, you can choose to do so in your `IActiveHealthCheckPolicy` implementation.

Would it be possible to add optional suppression for the warning messages?  Even utilizing a custom `IActiveHealthCheckPolicy` my log still gets filled because of the refused connect.

PR #1706, opened 2022-05-10
#1667 Fix k8s informer after upgrades

### LINKED ISSUE #1667: Ingress Events Not Detected After Upgrade of KubernetesClient to v6

dpbevin:
* Updated Dockerfiles based on SDK change in `global.json`
* Fixed the "combined" Dockerfile
* Fixed an issue with the annotations for the default ingress class
* Fixes https://github.com/microsoft/reverse-proxy/issues/1667 by removing customized ResourceInformer logic (now using out of the box code from the `KubernetesClient` library

Fixes #1667


src/Kubernetes.Controller/OperatorFramework/KubernetesControllerExtensions.cs
```diff
@@ -1,36 +1,31 @@
...
 
     /// <summary>
     /// Registers the resource informer.
     /// </summary>
     /// <typeparam name="TResource">The type of the t related resource.</typeparam>
+    /// <typeparam name="TService">The implementation type of the resource informer.</typeparam>
     /// <param name="services">The services.</param>
     /// <returns>IServiceCollection.</returns>
-    public static IServiceCollection RegisterResourceInformer<TResource>(this IServiceCollection services)
+    public static IServiceCollection RegisterResourceInformer<TResource, TService>(this IServiceCollection services)
```
MihaZupan, line 24: Should this have a `where TService : IResourceInformer<TResource>` constraint?

samples/KuberenetesIngress.Sample/Combined/Dockerfile
```diff
@@ -5,11 +5,11 @@ WORKDIR /app
 EXPOSE 80
 EXPOSE 443
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0.100 AS publish
+FROM mcr.microsoft.com/dotnet/sdk:7.0.100-preview.3 AS publish
```
MihaZupan, line 8: These should still stay on 6, unless there's a reason we need 7 I'm missing?

src/Kubernetes.Controller/OperatorFramework/Informers/ResourceInformer.cs
```diff
@@ -188,6 +190,8 @@ public override async Task RunAsync(CancellationToken cancellationToken)
             }
         }
 
+        protected abstract Task<Rest.HttpOperationResponse<TListResource>> RetrieveResourceList(bool? watch = null, string resourceVersion = null, CancellationToken cancellationToken = default);
```
MihaZupan, line 193:
```suggestion
        protected abstract Task<Rest.HttpOperationResponse<TListResource>> RetrieveResourceListAsync(bool? watch = null, string resourceVersion = null, CancellationToken cancellationToken = default);
```

samples/KuberenetesIngress.Sample/Combined/Dockerfile
dpbevin: The SDK version in the `global.json` file got bumped to 7.x by someone else. Without this change the `docker build` command fails.
Note that the runtime for the Docker images remains on 6 (see line 1 of the Dockerfiles).

src/Kubernetes.Controller/OperatorFramework/Informers/ResourceInformer.cs
dpbevin: Will fix

src/Kubernetes.Controller/OperatorFramework/KubernetesControllerExtensions.cs
dpbevin: Yes, it should. Got spot.

samples/KuberenetesIngress.Sample/Combined/Dockerfile
dpbevin: As a separate PR, I'll see if I can set up a simple workflow to build the Dockerfiles. I'd probably look at something similar to the workflow used for the docs.

MihaZupan: Oh I see, if it's just the SDK that's fine.

PR #1716, opened 2022-05-13
Build Dockerfiles (no publish)

dpbevin:
Adds a GitHub workflow for building (but not publishing) the three main Dockerfiles in the samples.

This should pick up any problems when global.json is modified and the Dockerfiles are not.


MihaZupan (approved):
Thanks!

Issue #1713, opened 2022-05-12
Ship InMemoryConfigProvider


Tratcher:
https://github.com/microsoft/reverse-proxy/blob/main/testassets/ReverseProxy.Code/InMemoryConfigProvider.cs

InMemoryConfigProvider is convenient for providing config from custom sources. It's copied several times in tests and samples, and even in customer projects.
https://github.com/DuendeSoftware/BFF/blob/main/src/Duende.Bff.Yarp/InMemoryConfigProvider.cs

The main reason I haven't done this yet is because I haven't implemented an update pattern.

leastprivilege:
Yes it is useful. Please include it.

samsp-msft:
> The main reason I haven't done this yet is because I haven't implemented an update pattern.

I think this really is the main value add that would be needed. The current implementation is kind of based on providing new collections each time. Should the routes and clusters be configurable, with startBatch, endBatch semantics so that the collection is editable?

leastprivilege:
This would be a nice to have - but not essential IMO. There are other config stores available if you need that functionality. I see this more as a simple config as code approach.

PR #1725, opened 2022-05-17
Fix race condition for concurrent Updates in InMemoryConfigProvider

MihaZupan:
Not sure if we care about this scenario (multiple concurrent `Update`s), but since we're considering exposing it (#1713) it seems worth doing.

For reference the issue is that multiple `Update` calls may read the same `oldConfig` and one of the new configurations will never be signaled. If the proxy picks up that config object, it will stop responding to config updates.


Tratcher (approved):
Lol so many copies.

PR #1732, opened 2022-05-23
Ship InMemoryConfigProvider 

### LINKED ISSUE #1713: Ship InMemoryConfigProvider

Tratcher:
Fixes #1713

This ships InMemoryConfigProvider, removes redundant copies, and proposes an update pattern.


samples/ReverseProxy.Code.Sample/Startup.cs
```diff
@@ -58,7 +62,7 @@ private RouteConfig[] GetRoutes()
             {
                 new RouteConfig()
                 {
-                    RouteId = "route1",
+                    RouteId = "route" + new Random().Next(), // Forces a new route id each time GetRoutes is called.
```
MihaZupan, line 65:
```suggestion
                    RouteId = "route" + Random.Shared.Next(), // Forces a new route id each time GetRoutes is called.
```

src/ReverseProxy/Configuration/InMemoryConfigProviderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public static class InMemoryConfigProviderExtensions
+{
+    /// <summary>
+    /// Adds an InMemoryConfigProvider
+    /// </summary>
+    public static IReverseProxyBuilder LoadFromMemory(this IReverseProxyBuilder builder, IReadOnlyList<RouteConfig> routes, IReadOnlyList<ClusterConfig> clusters)
+    {
+        builder.Services.AddSingleton(new InMemoryConfigProvider(routes, clusters));
+        builder.Services.AddSingleton<IProxyConfigProvider>(s => s.GetRequiredService<InMemoryConfigProvider>());
```
MihaZupan, line 17:
```suggestion
        builder.Services.AddSingleton<IProxyConfigProvider>(new InMemoryConfigProvider(routes, clusters));
```

Isn't this equivalent?

src/ReverseProxy/Configuration/InMemoryConfigProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Extensions.Primitives;
+
+namespace Yarp.ReverseProxy.Configuration;
+
+/// <summary>
+/// Provides an implementation of IProxyConfigProvider to support config being generated by code.
+/// </summary>
+public class InMemoryConfigProvider : IProxyConfigProvider
```
MihaZupan, line 13: Should we start out with it being sealed? `Update` isn't virtual so deriving from it has limited value as-is.

MihaZupan (approved):

src/ReverseProxy/Configuration/InMemoryConfigProviderExtensions.cs
Tratcher: InMemoryConfigProvider is intentionally registered independently so you can resolve InMemoryConfigProvider without downcasting.

greenEkatherine (approved):

Issue #1736, opened 2022-05-24
Single bad ingress prevents other ingresses from working


specialforest:
### Describe the bug
K8 controller fails to handle one of ingresses and as a result YARP configuration is empty. 

### To Reproduce
Likely related to a service not having an endpoint because the corresponding pod is in a crash loop.

```
warn: Yarp.Kubernetes.Controller.Services.Reconciler[0]
      Uncaught exception occured while reconciling
      System.NullReferenceException: Object reference not set to an instance of an object.
         at Yarp.Kubernetes.Controller.Converters.YarpParser.HandleIngressRulePath(YarpIngressContext ingressContext, V1ServicePort servicePort, List`1 endpoints, IList`1 defaultSubsets, V1IngressRule rule, V1HTTPIngressPath path, YarpConfigContext configContext) in src\Kubernetes.Controller\Converters\YarpParser.cs:line 79
         at Yarp.Kubernetes.Controller.Converters.YarpParser.HandleIngressRule(YarpIngressContext ingressContext, List`1 endpoints, IList`1 defaultSubsets, V1IngressRule rule, YarpConfigContext configContext) in src\Kubernetes.Controller\Converters\YarpParser.cs:line 44
         at Yarp.Kubernetes.Controller.Converters.YarpParser.ConvertFromKubernetesIngress(YarpIngressContext ingressContext, YarpConfigContext configContext) in src\Kubernetes.Controller\Converters\YarpParser.cs:line 31
         at Yarp.Kubernetes.Controller.Services.Reconciler.ProcessAsync(CancellationToken cancellationToken) in src\Kubernetes.Controller\Services\Reconciler.cs:line 47
```

### Further technical details
Two issues here: 
1. Exception handling in Reconciler.cs makes me think that single bad ingress prevents others from working. Shall it be moved inside the loop?
2. NullReferenceException that I'm trying to debug.

specialforest:
Another issue I noticed here with WorkQueue. After the exception reconciler stops doing anything despite changes to K8 resources.

specialforest:
NullReferenceException  is caused by ports mismatch in service and ingress specs which makes servicePort variable to be null.

E.g.
Service spec has:
```yaml
ports:
- name: https
  port: 10000
  protocol: TCP
  targetPort: 89
```
and Ingress spec has:
```yaml
port:
  number: 443
```

specialforest:
@MihaZupan could you reopen this?

specialforest:
[...]

PR #1737, opened 2022-05-25
Gracefully handle port mismatch between Ingress and Service specs and missing Service

specialforest:
This change addresses NullReferenceException exceptions during reconciliation #1736.


MihaZupan:
There is a similar problem with the name too, right?
https://github.com/microsoft/reverse-proxy/blob/b84a69f782d599fdc7e37f513c1367cfc119303b/src/Kubernetes.Controller/Converters/YarpParser.cs#L42-L43

Should we just skip the `HandleIngressRulePath` if we can't find the matching service/port?

dpbevin:
@MihaZupan  If there is no matching service one line 42, I agree that there's little point continuing the checks.

I've also wondered whether `ServiceData` should remain a `struct` or if a `class` would be more suitable (same of the `struct Endpoints`).

PR #1743, opened 2022-05-31
Fix flaky ReconciliationContinuesOnReconcilerError unit test

specialforest:
The test was occasionally failing because of order of execution of the
program.


MihaZupan (approved):
Thanks.

> because of order of execution of the
program

Can you describe where the race happened before please? It's not immediately obvious to me why the changed version behaves better.

PR #1753, opened 2022-06-05
Fix `QueryValueParameter` example

bgrainger:
The example documentation produces the error `The proxy config is invalid. (Unknown transform: Append;QueryStringParameter)`.


MihaZupan (approved):
Thanks.

(this change is already in main - #1708)

PR #1759, opened 2022-06-10
Fix copy/paste error in docs

skovsende:
Looks like a copy/paste error in the docs for the yarp.ingress.kubernetes.io/backend-protocol annotation


MihaZupan (approved):
Thanks!

PR #1766, opened 2022-06-21
Initial draft of tunneling spec

samsp-msft:


docs/designs/yarp-tunneling.md
```diff
@@ -0,0 +1,158 @@
+# YARP Tunneling
+
+## Introduction
+While many organizations are moving their computing to the cloud, there are occasions where you need to be able to run some services in a local datacenter. The problem then is if you need to be able to communicate with those services from the cloud. Creating a VPN network connection to Azure or other cloud provider is possible, but usually involves a lot of red tape, and configuration.
+
+If all that the cloud needs to access is resources that are exposed over http, then a simpler solution is to have a gateway that will route traffic to the remote resource. Outbound connections over http are not usually blocked, so having a on-prem gateway make an outbound connection to the cloud, is the easiest way to establish the route.
+
+That is the principle of the gateway feature for YARP. You operate two instances of the YARP proxy service, configured as a tunnel. 
+
+[! Tunnel diagram](https://github.com/assets/95136/52d7491b-6e8a-4a2c-a51d-0734b3e41930)
```
gfoidl, line 10:
```suggestion
![Tunnel diagram](https://github.com/assets/95136/52d7491b-6e8a-4a2c-a51d-0734b3e41930)
```

docs/designs/yarp-tunneling.md
```diff
@@ -0,0 +1,165 @@
...
+builder.Services.AddTunnelServices();
+```
+
+The front end needs to have a tunnel endpoint that the backend will connect to. The endpoint should specify the name of the cluster as part of the URL, and a callback that is used to validate the connection is approved:
+ - Including the ClusterId in the URL enables the mechanism to be used for multiple clusters. It shouldn't be necessary for a single destination to be part of multiple clusters.
+ - Using a callback for authentication enables whatever scheme the proxy author(s) wish to use. The samples that we produce should be based around client certs as its the easiest way to have some form of secure shared secrets.
+
+``` C#
+app.MapReversProxy();
+app.MapTunnel(/tunnel/{ClusterId}, async (connectionContext, cluster) => {
```
davidfowl, line 54:
```suggestion
app.MapTunnel("/tunnel/{ClusterId}", async (connectionContext, cluster) => {
```

samsp-msft:
@blowdart & @GrabYourPitchforks - We'd appreciate an early review with respect to security.

jmezach:
Not entirely sure if this is the right place, but after reading this I really had one question: what does the configuration on the backend look like? There's an example of the frontend configuration, but not of the backend. Or maybe I'm misunderstanding how it works? Would adding a backend proxy as a cluster to the frontend simply expose any routes that the backend has from the frontend? I guess that's not entirely clear from the doc.

docs/designs/yarp-tunneling.md
```diff
@@ -0,0 +1,176 @@
...
+}
+```
+
+To ensure scalability, multiple BackEnd proxy instances should be able to create tunnel connections for the same cluster. When that happens, the load balancing rules for the cluster should apply, and balance between the active tunnels. Similarly multiple FrontEnds can be used to reduce the problems with a single point of failure.
+
+## BackEnd
+
+The BackEnd instance is the proxy that will reside on the same network as the resources that should be exposed. The BackEnd will need to be able to connect to those resources, and also be able to create a WebSocket connection to the FrontEnd proxy server(s) via whatever firewalls are between them.
+
+The BackEnd proxy is configured with routes and destinations that it wishes to expose to the front end. Security is maintained because only URLs matching its routes will be proxyable via it. This prevents attacks at the FrontEnd having arbitrary access to other resources on the BackEnd network - they need to be explicitly included in the BackEnd route table.
```
blowdart, line 113: _Security is maintained because only URLs matching its routes will be proxyable via it_

Addressing is arguable not a security feature, but at least this is opt-in. How are you validating routes?

docs/designs/yarp-tunneling.md
```diff
@@ -0,0 +1,176 @@
...
+var tunnelOptions = new TunnelOptions(){
+       TunnelClient = new SocketsHttpHandler(),
+       ClientCertificates = new X509CertificateCollection { cert }
+       };
+tunnelOptions.Headers.Add("MyJWTToken", tokenString);
+
+builder.WebHost.UseTunnelTransport(url, tunnelOptions);
+```
+
+**Issue** Do we need a callback for the tunnel for the client to validate the server, or do we rely on TLS and a valid server certificate that must match the URL used by the tunnel?
```
blowdart, line 133: Add a validation callback to allow for extra authentication/validation like cert pinning.

docs/designs/yarp-tunneling.md
```diff
@@ -0,0 +1,176 @@
...
+
+> Note: The FrontEnd proxy will not be aware of the actual destinations that serve resources - each BackEnd should have its own cluster definition for the actual destinations, and so can include multiple servers for any route/cluster combination.
+
+- A BackEnd proxy should be able to create tunnels to multiple FrontEnds. The tunnels can be to related FrontEnd proxies that are sharing the same load, or to FrontEnds in different cloud deployments. This enables the Front Ends to be very specific to particular deployments - and have constrained v-Lan configurations in the cloud. This limits the possibility for other connections to the proxy that may cause security issues.
+
+## Authentication
+
+The authentcation options for ASP.NET are diverse, and IT departments will likely have their own conditions on what is required to be able to secure a tunnel. So rather than trying to implement the combinatorial matrix of what customers could need, we should use a callback so that the proxy author can decide.
```
blowdart, line 148: _IT departments will likely have their own conditions on what is required to be able to secure a tunnel_

Your intro says that this is aimed at avoiding the red tape IT requirements bring, yet now you acknowledge them.

blowdart (commented):
As someone who used to work in corporate IT, thanks, I hate it.

Lots of comments inline.

docs/designs/yarp-tunneling.md
```diff
@@ -0,0 +1,176 @@
+# YARP Tunneling
+
+## Introduction
+While many organizations are moving their computing to the cloud, there are occasions where you need to be able to run some services in a local datacenter. The problem then is if you need to be able to communicate with those services from the cloud. Creating a VPN network connection to Azure or other cloud provider is possible, but usually involves a lot of red tape, and configuration.
```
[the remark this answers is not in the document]
samsp-msft: Revised a bit, but this is precisely what Azure Relay is supposed to solve. The difference here is that using a proxy as the gateway removes the need to update the backend servers, which is required with Relay.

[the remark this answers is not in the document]
blowdart: Which does mean these things aren't hidden from view, which is my concern here. This can be fixed by logging.

Tratcher (approved):

PR #1769, opened 2022-06-22
Add .NET 7.0 support

### LINKED ISSUE #770: Pass CancellationToken to HttpContent.ReadAsStreamAsync for .Net 5.0
A new overload for `HttpContent.ReadAsStreamAsync`  is introduced in .Net 5.0 which gets `CancellationToken`. Should it be used?

https://github.com/microsoft/reverse-proxy/blob/b734f195b55d0ba623caf88262342d9068402042/src/ReverseProxy/Service/Proxy/HttpProxy.cs#L481
https://github.com/microsoft/reverse-proxy/blob/b734f195b55d0ba623caf88262342d9068402042/src/ReverseProxy/Service/Proxy/HttpProxy.cs#L533

### LINKED ISSUE #1722: Remove Cookie workaround for .NET 7
https://github.com/microsoft/reverse-proxy/pull/451/files#diff-0a4324a1a2cfc979ba03a1cd37836a44d3184f221cf82a121cad58dac3423d3bR419

This has been fixed in Kestrel so the mitigation can be #if removed for .NET 7
https://github.com/dotnet/aspnetcore/pull/41591

### LINKED ISSUE #1727: Update TFM list in YARP 2.0
* .NET 5 is out of support. 
* Core 3.1 will be out of support close to when .NET 7 ships. 

So we should reconsider the TFM map, and any issues that may be easier if we don't need to support 3.1.

Tratcher:
- Fixes #1727
  - Handle some obsolete and nullable changes in 7.0
- Fixes #1722
- Fixes #770
- Also updated the 6.0 runtime version to 6.0.6


wfurt (approved):
LGTM - but I don't know much about yarp.
Lot of files to touch. This may be rare but I'm wondering of some top level file would be sufficient.

Issue #1770, opened 2022-06-22
WebSockets not closed gracefully


Tratcher:
### Describe the bug
After the client and server have exchanged graceful WS Close messages, YARP isn't gracefully closing the connection. After a few seconds the client times out and closes the connection. This causes the client (E.g. Chrome) to report a WS error (1006) rather than a graceful close (1000).

### To Reproduce
1. Create a WebSocket connection through YARP
2. Close it gracefully from both sides
3. Observe a multi-second pause and then the client disconnects and may report an error

This reproduces with YARP running on Kestrel, HttpSys, and IIS Express. The client can be chrome, edge or even HttpClient, though that doesn't report an error.

It also reproduces in our WebSocket functional tests, causing them to take 1s or longer, vs a few ms when I tell YARP to cancel the request read.
https://github.com/microsoft/reverse-proxy/blob/b9f43dd4b383e7c2e1b222c0dd6587a848d88a56/test/ReverseProxy.FunctionalTests/WebSocketTests.cs#L26

I think the problem is here:
https://github.com/microsoft/reverse-proxy/blob/b9f43dd4b383e7c2e1b222c0dd6587a848d88a56/src/ReverseProxy/Forwarder/HttpForwarder.cs#L593

After the client and destination have exchanged close messages the destination closes the connection. YARP sees the response copy complete, but then waits for the request side to be closed from the client. Instead it should cancel the ongoing read.

### Further technical details
Version: 1.1
Reported by an internal partner

Recommend we patch this in 1.1

PR #1773, opened 2022-06-23
Close WebSockets gracefully 

Tratcher:
#1770

The client expects the server to close the socket first. Kestrel will only close the underlying socket once YARP exits. YARP isn't exiting because it's still trying to read from the client. When a WebSocket response copy loop finishes gracefully YARP needs to cancel the request copy loop so we can quickly clean things up. Ignore any errors caused by this cancellation.

Merge #1774 first

### DIFF
--- a/samples/SampleServer/wwwroot/index.html
+++ b/samples/SampleServer/wwwroot/index.html
@@ -1,0 +1,151 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8" />
+    <title></title>
+    <style>
+        table { border: 0 }
+        .commslog-data { font-family: Consolas, Courier New, Courier, monospace; }
+        .commslog-server { background-color: red; color: white }
+        .commslog-client { background-color: green; color: white }
+    </style>
+</head>
+<body>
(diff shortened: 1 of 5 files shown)


samples/SampleServer/wwwroot/index.html
```diff
@@ -0,0 +1,151 @@
...
+        .commslog-data { font-family: Consolas, Courier New, Courier, monospace; }
+        .commslog-server { background-color: red; color: white }
+        .commslog-client { background-color: green; color: white }
+    </style>
+</head>
+<body>
+    <h1>WebSocket Test Page</h1>
+    <p id="stateLabel">Ready to connect...</p>
+    <div>
+        <label for="connectionUrl">WebSocket Server URL:</label>
```
greenEkatherine, line 17: I'm wondering, do we want to set up sample server with websockets by default? If so, please update https://github.com/microsoft/reverse-proxy/tree/main/samples/SampleServer#readme

greenEkatherine: discussed offline

greenEkatherine (approved):

PR #1790, opened 2022-07-01
Close WebSockets gracefully

### LINKED ISSUE #1770: WebSockets not closed gracefully

Tratcher:
Forward port of #1773 into main. Contributes to #1770.

TODO: I'm seeing some failures in WebSocketTelemetryTests that didn't show up in the original PR, but are now showing up locally in both 1.1 and main. 🤷 
- I partially understand now. Before a server graceful disconnect presented as a client timeout. Now YARP cancels the reads and suppresses the error, causing the telemetry to report Unknown.
- The telemetry can track reads and know when read returns 0, but it can't know when the write stream is closed unless it tries to write to it.

The test changes should be backported to 1.1.

### DIFF
--- a/samples/SampleServer/Controllers/WebSocketsController.cs
+++ b/samples/SampleServer/Controllers/WebSocketsController.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 using System;
@@ -36,7 +36,9 @@ namespace SampleServer.Controllers
         {
             if (!HttpContext.WebSockets.IsWebSocketRequest)
             {
-                HttpContext.Response.StatusCode = StatusCodes.Status400BadRequest;
+                HttpContext.Response.ContentType = "text/html";
+                await HttpContext.Response.SendFileAsync("./wwwroot/index.html");
+                return;
             }
 
             using (var webSocket = await HttpContext.WebSockets.AcceptWebSocketAsync())
--- a/samples/SampleServer/wwwroot/index.html
+++ b/samples/SampleServer/wwwroot/index.html
@@ -1,0 +1,151 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8" />
+    <title></title>
+    <style>
+        table { border: 0 }
+        .commslog-data { font-family: Consolas, Courier New, Courier, monospace; }
+        .commslog-server { background-color: red; color: white }
+        .commslog-client { background-color: green; color: white }
+    </style>
+</head>
+<body>
+    <h1>WebSocket Test Page</h1>
+    <p id="stateLabel">Ready to connect...</p>
+    <div>
+        <label for="connectionUrl">WebSocket Server URL:</label>
+        <input id="connectionUrl" />
+        <button id="connectButton" type="submit">Connect</button>
+    </div>
+    <div>
+        <label for="sendMessage">Message to send:</label>
+        <input id="sendMessage" disabled />
+        <button id="sendButton" type="submit" disabled>Send</button>
+        <button id="closeButton" disabled>Close Socket</button>
+    </div>
+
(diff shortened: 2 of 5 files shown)


test/ReverseProxy.FunctionalTests/WebSocketsTelemetryTests.cs
```diff
@@ -90,15 +90,15 @@ public enum Behavior
...
     [InlineData(Behavior.ClosesConnection, Behavior.SendsClose_ClosesConnection, WebSocketCloseReason.ServerGracefulClose, WebSocketCloseReason.ClientDisconnect)]
     [InlineData(Behavior.ClosesConnection, Behavior.SendsClose_WaitsForClose, WebSocketCloseReason.ServerGracefulClose, WebSocketCloseReason.ClientDisconnect)]
     // One side closes the connection while the other is waiting for messages
     [InlineData(Behavior.ClosesConnection, Behavior.WaitsForClose_SendsClose, WebSocketCloseReason.ClientDisconnect)]
     [InlineData(Behavior.ClosesConnection, Behavior.WaitsForClose_ClosesConnection, WebSocketCloseReason.ClientDisconnect)]
-    [InlineData(Behavior.WaitsForClose_SendsClose, Behavior.ClosesConnection, WebSocketCloseReason.ServerDisconnect)]
-    [InlineData(Behavior.WaitsForClose_ClosesConnection, Behavior.ClosesConnection, WebSocketCloseReason.ServerDisconnect)]
+    [InlineData(Behavior.WaitsForClose_SendsClose, Behavior.ClosesConnection, WebSocketCloseReason.ServerDisconnect, WebSocketCloseReason.Unknown)]
+    [InlineData(Behavior.WaitsForClose_ClosesConnection, Behavior.ClosesConnection, WebSocketCloseReason.ServerDisconnect, WebSocketCloseReason.Unknown)]
```
MihaZupan, line 101: We should fix this case in `WebSocketsTelemetryStream` instead of changing the test.
These `Unknown` results should be `ServerDisconnect`s - this is an observable product bug in telemetry otherwise.

From offline discussion:

> Before if the destination disconnected while the client was still sending data (at least keep-alives), YARP would blow up when trying to write to the destination, and we would assign the blame to the destination.


> Now you have a race where if YARP is fast enough, it will set cancelReads = true and swallow the destination write failure exception. In this case the telemetry has no idea what's going on since neither side sent a WS close, and there is no forwarder error either.

> What may be happening in CI is that the machines are just slower so the write failure happens to be observed before we decide to cancel reads.


> We could change [this case of GetCloseReason](https://github.com/microsoft/reverse-proxy/blob/9d612d4c34f390a33bc5bdaa5ec1bac6f8251015/src/ReverseProxy/WebSocketsTelemetry/WebSocketsTelemetryStream.cs#L62-L64) to `ServerDisconnect` since it should really only be happening in this new situation where the server disconnected and we swallowed the error.

MihaZupan (approved):

Issue #1799, opened 2022-07-09
Raise an Error in `ConfigureHttpClient` when custom `IForwarderHttpClientFactory` is added


Kahbazi:
### What should we add or change to make your life better?
Throw an exception in `ConfigureHttpClient` when custom `IForwarderHttpClientFactory` is added stating that "You couldn't call `ConfigureHttpClient` when another type is added for `IForwarderHttpClientFactory`. "
### Why is this important to you?
I had `ConfigureHttpClient` in my code to configure `SslOptions` and lately I added a custom `IForwarderHttpClientFactory`. Now this wouldn't work because `ConfigureHttpClient` would add its own `IForwarderHttpClientFactory` and ignore the `DiagnosticsHandlerFactory`. I couldn't figure out why `DiagnosticsHandlerFactory` wasn't working until I checked the source code.

I suggest to throw an exception in `ConfigureHttpClient` if any other implementation of `IForwarderHttpClientFactory` beside `ForwarderHttpClientFactory` is added, in order to reduce the debugging time for others.

Btw this is my code:
```csharp
services.AddSingleton<IForwarderHttpClientFactory, DiagnosticsHandlerFactory>();
services.AddReverseProxy()
    .ConfigureHttpClient((context, handler) =>
    {
        handler.SslOptions = ...
    }
```

PR #1805, opened 2022-07-15
Throw an exception in ConfigureHttpClient when custom IForwarderHttpClientFactory is added

Kahbazi:
Fixes https://github.com/microsoft/reverse-proxy/issues/1799

### DIFF
--- a/src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
+++ b/src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Linq;
 using System.Net.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection.Extensions;
@@ -141,6 +142,15 @@ public static class ReverseProxyServiceCollectionExtensions
         if (configure is null)
         {
             throw new ArgumentNullException(nameof(configure));
+        }
+
+        var service = builder.Services.FirstOrDefault(service => service.ServiceType == typeof(IForwarderHttpClientFactory));
+        if (service is not null)
+        {
+            if (service.ImplementationType != typeof(ForwarderHttpClientFactory))
+            {
+                throw new InvalidOperationException($"ConfigureHttpClient will override the custom IForwarderHttpClientFactory type.");
+            }
         }
 
         builder.Services.AddSingleton<IForwarderHttpClientFactory>(services =>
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -143,6 +144,15 @@ public static IReverseProxyBuilder ConfigureHttpClient(this IReverseProxyBuilder
             throw new ArgumentNullException(nameof(configure));
         }
 
+        var service = builder.Services.FirstOrDefault(service => service.ServiceType == typeof(IForwarderHttpClientFactory));
```
Tratcher, line 147: This only works if your custom service is added first. You'd have the opposite problem if you added it after. Do you have proposals to mitigate that?

MihaZupan: What if we accept an `IEnumrable<IForwarderHttpClientFactory>` in `ProxyConfigManager` and throw if there's more than 1?

Kahbazi: > This only works if your custom service is added first. You'd have the opposite problem if you added it after. Do you have proposals to mitigate that?

Yeah, you're right. But it's better than nothing. For this to work if the service is added after, we could implement our own `IServiceCollection` which would be a wrapper to the default one and assign it to `IReverseProxyBuilder`, but I don't like this idea that much. It's way too complicated for this small problem and one could always add the service to the original `IServiceCollection`.

Maybe an Analyzer could be useful too.

> What if we accept an IEnumrable<IForwarderHttpClientFactory> in ProxyConfigManager and throw if there's more than 1?

This wouldn't work, because this code would have add two `IForwarderHttpClientFactory` already. Perhaps `ConfigureHttpClient` could remove the `IForwarderHttpClientFactory` service if it's the default. 
```csharp
ServiceCollection()
            .AddReverseProxy()
            .ConfigureHttpClient((_, _) => { });
```
But it feels weird to check it on `ProxyConfigManager` because it doesn't seems to be this class responsibility.

MihaZupan: > > What if we accept an IEnumrable in ProxyConfigManager and throw if there's more than 1?

> This wouldn't work, because this code would have add two IForwarderHttpClientFactory already.

How about something like this https://github.com/microsoft/reverse-proxy/commit/c1f6b5048a2ab6e88b94eaa41564f6cb4e616c30?

Tratcher: Triage: This is simple enough and provides a usability improvement. We'll take it as is, plus a comment acknowledging the gap.
```suggestion
        // Avoid overriding any other custom factories. This does not handle the case where a IForwarderHttpClientFactory
        // is registered after this call.
        var service = builder.Services.FirstOrDefault(service => service.ServiceType == typeof(IForwarderHttpClientFactory));
```

Tratcher (approved):

Tratcher:
Thanks @Kahbazi

Issue #1697, opened 2022-05-06
Add NotExists to HeaderMatchMode enum


mak0t0san:
### What should we add or change to make your life better?
Would like the ability to match only when a particular header is not present. This could be done by adding a `NotExists` value to `HeaderMatchMode` and implement code to handle it.

### Why is this important to you?
In my particular scenario we have an API application that runs in multiple types of environments, such as containerized environments as well as a dedicate virtual machine.  There are a couple of endpoints that don't work in the containerized environment so we use YARP to reverse-proxy those requests to the VM environment.  We've also had to add some extra deployment steps so that we can set different configuration values for the containerized version and the VM version. I would like to avoid having to setup extra deployment steps by having YARP add a header so that servers downstream can know that it's being forwarded the request.  Simply checking for the presence of a `X-Forwarded*` header is not sufficient as the servers upstream are behind load balancers that already get this header.

Here's what I want to do:

- Load-Balancer gets request, forwards to containerized environment
- Containerized environment receives request.  YARP configuration will reverse proxy some endpoints if a custom header does NOT exist and will transform and add a custom header
- VM environment receives request from YARP with new custom header added. This has the same exact files, including YARP configuration as the upstream servers.  However, the YARP rule will only match if the new header does NOT exist but the upstream server added the header so the YARP rule won't match and the code in the app will handle the request.

karelz:
Triage: Not high demand so far (please upvote top post if you want it) - moving to Backlog for now.
It should be fairly straightforward and we would take a contribution.

kevbry:
@karelz I needed this as well - PR submitted.

MihaZupan:
Added in #1806

MihaZupan:
Reopening to track adding docs

PR #1806, opened 2022-07-15
Add HeaderMatchMode.NotExists, enable matchers that only match if the…

kevbry:
Addresses #1697 by adding HeaderMatchMode.NotExists. Inverse of HeaderMatchMode.Exists, being that it only matches if the header is not present at all.

Have added a few tests, but had trouble deciphering the intent behind some of the tests in HeaderMatcherPolicyTests.cs and so may have missed some cases. Particularly the sortorder and "_AppliesScenarios" tests were not added to as I couldn't figure out what the intent of them was.

One failing test Yarp.ReverseProxy.FunctionalTests.HttpProxyCookieTests_Http2::ProxyAsync_RequestWithCookieHeaders is also failing in main, so I don't think I broke it.

Thanks for a great piece of software.

### DIFF
--- a/src/ReverseProxy/Configuration/HeaderMatchMode.cs
+++ b/src/ReverseProxy/Configuration/HeaderMatchMode.cs
@@ -44,7 +44,7 @@ public enum HeaderMatchMode
     Exists,
 
     /// <summary>
-    /// The header must not exist
+    /// The header must not exist.
     /// </summary>
-    NotExists
+    NotExists,
 }
--- a/src/ReverseProxy/Configuration/HeaderMatchMode.cs
+++ b/src/ReverseProxy/Configuration/HeaderMatchMode.cs
@@ -42,4 +42,9 @@ public enum HeaderMatchMode
     /// The header must exist and contain any non-empty value.
     /// </summary>
     Exists,
+
+    /// <summary>
+    /// The header must not exist
+    /// </summary>
+    NotExists
 }
--- a/src/ReverseProxy/Routing/HeaderMatcherPolicy.cs
+++ b/src/ReverseProxy/Routing/HeaderMatcherPolicy.cs
@@ -70,12 +70,19 @@ internal sealed class HeaderMatcherPolicy : MatcherPolicy, IEndpointComparerPoli
 
             foreach (var matcher in matchers)
             {
-                if (headers.TryGetValue(matcher.Name, out var requestHeaderValues) &&
+                var headerExistsInRequest = headers.TryGetValue(matcher.Name, out var requestHeaderValues);
+                if ( headerExistsInRequest &&
                     !StringValues.IsNullOrEmpty(requestHeaderValues))
                 {
                     if (matcher.Mode is HeaderMatchMode.Exists)
                     {
                         continue;
+                    }
+
+                    if (matcher.Mode is HeaderMatchMode.NotExists)
+                    {
+                        candidates.SetValidity(i, false);
+                        break;
                     }
 
                     if (matcher.Mode is HeaderMatchMode.ExactHeader or HeaderMatchMode.HeaderPrefix
@@ -84,6 +91,10 @@ internal sealed class HeaderMatcherPolicy : MatcherPolicy, IEndpointComparerPoli
                     {
                         continue;
                     }
+                }
+                else if ( matcher.Mode is HeaderMatchMode.NotExists && !headerExistsInRequest)
+                {
+                    continue;
                 }
 
                 candidates.SetValidity(i, false);
--- a/src/ReverseProxy/Routing/HeaderMatcherPolicy.cs
+++ b/src/ReverseProxy/Routing/HeaderMatcherPolicy.cs
@@ -71,8 +71,7 @@ internal sealed class HeaderMatcherPolicy : MatcherPolicy, IEndpointComparerPoli
             foreach (var matcher in matchers)
             {
                 var headerExistsInRequest = headers.TryGetValue(matcher.Name, out var requestHeaderValues);
-                if ( headerExistsInRequest &&
-                    !StringValues.IsNullOrEmpty(requestHeaderValues))
+                if (headerExistsInRequest && !StringValues.IsNullOrEmpty(requestHeaderValues))
                 {
                     if (matcher.Mode is HeaderMatchMode.Exists)
                     {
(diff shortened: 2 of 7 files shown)


src/ReverseProxy/Routing/HeaderMatcherPolicy.cs
```diff
@@ -70,21 +70,32 @@ public Task ApplyAsync(HttpContext httpContext, CandidateSet candidates)
 
             foreach (var matcher in matchers)
             {
-                if (headers.TryGetValue(matcher.Name, out var requestHeaderValues) &&
+                var headerExistsInRequest = headers.TryGetValue(matcher.Name, out var requestHeaderValues);
+                if ( headerExistsInRequest &&
                     !StringValues.IsNullOrEmpty(requestHeaderValues))
```
MihaZupan, line 75:
```suggestion
                if (headerExistsInRequest && !StringValues.IsNullOrEmpty(requestHeaderValues))
```

src/ReverseProxy/Routing/HeaderMatcherPolicy.cs
```diff
@@ -70,21 +70,32 @@ public Task ApplyAsync(HttpContext httpContext, CandidateSet candidates)
...
+                    }
+
                     if (matcher.Mode is HeaderMatchMode.ExactHeader or HeaderMatchMode.HeaderPrefix
                         ? TryMatchExactOrPrefix(matcher, requestHeaderValues)
                         : TryMatchContainsOrNotContains(matcher, requestHeaderValues))
                     {
                         continue;
                     }
                 }
+                else if ( matcher.Mode is HeaderMatchMode.NotExists && !headerExistsInRequest)
```
MihaZupan, line 95:
```suggestion
                else if (matcher.Mode is HeaderMatchMode.NotExists && !headerExistsInRequest)
```

test/ReverseProxy.Tests/Routing/HeaderMatcherPolicyTests.cs
```diff
@@ -150,18 +150,21 @@ public void AppliesToEndpoints_NoMetadata_DoesNotApply()
...
-    [InlineData(null, false)]
-    [InlineData("", false)]
-    [InlineData("abc", true)]
-    public async Task ApplyAsync_MatchingScenarios_AnyHeaderValue(string incomingHeaderValue, bool shouldMatch)
+    [InlineData(null, HeaderMatchMode.Exists, false)]
+    [InlineData("", HeaderMatchMode.Exists, false)]
+    [InlineData("abc", HeaderMatchMode.Exists, true)]
+    [InlineData(null, HeaderMatchMode.NotExists,true)]
+    [InlineData("", HeaderMatchMode.NotExists, false)]
+    [InlineData("abc", HeaderMatchMode.NotExists,false)]
```
MihaZupan, line 158:
```suggestion
    [InlineData(null, HeaderMatchMode.NotExists, true)]
    [InlineData("", HeaderMatchMode.NotExists, false)]
    [InlineData("abc", HeaderMatchMode.NotExists, false)]
```

MihaZupan (approved):
Thank you for your contribution!

> Particularly the sortorder and "_AppliesScenarios" tests were not added to as I couldn't figure out what the intent of them was.

No worries, your change does not impact these tests.
They only care about the number of matchers, not their type.

src/ReverseProxy/Configuration/HeaderMatchMode.cs
```diff
@@ -42,4 +42,9 @@ public enum HeaderMatchMode
     /// The header must exist and contain any non-empty value.
     /// </summary>
     Exists,
+
+    /// <summary>
+    /// The header must not exist
```
MihaZupan, line 47:
```suggestion
    /// The header must not exist.
```

src/ReverseProxy/Configuration/HeaderMatchMode.cs
```diff
@@ -42,4 +42,9 @@ public enum HeaderMatchMode
...
+    /// </summary>
+    NotExists
```
MihaZupan, line 49:
```suggestion
    /// The header must not exist.
    /// </summary>
    NotExists,
```

src/ReverseProxy/Routing/HeaderMatcherPolicy.cs
```diff
@@ -70,21 +70,31 @@ public Task ApplyAsync(HttpContext httpContext, CandidateSet candidates)
 
             foreach (var matcher in matchers)
             {
-                if (headers.TryGetValue(matcher.Name, out var requestHeaderValues) &&
-                    !StringValues.IsNullOrEmpty(requestHeaderValues))
+                var headerExistsInRequest = headers.TryGetValue(matcher.Name, out var requestHeaderValues);
```
Tratcher, line 73: headerExistsInRequest should include !StringValues.IsNullOrEmpty. An empty header has no semantic meaning in almost all cases.

MihaZupan: I would be surprised if I set `NotExists` and it still matched on a request that does have the header (albeit empty).

Tratcher: I guess that the difference between literally empty vs semantically empty. I doubt it will make a difference in practice. Nevermind.

kevbry: > I would be surprised if I set NotExists and it still matched on a request that does have the header (albeit empty).

That was my thinking in breaking the two portions apart. Also the hope of using this to help glue together weird legacy software that does things like send empty Accept header to a service with an incompatible default.

kevbry: Thinking about it a little further, I don't see a way to express "match only if the specified header is present, but empty". My use-case didn't need that, and I assumed that if I did at some point I could just use ExactHeader or PrefixHeader to an empty string. Looks like that existing IsNullOrEmpty check would prevent those from matching in that scenario.

I doubt it matters, just a note.

Kahbazi:
@MihaZupan @Tratcher [The docs](https://github.com/microsoft/reverse-proxy/blob/90a10856e558060ffb85499a7a30ed5a3a2b4786/docs/docfx/articles/header-routing.md) needs to be updated.

Issue #1762, opened 2022-06-16
Filter Content-Length: 0 from responses that don't allow it


Tratcher:
Some status codes like 101 do not allow a response body, or even declaring `Content-Length: 0`. Kestrel will throw in these scenarios.
```
System.InvalidOperationException: Response Content-Length mismatch: too many bytes written (90 of 0).
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.VerifyAndUpdateWrite(Int32 count)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.WritePipeAsync(ReadOnlyMemory`1 data, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseStream.WriteAsync(ReadOnlyMemory`1 source, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpUpgradeStream.WriteAsync(ReadOnlyMemory`1 source, CancellationToken cancellationToken)
   at Yarp.ReverseProxy.Forwarder.StreamCopier.CopyAsync(Stream input, Stream output, Int64 promisedContentLength, StreamCopierTelemetry telemetry, ActivityCancellationTokenSource activityToken, CancellationToken cancellation)
```

YARP should remove `Content-Length: 0` from such responses to mitigate misbehaving destination servers.

Also check 204, etc.

See https://github.com/dotnet/aspnetcore/issues/40882#issuecomment-1082183455

karelz:
At least one customer who hit it was with Python.

PR #1813, opened 2022-07-27
Remove Content-Length: 0 on responses that don't allow bodies

adityamandaleeka:
Fix https://github.com/microsoft/reverse-proxy/issues/1762

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpTransformer.cs
+++ b/src/ReverseProxy/Forwarder/HttpTransformer.cs
@@ -41,8 +41,7 @@ public class HttpTransformer
             // A 1xx response is terminated by the end of the header section; it cannot contain content
             // or trailers.
             // See https://www.rfc-editor.org/rfc/rfc9110.html#section-15.2-2
-            HttpStatusCode.Continue => true,
-            HttpStatusCode.SwitchingProtocols => true,
+            >= HttpStatusCode.Continue and < HttpStatusCode.OK => true,
             // A 204 response is terminated by the end of the header section; it cannot contain content
             // or trailers.
             // See https://www.rfc-editor.org/rfc/rfc9110.html#section-15.3.5-5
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Forwarder/HttpTransformer.cs
```diff
@@ -32,6 +34,23 @@ public class HttpTransformer
     /// </summary>
     protected HttpTransformer() { }
 
+    // Status codes that don't include a response body.
+    private static readonly HttpStatusCode[] BodylessStatusCodes = {
```
Kahbazi, line 38: How about using a HashSet instead of an Array? I think it might be better in performance when checking if status exists in the list.

gfoidl: I'd go with plain `int`-comparisons for max-perf here. Something like
```c#
private static bool IsBodyless(HttpStatusCode statusCode) => statusCode switch
{
    < HttpStatusCode.OK => true,
    HttpStatusCode.NoContent or HttpStatusCode.ResetContent => true,
    _ => false
};
```
as this method would be used in only one place, maybe it's also worth to put `[MethodImpl(MethodImplOptions.AggressiveInlining)]` on it.
Could also be a local function.

adityamandaleeka: Yea that's better. Updated.

src/ReverseProxy/Forwarder/HttpTransformer.cs
```diff
@@ -32,6 +34,26 @@ public class HttpTransformer
     /// </summary>
     protected HttpTransformer() { }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static bool IsBodylessStatusCode(HttpStatusCode statusCode) =>
+        statusCode switch
+        {
+            // A 1xx response is terminated by the end of the header section; it cannot contain content
```
MihaZupan, line 41: Shouldn't this check for the whole 100-199 range?

adityamandaleeka: Done.

MihaZupan (approved):

PR #1814, opened 2022-07-29
Use MediaTypeHeaderValue.MatchesMediaType for content type check

adityamandaleeka:
Noticed this TODO item while looking through the sources. The mentioned aspnetcore issue was addressed nearly two years ago :smile:

Benchmarking indicates that the perf difference here is negligible, but this is definitely cleaner from a maintenance standpoint.

### DIFF
--- a/src/ReverseProxy/Forwarder/ProtocolHelper.cs
+++ b/src/ReverseProxy/Forwarder/ProtocolHelper.cs
@@ -33,6 +33,8 @@ internal static class ProtocolHelper
     /// Checks whether the provided content type header value represents a gRPC request.
     /// </summary>
     public static bool IsGrpcContentType(string? contentType) =>
-        MediaTypeHeaderValue.TryParse(contentType, out var mediaType)
+        contentType is not null &&
+        && contentType.StartsWith(GrpcContentType, StringComparison.OrdinalIgnoreCase)
+        && MediaTypeHeaderValue.TryParse(contentType, out var mediaType)
         && mediaType.MatchesMediaType(GrpcContentType);
 }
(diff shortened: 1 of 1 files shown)


src/ReverseProxy/Forwarder/ProtocolHelper.cs
```diff
@@ -28,43 +29,10 @@ public static string GetVersionPolicy(HttpVersionPolicy policy)
...
-        {
-            // Accept any message format. Marshaller could be set to support third-party formats
-            return true;
-        }
-
-        return false;
-    }
+    public static bool IsGrpcContentType(string? contentType) =>
+        MediaTypeHeaderValue.TryParse(contentType, out var mediaType)
+        && mediaType.MatchesMediaType(GrpcContentType);
```
MihaZupan, line 37: Can we keep the `StartsWith` to avoid allocating the `MediaTypeHeaderValue` in the common case? 😃

```suggestion
        contentType is not null &&
        && contentType.StartsWith(GrpcContentType, StringComparison.OrdinalIgnoreCase)
        && MediaTypeHeaderValue.TryParse(contentType, out var mediaType)
        && mediaType.MatchesMediaType(GrpcContentType);
```

MihaZupan (approved):

src/ReverseProxy/Forwarder/ProtocolHelper.cs
```diff
@@ -28,43 +29,12 @@ public static string GetVersionPolicy(HttpVersionPolicy policy)
...
-        if (nextChar == '+')
-        {
-            // Accept any message format. Marshaller could be set to support third-party formats
-            return true;
-        }
-
-        return false;
-    }
+    public static bool IsGrpcContentType(string? contentType) =>
+        contentType is not null &&
```
MihaZupan, line 36:
```suggestion
        contentType is not null
```
Oops

PR #1831, opened 2022-08-17
Throw on duplicate Route IDs

### LINKED ISSUE #1830: Add a check for duplicate route IDs in provider(s)
Hi all,

If you are loading from multiple configuration sources and have a duplicate route ID in one of your configs, YARP is going to take the last one which is loaded and overwrite the first route ID.

I would like to add a validation check to inform the user that he has a duplicate route ID.

I checked YARP documentation and closest I found is **IProxyConfigFilter**, which partially does the job. However, it becomes tricky when configurations are reloaded. 

Also, looked at IProxyStateLookup and IProxyConfigProvider which has IProxyConfig and IChangeToken, but so far could not find an elegant solution.

I wonder, what would be the best way to achieve what I am trying to do? 

Many thanks in advance

Update:

have been looking more at a code, and I think it's might be a bug there, since looks like it suppose to check for duplicate routes but **seenRouteIds** collection is never updated

image

MihaZupan:
Fixes #1830

### DIFF
--- a/src/ReverseProxy/Management/ProxyConfigManager.cs
+++ b/src/ReverseProxy/Management/ProxyConfigManager.cs
@@ -321,7 +321,7 @@ internal sealed class ProxyConfigManager : EndpointDataSource, IProxyStateLookup
             return (Array.Empty<RouteConfig>(), Array.Empty<Exception>());
         }
 
-        var seenRouteIds = new HashSet<string>();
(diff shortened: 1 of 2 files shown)


Tratcher (approved):
Hmm, when did that break...

MihaZupan:
Back in PR #1 😆

Issue #1678, opened 2022-04-28
Consider setting a finite ConnectTimeout by default


MihaZupan:
Starting with 6.0, the consequences of a connection attempt that takes a long time to fail, or somehow becomes stuck are much higher than before (memory leaks, more failed requests ...).

In .NET 7, we made [a change in the connection pool](https://github.com/dotnet/runtime/pull/71785) where we will now cancel a connection attempt 5 seconds after the request that initiated it completes (or is canceled).

The main mitigation for this in 6.0 is to set `SocketsHttpHandler.ConnectTimeout` to a finite value.  We should consider setting such a default in YARP.
My preference would be that we make the change regardless of target framework (also change it for 7.0).

PR #1833, opened 2022-08-17
Bug fix for connection leak in SocketsHttpHandler

mahpate:
Setting SocketsHttpHandler.ConnectTimeout  to 15 seconds as default value in .Net 7.
See https://github.com/dotnet/runtime/issues/66673#issuecomment-1073811798

### DIFF
--- a/src/ReverseProxy/Forwarder/ForwarderHttpClientFactory.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderHttpClientFactory.cs
@@ -48,7 +48,7 @@ public class ForwarderHttpClientFactory : IForwarderHttpClientFactory
             AutomaticDecompression = DecompressionMethods.None,
             UseCookies = false,
             ActivityHeadersPropagator = new ReverseProxyPropagator(DistributedContextPropagator.Current),
-
+            ConnectTimeout = TimeSpan.FromSeconds(15), // See: https://github.com/dotnet/runtime/issues/66673#issuecomment-1073811798
             // NOTE: MaxResponseHeadersLength = 64, which means up to 64 KB of headers are allowed by default as of .NET Core 3.1.
         };
 


Tratcher:
@Miha, the change to SocketsHttpHandler's ConnectTimeout default was reverted, correct?

MihaZupan:
Yes, that change was reverted. The default `ConnectTimeout` stayed the same (infinite).

We went a different route in runtime with https://github.com/dotnet/runtime/pull/71785. A connection attempt will now be canceled 5 seconds after the request that started it is completed/canceled.
In practice, this should have a similar effect to `ConnectTimeout`, but without impacting the exception type surfaced to requests (a stalled connection attempt won't show up as a `ConnectTimeout` by default).

The issue mentioned in the linked comment about wasting connection attempts on canceled requests is also no longer a problem as we backported mitigations for that to 6.0 (https://github.com/dotnet/runtime/issues/66990).

---

That said, the reason the change to 15 seconds was reverted is not that the change is bad in itself.
We reverted it because users were expecting a different kind of exception (from the Sockets layer) and started getting `ConnectTimeout` exceptions instead. This broke code that was doing retries based on the kind of exception.

YARP is in a different situation here given that we are the ones calling into `HttpClient` to make requests. While someone could have implemented retry logic in a similar way, it's much less likely.

I would be okay with changing the default to 15s in YARP, but I don't think we should be linking to that comment as justification. The reason for doing so is not to avoid "connection leaks", but to avoid potential memory leaks/higher request failure rates.

MihaZupan:
For reference, this is an issue tracking making such changes in YARP: #1678

src/ReverseProxy/Forwarder/ForwarderHttpClientFactory.cs
```diff
@@ -48,7 +48,7 @@ public HttpMessageInvoker CreateClient(ForwarderHttpClientContext context)
             AutomaticDecompression = DecompressionMethods.None,
             UseCookies = false,
             ActivityHeadersPropagator = new ReverseProxyPropagator(DistributedContextPropagator.Current),
-
+            ConnectTimeout = TimeSpan.FromSeconds(15), // See: https://github.com/dotnet/runtime/issues/66673#issuecomment-1073811798
```
MihaZupan, line 51:
```suggestion
            ConnectTimeout = TimeSpan.FromSeconds(15),
```

MihaZupan (commented):
@mahpate can you please also update this property in other places where we specify these defaults:
- [Direct forwarding sample](https://github.com/microsoft/reverse-proxy/blob/main/samples/ReverseProxy.Direct.Sample/Startup.cs#L43)
- [HttpClient docs](https://github.com/microsoft/reverse-proxy/blob/9253842eea4b702489c9e4559a98f9d04d52c853/docs/docfx/articles/http-client-config.md?plain=1#L212)
- [Direct forwarding docs](https://github.com/microsoft/reverse-proxy/blob/9253842eea4b702489c9e4559a98f9d04d52c853/docs/docfx/articles/direct-forwarding.md?plain=1#L55)

MihaZupan:
@mahpate are you still interested in making this change?

MihaZupan:
@mahpate are you able to apply the recommended fixes here?

Issue #1840, opened 2022-08-22
FunctionalTests not running on command line build or CI


Tratcher:
The functional test project https://github.com/microsoft/reverse-proxy/blob/main/test/ReverseProxy.FunctionalTests/Yarp.ReverseProxy.FunctionalTests.csproj isn't running when testing from the command line or CI. It's built but not executed.

I noticed because Yarp.ReverseProxy.HttpProxyCookieTests_Http2.ProxyAsync_RequestWithCookieHeaders started failing in VS after a recent SDK update. We need to react to a aspnetcore bugfix.

adityamandaleeka:
> The functional test project https://github.com/microsoft/reverse-proxy/blob/main/test/ReverseProxy.FunctionalTests/Yarp.ReverseProxy.FunctionalTests.csproj isn't running when testing from the command line or CI. It's built but not executed.

`dotnet test` in `test/ReverseProxy.FunctionalTests` seems to run them just fine for me. Are you using a different workflow?

adityamandaleeka:
Oh, I see. `build.[cmd|sh] -test` doesn't discover them.

adityamandaleeka:
lol
 https://github.com/dotnet/arcade/blob/e7ede87875f41a9b3df898ae08da5ebc96e24f56/src/Microsoft.DotNet.Arcade.Sdk/tools/Tests.props#L17

PR #1841, opened 2022-08-23
Treat FunctionalTests project as tests

adityamandaleeka:
Fix https://github.com/microsoft/reverse-proxy/issues/1840

### DIFF
--- a/test/ReverseProxy.FunctionalTests/HttpProxyCookieTests.cs
+++ b/test/ReverseProxy.FunctionalTests/HttpProxyCookieTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Net;
 using System.Net.Http;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
@@ -114,9 +115,13 @@ public class HttpProxyCookieTests_Http1 : HttpProxyCookieTests
     }
 }
 
+[ConditionalClass(typeof(HttpProxyCookieTests_Http2), nameof(HttpProxyCookieTests_Http2.IsTestSupported))]
 public class HttpProxyCookieTests_Http2 : HttpProxyCookieTests
 {
     public override HttpProtocols HttpProtocol => HttpProtocols.Http2;
+
+    // HTTP/2 versions of these tests doesn't work on macOS.
+    public static bool IsTestSupported => !RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
 
     // HttpClient for H/2 will use different header frames for cookies from a container and message headers.
     // It will first send message header cookie and than the container one and we expect them in the order of cookieA;cookieB.
--- a/test/ReverseProxy.FunctionalTests/HttpProxyCookieTests.cs
+++ b/test/ReverseProxy.FunctionalTests/HttpProxyCookieTests.cs
@@ -30,12 +30,7 @@ public abstract class HttpProxyCookieTests
     public abstract HttpProtocols HttpProtocol { get; }
     public abstract Task ProcessHttpRequest(Uri proxyHostUri);
 
-    // Disabled on macOS because the HTTP/2 version of this test doesn't
-    // work on macOS. This should be refactored to re-enable on macOS for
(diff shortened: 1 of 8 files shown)


adityamandaleeka:
Good, the tests are running and failing (as expected until they are fixed).

test/ReverseProxy.FunctionalTests/Yarp.ReverseProxy.FunctionalTests.csproj
```diff
@@ -4,6 +4,12 @@
     <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
     <OutputType>Library</OutputType>
     <RootNamespace>Yarp.ReverseProxy</RootNamespace>
+
+    <!--
+      Arcade test detection looks for known suffixes like '.Tests' or '.UnitTests'
+      so we need to explicitly specify that this is a test project.
+      -->
+    <IsUnitTestProject>true</IsUnitTestProject>
```
Tratcher, line 12: Could this be updated centrally in the props file?
https://github.com/microsoft/reverse-proxy/blob/be3ae1da5cc6bb840b2a78f463e2957300d24c7a/test/Directory.Build.props#L7

adityamandaleeka: I suppose we could, though it would be annoying if we want to add things in here that aren't meant to run every CI (e.g. long running tests, performance tests, etc.).

Tratcher:
Now to fix the test 😁

Tratcher:
@MihaZupan can you take a look at the flaky WebSocketsTelemetry test? It passes for me sometimes locally.

MihaZupan:
@Tratcher is that the same WebSocket test as in #1790 (I can't get it to fail locally)?

From our offline conversation, I believe that test could be flaky without the fix.
> Before if the destination disconnected while the client was still sending data (at least keep-alives), YARP would blow up when trying to write to the destination, and we would assign the blame to the destination.

> Now you have a race where if YARP is fast enough, it will set `cancelReads = true` and swallow the destination write failure exception. In this case the telemetry has no idea what's going on since neither side sent a WS close, and there is no forwarder error either.
> What may be happening in CI is that the machines are just slower so the write failure happens to be observed before we decide to cancel reads.

> We could change [this case of GetCloseReason](https://github.com/microsoft/reverse-proxy/blob/9d612d4c34f390a33bc5bdaa5ec1bac6f8251015/src/ReverseProxy/WebSocketsTelemetry/WebSocketsTelemetryStream.cs#L62-L64) to ServerDisconnect since it should really only be happening in this new situation where the server disconnected and we swallowed the error.

adityamandaleeka:
We had a few tests of the newly enabled tests failing on macOS since HTTP/2 over TLS isn't supported there.

Turns out we didn't have a straightforward way to disable tests on certain OS's in this repo yet. The `ConditionalTheory` attribute addresses this (another approach would be upgrading to xUnit 2.4.2 which has `Assert.SkipWhen`). I did have to bring in a new reference to Microsoft.DotNet.XUnitExtensions to make the `ConditionalTheory` work though.

I'll seek feedback from the Arcade folks about whether this is the right thing to do. We could also just roll our own attribute and throw it in the common test utilities if this ends up being too complicated, but I wanted to first try using the shared components.

test/ReverseProxy.FunctionalTests/HttpProxyCookieTests.cs
```diff
@@ -29,7 +30,7 @@ public abstract class HttpProxyCookieTests
     public abstract HttpProtocols HttpProtocol { get; }
     public abstract Task ProcessHttpRequest(Uri proxyHostUri);
 
-    [Fact]
+    // [Fact] // Disabled. See https://github.com/microsoft/reverse-proxy/issues/1884
```
Tratcher, line 33:
```suggestion
    [Fact(Skip = "https://github.com/microsoft/reverse-proxy/issues/1884" )]
```

Tratcher (approved):

Tratcher:
Fixes https://github.com/microsoft/reverse-proxy/issues/1884

Issue #1776, opened 2022-06-25
Set Rate limiter schema for Routes


Kahbazi:
### What should we add or change to make your life better?
The new RateLimiter middleware per endpoint is currently [in draft](https://github.com/dotnet/aspnetcore/pull/42417/files) and it is worth being able to set different policy for YARP routes for it when it's merged.
```diff
public sealed record RouteConfig
{
+     public string? RateLimiterPolicy { get; init; }
}

  "ReverseProxy": {
    "Routes": {
      "route1" : {
        "ClusterId": "cluster1",
+      "RateLimiterPolicy": "customPolicy",
        "Match": {
          "Hosts": [ "localhost" ]
        },
      }
    },
    "Clusters": {
      "cluster1": {
        "Destinations": {
          "cluster1/destination1": {
            "Address": "https://localhost:10001/"
          }
        }
      }
    }
  }
}
```
### Why is this important to you?

Tratcher:
Also tracked by https://github.com/dotnet/aspnetcore/issues/37387

karelz:
Triage: We need to light up .NET 7.0 feature like this one.

Tratcher:
Unblocked by https://github.com/dotnet/aspnetcore/blob/a633e8821026868f180be736112f35e6f63159e8/src/Middleware/RateLimiting/src/EnableRateLimitingAttribute.cs.

We'll need to update the runtime/sdk in global.json to consume this.

Tratcher:
@Kahbazi want to try this now? It should be almost a copy of the auth feature.

Kahbazi:
@Tratcher Yes I will take this one.

Kahbazi:
Ok! I started to implement this and I hit some design issues. `RateLimitingMiddleware` is not in the ASP.Net Core framework. It has its own nuget package. What is the approach here?

1. Add reference to [Microsoft.AspNetCore.RateLimiting](https://www.nuget.org/packages/Microsoft.AspNetCore.RateLimiting/) in YARP. This is an easy but maybe not so good solution. 
2. Create a new package `Microsoft.ReverseProxy.RateLimiting`. This would have reference to the middleware but then extending RouteInfo (https://github.com/microsoft/reverse-proxy/pull/1796) might be challenging.

Tratcher:
It should be in the framework for rc1. You'll need a current runtime/SDK dependency to try that.
[...]

PR #1850, opened 2022-09-02
Update the sdk and runtimes to 7.0 rc.2 and 6.0.8

Tratcher:
@Kahbazi this should unblock working on the rate limiting feature https://github.com/microsoft/reverse-proxy/issues/1776


MihaZupan (approved):
CI doesn't seem to like us anymore

PR #1863, opened 2022-09-16
Update code sample of FirstUnsuccessfulResponseHealthPolicy

ricardodemauro:
Previous version doesn't work anymore. This is based on TransportFailureRateHealthPolicy.cs source code.


MihaZupan:
Can you please elaborate on how the current sample doesn't work?

The document already contains a sample for passive health checks below. If something is wrong with the active health sample, we would want to fix that instead of replacing it with another passive sample.

MihaZupan:
Closing as we don't understand the intent of the current change. Feel free to reopen the PR if you wish to continue working on this.

Issue #1254, opened 2021-09-19
Kubernetes Ingress Controller Wishlist


dpbevin:
### What should we add or change to make your life better?

I've captured some features that would be good for improving the Kubernetes Ingress controller code. I'm more than happy to contribute heavily in this space, as I had a previous Ingress Controller project based on Yarp with some of these items.

- [ ] Support ingresses of same name in multiple namespaces. Reasonably simple - cluster and route IDs don't include namespace, so clashes are possible.
- [ ] Single deployable. Combine the "sample ingress" project and the "controller" project into a single deployable, making it easy to set up and manage.
- [ ] Path rewrite via ingress annotations
- [ ] Default (fallback) TLS support for all ingresses, utilizing a Kube secret.
- [ ] SNI-based TLS support on a per-ingress basis. Again, [based on Kube approach](https://v1-20.docs.kubernetes.io/docs/concepts/services-networking/ingress/#tls).
- [ ] Follow newer [IngressClass specification](https://v1-20.docs.kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class). Allows multiple ingress controllers to be present in the same cluster. Not just other ingress technologies but potentially multiple yarp-based controllers for different usages (e.g. public vs internal ingresses).
- [ ] - [ ] Report "scheduled for sync", etc. events to ingress resources to identify that Yarp controller is managing specific ingresses. Not really well documented from a Kube point of view but something that other ingress controller do.
- [ ] Docker image and Helm Chart
- [ ] Session affinity via ingress annotations
- [ ] OAuth enforcement for backend services that don't have their own auth. Similar to https://github.com/oauth2-proxy/oauth2-proxy.
- [ ] GRPC-routing
- [ ] OpenTracing support. Great for diagnosing issues
- [ ] Request size limiting via ingress annotations
- [ ] Header-based canary routing via ingress annotations
- [ ] Rate limiting via ingress annotations

### Why?

To make this project a great option for Kubernetes ingress!

samsp-msft:
We welcome contributions in this space as k8s is not our area of expertise, and we are heads down on trying to finish 1.0 for YARP.

I have questions on a couple of features:

> * Single deployable. Combine the "sample ingress" project and the "controller" project into a single deployable, making it easy to set up and manage

We have found with the Service Fabric equivalent, that when you have a number of proxy instances querying the backend for changes, that can create quite a bit of load, which is why we are thinking that a separate instance for the controller makes more sense.

[...]

PR #1874, opened 2022-09-27
check the ingress for tls and compare to secrets

nquandt:
#1254 TLS from ingress.. 

I'm not sure if there was a different plan, but I just used the .GetIngresses() inside of the Update() method for secrets and compared...

Unit Tests would need to be made.. Feel free to scrap this

### DIFF
--- a/src/Kubernetes.Controller/Caching/IngressCache.cs
+++ b/src/Kubernetes.Controller/Caching/IngressCache.cs
@@ -121,12 +121,29 @@ public class IngressCache : ICache
         var namespacedName = NamespacedName.From(secret);
         _logger.LogDebug("Found secret '{NamespacedName}'. Checking against default {CertificateSecretName}", namespacedName, _options.DefaultSslCertificate);
 
-        if (!string.Equals(namespacedName.ToString(), _options.DefaultSslCertificate, StringComparison.OrdinalIgnoreCase))
+        
+        var hosts = new List<string>();
+
+        foreach (var ingress in Namespace(secret.Namespace()).GetIngresses())
+        {
+            foreach (var tls in ingress.Spec.Tls)
+            {
+                if (tls.SecretName == secret.Name())
+                {
+                    foreach (var host in tls.Hosts)
+                    {
+                        hosts.Add(host);
+                    }
+                }
+            }
+        }
+
+        if (!string.Equals(namespacedName.ToString(), _options.DefaultSslCertificate, StringComparison.OrdinalIgnoreCase) && hosts.Any() == false)
         {
             return;
         }
 
-        _logger.LogInformation("Found secret `{NamespacedName}` to use as default certificate for HTTPS traffic", namespacedName);
+        _logger.LogInformation("Found secret `{NamespacedName}` to use as certificate for HTTPS traffic", namespacedName);
 
         var certificate = _certificateHelper.ConvertCertificate(namespacedName, secret);
         if (certificate is null)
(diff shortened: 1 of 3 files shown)


nquandt:

MihaZupan:
> I'm not sure if there was a different plan, but I just used the .GetIngresses() inside of the Update() method for secrets and compared...

It seems like the reasonable thing to do to me. @dpbevin did you have a different approach in mind here?

src/Kubernetes.Controller/Caching/IngressCache.cs
```diff
@@ -136,7 +153,15 @@ public void Update(WatchEventType eventType, V1Secret secret)
 
         if (eventType == WatchEventType.Added || eventType == WatchEventType.Modified)
         {
-            _certificateSelector.AddCertificate(namespacedName, certificate);
```
dpbevin, line 139: My thinking was that only the default certificate would be added directly to the selector here because there should only be one of them.

In theory, there could be many secrets used for TLS in a system and we may want to update all referenced certificates in one go. My thinking is that this would be part of the [Reconciler](https://github.com/microsoft/reverse-proxy/blob/main/src/Kubernetes.Controller/Services/Reconciler.cs) since this updates the Yarp routing in one go too. Drip-feeding in certificates one-by-one feels "chatty" to me.

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,48 @@
...
 using System.Security.Cryptography.X509Certificates;
 using Microsoft.AspNetCore.Connections;
 
 namespace Yarp.Kubernetes.Controller.Certificates;
 
 internal class ServerCertificateSelector : IServerCertificateSelector
 {
-    private X509Certificate2 _defaultCertificate;
+    private const string FALLBACK_DOMAIN = "_default";
+    private Dictionary<NamespacedName, X509Certificate2> _defaultCertificate = new Dictionary<NamespacedName, X509Certificate2>();
```
dpbevin, line 14: There would only ever be one default certificate, so I'm not fully understanding why this needs to be a dictionary.

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,48 @@
...
+    private Dictionary<string, NamespacedName> _mapping = new Dictionary<string, NamespacedName>();
 
     public void AddCertificate(NamespacedName certificateName, X509Certificate2 certificate)
     {
-        _defaultCertificate = certificate;
+        AddCertificate(certificateName, certificate, null);
+    }
+
+    public void AddCertificate(NamespacedName certificateName, X509Certificate2 certificate, string domainName)
```
dpbevin, line 22: Following on from my comment above about updating certs all in one go, I'd envisaged there would be an UpdateCertificates method that processes all non-default certificates.

While using the host name from the Ingress resource, the certificates themselves might contain wildcards and/or Subject Alternate Names. I think it's important to validate the certificate before adding it to the dictionary. I found that the Service Fabric version of Yarp appears to do a really good job of this. See https://github.com/microsoft/service-fabric-yarp/blob/main/src/YarpProxy.Service/Service/Security/ServerCertificateBinding/SniServerCertificateSelector.cs

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,48 @@
...
     }
 
     public X509Certificate2 GetCertificate(ConnectionContext connectionContext, string domainName)
     {
-        return _defaultCertificate;
+        if (!_mapping.TryGetValue(domainName ?? FALLBACK_DOMAIN, out var mapp)) {
+            return _defaultCertificate[_mapping[FALLBACK_DOMAIN]];
+        }
+
+        return _defaultCertificate[mapp];
```
dpbevin, line 37: If `_defaultCertificate` was empty, wouldn't this throw? Returning null is valid but throwing isn't.

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,48 @@
...
     }
 
     public void RemoveCertificate(NamespacedName certificateName)
     {
-        _defaultCertificate = null;
+        var keys = _mapping.Keys.Where(x => _mapping[x] == certificateName);
```
dpbevin, line 42: Definitely need good testing around removal because of the one-to-many relationship between certificates and ingresses.

nquandt:
At the time of submitting this PR I was not aware of the Reconciler strategy... Closing this, thanks again for reviewing

PR #1899, opened 2022-10-17
Add AddMetricsConsumer helper to TelemetryConsumption

MihaZupan:
I am writing networking telemetry docs and it bothered me that we have very convenient helpers for `ITelemetryConsumer*`s, but nothing for `IMetricsConsumer<T>`.

So while you can write code like this for events:
```c#
services.AddTelemetryConsumer<MyTelemetryConsumer>();
```
you have to write this for metrics:
```c#
services.AddSingleton<IMetricsConsumer<SocketsMetrics>, MetricsConsumer>();
services.AddTelemetryListeners();
```
or even this if you implement multiple interfaces:
```c#
services.AddSingleton<MetricsConsumer>();
services.AddSingleton(services => (IMetricsConsumer<ForwarderMetrics>)services.GetRequiredService<MetricsConsumer>());
services.AddSingleton(services => (IMetricsConsumer<KestrelMetrics>)services.GetRequiredService<MetricsConsumer>());
services.AddSingleton(services => (IMetricsConsumer<SocketsMetrics>)services.GetRequiredService<MetricsConsumer>());
services.AddTelemetryListeners();
```

This PR adds `AddMetricsConsumer` helpers that do the same for metrics as `AddTelemetryConsumer` does for events.

### DIFF
--- a/src/TelemetryConsumption/EventListenerService.cs
+++ b/src/TelemetryConsumption/EventListenerService.cs
@@ -98,7 +98,7 @@ internal abstract class EventListenerService<TService, TTelemetryConsumer, TMetr
             return;
         }
 
-        var eventLevel = enableEvents ? EventLevel.Verbose : EventLevel.Critical;
+        var eventLevel = enableEvents ? EventLevel.Informational : EventLevel.Critical;
         var arguments = enableMetrics ? new Dictionary<string, string?> { { "EventCounterIntervalSec", MetricsOptions.Interval.TotalSeconds.ToString() } } : null;
 
         EnableEvents(eventSource, eventLevel, EventKeywords.None, arguments);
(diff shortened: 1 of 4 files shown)


src/TelemetryConsumption/EventListenerService.cs
```diff
@@ -98,7 +98,7 @@ private void EnableEventSource(EventSource eventSource)
             return;
         }
 
-        var eventLevel = enableEvents ? EventLevel.Verbose : EventLevel.Critical;
+        var eventLevel = enableEvents ? EventLevel.Informational : EventLevel.Critical;
```
Tratcher, line 101: ?

Tratcher (approved):

src/TelemetryConsumption/EventListenerService.cs
MihaZupan: This doesn't change anything currently, but it can make a significant difference if we ever add debug/trace events to the existing `EventSource`s.
Since this library only knows about the current (informational) events, it would only waste CPU on more verbose events.

If we ever want to consume more verbose events with this library, we will adjust this flag as needed.

PR #1914, opened 2022-11-01
[main] Add CodeQL3000 to microsoft-reverse-proxy-official

dougbu:
- add a new weekly schedule to keep database and tools up to database
- add top-level parameter enabling CodeQL3000 in manual builds
- add CodeQL3000 variables
- add CodeQL3000 tasks
  - not auto-injected because that happens only in builds of the default branch
- add tag indicating CodeQL3000 tasks did useful work
  - ignore no-op task executions
- add a tsaoptions.json file
  - cribbed values from eng/sdl-tsa-vars.config file

nits:
- disable auto-injection of build steps we or Arcade do explicitly
- remove irrelevant variables in test jobs

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,9 +7,7 @@ schedules:
   displayName: "Run CodeQL3000 weekly, Monday at 2:00 AM PDT"
   branches:
     include:
-    - release/1.0
-    - release/1.1
-    - release/latest
+    - release/*
     - main
   always: true
 
@@ -86,8 +84,6 @@ stages:
           # Component governance and SBOM creation are not needed here. Disable what Arcade would inject.
           disableComponentGovernance: true
           enableSbom: false
-          # CodeQL3000 extends build duration.
-          timeoutInMinutes: 180
         variables:
         - ${{ if eq(variables.runCodeQL3000, 'true') }}:
           - _OfficialBuildArgs: /p:Test=false /p:Sign=false /p:Pack=false /p:Publish=false /p:UseSharedCompilation=false
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,6 +2,27 @@
 # See https://docs.microsoft.com/azure/devops/pipelines/yaml-schema for details
 #
 
+schedules:
+- cron: 0 9 * * 1
+  displayName: "Run CodeQL3000 weekly, Monday at 2:00 AM PDT"
+  branches:
+    include:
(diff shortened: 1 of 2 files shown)


dougbu:
/cc @wtgodbe

@Tratcher note I have no permissions in this repo and will not be able to merge on my own.

Am however able to run internal test builds. See [#20221101.1](https://dev.azure.com/dnceng/internal/_build/results?buildId=2034344&view=results) because that enables the CodeQL3000 tasks.

azure-pipelines.yml
```diff
@@ -2,6 +2,27 @@
...
 
+schedules:
+- cron: 0 9 * * 1
+  displayName: "Run CodeQL3000 weekly, Monday at 2:00 AM PDT"
+  branches:
+    include:
+    - release/1.0
+    - release/1.1
+    - release/latest
+    - main
```
Tratcher, line 13: Does this need to be manually maintained when we add and remove branches? If so, please add a note to https://github.com/microsoft/reverse-proxy/blob/main/docs/operations/Branching.md

Tratcher (approved):

azure-pipelines.yml
dougbu: Hmm, I'll use a wildcard instead to avoid ongoing maintenance.

azure-pipelines.yml
```diff
@@ -2,6 +2,27 @@
...
 #
 
+schedules:
+- cron: 0 9 * * 1
+  displayName: "Run CodeQL3000 weekly, Monday at 2:00 AM PDT"
+  branches:
+    include:
+    - release/1.0
+    - release/1.1
+    - release/latest
```
dougbu, line 12:
```suggestion
    - release/*
```

azure-pipelines.yml
```diff
@@ -53,11 +82,30 @@ stages:
           ${{ if ne(variables['System.TeamProject'], 'public') }}:
             name: NetCore1ESPool-Internal
             demands: ImageOverride -equals windows.vs2019.amd64
+        ${{ if eq(variables.runCodeQL3000, 'true') }}:
+          # Component governance and SBOM creation are not needed here. Disable what Arcade would inject.
+          disableComponentGovernance: true
+          enableSbom: false
+          # CodeQL3000 extends build duration.
+          timeoutInMinutes: 180
```
dougbu, line 90: Test build took much less time than I expected
```suggestion
```

Issue #1893, opened 2022-10-13
No way to reject a route


CobusKruger:
### Describe the bug
Let's say I want to have an SPA served from https://example.com, and its APIs from https://example.com/api/name-of-api.

I would imagine setting up the two routes like this:

```json
[
    {
        "RouteId": "app",
        "Match": {
            "Path": "/{**any}"
        },
        "ClusterId": "app"
    },
    {
        "RouteId": "first-api",
        "Match": {
            "Path": "/api/first-api/{*any}"
        },
        "ClusterId": "first-api"
    },
    {
        "RouteId": "second-api",
        "Match": {
            "Path": "/api/second-api/{*any}"
        },
        "ClusterId": "second-api"
    }
]
```

All three of these URLs will work as intended:
https://example.com/
https://example.com/api/first-api
https://example.com/api/second-api

But this URL will also be accepted, and quietly return the same as https://example.com
https://example.com/api/third-and-totally-invalid-api

Ideally, I would want to instruct the proxy to reject that route with a 404, perhaps with something like this:

```typescript
    {
        "RouteId": "invalid-api",
        "Match": {
            "Path": "/api/{*any}"
        },
        StatusCode: 404
    },
```

### To Reproduce
Listed in the description

### Further technical details

Yarp.ReverseProxy 1.1.1
On Windows 10, .NET 6

Tratcher:
Similar request: https://github.com/microsoft/reverse-proxy/issues/1868

Workaround:
- That refusal route can be defined in code and sent to an endpoint that produces a 404.

karelz:
Triage: It is trivial in code. However, it will be interesting in a single-exe scenario.
We need to do a bit of design - e.g. which error codes we want to handle this way.

Tratcher:
Another ask: https://github.com/microsoft/reverse-proxy/discussions/1928
- Via config, add a response header to all requests that don't match a route. A low priority wildcard route could do this.

Issue #1868, opened 2022-09-21
Cluster "Action" for given routes


tswalker:
### What should we add or change to make your life better?
It would be nice to leverage the HttpClient for an inbound route for generating immediate responses using the appsettings.json file with a new cluster parameter called "Action" (or similar).  As "Destinations" defines how the forwarder handles the request, "Action" would permit (with appropriate properties) an immediate response without the need to handle that path within Program.cs and additional middleware.

### Why is this important to you?
Currently, I need to check the path of the request for a route (say "myspecialroute") in middleware to generate a response and terminate the chain.  Using something simple like:

```c#
app.Use((context, next) =>
{
	var path = context.Request.Path;
	if (path.HasValue && path.Value.Contains("myspecialroute"))
	{
		context.Response.StatusCode = StatusCodes.Status400BadRequest;
		return context.Response.WriteAsync("This integration is not supported.");
	}
	return next();
});
```

Having this 'actionable' configured in appsettings would be nice as if the route changed, a simple config file update would do the trick versus having to re-program the middleware.

Tratcher:
You could also define these special routes in AspNet MVC/Razor/minimal. YARP shares the route table with the rest of the app.

What capabilities would you expect an 'Action' to have, and how would those be described? Can you give some example JSON definitions?

karelz:
Triage: 12 days no response, closing as not actionable. If/when there is more info, feel free to reopen. Thanks!

Issue #1670, opened 2022-04-25
Mandatory Request header validation


askids:
hi,

Is there any existing feature available within YARP that allows me to enforce checking for presence of mandatory request headers on incoming requests?

Thanks!

Tratcher:
Not directly. You can do this with middleware: https://microsoft.github.io/reverse-proxy/articles/middleware.html

MihaZupan:
For simple validation on a per-route basis, you can use [header routing](https://microsoft.github.io/reverse-proxy/articles/header-routing.html).
The conditions are limited to basic operations such as "header foo exists", "header bar contains myValue".
If you must do more validation on the value, a custom middleware is the way to go as Chris suggested.

askids:
I dont need to do header based routing. I just need to check if the header exists or not., when I receive a call for a route. if the header is not present, I want to throw a 400 bad request error. Looks like I will need to use custom middleware then.

Issue #187, opened 2020-05-22
Support to configure static file resources


expcat:
### What should we add or change to make your life better?
Support to configure static file resources.
### Why is this important to you?
Avoid using Nginx on the front end.

Tratcher:
Since we use the AspNetCore infrastructure you can already use the existing StaticFile middleware for this. Can you try this with preview1 and let us know how it works for you?

When we create a pre-built exe version of the proxy we'll want to consider how to enable static files via config.

samsp-msft:
Triage: Will need config for exe scenario.

samsp-msft:
Doc so we can close it

Tratcher:
@samsp-msft this should be combine with the exe scenario.

PR #1923, opened 2022-11-05
Enable request transforms to reject requests 

### LINKED ISSUE #1701: Request/response validators
YARP has request/response transforms. They are useful for what they are.

They do not have the concept of failing a request or response when certain validation criteria are not satisfied. It would be nice to have a similar way to plugin validators that are able to return error responses to the caller.

Two use cases we have

1) we want to make sure a certain OAuth access token is available to forward to a remote API. If that token is not available, YARP should return a 401 straight away without proxying to the backend
2) require the existence of certain headers (e.g. for antiforgery protection). Again the request should fail immediately.

I know that this could be achieved via middleware - but a mechanism similar to transforms would be less intrusive from configuration point of view.

Tratcher:
Fixes #1701 @leastprivilege 

If a request transform encounters a problem right now the only thing it can do is throw. A few customers have asked how to generate a response and exit gracefully. Response transforms can already do this by returning false to prevent us from proxying the body.
- https://github.com/microsoft/reverse-proxy/issues/1701
- https://github.com/microsoft/reverse-proxy/issues/1670
- https://github.com/microsoft/reverse-proxy/discussions/1724

`HttpTransformer.TransformRequestAsync` has a signature that can't return any additional information without an API breaking change. Instead, we can detect if a transform generated a response by checking HttpResponse.StatusCode or HasStarted. In this case we'll stop trying to proxy the request and exit gracefully. If someone had modified the StatusCode before this change it would have been overwritten when proxying the response. If they had started the response it would have failed when trying to proxy the response later.

There's a series of related requests that I checked if could be solved in the same way but decided against it. I think these can be handled at the config and routing layer with alternate endpoints, there's no need for them to execute the proxy pipeline / load balancing, session affinity, etc. I'll send a separate PR for those.
- https://github.com/microsoft/reverse-proxy/issues/1893
- https://github.com/microsoft/reverse-proxy/issues/1868
- https://github.com/microsoft/reverse-proxy/issues/187

TODO: Update docs

### DIFF
--- a/samples/ReverseProxy.Auth.Sample/TokenService.cs
+++ b/samples/ReverseProxy.Auth.Sample/TokenService.cs
@@ -1,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Yarp.Sample
+{
+    internal class TokenService
+    {
+        internal Task<string> GetAuthTokenAsync(ClaimsPrincipal user)
+        {
+            // we only have tokens for bob
+            if (string.Equals("Bob", user.Identity.Name))
+            {
+                return Task.FromResult("valid");
+            }
+            return Task.FromResult<string>(null);
+        }
+    }
+}
--- a/samples/ReverseProxy.Auth.Sample/TokenService.cs
+++ b/samples/ReverseProxy.Auth.Sample/TokenService.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using System.Security.Claims;
 using System.Threading.Tasks;
 
@@ -13,7 +14,7 @@ namespace Yarp.Sample
             // we only have tokens for bob
             if (string.Equals("Bob", user.Identity.Name))
             {
-                return Task.FromResult("valid");
+                return Task.FromResult(Guid.NewGuid().ToString());
             }
             return Task.FromResult<string>(null);
         }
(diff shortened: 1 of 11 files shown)


samples/ReverseProxy.Auth.Sample/TokenService.cs
```diff
@@ -0,0 +1,21 @@
...
+        internal Task<string> GetAuthTokenAsync(ClaimsPrincipal user)
+        {
+            // we only have tokens for bob
+            if (string.Equals("Bob", user.Identity.Name))
+            {
+                return Task.FromResult("valid");
+            }
+            return Task.FromResult<string>(null);
+        }
+    }
```
gfoidl, line 20: Although a sample, but it should show a perf-concise way with
```suggestion
    internal class TokenService
    {
        private static readonly Task<string> s_validTask = Task.FromResult("valid");
        
        internal Task<string> GetAuthTokenAsync(ClaimsPrincipal user)
        {
            // we only have tokens for bob
            if (string.Equals("Bob", user.Identity.Name))
            {
                return s_validTask;
            }
            return Task.FromResult<string>(null);
        }
    }
```
to save the repeated task allocation (72B on 64-bit)?

It's more an educational thing for people that aren't aware of such things.

Tratcher: It shouldn't actually be a static value for security reasons. Maybe I should change it to new Guid to clarify that.

gfoidl: 👍🏻 even better!

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -94,6 +94,11 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
         _ = requestConfig ?? throw new ArgumentNullException(nameof(requestConfig));
         _ = transformer ?? throw new ArgumentNullException(nameof(transformer));
 
+        if (context.Response.StatusCode != StatusCodes.Status200OK || context.Response.HasStarted)
```
MihaZupan, line 97: Can we add a shared helper that does this check?

MihaZupan (approved):

PR #1940, opened 2022-11-28
log proxy errors as warnings

luk355:
Logging proxy errors as warning helps to separate proxy errors from other log informational events. Proxy errors are useful to troubleshoot any errors.

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -779,7 +779,7 @@ internal sealed class HttpForwarder : IHttpForwarder
             "Proxying to {targetUrl} {version} {versionPolicy} {isStreaming}");
 
         private static readonly Action<ILogger, ForwarderError, string, Exception> _proxyError = LoggerMessage.Define<ForwarderError, string>(
-            LogLevel.Information,
+            LogLevel.Warning,
             EventIds.ForwardingError,
             "{error}: {message}");
 


Tratcher:
This log level was intentional since many of these errors will happen due to external client or destination issues rather than proxy issues.

Please open issues for discussion prior to submitting PRs.

Are you aware that you can filter log outputs by category in the logging configuration?

luk355:
Yes, I am aware that filtering can be done based on the category (`Yarp.ReverseProxy.Forwarder.HttpForwarder` in this case), however, we really want to keep log events related to the processing errors and omit the remaining log events from this class/category.

A solution could be to log these errors under separate category i.e. `Yarp.ReverseProxy.Forwarder.HttpForwarder.ProxyErrors` but that would require injection of new logger instance. I would argue how many of those proxy error events does not need user attention. Surely, you want to know about the cases when the end host is unreachable or connection is reset  - thus justifying the Warning loglevel.

If you want, we can move the discussion into issues...

Tratcher:
I reviewed this with some of our large customer deployments and they said this change would produce too much noise for them. These are not the kind of errors they're concerned about monitoring at scale, especially since they get the return value from HttpForwarder and can do more granular processing themselves. Have you tried that aproach?
https://microsoft.github.io/reverse-proxy/articles/middleware.html#error-handling

samsp-msft:
I would suggest if we do something here, it would be about changing the category so that customers can decide at a more granular level which items they want to track and which to ignore.

Tratcher:
Since we won't be taking this as is, please file an issue and we can discuss alternatives there.

PR #1947, opened 2022-12-01
Make MaxRequestBodySize configurable 

### LINKED ISSUE #640: Request body too large
For reference, we are using Microsoft.ReverseProxy 1.0.0-preview.7.20562.2 in an empty ASP.NET Web Application using .NET 5, running with Kestrel

On larger requests we receive the following error:

RequestBodyClient: "The client reported an error when copying the request body." (1474d177)
System.AggregateException: One or more errors occurred. (Request body too large.) (Error while copying content to a stream.)

Is it possible to increase the size and how can we do this?

Tratcher:
Fixes #640

The servers all have a configurable MaxRequestBodySize limit that defaults to ~30MB. This can be overridden per request with a feature. The most common case is to raise it for certain endpoints that upload big files, or to disable it for streaming scenarios. Some scenarios automatically disable it like WebSockets or gRPC streaming. Others need to be configured.

There is .NET 8 work planned to centralize this into endpoint metadata and middleware, but we want to support this for 6 and 7 in the meantime. https://github.com/dotnet/aspnetcore/issues/40452

### DIFF
--- a/docs/docfx/articles/config-files.md
+++ b/docs/docfx/articles/config-files.md
@@ -130,7 +130,7 @@ For additional fields see [ClusterConfig](xref:Yarp.ReverseProxy.Configuration.C
         // matches /something/* and routes to "allclusterprops"
         "ClusterId": "allclusterprops", // Name of one of the clusters
         "Order" : 100, // Lower numbers have higher precedence
-        "MaxRequestBodySize" : 1000000, // In bytes. An optional override of the server's limit (30MB). Set to -1 to disable.
+        "MaxRequestBodySize" : 1000000, // In bytes. An optional override of the server's limit (30MB default). Set to -1 to disable.
         "Authorization Policy" : "Anonymous", // Name of the policy or "Default", "Anonymous"
         "CorsPolicy" : "Default", // Name of the CorsPolicy to apply to this route or "Default", "Disable"
         "Match": {
--- a/docs/docfx/articles/config-files.md
+++ b/docs/docfx/articles/config-files.md
@@ -130,6 +130,7 @@ For additional fields see [ClusterConfig](xref:Yarp.ReverseProxy.Configuration.C
         // matches /something/* and routes to "allclusterprops"
         "ClusterId": "allclusterprops", // Name of one of the clusters
         "Order" : 100, // Lower numbers have higher precedence
+        "MaxRequestBodySize" : 1000000, // In bytes. An optional override of the server's limit (30MB). Set to -1 to disable.
         "Authorization Policy" : "Anonymous", // Name of the policy or "Default", "Anonymous"
         "CorsPolicy" : "Default", // Name of the CorsPolicy to apply to this route or "Default", "Disable"
         "Match": {
--- a/src/ReverseProxy/Limits/LimitsMiddleware.cs
+++ b/src/ReverseProxy/Limits/LimitsMiddleware.cs
@@ -11,8 +11,9 @@ namespace Yarp.ReverseProxy.Limits;
 
 /// <summary>
 /// Updates request limits based on route config. This is implemented as middleware at the end of the proxy
-/// pipeline so that apps can call ReassignProxyRequest to move the request to a different route before limits are applied.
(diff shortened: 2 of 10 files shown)


src/ReverseProxy/Limits/LimitsMiddleware.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.Logging;
+
+namespace Yarp.ReverseProxy.Limits;
+
+/// <summary>
+/// Updates request limits based on route config. This is implemented as middleware at the end of the proxy
+/// pipeline so that apps can call ReassignProxyRequest to move the request to a different route before limits are applied.
+/// This may be replaced in the future by route metadata and an aspnetcore middleware https://github.com/dotnet/aspnetcore/issues/40452.
```
MihaZupan, line 15: I realize you did it this way to respect changes made by `ReassignProxyRequest`, but doesn't this mean that the behavior will differ if we ever want to replace it with the eventual aspnetcore middleware?

Right now, we'll honor the YARP-specific config on whichever YARP route you specify.
Afterward, it would honor the config for the endpoint that initially accepted the request.

MihaZupan (approved):

src/ReverseProxy/Limits/LimitsMiddleware.cs
Tratcher: We're still not sure when/where in the pipleline the new aspnetcore will read the route metadata and apply the limits. Presumably somewhere between then the endpoint is assigned and when it's executed. In that case it would already have run before the proxy pipeline. ReassignProxyRequest gets problematic because you're right, it replaces the proxy route, not the endpoint. Could/should it look up and replace the endpoint?

Tratcher: The more I think about this, I realize that ReassignProxyRequest means we can't ever unify with the proposed aspnetcore feature. That feature would need to apply the limits before the proxy pipeline runs. If ReassignProxyRequest changed the route then the limites would need to be changed again. Problem: If route1 had a specific limit and route2 didn't, route1's limit would be used even after switching to route2 because we wouldn't know what to set the limit back to, we don't have access to the server's original limit.

The limit can only be changed once, so it needs to happen late in the pipeline. I'll update the code comments, but I think the current implementation is the best we'll manage.

docs/docfx/articles/config-files.md
```diff
@@ -130,6 +130,7 @@ For additional fields see [ClusterConfig](xref:Yarp.ReverseProxy.Configuration.C
         // matches /something/* and routes to "allclusterprops"
         "ClusterId": "allclusterprops", // Name of one of the clusters
         "Order" : 100, // Lower numbers have higher precedence
+        "MaxRequestBodySize" : 1000000, // In bytes. An optional override of the server's limit (30MB). Set to -1 to disable.
```
Tratcher, line 133:
```suggestion
        "MaxRequestBodySize" : 1000000, // In bytes. An optional override of the server's limit (30MB default). Set to -1 to disable.
```

src/ReverseProxy/Limits/LimitsMiddleware.cs
Tratcher, line 15:
```suggestion
/// Updates request limits based on route config. This is implemented as middleware at the end of the proxy
/// pipeline so that apps can call ReassignProxyRequest to move the request to a different route before limits are applied. While similar to a proposed aspnetcore feature (https://github.com/dotnet/aspnetcore/issues/40452),
/// the possibility of reassigning routes means we need to apply this limit very late. Trying to apply it twice could
/// result in unexpected behavior like being unable to set it back to the server default.
```

MihaZupan: Yes, if we want to respect `ReassignProxyRequest` (and I agree we should), we'll have to stick with our own implementation.
Updated comment LGTM

Issue #1950, opened 2022-12-03
Support for addition of Multiple Host Domains in K8s Ingress Rules


unknownstring:
### Yarp.Kubernetes.Controller
We currently use Nginx as Ingress controller for our endpoints in K8s. However, there are a few shortcomings and as our technology stack is .NET based, we consider using Yarp instead. 

Our Nginx rules are 'project' based, hence we use subdomains to organize and target individual related services.

Ingress manifest with multiple host domains for one external IP:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: repro-1-ingress
  namespace: foo
spec:
  ingressClassName: yarp-ingress
  rules:
    - host: 'subdomain1.example.com'
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: repro-1-service
                port:
                  name: http
    - host: 'subdomain2.example.com'
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: repro-2-service
                port:
                  name: http
```

Current workaround, split into two ingress manifests:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: repro-1-ingress-for-subdomain1
  namespace: foo
spec:
  ingressClassName: yarp-ingress
  rules:
    - host: 'subdomain1.example.com'
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: repro-1-service
                port:
                  name: http
```
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: repro-1-ingress-for-subdomain2
  namespace: foo
spec:
  ingressClassName: yarp-ingress
  rules:
    - host: 'subdomain2.example.com'
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: repro-2-service
                port:
                  name: http
```

### Include host domain as part of the route naming
Due to Ingress manifest parsing and Yarp route id naming, this is not currently possible. 

Hence, suggestion is to include host domain as part of the route naming. This reduces duplicating Ingress manifests and additionally bundles related host domains in Azure Ingress overview.

MihaZupan:
Fixed by #1951

PR #1951, opened 2022-12-03
Add support for addition of Multiple Host Domains to K8s Ingress Rules

unknownstring:
Include Host name in Yarp Route Id, when parsing Ingress Manifest Rules.

Fixes issue as described in #1950

### DIFF
--- a/src/Kubernetes.Controller/Converters/YarpParser.cs
+++ b/src/Kubernetes.Controller/Converters/YarpParser.cs
@@ -101,7 +101,7 @@ internal static class YarpParser
                         Headers = ingressContext.Options.RouteHeaders
                     },
                     ClusterId = cluster.ClusterId,
-                    RouteId = $"{ingressContext.Ingress.Metadata.Name}.{ingressContext.Ingress.Metadata.NamespaceProperty}:{path.Path}",
+                    RouteId = $"{ingressContext.Ingress.Metadata.Name}.{ingressContext.Ingress.Metadata.NamespaceProperty}:{host}{path.Path}",
                     Transforms = ingressContext.Options.Transforms,
                     AuthorizationPolicy = ingressContext.Options.AuthorizationPolicy,
                     CorsPolicy = ingressContext.Options.CorsPolicy,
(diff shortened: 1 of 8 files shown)


unknownstring:
>

unknownstring:
@microsoft-github-policy-service agree company="GN Audio A/S"

MihaZupan:
Thanks for looking to contribute here!

> Our Nginx rules are 'project' based, hence we use subdomains to organize and target individual related services.

> Due to Ingress manifest parsing and Yarp route id naming, this is not currently possible.

> Hence, suggestion is to include host domain as part of the route naming. This reduces duplicating Ingress manifests and additionally bundles related host domains in Azure Ingress overview.

Can you elaborate on how including the host into the route id simplifies this for you?

This `RouteId` is only visible internally to the YARP process. Are you using custom middleware to react to this ID?

unknownstring:
> Thanks for looking to contribute here!
> 
> > Our Nginx rules are 'project' based, hence we use subdomains to organize and target individual related services.
> 
> > Due to Ingress manifest parsing and Yarp route id naming, this is not currently possible.
> 
> > Hence, suggestion is to include host domain as part of the route naming. This reduces duplicating Ingress manifests and additionally bundles related host domains in Azure Ingress overview.
> 
> Can you elaborate on how including the host into the route id simplifies this for you?
> 
> This `RouteId` is only visible internally to the YARP process. Are you using custom middleware to react to this ID?

Hi!

it's more a question of duplicate RouteIds and less about visibility. 

In the current implementation, adding multiple hosts to rules for a given Ingress parses duplicate RouteIds to YARP, as RouteId is based on _Ingress name and path only_. 

Hence the deployment of Ingress to K8s fails during validation.

dpetrovych:
Hi @MihaZupan , I'm co-authoring this change.
We are not using RouteId in any way. By including `{host}` we want to avoid `RouteId` duplicates when using quite standard and valid k8s configuration for [Name based virtual hosting](https://kubernetes.io/docs/concepts/services-networking/ingress/#name-based-virtual-hosting).

So, in the test example _test/Kubernetes.Tests/testassets/multiple-hosts/routes.json_ before change all routes would have `RouteId: repro-1-ingress.foo:/`, this won't pass validation in https://github.com/microsoft/reverse-proxy/blob/9253842eea4b702489c9e4559a98f9d04d52c853/src/ReverseProxy/Management/ProxyConfigManager.cs#L373

Same error we first faced in one of k8s deployments:
image

MihaZupan:
Thank you for the explanation, that makes sense.

MihaZupan (approved):

PR #1953, opened 2022-12-06
Fix startup race condition in K8s IngressController

archonandrewhunt:
There is currently a race condition in the loading of K8s resources through the resource informers. If the ingress resources load before the ingress classes, and they have an ingress class name specified, then they are not considered for config generation. Similarly, if you have your yarp ingress class defined as default, but ingresses load first, then they are also not considered because the controller is not yet aware of what the default controller should be.

This fix guarantees the ingress classes will be loaded first prior to loading all other resource types. Please let me know if you would like this done a different way, I am happy to make any changes to help see this contribution through.

### DIFF
--- a/src/Kubernetes.Controller/Caching/ICache.cs
+++ b/src/Kubernetes.Controller/Caching/ICache.cs
@@ -23,4 +23,5 @@ public interface ICache
     bool TryGetReconcileData(NamespacedName key, out ReconcileData data);
     void GetKeys(List<NamespacedName> keys);
     IEnumerable<IngressData> GetIngresses();
+    bool IsYarpIngress(IngressData ingress);
 }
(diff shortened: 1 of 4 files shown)


archonandrewhunt:
fyi @dpbevin

MihaZupan:
Thanks for spotting this and looking to contribute!

Doesn't this race also exist after we've completed the initial load?
That is if we were to receive ingress class updates from k8s sometime after we're fully loaded, could we potentially drop some ingress resources in between?

archonandrewhunt:
It is possible I suppose if you update ingress class and ingresses simultaneously since the watches aren't guaranteed to happen in any order at the moment.

A more robust fix here may be to just move the ingress class checking logic to the reconciler rather than before the ingresses are stored in the cache in `IngressCache`. I could explore that instead if you think that makes more sense.

MihaZupan:
> A more robust fix here may be to just move the ingress class checking logic to the reconciler rather than before the ingresses are stored in the cache in IngressCache.

Your current change is obviously already an improvement, but I think moving this to the reconciler would be better long term. It does mean we may keep a few more ingress instances around in memory, though I assume that's fine.

MihaZupan (approved):
Thanks. This is an even nicer change than what I'd expected we'd need.

Issue #1375, opened 2021-11-11
WebSockets over HTTP/2


Tratcher:
We've heard some interest express for WebSockets support over HTTP/2 in YARP. This would require support in ASP.NET Core (Kestrel) and HttpClient. The protocol is quite different from original WebSockets.

RFC: https://datatracker.ietf.org/doc/html/rfc8441
Kestrel: https://github.com/dotnet/aspnetcore/issues/7801
HttpClient: https://github.com/dotnet/runtime/issues/69669

We'd also need to worry about protocol upgrades/downgrades between front-end and back-end.

karelz:
Triage: We have interested partner team, but not yet committed. It helps with COGS.
We would need feature work in .NET Runtime (7.0+).

Forwarding is directly through `SocketsHttpHandler`, so the API pattern might be different from current `ClientWebSocket`.

adityamandaleeka:
This should be happening for .NET 7 on both the server and client side, so let's make YARP take advantage of that in 2.0

Tratcher:
@greenEkatherine as the first step here, can you test YARP 1.1 on .NET 7 with an Http/2 WebSocket client and capture the error here in this issue?

serpent5:
I'm running YARP 1.1 with .NET 7, attempting to proxy the Angular CLI development server's WebSocket connection. Here's some information that might be helpful:

Headers:
```
{
  "method": "GET",
  "url": "wss://localhost:5001/ng-cli-ws",
  "httpVersion": "",
  "headers": [
    {
      "name": "Pragma",
      "value": "no-cache"
    },
    {
      "name": "Origin",
      "value": "https://localhost:5001"
    },
    {
      "name": "Accept-Encoding",
      "value": "gzip, deflate, br"
    },
    {
      "name": "Host",
      "value": "localhost:5001"
    },
    {
      "name": "Accept-Language",
      "value": "en-GB,en-US;q=0.9,en;q=0.8"
    },
    {
      "name": "User-Agent",
      "value": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/107.0.0.0 Safari/537.36 Edg/107.0.1418.42"
    },
    {
      "name": "Upgrade",
      "value": "websocket"
    },
    {
      "name": "Cache-Control",
      "value": "no-cache"
    },
    {
      "name": "Connection",
      "value": "Upgrade"
    },
    {
[...]

PR #1956, opened 2022-12-08
Refactor functional test environment

Tratcher:
I was adding new functional tests for H2WS (https://github.com/microsoft/reverse-proxy/issues/1375) but found the structure of the TestEnvironment difficult to add additional properties to due to the complex constructors. I've re-factored it to use properties instead.


MihaZupan (approved):
> complex constructors

And they've only grown over time :)

PR #1967, opened 2022-12-18
Added support for AspNetCore 7 rate limiting

mburumaxwell:
Add support for Rate Limiting available from AspNetCore 7.0.

Caveats:
- The repo/libraries support both net6 and net7 using `#if NET7_0_OR_GREATER` which can potentially be a problem for maintenance.
- There is no means to validate the policy actually exists without the prescence of an interface/class similar to `IAuthorizationPolicyProvider` or `ICorsPolicyProvider`. The `RateLimitingMiddleware` uses `RateLimiterOptions.PolicyMap`  but it is internal. Hopefully something will be available in AspNetCore 8 to allow validation. A `TODO` exists in the validation just tracking and as a reminder.
- Not possible to use `default` similar to `AuthorizationPolicy` and `CorsPolicy` because the `EnableRateLimitingAttribute` requires a `policyName`. Default policy would have to be configured in the serivces setup:
    `services.AddRateLimiter(options => { /* conf stuff here */ });`

Fixes: #1776

### DIFF
--- a/docs/docfx/articles/rate-limiting.md
+++ b/docs/docfx/articles/rate-limiting.md
@@ -7,7 +7,12 @@ The reverse proxy can be used to rate-limit requests before they are proxied to 
 
 ## Defaults
 
-No rate limiting is performed on requests unless enabled in the route or application configuration.
+No rate limiting is performed on requests unless enabled in the route or application configuration. However, the Rate Limiting middleware (`app.UseRateLimiter()`) can apply a default limiter applied to all routes, and this doesn't require any opt-in from the config.
+
+Example:
+```
+builder.Services.AddRateLimiter(options => options.GlobalLimiter = globalLimiter);
+```
 
 ## Configuration
 Rate Limiter policies can be specified per route via [RouteConfig.RateLimiterPolicy](xref:Yarp.ReverseProxy.Configuration.RouteConfig) and can be bound from the `Routes` sections of the config file. As with other route properties, this can be modified and reloaded without restarting the proxy. Policy names are case insensitive.
--- a/docs/docfx/articles/rate-limiting.md
+++ b/docs/docfx/articles/rate-limiting.md
@@ -10,7 +10,7 @@ The reverse proxy can be used to rate-limit requests before they are proxied to 
 No rate limiting is performed on requests unless enabled in the route or application configuration. However, the Rate Limiting middleware (`app.UseRateLimiter()`) can apply a default limiter applied to all routes, and this doesn't require any opt-in from the config.
 
 Example:
-```
+```c#
 builder.Services.AddRateLimiter(options => options.GlobalLimiter = globalLimiter);
 ```
 
@@ -46,7 +46,7 @@ Example:
 [RateLimiter policies](https://learn.microsoft.com/aspnet/core/performance/rate-limit) are an ASP.NET Core concept that the proxy utilizes. The proxy provides the above configuration to specify a policy per route and the rest is handled by existing ASP.NET Core rate limiting middleware.
 
 RateLimiter policies can be configured in Startup.ConfigureServices as follows:
-```
+```c#
 public void ConfigureServices(IServiceCollection services)
 {
     services.AddRateLimiter(options =>
@@ -64,7 +64,7 @@ public void ConfigureServices(IServiceCollection services)
 
 In Startup.Configure add the RateLimiter middleware between Routing and Endpoints.
 
-```
+```c#
 public void Configure(IApplicationBuilder app)
 {
     app.UseRouting();
--- a/docs/docfx/articles/rate-limiting.md
+++ b/docs/docfx/articles/rate-limiting.md
@@ -1,0 +1,78 @@
+# Rate Limiting
+
+## Introduction
+The reverse proxy can be used to rate-limit requests before they are proxied to the destination servers. This can reduce load on the destination servers, add a layer of protection, and ensure consistent policies are implemented across your applications.
+
+## Defaults
+
+No rate limiting is performed on requests unless enabled in the route or application configuration.
+
+## Configuration
+Rate Limiter policies can be specified per route via [RouteConfig.RateLimiterPolicy](xref:Yarp.ReverseProxy.Configuration.RouteConfig) and can be bound from the `Routes` sections of the config file. As with other route properties, this can be modified and reloaded without restarting the proxy. Policy names are case insensitive.
+
+Example:
+```JSON
+{
+  "ReverseProxy": {
+    "Routes": {
+      "route1" : {
+        "ClusterId": "cluster1",
+        "RateLimiterPolicy": "customPolicy",
+        "Match": {
+          "Hosts": [ "localhost" ]
+        },
+      }
+    },
+    "Clusters": {
+      "cluster1": {
+        "Destinations": {
+          "cluster1/destination1": {
+            "Address": "https://localhost:10001/"
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+[RateLimiter policies](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit) are an ASP.NET Core concept that the proxy utilizes. The proxy provides the above configuration to specify a policy per route and the rest is handled by existing ASP.NET Core rate limiting middleware.
+
+RateLimiter policies can be configured in Startup.ConfigureServices as follows:
+```
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddRateLimiter(options =>
+    {
+        options.AddFixedWindowLimiter("customPolicy", opt =>
+        {
+            opt.PermitLimit = 4;
+            opt.Window = TimeSpan.FromSeconds(12);
+            opt.QueueProcessingOrder = QueueProcessingOrder.OldestFirst;
+            opt.QueueLimit = 2;
+        });
+    });
+}
+```
+
+In Startup.Configure add the RateLimiter middleware between Routing and Endpoints.
+
+```
+public void Configure(IApplicationBuilder app)
+{
+    app.UseRouting();
+
+    app.UseRateLimiter();
+
+    app.UseEndpoints(endpoints =>
+    {
+        endpoints.MapReverseProxy();
+    });
+}
+```
+
+See the [Rate Limiting](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit) docs for setting up your preferred kind of rate limiting.
+
+### Disable CORS
+
+Specifying the value `disable` in a route's `RateLimiterPolicy` parameter means the rate limiter middleware will not rate limit requests.
--- a/docs/docfx/articles/rate-limiting.md
+++ b/docs/docfx/articles/rate-limiting.md
@@ -43,7 +43,7 @@ Example:
 }
 ```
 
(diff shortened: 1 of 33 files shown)


Tratcher:
> * There is no means to validate the policy actually exists without the prescence of an interface/class similar to `IAuthorizationPolicyProvider` or `ICorsPolicyProvider`. The `RateLimitingMiddleware` uses `RateLimiterOptions.PolicyMap`  but it is internal. Hopefully something will be available in AspNetCore 8 to allow validation. A `TODO` exists in the validation just tracking and as a reminder.

Is there an issue tracking that in AspNetCore?

docs/docfx/articles/rate-limiting.md
```diff
@@ -0,0 +1,78 @@
+# Rate Limiting
```
Tratcher, line 1: Add this to the article list: https://github.com/microsoft/reverse-proxy/blob/main/docs/docfx/articles/toc.yml

docs/docfx/articles/rate-limiting.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Introduction
+The reverse proxy can be used to rate-limit requests before they are proxied to the destination servers. This can reduce load on the destination servers, add a layer of protection, and ensure consistent policies are implemented across your applications.
```
Tratcher, line 4: This feature is only available when using .NET 7.0.

docs/docfx/articles/rate-limiting.md
```diff
@@ -0,0 +1,78 @@
...
+            "Address": "https://localhost:10001/"
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+[RateLimiter policies](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit) are an ASP.NET Core concept that the proxy utilizes. The proxy provides the above configuration to specify a policy per route and the rest is handled by existing ASP.NET Core rate limiting middleware.
```
Tratcher, line 39:
```suggestion
[RateLimiter policies](https://learn.microsoft.com/aspnet/core/performance/rate-limit) are an ASP.NET Core concept that the proxy utilizes. The proxy provides the above configuration to specify a policy per route and the rest is handled by existing ASP.NET Core rate limiting middleware.
```

src/ReverseProxy/Configuration/ConfigValidator.cs
```diff
@@ -287,6 +288,47 @@ private async ValueTask ValidateAuthorizationPolicyAsync(IList<Exception> errors
...
 
+    private ValueTask ValidateRateLimiterPolicyAsync(IList<Exception> errors, string? rateLimiterPolicyName, string routeId)
+    {
+        if (string.IsNullOrEmpty(rateLimiterPolicyName))
+        {
+            //return;
+            return ValueTask.CompletedTask;
+        }
+
+        // TODO: update this once AspNetCore provides a mechanism to validate the RateLimiter policies (maybe .NET8?)
```
Tratcher, line 299: Link to an issue in AspNetCore for this.

docs/docfx/articles/rate-limiting.md
```diff
@@ -0,0 +1,78 @@
...
+        endpoints.MapReverseProxy();
+    });
+}
+```
+
+See the [Rate Limiting](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit) docs for setting up your preferred kind of rate limiting.
+
+### Disable CORS
+
+Specifying the value `disable` in a route's `RateLimiterPolicy` parameter means the rate limiter middleware will not rate limit requests.
```
Tratcher, line 78:
```suggestion
Specifying the value `disable` in a route's `RateLimiterPolicy` parameter means the rate limiter middleware will not apply any policies to this route, even the default policy.
```

docs/docfx/articles/rate-limiting.md
```diff
@@ -0,0 +1,78 @@
+# Rate Limiting
+
+## Introduction
+The reverse proxy can be used to rate-limit requests before they are proxied to the destination servers. This can reduce load on the destination servers, add a layer of protection, and ensure consistent policies are implemented across your applications.
+
+## Defaults
+
+No rate limiting is performed on requests unless enabled in the route or application configuration.
```
Tratcher, line 8: Clarify that the middleware can have a default limiter applied to all routes, and this doesn't require any opt-in from the config.

src/ReverseProxy/Configuration/ConfigValidator.cs
```diff
@@ -287,6 +288,47 @@ private async ValueTask ValidateAuthorizationPolicyAsync(IList<Exception> errors
         }
     }
 
+    private ValueTask ValidateRateLimiterPolicyAsync(IList<Exception> errors, string? rateLimiterPolicyName, string routeId)
+    {
+        if (string.IsNullOrEmpty(rateLimiterPolicyName))
```
Tratcher, line 293: would it be clearer and more consistent if we supported a value of 'default' here and just no-op'd?

Tratcher (commented):
Very nice

docs/docfx/articles/rate-limiting.md
mburumaxwell: Done

src/ReverseProxy/Configuration/ConfigValidator.cs
mburumaxwell: Done

docs/docfx/articles/rate-limiting.md
```diff
@@ -0,0 +1,85 @@
...
+        endpoints.MapReverseProxy();
+    });
+}
+```
+
+See the [Rate Limiting](https://learn.microsoft.com/aspnet/core/performance/rate-limit) docs for setting up your preferred kind of rate limiting.
+
+### Disable Rate Limiting
+
+Specifying the value `disable` in a route's `RateLimiterPolicy` parameter means the rate limiter middleware will not  apply any policies to this route, even the default policy.
```
Kahbazi, line 85:
```suggestion
Specifying the value `disable` in a route's `RateLimiterPolicy` parameter means the rate limiter middleware will not apply any policies to this route, even the default policy.
```

src/ReverseProxy/Configuration/ConfigValidator.cs
```diff
@@ -287,6 +288,60 @@ private async ValueTask ValidateAuthorizationPolicyAsync(IList<Exception> errors
...
+            //return;
+            return ValueTask.CompletedTask;
+        }
+
+        // TODO: update this once AspNetCore provides a mechanism to validate the RateLimiter policies https://github.com/dotnet/aspnetcore/issues/45684
+
+        if (string.Equals(RateLimitingConstants.Default, rateLimiterPolicyName, StringComparison.OrdinalIgnoreCase))
+        {
+#if NET7_0_OR_GREATER
+            //var policy = await _rateLimiterPolicyProvider.GetPolicyAsync(rateLimiterPolicyName);
```
Kahbazi, line 304: Since this won't be added in 7, should it use reflection to get the policies? 😟

src/ReverseProxy/Configuration/RouteConfig.cs
```diff
@@ -44,6 +44,13 @@ public sealed record RouteConfig
     /// </summary>
     public string? AuthorizationPolicy { get; init; }
 
+    /// <summary>
+    /// The name of the RateLimiterPolicy to apply to this route.
+    /// If not set then only the GlobalLimiter will apply.
+    /// Set to "Disable" to disable rate limiting for this route.
```
Tratcher, line 50:
```suggestion
    /// Set to "Disable" to disable rate limiting for this route.
    /// Set to "Default" or leave empty use the global rate limits, if any.
```

Tratcher (approved):

src/ReverseProxy/Configuration/ConfigValidator.cs
Tratcher: Sure, see how messy that gets.

Tratcher: @mburumaxwell did you want to try to do this validation via reflection?

mburumaxwell: Done

MihaZupan (approved):
Thanks!

src/ReverseProxy/Configuration/IYarpRateLimiterPolicyProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+#endif
+using Microsoft.Extensions.Options;
+
+namespace Yarp.ReverseProxy.Configuration;
+
+// TODO: update this once AspNetCore provides a mechanism to validate the RateLimiter policies https://github.com/dotnet/aspnetcore/issues/45684
+
+#if NET7_0_OR_GREATER
+
+internal interface IYarpRateLimiterPolicyProvider
```
Kahbazi, line 18: Since this service is internal, is there any reason for an interface? I suggest removing the interface and use `YarpRateLimiterPolicyProvider` directly.

src/ReverseProxy/Configuration/IYarpRateLimiterPolicyProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        _rateLimiterOptions = rateLimiterOptions?.Value ?? throw new ArgumentNullException(nameof(rateLimiterOptions));
+
+        var type = typeof(RateLimiterOptions);
+        var flags = BindingFlags.Instance | BindingFlags.NonPublic;
+        _policyMap = (System.Collections.IDictionary)type.GetProperty("PolicyMap", flags)!.GetValue(_rateLimiterOptions, null)!;
+        _unactivatedPolicyMap = (System.Collections.IDictionary)type.GetProperty("UnactivatedPolicyMap", flags)!.GetValue(_rateLimiterOptions, null)!;
+    }
+
+    public ValueTask<object?> GetPolicyAsync(string policyName)
```
Kahbazi, line 39: Does this need to be async?

mburumaxwell: Suspicion is that the implementation on the framework when there will by asynchrony because some policies may need activation asynchronously.

src/ReverseProxy/Configuration/IYarpRateLimiterPolicyProvider.cs
mburumaxwell: Just but a preference aligned with the hope that it is added to the framework sooner rather than later.

Tratcher: This is an OK way to hide the reflection logic for now. We can revisit when there's a new implementation available.

I think you can remove most of these #if's if you make IYarpRateLimiterPolicyProvider available and then provide an alternate implementation for Net60 that just returns null.

Tratcher:
@mburumaxwell I pushed some fixes to your branch so we can get this into the upcoming preview release. Thanks for helping out.

mburumaxwell:
> @mburumaxwell I pushed some fixes to your branch so we can get this into the upcoming preview release. Thanks for helping out.

Okay. Looking forward. Should the same workflow be used for output caching? (Also introduced in .NET7)

MihaZupan (approved):

PR #1970, opened 2022-12-20
fix MatchesPort when service port name and targetPort name not same

### LINKED ISSUE #1969: MatchesPort miss the endpoint port  when the service port name and targetPort name  not same
when the service.yaml port like this :

```yaml
apiVersion: v1
kind: Service
metadata:
  name: backend-service
  namespace: default
spec:
  ports:
  - port: 88
    protocol: TCP
    name: my-http
    targetPort: http
  selector:
    app.kubernetes.io/instance: backend
    app.kubernetes.io/name: helm-charts-backend
  sessionAffinity: None
  type: ClusterIP
```

the endpoints like this:
```yaml
apiVersion: v1
kind: Endpoints
metadata:
  name: backend-service
  namespace: default
subsets:
- addresses:
  - ip: 10.1.1.26
    nodeName: docker-desktop
    targetRef:
      kind: Pod
      name: backend-helm-charts-backend-6c95487d84-2vsll
      namespace: default
      uid: d70405a1-32e9-4dc3-8481-a40d05f2b0ac
  ports:
  - port: 80
    protocol: TCP
    name: my-http
```

This line will return false:
https://github.com/microsoft/reverse-proxy/blob/65af2875bb552402575d49a225937ae5fce274e4/src/Kubernetes.Controller/Converters/YarpParser.cs#L87

witskeeper:
Closes #1969


MihaZupan (approved):
Thanks! Mind adding a test for this as well?

witskeeper:
> Thanks! Mind adding a test for this as well?

done

PR #1975, opened 2022-12-22
internal SetPassiveAsync can't be used

NZE:
In the example of FirstUnsuccessfulResponseHealthPolicy, there are following issues need to be address with the code:
1. public void RequestProxied function was provided parameters with wrong order
2. SetPassiveAsync is internal function
3. IProxyErrorFeature cannot be found, should be IForwarderErrorFeature


Tratcher (approved):
Thanks

PR #1978, opened 2022-12-22
WebSockets over HTTP/2 

### LINKED ISSUE #1375: WebSockets over HTTP/2

Tratcher:
Fixes #1375

Before this change the forwarder assumed that WebSockets always had to be forwarded using HTTP/1.1. Now it will honor the version and policy settings to decide which outgoing protocol to use. Request and response headers are adapted as needed. If HTTP/2 doesn't work then it can re-try with HTTP/1.1.

Notes:
- SocketsHttpHandler auto-flushes writes for HTTP/1.1 WebSockets, but not for HTTP/2. We have to manually flush writes.
- Something is hanging when YARP tries HTTP/2 prior knowledge to a non-TLS HTTP/1.1 endpoint. https://github.com/dotnet/runtime/issues/80056
- HttpSys is used as a negative test server because it doesn't support H2WS. This requires TLS, so we use the ports registered for IIS Express's certificate.
- Added a lot of logging infrastructure
- Updated Docs


src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -270,23 +319,83 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
...
+            destinationRequest.VersionPolicy = HttpVersionPolicy.RequestVersionExact;
+            destinationRequest.Method = HttpMethod.Connect;
+            destinationRequest.Headers.Protocol = connectProtocol ?? "websocket";
+        }
+#endif
+        else
+        {
+            destinationRequest.Method = RequestUtilities.GetHttpMethod(context.Request.Method);
+            destinationRequest.Version = requestConfig?.Version ?? DefaultVersion;
+            destinationRequest.VersionPolicy = requestConfig?.VersionPolicy ?? DefaultVersionPolicy;
```
MihaZupan, line 397:
```suggestion
            destinationRequest.Version = outgoingVersion;
            destinationRequest.VersionPolicy = outgoingPolicy;
```

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -141,6 +143,45 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
...
+                    Log.RetryingWebSocketDowngradeNoHttp2(_logger);
+                }
+                else
+                {
+                    return await HandleRequestFailureAsync(context, requestContent, hre, transformer, activityCancellationSource);
+                }
+
+                // Trying again
+                activityCancellationSource.ResetTimeout();
+
```
MihaZupan, line 163:
```suggestion

                Debug.Assert(requestContent is null || !requestContent.Started);

```

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -141,6 +143,45 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
...
             }
+            catch (HttpRequestException hre) when (tryDowngradingH2WsOnFailure)
+            {
+                if (hre.Data.Contains("SETTINGS_ENABLE_CONNECT_PROTOCOL"))
+                {
+                    Log.RetryingWebSocketDowngradeNoConnect(_logger);
+                }
+                else if (hre.Data.Contains("HTTP2_ENABLED"))
+                {
+                    Log.RetryingWebSocketDowngradeNoHttp2(_logger);
```
MihaZupan, line 154:
```suggestion
                    Debug.Assert(hre.Data["HTTP2_ENABLED"] == false);
                    Log.RetryingWebSocketDowngradeNoHttp2(_logger);
```

src/ReverseProxy/Forwarder/ProtocolHelper.cs
```diff
@@ -37,4 +40,39 @@ public static string GetVersionPolicy(HttpVersionPolicy policy)
...
+        var success = Guid.NewGuid().TryWriteBytes(bytes);
+        Debug.Assert(success);
+        var secKey = Convert.ToBase64String(bytes);
+        return secKey;
+    }
+
+    /// <summary>
+    /// Creates the Accept response to a given security key for sending in or verifying the Sec-WebSocket-Accept header value.
+    /// </summary>
+    internal static string CreateSecWebSocketAccept(string? key)
```
MihaZupan, line 60: This should validate that the key is in the expected format (at least that the length matches) to avoid throwing an unhelpful exception from GetBytes below

MihaZupan (commented):
Nice! I take it the telemetry for H2 WebSockets also "just work ™️" because ASP.NET exposes them the same way as H1?

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -138,10 +134,233 @@ public async Task FalseUpgradeTest()
...
+        {
+            using var client = new ClientWebSocket();
+            var webSocketsTarget = uri.Replace("https://", "wss://").Replace("http://", "ws://");
+            var targetUri = new Uri(new Uri(webSocketsTarget, UriKind.Absolute), "websockets");
+#if NET7_0_OR_GREATER
+            using var invoker = CreateInvoker();
+            var wse = await Assert.ThrowsAsync<WebSocketException>(() => client.ConnectAsync(targetUri, invoker, cts.Token));
+#else
+            client.Options.RemoteCertificateValidationCallback = (_, _, _, _) => true;
+            var wse = await Assert.ThrowsAsync<WebSocketException>(() => await client.ConnectAsync(targetUri, cts.Token));
```
MihaZupan, line 314:
```suggestion
            var wse = await Assert.ThrowsAsync<WebSocketException>(() => client.ConnectAsync(targetUri, cts.Token));
```

I may be missing something, but shouldn't this break the build in CI?

src/ReverseProxy/Forwarder/HttpForwarder.cs
Tratcher: tryDowngradingH2WsOnFailure should mean there's never a request content.

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
Tratcher: ThrowsAsync awaits the returned task regardless.

Tratcher:
> Nice! I take it the telemetry for H2 WebSockets also "just work ™️" because ASP.NET exposes them the same way as H1?

Nope, had to add it there too 😁

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -141,6 +143,45 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
...
+                }
+                else
+                {
+                    return await HandleRequestFailureAsync(context, requestContent, hre, transformer, activityCancellationSource);
+                }
+
+                // Trying again
+                activityCancellationSource.ResetTimeout();
+
+                var config = requestConfig! with
```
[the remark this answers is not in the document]
Tratcher: It missed it so we do need this override.

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -138,10 +134,233 @@ public async Task FalseUpgradeTest()
...
+        await test.Invoke(async uri =>
+        {
+            using var client = new ClientWebSocket();
+            await SendWebSocketRequestAsync(client, uri, "HTTP/1.1", cts.Token);
+        }, cts.Token);
+    }
+
+    [Theory]
+    [InlineData(HttpVersionPolicy.RequestVersionExact, true)]
+    // [InlineData(HttpVersionPolicy.RequestVersionExact, false)] HttpClient bug causes this to time out?
```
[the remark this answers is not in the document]
Tratcher: Yeah, I need to dig into this. It doesn't directly affect this feature, only this negative test case.

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -135,17 +137,51 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
...
+                    if (hre.Data.Contains("SETTINGS_ENABLE_CONNECT_PROTOCOL"))
+                    {
+                        Debug.Assert(false == (bool?)hre.Data["SETTINGS_ENABLE_CONNECT_PROTOCOL"]);
+                        Log.RetryingWebSocketDowngradeNoConnect(_logger);
+                    }
+                    else if (hre.Data.Contains("HTTP2_ENABLED"))
+                    {
+                        Debug.Assert(false == (bool?)hre.Data["HTTP2_ENABLED"]);
+                        Log.RetryingWebSocketDowngradeNoHttp2(_logger);
+                    }
```
davidfowl, line 157: Add big comments here. This is ugly. @MihaZupan can we improve this with more structured data on the exception?

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -135,17 +137,51 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
...
-                // Reset the timeout since we received the response headers.
-                activityCancellationSource.ResetTimeout();
+                try
+                {
+                    destinationResponse = await httpClient.SendAsync(destinationRequest, activityCancellationSource.Token);
+                }
+                catch (HttpRequestException hre) when (tryDowngradingH2WsOnFailure)
+                {
+                    Debug.Assert(requestContent is null);
+                    if (hre.Data.Contains("SETTINGS_ENABLE_CONNECT_PROTOCOL"))
```
Tratcher, line 148:
```suggestion
                    // This is how SocketsHttpHandler communicates to us that we tried a HTTP/2 extension that wasn't
                    // enabled by the server. We should retry on HTTP/1.1.
                    if (hre.Data.Contains("SETTINGS_ENABLE_CONNECT_PROTOCOL"))
```

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -135,17 +137,51 @@ public HttpForwarder(ILogger<HttpForwarder> logger, IClock clock)
...
+                    {
+                        Debug.Assert(false == (bool?)hre.Data["SETTINGS_ENABLE_CONNECT_PROTOCOL"]);
+                        Log.RetryingWebSocketDowngradeNoConnect(_logger);
+                    }
+                    else if (hre.Data.Contains("HTTP2_ENABLED"))
```
Tratcher, line 153:
```suggestion
                    // This is how SocketsHttpHandler communicates to us that we tried HTTP/2, but the server only supports 
                    // HTTP/1.x (as determined by ALPN). We'll only get this when using TLS/https. Retry on HTTP/1.1.
                    // We don't let SocketsHttpHandler downgrade automatically for us because we need to send different headers.
                    else if (hre.Data.Contains("HTTP2_ENABLED"))
```

src/ReverseProxy/Forwarder/HttpForwarder.cs
MihaZupan: I'm not a fan of it either, but we don't yet have a good story for exposing programmatically-consumable/strongly-typed information on exceptions. It's certainly worth thinking about this case too as part of https://github.com/dotnet/runtime/issues/76644#issuecomment-1367521636, which we're (hopefully) looking into in .NET 8.

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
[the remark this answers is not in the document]
Tratcher: https://github.com/dotnet/runtime/issues/80056

Tratcher, line 293:
```suggestion
    // [InlineData(HttpVersionPolicy.RequestVersionExact, false)] // Times out. https://github.com/dotnet/runtime/issues/80056
```

test/ReverseProxy.FunctionalTests/WebSocketTests.cs
```diff
@@ -138,10 +135,276 @@ public async Task FalseUpgradeTest()
...
+        await test.Invoke(async uri =>
+        {
+            using var client = new ClientWebSocket();
+            await SendWebSocketRequestAsync(client, uri, "HTTP/1.1", cts.Token);
+        }, cts.Token);
+    }
+
+    [Theory]
+    [InlineData(HttpVersionPolicy.RequestVersionExact, true)]
+    // [InlineData(HttpVersionPolicy.RequestVersionExact, false)] // Times out. https://github.com/dotnet/runtime/issues/80056
```
MihaZupan, line 293:
```suggestion
#if NET8_0_OR_GREATER
    [InlineData(HttpVersionPolicy.RequestVersionExact, false)] // Times out. https://github.com/dotnet/runtime/issues/80056
#endif
```
I have a feeling we'll forget to re-enable it otherwise

src/ReverseProxy/Utilities/EventIds.cs
```diff
@@ -64,4 +64,7 @@ internal static class EventIds
     public static readonly EventId Http10RequestVersionDetected = new EventId(58, "Http10RequestVersionDetected");
     public static readonly EventId NotForwarding = new EventId(59, "NotForwarding");
     public static readonly EventId MaxRequestBodySizeSet = new EventId(60, "MaxRequestBodySizeSet");
+    public static readonly EventId RetryingWebSocketDowngradeNoConnect = new EventId(61, "RetryingWebSocketDowngradeNoConnect");
+    public static readonly EventId RetryingWebSocketDowngradeNoHttp2 = new EventId(62, "RetryingWebSocketDowngradeNoHttp2");
+    public static readonly EventId InvalidSecWebSocketKeyHeader = new EventId(63, "RetryingWebSocketDowngradeNoHttp2");
```
MihaZupan, line 69:
```suggestion
    public static readonly EventId InvalidSecWebSocketKeyHeader = new EventId(63, "InvalidSecWebSocketKeyHeader");
```

src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -642,6 +771,64 @@ ForwarderError ReportResult(HttpContext context, bool request, StreamCopyResult
...
+    private ForwarderError FixupUpgradeResponseHeaders(HttpContext context, HttpResponseMessage response, bool isHttp2Request)
+    {
+        if (isHttp2Request)
+        {
+            // H2 <- H1 Validate & remove the Sec-WebSocket-Accept header.
+            if (response.Version != HttpVersion.Version20)
+            {
+                var key = response.RequestMessage!.Headers.GetValues(HeaderNames.SecWebSocketKey);
+                var accept = context.Response.Headers[HeaderNames.SecWebSocketAccept];
+                var expectedAccept = ProtocolHelper.CreateSecWebSocketAccept(key.FirstOrDefault());
```
MihaZupan, line 784:
```suggestion
                var success = RequestUtilities.TryGetValues(response.RequestMessage!.Headers, HeaderNames.SecWebSocketKey, out var key);
                Debug.Assert(success);
                var accept = context.Response.Headers[HeaderNames.SecWebSocketAccept];
                var expectedAccept = ProtocolHelper.CreateSecWebSocketAccept(key.ToString());
```

MihaZupan (approved):

PR #1979, opened 2022-12-23
Add extension methods for adding direct forwarding to the ASP.NET Core request pipeline (#1972)

### LINKED ISSUE #1972: Contributing Extensions Methods for Easier in Code Configuration of Direct Forwarding

Currently, [configuring direct forwarding in code](https://microsoft.github.io/reverse-proxy/articles/direct-forwarding.html#update-startup) is quite verbose. I could contribute extension methods which could simplify that. An example:

```cs
public static class EndpointRouteBuilderExtensions
{
    private static HttpMessageInvoker? _httpClient;

    private static HttpMessageInvoker HttpClient
    {
        get
        {
            if (_httpClient is null)
            {
                 _httpClient = new HttpMessageInvoker(new SocketsHttpHandler()
                {
                    UseProxy = false,
                    AllowAutoRedirect = false,
                    UseCookies = false,
                    ActivityHeadersPropagator = new ReverseProxyPropagator(DistributedContextPropagator.Current)
                });
            }

            return _httpClient;
        }
    }

    public static void MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string serviceUrl)
    {
        var forwarder = endpoints.ServiceProvider.GetRequiredService<IHttpForwarder>();
        var requestConfig = new ForwarderRequestConfig { ActivityTimeout = TimeSpan.FromMilliseconds(500) };

        endpoints.Map(pattern, async httpContext =>
        {
            var error = await forwarder.SendAsync(httpContext, serviceUrl, HttpClient, requestConfig, HttpTransformer.Default);

            if (error != ForwarderError.None)
            {
                var errorFeature = httpContext.GetForwarderErrorFeature();
                throw new Exception("An error has occurred while forwarding the request", errorFeature?.Exception);
            }
        });
    }
}
```

A combination of overloads should probably be established as well to enable passing instances of `HttpMessageInvoker`, `ForwarderRequestConfig`, `HttpTransformer`, or maybe some of their properties (like timeout).

In cases where YARP is used as an integrated part of the solution (instead of being a black-box generic proxy component), it's sometimes preferred to configure routing via code. This allows to expose only necessary options instead of entire proxy logic.

cc @Tratcher

tpeczek:
Closes #1972

### DIFF
--- a/src/ReverseProxy/Forwarder/DirectForwardingHttpClientProvider.cs
+++ b/src/ReverseProxy/Forwarder/DirectForwardingHttpClientProvider.cs
@@ -1,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Net.Http;
+using Yarp.ReverseProxy.Configuration;
+
+namespace Yarp.ReverseProxy.Forwarder;
+
+internal class DirectForwardingHttpClientProvider
+{
+    public HttpMessageInvoker HttpClient { get; }
+
+    public DirectForwardingHttpClientProvider() : this(new ForwarderHttpClientFactory()) { }
+
+    public DirectForwardingHttpClientProvider(IForwarderHttpClientFactory factory)
+    {
+        HttpClient = factory.CreateClient(new ForwarderHttpClientContext
+        {
+            NewConfig = HttpClientConfig.Empty
+        });
+    }
+}
--- a/src/ReverseProxy/Forwarder/DirectForwardingHttpClientProvider.cs
+++ b/src/ReverseProxy/Forwarder/DirectForwardingHttpClientProvider.cs
@@ -6,7 +6,7 @@ using Yarp.ReverseProxy.Configuration;
 
 namespace Yarp.ReverseProxy.Forwarder;
 
-internal class DirectForwardingHttpClientProvider
+internal sealed class DirectForwardingHttpClientProvider
 {
     public HttpMessageInvoker HttpClient { get; }
 
--- a/src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
+++ b/src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
@@ -30,6 +30,7 @@ public static class ReverseProxyServiceCollectionExtensions
     {
         services.TryAddSingleton<IClock, Clock>();
         services.TryAddSingleton<IHttpForwarder, HttpForwarder>();
+        services.TryAddSingleton<DirectForwardingHttpClientProvider>();
         services.TryAddSingleton<ITransformBuilder, TransformBuilder>();
         return services;
     }
--- a/src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
+++ b/src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
@@ -30,8 +30,10 @@ public static class ReverseProxyServiceCollectionExtensions
     {
         services.TryAddSingleton<IClock, Clock>();
         services.TryAddSingleton<IHttpForwarder, HttpForwarder>();
-        services.TryAddSingleton<DirectForwardingHttpClientProvider>();
         services.TryAddSingleton<ITransformBuilder, TransformBuilder>();
+
+        services.AddSingleton<DirectForwardingHttpClientProvider>();
+
         return services;
     }
 
--- a/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
+++ b/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
@@ -3,12 +3,9 @@
 
 using System;
 using System.Net.Http;
-using System.Runtime.CompilerServices;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 using Yarp.ReverseProxy.Forwarder;
-using Yarp.ReverseProxy.Configuration;
 
 namespace Microsoft.AspNetCore.Builder;
 
@@ -17,8 +14,6 @@ namespace Microsoft.AspNetCore.Builder;
 /// </summary>
 public static class DirectForwardingIEndpointRouteBuilderExtensions
 {
-    private static readonly ConditionalWeakTable<IServiceProvider, HttpMessageInvoker> _httpClients = new();
-
     /// <summary>
     /// Adds direct forwarding of HTTP requests that match the specified pattern to a specific destination using default configuration for the outgoing request, default transforms, and default HTTP client.
     /// </summary>
@@ -40,9 +35,9 @@ public static class DirectForwardingIEndpointRouteBuilderExtensions
     /// </summary>
     public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string destinationPrefix, ForwarderRequestConfig requestConfig, HttpTransformer transformer)
     {
-        var httpClient = GetHttpClient(endpoints.ServiceProvider);
+        var httpClientProvider = endpoints.ServiceProvider.GetRequiredService<DirectForwardingHttpClientProvider>();
 
-        return endpoints.MapForwarder(pattern, destinationPrefix, requestConfig, transformer, httpClient);
+        return endpoints.MapForwarder(pattern, destinationPrefix, requestConfig, transformer, httpClientProvider.HttpClient);
     }
 
     /// <summary>
@@ -62,22 +57,5 @@ public static class DirectForwardingIEndpointRouteBuilderExtensions
         {
             await forwarder.SendAsync(httpContext, destinationPrefix, httpClient, requestConfig, transformer);
         });
-    }
-
-    private static HttpMessageInvoker GetHttpClient(IServiceProvider serviceProvider)
-    {
-        lock (_httpClients)
-        {
-            return _httpClients.GetValue(serviceProvider, static serviceProvider =>
-            {
-                var httpClientFactory = serviceProvider.GetService<IForwarderHttpClientFactory>()
-                    ?? new ForwarderHttpClientFactory(serviceProvider.GetRequiredService<ILogger<ForwarderHttpClientFactory>>());
-
-                return httpClientFactory.CreateClient(new ForwarderHttpClientContext
-                {
-                    NewConfig = HttpClientConfig.Empty
-                });
-            });
-        }
     }
 }
--- a/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
+++ b/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
@@ -50,11 +50,11 @@ public static class DirectForwardingIEndpointRouteBuilderExtensions
     /// </summary>
     public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string destinationPrefix, ForwarderRequestConfig requestConfig, HttpTransformer transformer, HttpMessageInvoker httpClient)
     {
-        ArgumentNullException.ThrowIfNull(endpoints, nameof(endpoints));
-        ArgumentNullException.ThrowIfNull(destinationPrefix, nameof(destinationPrefix));
-        ArgumentNullException.ThrowIfNull(httpClient, nameof(httpClient));
-        ArgumentNullException.ThrowIfNull(requestConfig, nameof(requestConfig));
-        ArgumentNullException.ThrowIfNull(transformer, nameof(transformer));
+        ArgumentNullException.ThrowIfNull(endpoints);
+        ArgumentNullException.ThrowIfNull(destinationPrefix);
+        ArgumentNullException.ThrowIfNull(httpClient);
+        ArgumentNullException.ThrowIfNull(requestConfig);
+        ArgumentNullException.ThrowIfNull(transformer);
 
         var forwarder = endpoints.ServiceProvider.GetRequiredService<IHttpForwarder>();
 
--- a/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
+++ b/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
@@ -1,7 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System;
 using System.Net.Http;
+using System.Runtime.CompilerServices;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
@@ -15,6 +17,8 @@ namespace Microsoft.AspNetCore.Builder;
 /// </summary>
 public static class DirectForwardingIEndpointRouteBuilderExtensions
 {
+    private static readonly ConditionalWeakTable<IServiceProvider, HttpMessageInvoker> _httpClients = new();
+
     /// <summary>
     /// Adds direct forwarding of HTTP requests that match the specified pattern to a specific destination using default configuration for the outgoing request, default transforms, and default HTTP client.
     /// </summary>
@@ -36,13 +40,7 @@ public static class DirectForwardingIEndpointRouteBuilderExtensions
     /// </summary>
     public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string destinationPrefix, ForwarderRequestConfig requestConfig, HttpTransformer transformer)
     {
-        var httpClientFactory = endpoints.ServiceProvider.GetService<IForwarderHttpClientFactory>()
-            ?? new ForwarderHttpClientFactory(endpoints.ServiceProvider.GetRequiredService<ILogger<ForwarderHttpClientFactory>>());
-
-        var httpClient = httpClientFactory.CreateClient(new ForwarderHttpClientContext
-        {
-            NewConfig = HttpClientConfig.Empty
-        });
+        var httpClient = GetHttpClient(endpoints.ServiceProvider);
 
         return endpoints.MapForwarder(pattern, destinationPrefix, requestConfig, transformer, httpClient);
     }
@@ -52,11 +50,34 @@ public static class DirectForwardingIEndpointRouteBuilderExtensions
     /// </summary>
     public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string destinationPrefix, ForwarderRequestConfig requestConfig, HttpTransformer transformer, HttpMessageInvoker httpClient)
     {
+        ArgumentNullException.ThrowIfNull(endpoints, nameof(endpoints));
+        ArgumentNullException.ThrowIfNull(destinationPrefix, nameof(destinationPrefix));
+        ArgumentNullException.ThrowIfNull(httpClient, nameof(httpClient));
+        ArgumentNullException.ThrowIfNull(requestConfig, nameof(requestConfig));
+        ArgumentNullException.ThrowIfNull(transformer, nameof(transformer));
+
         var forwarder = endpoints.ServiceProvider.GetRequiredService<IHttpForwarder>();
 
         return endpoints.Map(pattern, async httpContext =>
         {
             await forwarder.SendAsync(httpContext, destinationPrefix, httpClient, requestConfig, transformer);
         });
+    }
+
+    private static HttpMessageInvoker GetHttpClient(IServiceProvider serviceProvider)
+    {
+        lock (_httpClients)
+        {
+            return _httpClients.GetValue(serviceProvider, static serviceProvider =>
+            {
+                var httpClientFactory = serviceProvider.GetService<IForwarderHttpClientFactory>()
+                    ?? new ForwarderHttpClientFactory(serviceProvider.GetRequiredService<ILogger<ForwarderHttpClientFactory>>());
+
+                return httpClientFactory.CreateClient(new ForwarderHttpClientContext
+                {
+                    NewConfig = HttpClientConfig.Empty
+                });
+            });
+        }
     }
 }
--- a/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
+++ b/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
@@ -1,14 +1,12 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-using System;
-using System.Diagnostics;
-using System.Net;
 using System.Net.Http;
-using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Yarp.ReverseProxy.Forwarder;
+using Yarp.ReverseProxy.Configuration;
 
 namespace Microsoft.AspNetCore.Builder;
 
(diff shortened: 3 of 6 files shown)


tpeczek:

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        get
+        {
+            _defaultTransport ??= new HttpMessageInvoker(new SocketsHttpHandler()
+            {
+                UseProxy = false,
+                AllowAutoRedirect = false,
+                AutomaticDecompression = DecompressionMethods.None,
+                UseCookies = false,
+                ActivityHeadersPropagator = new ReverseProxyPropagator(DistributedContextPropagator.Current)
```
davidfowl, line 32: Is this the YARP default?

tpeczek: This is in line with what [CreateClient](https://github.com/microsoft/reverse-proxy/blob/65af2875bb552402575d49a225937ae5fce274e4/src/ReverseProxy/Forwarder/ForwarderHttpClientFactory.cs#L44) does by default. Using `IForwarderHttpClientFactory` here might be a better alternative - I would need some guidance around instantiating `ForwarderHttpClientContext` for this context.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// </summary>
+    public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string serviceUrl, ForwarderRequestConfig requestConfig, HttpTransformer transforms)
+    {
+        return endpoints.MapForwarder(pattern, serviceUrl, requestConfig, transforms, DefaultTransport);
+    }
+
+    /// <summary>
+    /// Adds direct forwarding of HTTP requests that match the specified pattern to a specific destination using customized configuration for the outgoing request, customized transforms, and customized HTTP client.
+    /// </summary>
+    public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string serviceUrl, ForwarderRequestConfig requestConfig, HttpTransformer transforms, HttpMessageInvoker transport)
```
MihaZupan, line 66:
```suggestion
    public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string destinationPrefix, ForwarderRequestConfig requestConfig, HttpTransformer transformer, HttpMessageInvoker httpClient)
```

These should match the naming from https://github.com/microsoft/reverse-proxy/blob/main/src/ReverseProxy/Forwarder/IHttpForwarder.cs

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        return endpoints.Map(pattern, async httpContext =>
+        {
+            var error = await forwarder.SendAsync(httpContext, serviceUrl, transport, requestConfig, transforms);
+
+            if (error != ForwarderError.None)
+            {
+                var errorFeature = httpContext.GetForwarderErrorFeature();
+                throw new Exception("An error has occurred while forwarding the request", errorFeature?.Exception);
+            }
```
MihaZupan, line 78:
```suggestion
            await forwarder.SendAsync(httpContext, destinationPrefix, httpClient, requestConfig, transformer);
```

I don't think we should throw within the pipeline again. We'll already log the error to `ILogger` if one occurred.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
Tratcher: My bad 😁

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+namespace Microsoft.AspNetCore.Builder;
+
+/// <summary>
+/// Extension methods for <see cref="IEndpointRouteBuilder"/> used to add direct forwarding to the ASP.NET Core request pipeline.
+/// </summary>
+public static class DirectForwardingIEndpointRouteBuilderExtensions
+{
+    private static HttpMessageInvoker? _defaultTransport;
+
+    private static HttpMessageInvoker DefaultTransport
```
Tratcher, line 22: @MihaZupan any concerns about having a static instance here?

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
Tratcher: Try: `new ForwarderHttpClientContext().CreateClient(new ForwarderHttpClientContext());`. I don't think any of the values are required.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
MihaZupan: It should be fine that we're using a single instance here. If you were to define a config with a single cluster and never changed it, you would effectively get the same thing.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        get
+        {
+            _defaultTransport ??= new HttpMessageInvoker(new SocketsHttpHandler()
```
MihaZupan, line 26: This should guard against allocating (and not disposing of) multiple instances in a race condition.
The simplest option is likely to just allocate the instance upfront (have a `static readonly` field).

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
tpeczek: It seems to require `NewConfig` for which I can provide `HttpClientConfig.Empty`, so I'm going with this unless we decide otherwise in different thread.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
tpeczek: Removed 😊.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
tpeczek: In latest commit I've changed this to use `IForwarderHttpClientFactory` (from DI if available) with basically empty `ForwarderHttpClientContext`. Let me know if you want me to revert that to a static instance (which I would grab from *new* `ForwarderHttpClientFactory`).

MihaZupan: Honoring custom `IForwarderHttpClientFactory` instances is even better, but as-is it has the issue that each `MapForwarder` call will allocate its own handler instance, and you'll end up with more connection pools than you might need.

It may be worth caching the instance based on the service provider, e.g.
```c#
private static readonly ConditionalWeakTable<IServiceProvider, HttpMessageInvoker> _invokers = new();

private static HttpMessageInvoker GetInvoker(IServiceProvider serviceProvider)
{
    lock (_invokers)
    {
        return _invokers.GetValue(serviceProvider, static serviceProvider =>
        {
            var httpClientFactory = serviceProvider.GetService<IForwarderHttpClientFactory>()
                ?? new ForwarderHttpClientFactory(serviceProvider.GetRequiredService<ILogger<ForwarderHttpClientFactory>>());

            return httpClientFactory.CreateClient(new ForwarderHttpClientContext
            {
                NewConfig = HttpClientConfig.Empty
            });
        });
    }
}

public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string destinationPrefix, ForwarderRequestConfig requestConfig, HttpTransformer transformer)
{
    return endpoints.MapForwarder(pattern, destinationPrefix, requestConfig, transformer, GetInvoker(endpoints.ServiceProvider));
}
```

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        return endpoints.MapForwarder(pattern, destinationPrefix, requestConfig, transformer, httpClient);
+    }
+
+    /// <summary>
+    /// Adds direct forwarding of HTTP requests that match the specified pattern to a specific destination using customized configuration for the outgoing request, customized transforms, and customized HTTP client.
+    /// </summary>
+    public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string destinationPrefix, ForwarderRequestConfig requestConfig, HttpTransformer transformer, HttpMessageInvoker httpClient)
+    {
+        var forwarder = endpoints.ServiceProvider.GetRequiredService<IHttpForwarder>();
```
MihaZupan, line 55: Can you please add null checks here so that we throw early, before we start processing requests?

tpeczek: Null checks added.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
tpeczek: Implemented according to the suggestion 🙂

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// <summary>
+    /// Adds direct forwarding of HTTP requests that match the specified pattern to a specific destination using customized configuration for the outgoing request, customized transforms, and customized HTTP client.
+    /// </summary>
+    public static IEndpointConventionBuilder MapForwarder(this IEndpointRouteBuilder endpoints, string pattern, string destinationPrefix, ForwarderRequestConfig requestConfig, HttpTransformer transformer, HttpMessageInvoker httpClient)
+    {
+        ArgumentNullException.ThrowIfNull(endpoints, nameof(endpoints));
+        ArgumentNullException.ThrowIfNull(destinationPrefix, nameof(destinationPrefix));
+        ArgumentNullException.ThrowIfNull(httpClient, nameof(httpClient));
+        ArgumentNullException.ThrowIfNull(requestConfig, nameof(requestConfig));
+        ArgumentNullException.ThrowIfNull(transformer, nameof(transformer));
```
MihaZupan, line 57:
```suggestion
        ArgumentNullException.ThrowIfNull(endpoints);
        ArgumentNullException.ThrowIfNull(destinationPrefix);
        ArgumentNullException.ThrowIfNull(httpClient);
        ArgumentNullException.ThrowIfNull(requestConfig);
        ArgumentNullException.ThrowIfNull(transformer);
```
Nit: `ThrowIfNull` is using `CallerArgumentExpression` so you don't have to specify these manually

MihaZupan (approved):
We'll want to update docs/samples/tests as well, but implementation LGTM, thanks!

tpeczek:
> We'll want to update docs/samples/tests as well, but implementation LGTM, thanks!

@MihaZupan I can update code snippet in `direct-forwarding.md` and `ReverseProxy.Direct.Sample` to use extension methods. I also see one tests asset (`ReverseProxy.Direct`) which could be using them. I can update those, is there anything else you would have in mind?

Tratcher:
> > We'll want to update docs/samples/tests as well, but implementation LGTM, thanks!
> 
> @MihaZupan Miha Zupan FTE I can update code snippet in `direct-forwarding.md` and `ReverseProxy.Direct.Sample` to use extension methods. I also see one tests asset (`ReverseProxy.Direct`) which could be using them. I can update those, is there anything else you would have in mind?

Please show both options when you update these.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
tpeczek: Adjusted 🙂

tpeczek:
> > > We'll want to update docs/samples/tests as well, but implementation LGTM, thanks!
> > 
> > 
> > MihaZupan I can update code snippet in `direct-forwarding.md` and `ReverseProxy.Direct.Sample` to use extension methods. I also see one tests asset (`ReverseProxy.Direct`) which could be using them. I can update those, is there anything else you would have in mind?
> 
> Please show both options (old and new) when you update these. We still want people to understand how to use the lowest level direct forwarding.

I've pushed the first attempt, please let me know what to adjust.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+using Yarp.ReverseProxy.Configuration;
+
+namespace Microsoft.AspNetCore.Builder;
+
+/// <summary>
+/// Extension methods for <see cref="IEndpointRouteBuilder"/> used to add direct forwarding to the ASP.NET Core request pipeline.
+/// </summary>
+public static class DirectForwardingIEndpointRouteBuilderExtensions
+{
+    private static readonly ConditionalWeakTable<IServiceProvider, HttpMessageInvoker> _httpClients = new();
```
davidfowl, line 20: Can we avoid this?

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+        });
+    }
+
+    private static HttpMessageInvoker GetHttpClient(IServiceProvider serviceProvider)
+    {
+        lock (_httpClients)
+        {
+            return _httpClients.GetValue(serviceProvider, static serviceProvider =>
+            {
+                var httpClientFactory = serviceProvider.GetService<IForwarderHttpClientFactory>()
```
davidfowl, line 73: Why would it be null? Can we tell the user to call an API that will add it to the container?

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
MihaZupan: What would you suggest?
We want to reuse handler instances between `MapForwarder` calls in order to reuse the connection pool effectively.

davidfowl: https://github.com/microsoft/reverse-proxy/pull/1979#discussion_r1059157846

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
MihaZupan: This will always be included when you call `AddReverseProxy`/`AddProxy`, so this can be GetRequiredService instead (or throw an exception if `GetServices` returns null).

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
MihaZupan: I don't think that's the same issue. The `IForwarderHttpClientFactory` doesn't store the instances it creates, you still have to keep state around to avoid creating multiple clients.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
davidfowl: That sounds like a good plan 👍🏾

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
davidfowl: Why aren't we using a singleton then? What am I missing?

MihaZupan: Do you mean why this is a `ConditionalWeakTable<IServiceProvider, HttpMessageInvoker>` instead of a lazily-initialized static `HttpMessageInvoker` field?
You may provide different client factory implementations to different service collections in the same process (in tests if nowhere else).

davidfowl: Why not something in the DI container?

MihaZupan: I don't understand what you mean by that.

davidfowl: OK I re-read the code. I would say the requirement here is to end up with a singleton `HttpMesageInvoker` that can be used for these methods:

```C#
internal class DirectForwardingHttpInvoker
{
    public DirectForwardingHttpInvoker(IForwarderHttpClientFactory factory)
    {
          Invoker = factory.CreateClient(new ForwarderHttpClientContext
          {
              NewConfig = HttpClientConfig.Empty
          });
    }
    public HttpMessageInvoker Invoker { get; }
}
```

We'd register this by default with the call to AddReverseProxy/AddHttpForwarder etc. Then we just resolve `DirectForwardingHttpInvoker` and use the Invoker instance.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
tpeczek: The documentation and samples for direct forwarding are using only `AddHttpForwarder` which doesn't register `IForwarderHttpClientFactory`. So I guess the documentation would have to be adjusted to inform that you can use `AddHttpForwarder` only if you are providing `HttpMessageInvoker` and must use `AddProxy` when not. Would that be correct?

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
MihaZupan: Ah, I see, yeah that would work too.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
MihaZupan: Going with David's recommendation https://github.com/microsoft/reverse-proxy/pull/1979#discussion_r1059287214, you could make this work even if the user only called `AddHttpForwarder`.

We can make the factory parameter in `DirectForwardingHttpInvoker`'s ctor optional, falling back to the default factory if one wasn't set.

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
tpeczek: Do you want to have an interface for this class (assuming it would be only used in those extension methods it may not be needed for testability) and if yes, do you want that interface to be public or internal?

tpeczek: I went for implementation without an interface.

src/ReverseProxy/Forwarder/DirectForwardingHttpClientProvider.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Net.Http;
+using Yarp.ReverseProxy.Configuration;
+
+namespace Yarp.ReverseProxy.Forwarder;
+
+internal class DirectForwardingHttpClientProvider
```
Kahbazi, line 9:
```suggestion
internal sealed class DirectForwardingHttpClientProvider
```

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
Kahbazi: > OK I re-read the code. I would say the requirement here is to end up with a singleton `HttpMesageInvoker` that can be used for these methods:

@davidfowl Uhmm, shouldn't we get a new `HttpClient` instance from `IHttpClientFactory` each time the route hits? 
Quoting from the [docs ](https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#issues-with-the-original-httpclient-class-available-in-net):

> Another issue that developers run into is when using a shared instance of HttpClient in long-running processes. In a situation where the HttpClient is instantiated as a singleton or a static object, it fails to handle the DNS changes as described in this [issue](https://github.com/dotnet/runtime/issues/18348) of the dotnet/runtime GitHub repository.

src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs
```diff
@@ -30,6 +30,7 @@ public static IServiceCollection AddHttpForwarder(this IServiceCollection servic
     {
         services.TryAddSingleton<IClock, Clock>();
         services.TryAddSingleton<IHttpForwarder, HttpForwarder>();
+        services.TryAddSingleton<DirectForwardingHttpClientProvider>();
```
Kahbazi, line 33: Since `DirectForwardingHttpClientProvider` is internal and have no interface, there's no intent to replace this service with another service. Perhaps we could `AddSingleton` is enough in this situation.
```suggestion
        services.AddSingleton<DirectForwardingHttpClientProvider>();
```

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
MihaZupan: This is really a separate discussion from this PR - you get the same behavior if you use other `Map*` overloads as you're effectively using a single instance for a given cluster.

YARP is generally used as part of the infrastructure where the HttpClient isn't talking to external services. We can document that setting `PooledConnectionLifetime` is an option if you rely on DNS changes for destinations, but I don't think we should touch the defaults as that would impact overall proxy latency.
In any case, YARP should not use `IHttpClientFactory` as we want full control over the lifetime of handlers.

src/ReverseProxy/Forwarder/DirectForwardingHttpClientProvider.cs
tpeczek: Adjusted 🙂

src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs
tpeczek: Pushed the proposed small adjustments.

tpeczek:
Please let me know if any further updates to tests/samples/docs are needed.

Tratcher (approved):

Tratcher:
Thanks for this!

tpeczek:
> Thanks for this!

You're welcome, happy to repeat that in future if I happen to have an interesting idea 🙂

Tratcher:
There are a few here: https://github.com/microsoft/reverse-proxy/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22

PR #1984, opened 2023-01-04
Drop the Strict-Transport-Security response header 

### LINKED ISSUE #1862: Duplicated Strict-Transport-Security response header
### Describe the bug
We return a duplicated `Strict-Transport-Security` response header when using the HSTS middleware (`UseHsts`) **and** the destination also includes a `Strict-Transport-Security` response header.

Producing this duplicated response header can be problematic, though it is generally benign (see: [Is it fine to use duplicate response header with same value?](https://stackoverflow.com/a/47902496))

### To Reproduce
Proxy setup (pseudo-code):
```cs
app.UseHsts();
app.UseRouting();
app.UseEndpoints(endpoints => endpoints.MapReverseProxy());
```

Destination setup:
- Returns a `Strict-Transport-Security` response header.

### Suggested fix

Do not return a duplicated Strict-Transport-Security value. Perhaps YARP should IGNORE the destination's response header when HSTS is already applied to the current response -- since the YARP-based server already owns TLS termination, it seems reasonable that it should also own the HSTS response.

### Further technical details

Repro's with:
- .NET 6.0.8
- YARP 1.1.1

Tratcher:
Fixes #1862

Strict-Transport-Security is a per-hop response header. Since we terminate TLS, we should not proxy this header.

### DIFF
--- a/docs/docfx/articles/header-guidelines.md
+++ b/docs/docfx/articles/header-guidelines.md
@@ -9,7 +9,7 @@ Headers are a very important part of processing HTTP requests and each have thei
 
 ## YARP header filtering
 
-YARP automatically removes request and response headers that could impact its ability to forward a request correctly, or that may be used maliciously to bypass features of the proxy. A complete list can be found [here](https://github.com/microsoft/reverse-proxy/blob/b0a24521b269c030c50617f9fc56be9b8a3fe247/src/ReverseProxy/Forwarder/RequestUtilities.cs#L65-L81), with some highlights described below.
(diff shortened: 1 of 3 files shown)


Kahbazi:
Should the doc be updated as well to include this header? https://github.com/microsoft/reverse-proxy/blob/4f46326e6a73ecc1785bc07040491600ba2d312b/docs/docfx/articles/header-guidelines.md#yarp-header-filtering

MihaZupan (approved):
Can you update
https://github.com/microsoft/reverse-proxy/blob/07185a455a925f266add684b9bb8c1e4e7aad3a0/src/ReverseProxy/Forwarder/RequestUtilities.cs#L63
to `18` as well?

PR #1985, opened 2023-01-04
Enable direct cancellation for IHttpForwarder 

### LINKED ISSUE #1542: IHttpForwarder and cancellation
IHttpForwarder.SendAsync does not directly accept a CancellationToken to abort it. It supports canceling if the ForwarderRequestConfig.ActivityTimeout expires, or if HttpContext.RequestAborted triggers (client disconnect).
https://github.com/microsoft/reverse-proxy/blob/79c88ce07f9b88a283e10fc4cc45287b2efa6750/src/ReverseProxy/Forwarder/IHttpForwarder.cs#L25-L26

A customer that wanted to add their own cancellation condition did so by replacing HttpContext.RequestAborted. However, this caused an issue where they were leaking resources because they were no longer responding to the original RequestAborted token. The fix was to chain the two tokens.

Feedback: Having a CancellationToken parameter on IHttpForwarder.SendAsync would make this less error prone.

Tratcher:
Fixes #1542

Some customers using IHttpForwarder directly have asked for the ability to pass their own cancellation token. Today they have to replace HttpContext.RequestAborted which can be error prone.

This uses a Default Interface Method to avoid a breaking change.

### DIFF
--- a/src/ReverseProxy/Forwarder/ForwarderRequestConfig.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderRequestConfig.cs
@@ -3,11 +3,13 @@
 
 using System;
 using System.Net.Http;
+using System.Threading;
+using Microsoft.AspNetCore.Http;
 
 namespace Yarp.ReverseProxy.Forwarder;
 
 /// <summary>
-/// Config for <see cref="IHttpForwarder.SendAsync"/>
+/// Config for <see cref="IHttpForwarder.SendAsync(HttpContext, string, HttpMessageInvoker, ForwarderRequestConfig, HttpTransformer, CancellationToken)"/>
 /// </summary>
 public sealed record ForwarderRequestConfig
 {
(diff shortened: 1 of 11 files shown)


MihaZupan (approved):

MihaZupan:
While on the topic, should we augment `ApplyAsync` methods on transforms to accept `CancellationToken`s as well?

Tratcher:
> While on the topic, should we augment `ApplyAsync` methods on transforms to accept `CancellationToken`s as well?

Updated. How does that look @MihaZupan?

MihaZupan (approved):
Looks good.

~~We should also add CTs to the new extension methods we just added - https://github.com/microsoft/reverse-proxy/blob/main/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs. I can do that as a followup.~~

Nvm :)

Tratcher:
> We should also add CTs to the new extension methods we just added - https://github.com/microsoft/reverse-proxy/blob/main/src/ReverseProxy/Routing/DirectForwardingIEndpointRouteBuilderExtensions.cs. I can do that as a followup.

? Those are route table wire-up methods, there's nothing to cancel. The one place we do call the forwarder would only ever be cancelled by the RequestAborted token.

PR #1987, opened 2023-01-05
Fix ConnectionClosed_BlameAttributedCorrectly test

MihaZupan:
I can't see this failing locally, but what I imagine is happening is that:
- The client sends the close message and closes the connection in close succession
- The proxy is reading from the client and receives the close message and the connection close at about the same time
- Kestrel signals `RequestAborted` in response to the connection close
- The proxy happens to observe the cancellation before reading the close message and tears everything down (this is a race - that's why the test is flaky)
- Because the close message never went through the request pipeline, the telemetry parser never saw it, so we don't consider the close as graceful


MihaZupan:
I'll keep re-running the CI a few times here

Tratcher (approved):

PR #1989, opened 2023-01-05
Change default cookie session affinity format

### LINKED ISSUE #1980: Affinity is broken with multiple YARP instances
### Describe the bug
Affinity is not working in YARP. When you have more than one YARP instance each YARP produces different hash values for the same destination machines. When the affinity is being resolved on a different YARP instance it fails. Hence Affinity is broken in YARP.

### To Reproduce
- Use 2 or more YARP instances, with 2 or more destinations in a cluster. 
- Define cookie affinity. 
- try to get to a single destination machine using the YARP produced cookie
- the affinity resolution is failing when reaching the YARP instance who did not create the affinity cookie.

### Further technical details
- The issue root source is that YARP is using data protection. 
- When data protection not configured. The default behavior will be to create a different key per YARP instance.
- The workaround is to define data protection with either no hashing, or with a distributed key across YARP instances
- This adds unnecessary dependency on the key storage place to allow affinity to work.

### Suggested fix
- Provide a default SHA256 one-way hashing for YARP
- Affinity will work out of the box across YARP instances
- Destinations names are protected
- No dependency on distributed cache and no need to define data protection manually.
- Still allows YARP users to use DI to replace the data protection algorithm if needed.

Tratcher:
Fixes #1980

This also addresses the doc issues for #1981, but some of those should be backported to the release/latest branch.

This adds two new cookie based session affinity policies, 'HashCookie' and 'ArrCookie' that use stable hashes rather than Data Protection, making it easier to deploy across multiple instances without additional configuration or external storage. 
- 'HashCookie' uses XxHash64 which is fast and compact. This is the new ***_default_*** policy.
- 'ArrCookie' uses SHA-256 and matches the format IIS's ARR uses, allowing YARP and ARR to be used together in the same environment.

_Breaking Change_: This is considered a behavioral breaking change because existing session cookies will not be honored on upgrade unless the old 'Cookie' policy is specified in config. The request will be re load balanced and the old cookie will be replaced. This format also makes fewer security guarantees about the protection level of the key data, the destination id. Sensitive data should not be used for destination ids.


blowdart:
One thing that occurred to me just now is the actual algorithm. As you're not using SHA256 for anything cryptographically related, you may want to consider one the of the non-cryptographic hashes we added a couple of versions ago. This was done so it would be a clear indication the hash wasn't be used for security purposes, and to enable folks to stop getting warnings for old algorithm use like SHA1. While SHA256 is probably not going away anytime soon, using anything from [System.IO.Hashing.NonCryptographicHashAlgorithm](https://learn.microsoft.com/en-us/dotnet/api/system.io.hashing.noncryptographichashalgorithm) would head that sort of code review off at the pass, and they'd be faster to generate as well.

blowdart (approved):
One thing that occurred to me just now is the actual algorithm. As you're not using SHA256 for anything cryptographically related, you may want to consider one the of the non-cryptographic hashes we added a couple of versions ago. This was done so it would be a clear indication the hash wasn't be used for security purposes, and to enable folks to stop getting warnings for old algorithm use like SHA1. While SHA256 is probably not going away anytime soon, using anything from [System.IO.Hashing.NonCryptographicHashAlgorithm](https://learn.microsoft.com/en-us/dotnet/api/system.io.hashing.noncryptographichashalgorithm) would head that sort of code review off at the pass, and they'd be faster to generate as well.

src/ReverseProxy/SessionAffinity/HashCookieSessionAffinityPolicy.cs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private string GetDestinationHash(DestinationState d)
+    {
+        return _hashes.GetValue(d, static d =>
+        {
+            // Matches the format used by ARR
+            var destinationIdBytes = Encoding.Unicode.GetBytes(d.DestinationId.ToLowerInvariant());
+            var hashBytes = SHA256.HashData(destinationIdBytes);
+            return Convert.ToHexString(hashBytes);
```
davidfowl, line 106:
```suggestion
            return Convert.ToHexString(hashBytes).ToLowerInvariant();
```

davidfowl: To lowercase looks nicer

eyalg-msft:
> One thing that occurred to me just now is the actual algorithm. As you're not using SHA256 for anything cryptographically related, you may want to consider one the of the non-cryptographic hashes we added a couple of versions ago. This was done so it would be a clear indication the hash wasn't be used for security purposes, and to enable folks to stop getting warnings for old algorithm use like SHA1. While SHA256 is probably not going away anytime soon, using anything from [System.IO.Hashing.NonCryptographicHashAlgorithm](https://learn.microsoft.com/en-us/dotnet/api/system.io.hashing.noncryptographichashalgorithm) would head that sort of code review off at the pass, and they'd be faster to generate as well.

The SHA256 is required to be compatible with ARR. This is an important requirement for products migrating from ARR. If we really want, we can allow selecting different algorithm. However, I do suggest keeping it simple, and to keep the ARR backward compatible version the default behavior.

src/ReverseProxy/SessionAffinity/HashCookieSessionAffinityPolicy.cs
eyalg-msft: This is needed to be compatible with ARR, also with other SHA256 I checked. Is there a way to get it lowercase from the SHA256,HashData to begin with?

docs/docfx/articles/session-affinity.md
```diff
@@ -24,7 +26,7 @@ Session affinity is configured per cluster according to the following configurat
         "<cluster-name>": {
             "SessionAffinity": {
                 "Enabled": "(true|false)", // defaults to 'false'
-                "Policy": "(Cookie|CustomHeader)", // defaults to 'Cookie'
+                "Policy": "(HashCookie|Cookie|CustomHeader)", // defaults to 'HashCookie'
```
eyalg-msft, line 29: Do we want to enable hashed header affinity respectively? If a customer wants to update its affinity from a cookie to a header, does it make sense that it will no longer work for him without changing the system design to include a shared storage for keys?  IMO we should be consistent with the chosen approach for header affinity as well. The customer can always choose the data protection version, but we should allow keeping the same hash paradigm for headers as well. What do you think?

src/ReverseProxy/SessionAffinity/HashCookieSessionAffinityPolicy.cs
```diff
@@ -0,0 +1,132 @@
...
+                SameSite = config.Cookie?.SameSite ?? SameSiteMode.Unspecified,
+                HttpOnly = config.Cookie?.HttpOnly ?? true,
+                MaxAge = config.Cookie?.MaxAge,
+                Domain = config.Cookie?.Domain,
+                IsEssential = config.Cookie?.IsEssential ?? false,
+                Secure = config.Cookie?.SecurePolicy == CookieSecurePolicy.Always || (config.Cookie?.SecurePolicy == CookieSecurePolicy.SameAsRequest && context.Request.IsHttps),
+                Expires = config.Cookie?.Expiration is not null ? _clock.GetUtcNow().Add(config.Cookie.Expiration.Value) : default(DateTimeOffset?),
+            };
+
+            context.Response.Cookies.Append(config.AffinityKeyName, affinityKey, affinityCookieOptions);
```
eyalg-msft, line 59: seems like duplication of cookie affinity general code. can we extract it to a base class to share the code?

src/ReverseProxy/SessionAffinity/HashCookieSessionAffinityPolicy.cs
blowdart: Why does this need to be compatible with ARR? Is there a case where you're running both?

(But really isn't a tolower adding a small performance hit that isn't needed just for looks?)

blowdart:
> 
> The SHA256 is required to be compatible with ARR. This is an important requirement for products migrating from ARR. If we really want, we can allow selecting different algorithm. However, I do suggest keeping it simple, and to keep the ARR backward compatible version the default behavior.

Are there really mixed setups where something could hit ARR and Yarp at the same time and you want routing to work the same in both? Or is the concern that a yarp cookie would upset ARR, in which case changing the cookie name is better.

Avoiding unnecessary SDL compliance issues is a huge plus for not using SHA256.

eyalg-msft:
> > The SHA256 is required to be compatible with ARR. This is an important requirement for products migrating from ARR. If we really want, we can allow selecting different algorithm. However, I do suggest keeping it simple, and to keep the ARR backward compatible version the default behavior.
> 
> Are there really mixed setups where something could hit ARR and Yarp at the same time and you want routing to work the same in both? Or is the concern that a yarp cookie would upset ARR, in which case changing the cookie name is better.
> 
> Avoiding unnecessary SDL compliance issues is a huge plus for not using SHA256, and the current, non-SHA approach isn't compatible with ARR either, so using something else puts you in the same incompatible format. The original issue did not talk about wanting strict ARR compatibility. Also using SHA256 doesn't ensure ARR compatibility at all, unless ARR works in the exact same way, hashing the exact same resource identifier, and that seems a fragile thing to base a decision on.

We had to check ARR code to validate, and we did already validate its working well it in FastFood.

src/ReverseProxy/SessionAffinity/HashCookieSessionAffinityPolicy.cs
davidfowl: > really isn't a tolower adding a small performance hit that isn't needed just for looks?)

It's cached/

blowdart:
> We had to check ARR code to validate, and we did already validate its working well it in FastFood.

That doesn't answer my question about what you mean by ARR compatible, and why it's needed.

docs/docfx/articles/session-affinity.md
Tratcher: Honestly I haven't seen much use of this approach. I'd rather not expand on it without customer demand. It's easy enough for customers to replicate what we've done here in the short term.

src/ReverseProxy/SessionAffinity/HashCookieSessionAffinityPolicy.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        return new(null, AffinityStatus.DestinationNotFound);
+    }
+
+    private string GetDestinationHash(DestinationState d)
+    {
+        return _hashes.GetValue(d, static d =>
+        {
+            // Matches the format used by ARR
+            var destinationIdBytes = Encoding.Unicode.GetBytes(d.DestinationId.ToLowerInvariant());
```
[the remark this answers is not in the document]
Tratcher: This matches what ARR does. I'm not too concerned about variation, it's unlikely within deployments.

Tratcher:
Updated plan (see the edited description above): Ship both XxHash64 and SHA-256 implementations, with XxHash64 as the default.

src/ReverseProxy/SessionAffinity/BaseHashCookieSessionAffinityPolicy.cs
```diff
@@ -0,0 +1,81 @@
...
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    public abstract string Name { get; }
+
+    public void AffinitizeResponse(HttpContext context, ClusterState cluster, SessionAffinityConfig config, DestinationState destination)
+    {
+        if (!config.Enabled.GetValueOrDefault())
+        {
+            throw new InvalidOperationException($"Session affinity is disabled for cluster.");
```
davidfowl, line 32:
```suggestion
            throw new InvalidOperationException("Session affinity is disabled for cluster.");
```

src/ReverseProxy/SessionAffinity/BaseHashCookieSessionAffinityPolicy.cs
```diff
@@ -0,0 +1,81 @@
...
+            return new(null, AffinityStatus.AffinityKeyNotSet);
+        }
+
+        foreach (var d in destinations)
+        {
+            var hashValue = GetDestinationHash(d);
+
+            if (affinityHash == hashValue)
+            {
+                context.Items[AffinityKeyId] = affinityHash;
```
Kahbazi, line 63: If I'm correct ,we don't need the affinity hash value. This is just a flag to not add it twice. So if we just stored a static `new object()` and set that, would it prevent boxing when setting `affinityHash` to context.Items?

docs/docfx/articles/session-affinity.md
```diff
@@ -71,7 +72,7 @@ new ClusterConfig
 ```
 
 ## Affinity Key
-Request to destination affinity is established via the affinity key identifying the target destination. That key can be stored on different request parts depending on the given session affinity implementation, but each request cannot have more than one such key. The exact key semantics is implementation dependent, in example the both of built-in `CookieSessionAffinityPolicy` and `CustomHeaderAffinityPolicy` are currently use `DestinationId` as the affinity key.
+Request to destination affinity is established via the affinity key identifying the target destination. That key can be stored on different request parts depending on the given session affinity implementation, but each request cannot have more than one such key. The exact key semantics is implementation dependent, but the built-in policies currently use `DestinationId` as the affinity key.
 
 The current design doesn't require a key to uniquely identify the single affinitized destination. It's allowed to establish affinity to a destination group. In this case, the exact destination to handle the given request will be determined by the load balancer.
```
MihaZupan, line 77: Am I misunderstanding this or is this just plain wrong? If the key is shared by more than 1 destination, we'll pick whichever one we happen to find first.

MihaZupan (approved):

docs/docfx/articles/session-affinity.md
Tratcher: The current implementation would pick the first one since we do a linear search, but the abstraction allows us to create keys that are not specific to a single destination. E.g. if you had sets of destinations a-1, a-2, b-1, b-2... you could make a policy that used the key `a` and returned both a-1 and a-2, then load balancing would choose between them.

davidfowl:
This is an excellent change!

davidfowl:
@Tratcher do we need to hardcode the ARR affinity cookie name for it to be compatible?

Tratcher:
> @Tratcher Chris Ross FTE do we need to hardcode the ARR affinity cookie name for it to be compatible? Or should we document that if the name isn't set appropriately, it won't be compatible?

That's at least worth documenting, what was the name?

YARP doesn't have a default cookie name for session affinity, it must be specified in config.

PR #1990, opened 2023-01-06
Add async ISessionAffinityPolicy APIs

### LINKED ISSUE #1974: ISessionAffinityPolicy should be async

FindAffinitizedDestinations and AffinitizeResponse could use async overloads to allow for external data store affinity lookups, or other operations like checking the body. These could be added in a non-breaking way as Default Interface Methods that call the sync APIs.

Example: David had to read the response body to get an identifier. He was also unable to use IDistributedCache.
https://github.com/davidfowl/SmartLoadBalancer/blob/1d43229daf09f836aabf7a78ce31658a207cf53b/Yarp.Ingress/SignalRSessionAffinity.cs

MihaZupan:
Closes #1974

### DIFF
--- a/src/ReverseProxy/SessionAffinity/AffinitizeTransform.cs
+++ b/src/ReverseProxy/SessionAffinity/AffinitizeTransform.cs
@@ -41,6 +41,6 @@ internal sealed class AffinitizeTransform : ResponseTransform
             proxyFeature.Route.Cluster,
             options,
             proxyFeature.ProxiedDestination,
-            context.HttpContext.RequestAborted);
+            context.CancellationToken);
     }
 }
--- a/src/ReverseProxy/SessionAffinity/AffinitizeTransform.cs
+++ b/src/ReverseProxy/SessionAffinity/AffinitizeTransform.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Yarp.ReverseProxy.Model;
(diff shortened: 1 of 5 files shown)


src/ReverseProxy/SessionAffinity/AffinitizeTransform.cs
```diff
@@ -31,8 +32,15 @@ public override ValueTask ApplyAsync(ResponseTransformContext context)
...
+
+        Debug.Assert(proxyFeature.Route.Cluster is not null);
+        Debug.Assert(proxyFeature.ProxiedDestination is not null);
+
+        return _sessionAffinityPolicy.AffinitizeResponseAsync(
+            context.HttpContext,
+            proxyFeature.Route.Cluster,
+            options,
+            proxyFeature.ProxiedDestination,
+            context.HttpContext.RequestAborted);
```
MihaZupan, line 44: This is why I was asking about adding `CancellationToken`s to transforms.
After #1542, there's no way to observe the new token provided to the forwarder from these APIs.

Tratcher: You can rebase on my changes in https://github.com/microsoft/reverse-proxy/pull/1985 to get the CT.

src/ReverseProxy/SessionAffinity/ISessionAffinityPolicy.cs
```diff
@@ -28,6 +30,22 @@ public interface ISessionAffinityPolicy
...
+    /// <summary>
+    /// Finds <see cref="DestinationState"/> to which the current request is affinitized by the affinity key.
+    /// </summary>
+    /// <param name="context">Current request's context.</param>
+    /// <param name="cluster">Current request's cluster.</param>
+    /// <param name="config">Affinity config.</param>
+    /// <param name="destinations"><see cref="DestinationState"/>s available for the request.</param>
+    /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+    /// <returns><see cref="AffinityResult"/> carrying the found affinitized destinations if any and the <see cref="AffinityStatus"/>.</returns>
+    ValueTask<AffinityResult> FindAffinitizedDestinationsAsync(HttpContext context, ClusterState cluster, SessionAffinityConfig config, IReadOnlyList<DestinationState> destinations, CancellationToken cancellationToken = default)
```
Tratcher, line 42: No need for defaults, we'll be the main caller and we want to make sure we remember the CT.
```suggestion
    ValueTask<AffinityResult> FindAffinitizedDestinationsAsync(HttpContext context, ClusterState cluster, SessionAffinityConfig config, IReadOnlyList<DestinationState> destinations, CancellationToken cancellationToken)
```

Tratcher (approved):
After updating the CT

davidfowl:
Very nice!

PR #1991, opened 2023-01-06
Set the default ConnectTimeout to a finite value

### LINKED ISSUE #1678: Consider setting a finite ConnectTimeout by default

MihaZupan:
Closes #1678

### DIFF
--- a/docs/docfx/articles/direct-forwarding.md
+++ b/docs/docfx/articles/direct-forwarding.md
@@ -52,7 +52,8 @@ public void Configure(IApplicationBuilder app, IHttpForwarder forwarder)
         AllowAutoRedirect = false,
         AutomaticDecompression = DecompressionMethods.None,
         UseCookies = false,
-        ActivityHeadersPropagator = new ReverseProxyPropagator(DistributedContextPropagator.Current)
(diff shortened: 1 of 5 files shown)


Tratcher (approved):
Should we add this to https://github.com/microsoft/reverse-proxy/blob/main/src/ReverseProxy/Configuration/HttpClientConfig.cs too?

MihaZupan:
I'm not sure it'd get much use. Do you think people would want to change it from the config to potentially lower it even further?

Tratcher:
> I'm not sure it'd get much use. Do you think people would want to change it from the config to potentially lower it even further?

Setting it by default might be enough. We can revisit later if there's demand.

Issue #1995, opened 2023-01-12
2.0-rc.1 release checklist


Tratcher:
See [Release.md](https://github.com/microsoft/reverse-proxy/blob/main/docs/operations/Release.md) for detailed instructions.

- [x] Ensure that `Version.props` has the version and `PreReleaseVersionLabel` updated to the expected value
    - #2001
- [x] Identify and validate the build on the `microsoft-reverse-proxy-official` pipeline
    - 50919f1dd3dfbacad3d62ce864ab6b572689f9c9
    - https://dev.azure.com/dnceng/internal/_build/results?buildId=2091559&view=results
    - `2.0.0-rc.1.23068.3`
- [x] Ensure there's a release branch created (see [Branching](https://github.com/microsoft/reverse-proxy/blob/main/docs/operations/Branching.md))
    - [`release/2.0.0-rc1`](https://github.com/microsoft/reverse-proxy/tree/release/2.0.0-rc1)
- [x] Ensure the `Version.props` has the version and `PreReleaseVersionLabel` updated to the next preview
    - #2003 
- [x] Release the build
    - https://www.nuget.org/packages/Yarp.ReverseProxy/2.0.0-rc.1.23068.3
    - https://www.nuget.org/packages/Yarp.Telemetry.Consumption/2.0.0-rc.1.23068.3
- [x] Tag the commit
    - https://github.com/microsoft/reverse-proxy/releases/tag/v2.0.0-rc.1
- [x] Draft release notes
- [x] Update samples to use updated package versions
    - #2009
- [x] Publish the docs
- [x] Publish release notes
- [x] Close the [old milestone](https://github.com/microsoft/reverse-proxy/milestones)
- [x] Announce on social media
  - https://twitter.com/_MihaZupan/status/1616230875468808197
- [x] Set the preview branch to protected
- [x] ~~Delete the [previous preview branch](https://github.com/microsoft/reverse-proxy/branches)~~
- [x] Request source code archival

MihaZupan:
I'll move it

MihaZupan:
`release/latest` is now `release/2.0.0-rc1`.
In case I messed something up, here's a backup of the old release/latest:
https://github.com/MihaZupan/reverse-proxy/tree/release-latest-copy-2022-01 😄

PR #2005, opened 2023-01-18
Set IsPackable to false for the k8s project

MihaZupan:
Brings back #941
This slipped through in #1642 when we rearranged all the k8s project files.

Contributes to #1995

I'll backport this to main as well.


Tratcher:
Rather than backport, merge this branch into main.

Do you need this to avoid shipping the packages? Can the release step just exclude them instead?

MihaZupan:
> Do you need this to avoid shipping the packages?

Yes

> Can the release step just exclude them instead?

I don't see anything in Azdo for it

Tratcher (approved):

PR #2006, opened 2023-01-18
Set IsPackable to false for the k8s project (#2005)

MihaZupan:
Port of #2005 to main


Tratcher (approved):
Do not squash or rebase, make an actual merge commit.

PR #2007, opened 2023-01-19
Merge main into release/latest

MihaZupan:
Also includes updating the package versions for samples - 1f5b456815a982332deff9098db8881edbb16909


MihaZupan:
I'm not sure what that gets us? Do you want to bring the Arcade changes from release/latest to the RC1 branch?

Tratcher:
release/2.0.0-rc.1 is the authoritative branch, release/latest is just a reference to latest + a few doc fixes. Alternatively we could force push release/2.0.0-rc.1 to release/latest if we don't think we'd loose anything.

MihaZupan:
> release/latest is just a reference to latest + a few doc fixes

That is what it is/would be even after this change, no? main is still effectively the same as rc1.
So you'd end up with
- main and rc1 being about the same (excluding https://github.com/microsoft/reverse-proxy/pull/2007/commits/196756ee493be9cffa4e14e476bbb883bd2499aa and 4ebab4284c81c2d4d3eff4ff854558de74c72b82)
- release/latest being the same as those + the existing docs fixes

And then we can backport any docs fixes back into main to avoid future merge conflicts.

Tratcher:
Ideally changes would consistently flow/merge forward into authoritative branches. E.g. latest -> rc.1 -> main. I'd also like rc.1 and latest to reference the same commit while we're maintaining rc.1.

MihaZupan:
> latest -> rc.1

This is the part we haven't been doing - we'd only port PRs from latest => main.
I'm okay with keeping the rc more up-to-date though - in that case I'll put up PRs to
1. Port fixes from release/latest to main
1. Port fixes from release/latest to rc1
1. Merge rc1 into release/latest

PR #2008, opened 2023-01-19
Update release/2.0.0-rc1 with Arcade changes from release/latest

MihaZupan:
#1995

These two changes are the only things where release/latest is ahead of main or rc1.

After this we should be able to force push the rc1 branch into release/latest.


Tratcher (approved):

MihaZupan:
For reference, these two commits are parts of #1922 and #1916

PR #2029, opened 2023-02-09
Fix edge-case error attribution bug on WS downgrade

MihaZupan:
We determine whether to report `Request` vs `RequestCreation` based on whether `destinationRequest is null` or not.
If an exception was thrown by the request transform for the downgrade request, it would be reported as `Request` instead of `RequestCreation`.


Tratcher (approved):
How did you notice this?

MihaZupan:
When I was experimenting with request pooling I ended up with the request never being null, and the existing RequestCreation test failed.

PR #2030, opened 2023-02-09
Refresh NuGet feeds

sebastienros:
net8.0 required for benchmarks
net5.0 removes as obsolete for this repos


MihaZupan (approved):
Thanks!

PR #2039, opened 2023-02-16
Use Arcade's pool-providers.yml

dougbu:
- import the variable template
- use the two variables it provides wherever we name the pool
  - future release branches will now automatically switch to the -Svc pools


Tratcher (approved):

Tratcher:
We froze the subscriptions because we were still building against .NET 7 and Arcade kept trying to push us to 8. We've released now, so it should be ok to re-enable the Arcade and runtime subscriptions.

Tratcher (approved):

dougbu:
> We froze the subscriptions because we were still building against .NET 7 and Arcade kept trying to push us to 8. We've released now, so it should be ok to re-enable the Arcade and runtime subscriptions.

And the aspnetcore sub too❔

Tratcher:
Sure, once the Arcade one goes through.

dougbu:
Arcade update complete and all subscriptions are now reenabled. This should work now&hellip;

Issue #1926, opened 2022-11-15
The Receiver's HttpClient should be extensible


Tratcher:
https://github.com/microsoft/reverse-proxy/blob/c39fcf8c1fc574dc94c61d4bbb4222f1cd35c656/src/Kubernetes.Controller/Protocol/Receiver.cs#L45

This HttpClient instance should be replaceable from the options to enable extensibility and testing.

Compare to:
https://github.com/dotnet/aspnetcore/blob/67384a0b6f302e222d62162c9073708024acebe9/src/Security/Authentication/Core/src/RemoteAuthenticationOptions.cs#L77

PR #2043, opened 2023-02-21
Make Receiver's HttpClient extensible

catcherwong:
For #1926 , make Receiver's HttpClient instance from ReceiverOptions.

### DIFF
--- a/src/Kubernetes.Controller/Protocol/Receiver.cs
+++ b/src/Kubernetes.Controller/Protocol/Receiver.cs
@@ -42,8 +42,6 @@ public class Receiver : BackgroundHostedService
 
     public override async Task RunAsync(CancellationToken cancellationToken)
     {
-        using var client = new HttpClient();
-
         while (!cancellationToken.IsCancellationRequested)
         {
             await _limiter.WaitAsync(cancellationToken).ConfigureAwait(false);
@@ -53,7 +51,7 @@ public class Receiver : BackgroundHostedService
 
             try
             {
-                using var stream = await client.GetStreamAsync(_options.ControllerUrl, cancellationToken).ConfigureAwait(false);
+                using var stream = await _options.Client.GetStreamAsync(_options.ControllerUrl, cancellationToken).ConfigureAwait(false);
                 using var reader = new StreamReader(stream, Encoding.UTF8, leaveOpen: true);
                 using var cancellation = cancellationToken.Register(stream.Close);
                 while (true)
--- a/src/Kubernetes.Controller/Protocol/Receiver.cs
+++ b/src/Kubernetes.Controller/Protocol/Receiver.cs
@@ -68,7 +68,7 @@ public class Receiver : BackgroundHostedService
             try
             {
                 var requestMessage = new HttpRequestMessage(HttpMethod.Get, _options.ControllerUrl);
-                var responseMessage = await _options.Client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
+                var responseMessage = await _options.Client.SendAsync(requestMessage, cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


catcherwong:

src/Kubernetes.Controller/Protocol/ReceiverOptions.cs
```diff
@@ -2,10 +2,17 @@
...
 
 public class ReceiverOptions
 {
     public Uri ControllerUrl { get; set; }
+
+    public TimeSpan ClientTimeout { get; set; } = TimeSpan.FromSeconds(60);
+
+    public HttpMessageHandler ClientHttpHandler { get; set; }
+
+    public HttpClient Client { get; set; } = default!;
```
MihaZupan, line 17:
```suggestion
    public HttpMessageInvoker? Client { get; set; }
```

It feels odd to have multiple settings for the same thing. It's also confusing that the timeout would apply if you set the client, but be ignored otherwise (I don't know why `RemoteAuthenticationOptions` was designed this way).

Can we simplify this to accepting the optional client, and change the type to `HttpMessageInvoker` instead of `HttpClient`, as that is what YARP users who do direct forwarding already use today?

MihaZupan (commented):
Thanks for looking to contribute. I would simplify the number of properties to configure, but otherwise LGTM.

src/Kubernetes.Controller/Protocol/Receiver.cs
```diff
@@ -53,7 +67,9 @@ public override async Task RunAsync(CancellationToken cancellationToken)
 
             try
             {
-                using var stream = await client.GetStreamAsync(_options.ControllerUrl, cancellationToken).ConfigureAwait(false);
+                var requestMessage = new HttpRequestMessage(HttpMethod.Get, _options.ControllerUrl);
+                var responseMessage = await _options.Client.SendAsync(requestMessage, cancellationToken).ConfigureAwait(false);
```
MihaZupan, line 71:
```suggestion
                var responseMessage = await _options.Client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
```

MihaZupan (approved):

src/Kubernetes.Controller/Protocol/Receiver.cs
```diff
@@ -53,7 +67,9 @@ public override async Task RunAsync(CancellationToken cancellationToken)
 
             try
             {
-                using var stream = await client.GetStreamAsync(_options.ControllerUrl, cancellationToken).ConfigureAwait(false);
+                var requestMessage = new HttpRequestMessage(HttpMethod.Get, _options.ControllerUrl);
+                var responseMessage = await _options.Client.SendAsync(requestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
```
MihaZupan, line 71:
```suggestion
                var responseMessage = await _options.Client.SendAsync(requestMessage, cancellationToken).ConfigureAwait(false);
```

Mb, missed that this wasn't HttpClient anymore

src/Kubernetes.Controller/Protocol/Receiver.cs
```diff
@@ -35,15 +38,26 @@ public class Receiver : BackgroundHostedService
 
         _options = options.Value;
 
+        if (_options.Client == null)
+        {
+            _options.Client = new(new SocketsHttpHandler
+            {
+                UseProxy = false,
+                AllowAutoRedirect = false,
+                AutomaticDecompression = DecompressionMethods.None,
```
Tratcher, line 47: @MihaZupan These are the settings we use elsewhere for proxying requests, but that's not applicable here is it, this is just used to retrieve config data? The defaults seem more applicable, especially for decompression.

MihaZupan: Most of these shouldn't matter, but I agree decompression may be useful to keep on by default.
#2054

PR #2054, opened 2023-02-28
Use default HttpClient settings in k8s Receiver

MihaZupan:


src/Kubernetes.Controller/Protocol/Receiver.cs
```diff
@@ -38,18 +38,11 @@ public class Receiver : BackgroundHostedService
...
-            {
-                UseProxy = false,
-                AllowAutoRedirect = false,
-                AutomaticDecompression = DecompressionMethods.None,
-                UseCookies = false,
-                ActivityHeadersPropagator = new ReverseProxyPropagator(DistributedContextPropagator.Current),
-                ConnectTimeout = TimeSpan.FromSeconds(15),
-            });
-        }
+            UseCookies = false,
```
Tratcher, line 43: Why not use cookies?

src/Kubernetes.Controller/Protocol/Receiver.cs
```diff
@@ -38,18 +38,11 @@ public class Receiver : BackgroundHostedService
 
         _options = options.Value;
 
-        if (_options.Client == null)
+        _options.Client ??= new HttpClient(new SocketsHttpHandler
```
Tratcher, line 41: Why an HttpClient vs an HttpMessageInvoker? You'll get different default behavior around buffering, timeouts, etc.

Tratcher (approved):

src/Kubernetes.Controller/Protocol/Receiver.cs
MihaZupan: Should be invoker

PR #2065, opened 2023-03-09
Fix an issue where destinations are compared against themselves

felipeogarcia:
The two calls to `Random.Next` may return the same index, so the comparison happens against the same destination. This behavior compromises the worst case protection this algorithm has because it might only use the worst destination, not offering opportunity of a better one. The fewer destinations we have, the more prominent this issue becomes.

### DIFF
--- a/src/ReverseProxy/LoadBalancing/PowerOfTwoChoicesLoadBalancingPolicy.cs
+++ b/src/ReverseProxy/LoadBalancing/PowerOfTwoChoicesLoadBalancingPolicy.cs
@@ -21,17 +21,28 @@ internal sealed class PowerOfTwoChoicesLoadBalancingPolicy : ILoadBalancingPolic
 
     public DestinationState? PickDestination(HttpContext context, ClusterState cluster, IReadOnlyList<DestinationState> availableDestinations)
     {
-        if (availableDestinations.Count == 0)
+        var destinationCount = availableDestinations.Count;
+        if (destinationCount == 0)
         {
             return null;
+        }
+        
+        if (destinationCount == 1)
+        {
+            return availableDestinations[0];
         }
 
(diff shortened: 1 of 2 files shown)


felipeogarcia:

MihaZupan (approved):
Thanks

src/ReverseProxy/LoadBalancing/PowerOfTwoChoicesLoadBalancingPolicy.cs
```diff
@@ -21,17 +21,28 @@ public PowerOfTwoChoicesLoadBalancingPolicy(IRandomFactory randomFactory)
...
-        var destinationCount = availableDestinations.Count;
         var random = _randomFactory.CreateRandomInstance();
-        var first = availableDestinations[random.Next(destinationCount)];
-        var second = availableDestinations[random.Next(destinationCount)];
+        var firstIndex = random.Next(destinationCount);
+        int secondIndex;
+        do
+        {
+            secondIndex = random.Next(destinationCount);
+        } while (firstIndex == secondIndex);
```
Tratcher, line 43: This fix suffers from the same issue as the original, a high collision rate in small sets. I think it'd be better to do a deterministic shift in those cases.
```
int secondIndex = random.Next(destinationCount);
if (secondIndex == firstIndex)
{
    secondIndex = (irstIndex + 1) % destinationCount;
}
```

I'll send a PR.

Tratcher: @MihaZupan convinced me that the +1 approach leads to non-trivial imbalance.

Even in the smallest case of two destinations where collisions will be common, the loop will only need to run a few times on average to break the tie. That's probably adequate.

PR #2098, opened 2023-04-12
Adding distinct activity for distributed tracing to YARP

samsp-msft:
Adding distinct tracing objects to YARP.

These can then be picked up via monitoring middleware such as Open Telemetry.

```c#
using Azure.Monitor.OpenTelemetry.AspNetCore;

var builder = WebApplication.CreateBuilder(args);
builder.Services.AddReverseProxy().LoadFromConfig(builder.Configuration.GetSection("ReverseProxy"));

var otel = builder.Services.AddOpenTelemetry();
otel.UseAzureMonitor(o =>
{
    o.ConnectionString = builder.Configuration["APPLICATIONINSIGHTS_CONNECTION_STRING"];
});
otel.WithTracing(t =>
{
    t.AddSource("Yarp.*");
});

var app = builder.Build();

app.MapReverseProxy();

app.Run();
```

The activity source is only active if something is listening to it, such as `t.AddSource("Yarp.*");`.

### DIFF
--- a/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
@@ -2,10 +2,9 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Threading.Tasks;
 using System.Diagnostics;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Connections;
 using Microsoft.Extensions.Logging;
 using Yarp.ReverseProxy.Model;
 using Yarp.ReverseProxy.Utilities;
--- a/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Connections;
 using Microsoft.Extensions.Logging;
 using Yarp.ReverseProxy.Model;
 using Yarp.ReverseProxy.Utilities;
@@ -64,11 +65,19 @@ internal sealed class ForwarderMiddleware
             throw new InvalidOperationException($"Chosen destination has no model set: '{destination.DestinationId}'");
         }
 
+        var activityForTracing = reverseProxyFeature.ActivityForTracing;
+        if (activityForTracing != null)
+        {
+            activityForTracing.AddTag("Route", route.Config.RouteId);
+            activityForTracing.AddTag("Cluster", cluster.ClusterId);
+            activityForTracing.AddTag("Destination", destination.DestinationId);
+        }
+
         try
         {
             cluster.ConcurrencyCounter.Increment();
             destination.ConcurrencyCounter.Increment();
-
+            activityForTracing?.Start();
             ForwarderTelemetry.Log.ForwarderInvoke(cluster.ClusterId, route.Config.RouteId, destination.DestinationId);
 
             var clusterConfig = reverseProxyFeature.Cluster;
@@ -79,6 +88,7 @@ internal sealed class ForwarderMiddleware
         {
             destination.ConcurrencyCounter.Decrement();
             cluster.ConcurrencyCounter.Decrement();
+            activityForTracing?.Stop();
         }
     }
 
--- a/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
@@ -43,7 +43,7 @@ internal sealed class ForwarderMiddleware
         var cluster = route.Cluster!;
 
 
-        var activityForTracing = reverseProxyFeature.ActivityForTracing;
+        var activityForTracing = ((ReverseProxyFeature)reverseProxyFeature).ActivityForTracing;
         activityForTracing?.AddTag("RouteId", route.Config.RouteId);
         activityForTracing?.AddTag("ClusterId", cluster.ClusterId);
 
@@ -55,7 +55,6 @@ internal sealed class ForwarderMiddleware
             context.Features.Set<IForwarderErrorFeature>(new ForwarderErrorFeature(ForwarderError.NoAvailableDestinations, ex: null));
             activityForTracing?.SetStatus(ActivityStatusCode.Error);
             activityForTracing?.AddTag("DestinationId", "No destinations available");
-            activityForTracing?.Stop();
             return;
         }
 
@@ -87,13 +86,11 @@ internal sealed class ForwarderMiddleware
                 clusterConfig.Config.HttpRequest ?? ForwarderRequestConfig.Empty, route.Transformer);
 
             activityForTracing?.SetStatus( (result == ForwarderError.None) ? ActivityStatusCode.Ok : ActivityStatusCode.Error);
-
         }
         finally
         {
             destination.ConcurrencyCounter.Decrement();
             cluster.ConcurrencyCounter.Decrement();
-            activityForTracing?.Stop();
         }
     }
 
--- a/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading.Tasks;
+using System.Diagnostics;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Connections;
 using Microsoft.Extensions.Logging;
@@ -77,12 +78,14 @@ internal sealed class ForwarderMiddleware
         {
             cluster.ConcurrencyCounter.Increment();
             destination.ConcurrencyCounter.Increment();
-            activityForTracing?.Start();
             ForwarderTelemetry.Log.ForwarderInvoke(cluster.ClusterId, route.Config.RouteId, destination.DestinationId);
 
             var clusterConfig = reverseProxyFeature.Cluster;
-            await _forwarder.SendAsync(context, destinationModel.Config.Address, clusterConfig.HttpClient,
+            var result = await _forwarder.SendAsync(context, destinationModel.Config.Address, clusterConfig.HttpClient,
                 clusterConfig.Config.HttpRequest ?? ForwarderRequestConfig.Empty, route.Transformer);
+
+            activityForTracing?.SetStatus( (result == ForwarderError.None) ? ActivityStatusCode.Ok : ActivityStatusCode.Error);
+
         }
         finally
         {
--- a/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
@@ -42,10 +42,9 @@ internal sealed class ForwarderMiddleware
         var route = context.GetRouteModel();
         var cluster = route.Cluster!;
 
-
-        var activityForTracing = ((ReverseProxyFeature)reverseProxyFeature).ActivityForTracing;
-        activityForTracing?.AddTag("RouteId", route.Config.RouteId);
-        activityForTracing?.AddTag("ClusterId", cluster.ClusterId);
+        var activity = context.GetYarpActivity();
+        activity?.AddTag("RouteId", route.Config.RouteId);
+        activity?.AddTag("ClusterId", cluster.ClusterId);
 
 
         if (destinations.Count == 0)
@@ -53,8 +52,8 @@ internal sealed class ForwarderMiddleware
             Log.NoAvailableDestinations(_logger, cluster.ClusterId);
             context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
             context.Features.Set<IForwarderErrorFeature>(new ForwarderErrorFeature(ForwarderError.NoAvailableDestinations, ex: null));
-            activityForTracing?.SetStatus(ActivityStatusCode.Error);
-            activityForTracing?.AddTag("DestinationId", "No destinations available");
+            activity?.SetStatus(ActivityStatusCode.Error);
+            activity?.AddTag("DestinationId", "No destinations available");
             return;
         }
 
@@ -67,7 +66,7 @@ internal sealed class ForwarderMiddleware
         }
 
         reverseProxyFeature.ProxiedDestination = destination;
-        activityForTracing?.AddTag("DestinationId", destination.DestinationId);
+        activity?.AddTag("DestinationId", destination.DestinationId);
 
         var destinationModel = destination.Model;
         if (destinationModel is null)
@@ -85,7 +84,7 @@ internal sealed class ForwarderMiddleware
             var result = await _forwarder.SendAsync(context, destinationModel.Config.Address, clusterConfig.HttpClient,
                 clusterConfig.Config.HttpRequest ?? ForwarderRequestConfig.Empty, route.Transformer);
 
-            activityForTracing?.SetStatus( (result == ForwarderError.None) ? ActivityStatusCode.Ok : ActivityStatusCode.Error);
+            activity?.SetStatus((result == ForwarderError.None) ? ActivityStatusCode.Ok : ActivityStatusCode.Error);
         }
         finally
         {
--- a/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
@@ -43,8 +43,8 @@ internal sealed class ForwarderMiddleware
         var cluster = route.Cluster!;
 
         var activity = context.GetYarpActivity();
-        activity?.AddTag("RouteId", route.Config.RouteId);
-        activity?.AddTag("ClusterId", cluster.ClusterId);
+        activity?.AddTag("proxy.route_id", route.Config.RouteId);
+        activity?.AddTag("proxy.cluster_id", cluster.ClusterId);
 
 
         if (destinations.Count == 0)
@@ -53,7 +53,7 @@ internal sealed class ForwarderMiddleware
             context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
             context.Features.Set<IForwarderErrorFeature>(new ForwarderErrorFeature(ForwarderError.NoAvailableDestinations, ex: null));
             activity?.SetStatus(ActivityStatusCode.Error);
-            activity?.AddTag("DestinationId", "No destinations available");
+            activity?.AddError("Proxy forwarding failed","No available destinations to route to");
             return;
         }
 
@@ -66,7 +66,7 @@ internal sealed class ForwarderMiddleware
         }
 
         reverseProxyFeature.ProxiedDestination = destination;
-        activity?.AddTag("DestinationId", destination.DestinationId);
+        activity?.AddTag("proxy.destination_id", destination.DestinationId);
 
         var destinationModel = destination.Model;
         if (destinationModel is null)
--- a/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
+++ b/src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
@@ -42,11 +42,20 @@ internal sealed class ForwarderMiddleware
         var route = context.GetRouteModel();
         var cluster = route.Cluster!;
 
+
+        var activityForTracing = reverseProxyFeature.ActivityForTracing;
+        activityForTracing?.AddTag("RouteId", route.Config.RouteId);
+        activityForTracing?.AddTag("ClusterId", cluster.ClusterId);
+
+
         if (destinations.Count == 0)
         {
             Log.NoAvailableDestinations(_logger, cluster.ClusterId);
             context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
             context.Features.Set<IForwarderErrorFeature>(new ForwarderErrorFeature(ForwarderError.NoAvailableDestinations, ex: null));
+            activityForTracing?.SetStatus(ActivityStatusCode.Error);
+            activityForTracing?.AddTag("DestinationId", "No destinations available");
+            activityForTracing?.Stop();
             return;
         }
 
@@ -59,19 +68,12 @@ internal sealed class ForwarderMiddleware
         }
 
         reverseProxyFeature.ProxiedDestination = destination;
+        activityForTracing?.AddTag("DestinationId", destination.DestinationId);
 
         var destinationModel = destination.Model;
         if (destinationModel is null)
         {
             throw new InvalidOperationException($"Chosen destination has no model set: '{destination.DestinationId}'");
-        }
-
-        var activityForTracing = reverseProxyFeature.ActivityForTracing;
-        if (activityForTracing != null)
-        {
-            activityForTracing.AddTag("Route", route.Config.RouteId);
-            activityForTracing.AddTag("Cluster", cluster.ClusterId);
-            activityForTracing.AddTag("Destination", destination.DestinationId);
         }
 
         try
--- a/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
+++ b/src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
@@ -10,6 +10,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Yarp.ReverseProxy.Model;
 using Yarp.ReverseProxy.Utilities;
+using System.Diagnostics;
 
 namespace Yarp.ReverseProxy.Health;
 
@@ -108,6 +109,8 @@ internal partial class ActiveHealthCheckMonitor : IActiveHealthCheckMonitor, ICl
             return;
         }
 
+        var activity = Observability.YarpActivitySource.CreateActivity("Yarp.ActiveHealthCheck", ActivityKind.Internal);
+
         Log.StartingActiveHealthProbingOnCluster(_logger, cluster.ClusterId);
 
         var allDestinations = cluster.DestinationsState.AllDestinations;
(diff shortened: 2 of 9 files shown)


src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
```diff
@@ -150,10 +153,11 @@ private async Task ProbeCluster(ClusterState cluster)
             }
 
             Log.StoppedActiveHealthProbingOnCluster(_logger, cluster.ClusterId);
+            activity?.Stop();
         }
     }
 
-    private async Task<DestinationProbingResult> ProbeDestinationAsync(ClusterState cluster, DestinationState destination, TimeSpan timeout)
+    private async Task<DestinationProbingResult> ProbeDestinationAsync(ClusterState cluster, DestinationState destination, TimeSpan timeout, Activity? healthCheckActivity)
```
samsp-msft, line 160: I think creating an activity will always parent it, so it may not be necessary to pass one in

src/ReverseProxy/Model/IReverseProxyFeature.cs
```diff
@@ -35,4 +36,6 @@ public interface IReverseProxyFeature
     /// The actual destination that the request was proxied to.
     /// </summary>
     DestinationState? ProxiedDestination { get; set; }
+
+    Activity? ActivityForTracing { get; }
```
samsp-msft, line 40: Adding the Activity to ProxyFeature seemed to be the best place to be able to pass it on

src/ReverseProxy/Utilities/Observability.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Diagnostics;
+
+namespace Yarp.ReverseProxy.Utilities
```
samsp-msft, line 8: This is all static. Should it be registered with services, or does that just get more complicated for the sake of it?

src/ReverseProxy/Model/IReverseProxyFeature.cs
MihaZupan: Why publicly expose it at all?

The lifetime/scope is odd - starts inside `ProxyPipelineInitializerMiddleware` and stops in `ForwarderMiddleware`?

Can we do something like this in the forwarder middleware instead?
```c#
using var activity = ActivitySource.CreateActivity("Forward", ActivityKind.Server);
activity?.AddTag(...);

await _forwarder.SendAsync(...);

activity?.SetAnyOtherInfo();
```
If someone wants to interact with the exact activity, they can still do so through `Activity.Current` inside transforms / via ActivityListener.

Nit: Should the activity kind be Client instead?

samsp-msft: I wanted to start it early so that it can have properties set on it throughout the pipeline - eg NYI is including if the destination was selected because of session affinity, or if session affinity is broken for that request.

You can't guarantee that somebody hasn't started another activity, which is why you can't just write to the current one,

samsp-msft: The activity kind is weird as its both a server and a client.

samsp-msft:
Result when seen in App Insights
image

src/ReverseProxy/Model/IReverseProxyFeature.cs
MihaZupan: If you want it to start early, I think you'd want it to end in the same scope as well.
Right now, you could modify the activity in middleware until you called next, but it'd be stopped by the time the execution came back to your middleware.
Also if we never got to the forwarder middleware, the activity would be started, but never stopped.

> You can't guarantee that somebody hasn't started another activity, which is why you can't just write to the current one

Sure, but does it need to be public?

MihaZupan: > eg NYI is including if the destination was selected because of session affinity, or if session affinity is broken for that request.

Is that something that we'd want to put onto the trace though?
It feels like something we should ensure is properly logged, and that those logs can be correlated back via the trace id.

samsp-msft: @noahfalk - this is something I am unclear on - what's the guidance for what should be set as properties of the activity, and what should we assume people will pickup from logging?

samsp-msft: > Sure, but does it need to be public?
Probably not

MihaZupan:
If you add `.AddHttpClientInstrumentation()` to the builder, you should see another dependency node at about the same time you see `Forward` now (assuming things aren't broken). That means you don't gain new timing information from the new activity.

Re: Tags like routeId, is there prior art in adding such information to existing Activities?
E.g. to the root one ASP.NET will create, or to the one `HttpClient` creates internally?

Things like that should be trivial to do now that `ActivityListener` is a thing
```c#
ActivitySource.AddActivityListener(new ActivityListener()
{
    ShouldListenTo = source => source.Name == "System.Net.Http",
    ActivityStarted = activity =>
    {
        string routeId = "..."; // Get that somehow (IHttpContextAccessor?)
        activity.AddTag("YarpRoute", routeId);
    }
});
```

src/ReverseProxy/Utilities/Observability.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    internal class Observability
+    {
+        private static ActivitySource _activitySource = new ActivitySource("Yarp.ReverseProxy");
```
WeihanLi, line 12:
```suggestion
        private static readonly ActivitySource _activitySource = new ActivitySource("Yarp.ReverseProxy");
```

src/ReverseProxy/Model/ReverseProxyFeature.cs
```diff
@@ -31,4 +33,12 @@ public IReadOnlyList<DestinationState> AvailableDestinations
 
     /// <inheritdoc/>
     public DestinationState? ProxiedDestination { get; set; }
+
+    public Activity? ActivityForTracing { get; init; }
+
+    public ReverseProxyFeature()
+    {
+        if (Observability.IsListening)
+        ActivityForTracing = (Observability.IsListening) ? Observability.YarpActivitySource.StartActivity("Proxy Forwarder", ActivityKind.Server) : null;
```
WeihanLi, line 42: Seems to be a duplicate condition?

src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
```diff
@@ -167,19 +174,28 @@ private async Task<DestinationProbingResult> ProbeDestinationAsync(ClusterState
             return new DestinationProbingResult(destination, null, ex);
         }
 
+        var probeActivity = Observability.YarpActivitySource.StartActivity("Proxy destination health check", ActivityKind.Client);
```
MihaZupan, line 177:
```suggestion
        using var probeActivity = Observability.YarpActivitySource.StartActivity("Proxy destination health check", ActivityKind.Client);
```
And then you can remove the calls to `Stop`

src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
```diff
@@ -108,6 +109,9 @@ private async Task ProbeCluster(ClusterState cluster)
             return;
         }
 
+        var activity = Observability.YarpActivitySource.StartActivity("Proxy cluster active health checks", ActivityKind.Consumer);
```
MihaZupan, line 112:
```suggestion
        using var activity = Observability.YarpActivitySource.StartActivity("Proxy cluster active health checks", ActivityKind.Consumer);
```

src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
```diff
@@ -68,17 +80,20 @@ public async Task Invoke(HttpContext context)
...
                 clusterConfig.Config.HttpRequest ?? ForwarderRequestConfig.Empty, route.Transformer);
+
+            activityForTracing?.SetStatus( (result == ForwarderError.None) ? ActivityStatusCode.Ok : ActivityStatusCode.Error);
+
         }
         finally
         {
             destination.ConcurrencyCounter.Decrement();
             cluster.ConcurrencyCounter.Decrement();
+            activityForTracing?.Stop();
```
MihaZupan, line 96: We shouldn't be stopping the activity here, but at the same scope where we're creating it.
If we want to start it as part of creating the proxy feature (so `ProxyPipelineInitializerMiddleware`), we should be stopping it there as well. This ensures both that the scope makes sense, and that the Activity is definitely stopped even if we never get to the `ForwarderMiddleware`.

This might mean something like
```c#
// In ProxyPipelineInitializerMiddleware
public Task Invoke(HttpContext context)
{
    var activity = ...;

    try
    {
        Task task = _next(context);
        return activity is null ? task : AwaitWithActivity(task, activity);
    }
    catch
    {
        activity?.Dispose();
        throw;
    }
}

private static async Task AwaitWithActivity(Task task, Activity activity)
{
    try
    {
        await task;
    }
    finally
    {
        activity.Dispose();
    }
}
```

src/ReverseProxy/Utilities/Observability.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Diagnostics;
+
+namespace Yarp.ReverseProxy.Utilities
+{
+    internal class Observability
```
MihaZupan, line 10:
```suggestion
    internal static class Observability
```

src/ReverseProxy/Utilities/Observability.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        private static ActivitySource _activitySource = new ActivitySource("Yarp.ReverseProxy");
+
+        public static ActivitySource YarpActivitySource => _activitySource;
```
MihaZupan, line 14: If these stay as internal helpers, I think all we need is
```c#
internal static class Observability
{
    public static readonly ActivitySource ActivitySource = new("Yarp.ReverseProxy");
}
```

src/ReverseProxy/Model/ReverseProxyFeature.cs
```diff
@@ -31,4 +33,12 @@ public IReadOnlyList<DestinationState> AvailableDestinations
 
     /// <inheritdoc/>
     public DestinationState? ProxiedDestination { get; set; }
+
+    public Activity? ActivityForTracing { get; init; }
```
MihaZupan, line 37: Storing this on our feature runs the risk of someone unintentionally dropping it if they have their own implementation of `IReverseProxyFeature`.
I also think it should be an internal implementation detail and therefore not a part of the interface.

We could
- Inject our own internal feature to the context at the time we create an Activity to flow it through, or
- Look at `Activity.Current` - we can have a helper along the lines of
  ```c#
  public static Activity? GetCurrentYarpActivity()
  {
      Activity? activity = Activity.Current;
      while (activity is not null && activity.Source != Observability.ActivitySource)
      {
          activity = activity.Parent;
      }
      return activity;
  }
  ```

src/ReverseProxy/SessionAffinity/SessionAffinityMiddleware.cs
```diff
@@ -75,10 +76,12 @@ private async Task InvokeInternal(HttpContext context, IReverseProxyFeature prox
                     // Policy reported the failure is unrecoverable and took the full responsibility for its handling,
                     // so we simply stop processing.
                     Log.AffinityResolutionFailedForCluster(_logger, cluster.ClusterId);
+                    proxyFeature.ActivityForTracing?.SetTag("Session Affinity", $"Failure in {policy.Name}");
                     return;
                 }
 
                 Log.AffinityResolutionFailureWasHandledProcessingWillBeContinued(_logger, cluster.ClusterId, failurePolicy.Name);
+                proxyFeature.ActivityForTracing?.SetTag("Session Affinity", $"Failure in {policy.Name}, recovered using {failurePolicy.Name}");
```
MihaZupan, line 84: I don't believe the tags added as part of session affinity belong on the activity, but should remain as logs only.
`AffinityResolutionFailedForCluster` is already set to `Error`.

If we think `AffinityResolutionFailureWasHandledProcessingWillBeContinued` is important enough, then we should consider increasing its level instead (currently `Debug`).

src/ReverseProxy/Model/ReverseProxyFeature.cs
samsp-msft: It was actually a triple condition as StartActivity is also conditional.

src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
samsp-msft: I moved it to ProxyPipelineInitializerMiddleware as that is initializing the ReverseProxyFeature,

src/ReverseProxy/Model/ReverseProxyFeature.cs
samsp-msft: I am storing it on the feature, but conditionally fetch it so it will fail gracefully if somebody uses their own IProxyFeature.

src/ReverseProxy/SessionAffinity/SessionAffinityMiddleware.cs
samsp-msft: Log correlation with tracing is not especially reliable. I think its important to know if the destination was not found as that will likely be a cause of failures, so makes it much clearer if that's the case.

MihaZupan: > Log correlation with tracing is not especially reliable

It should be. Wasn't that the point of https://github.com/dotnet/runtime/issues/34305?

src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
```diff
@@ -40,11 +42,19 @@ public async Task Invoke(HttpContext context)
         var route = context.GetRouteModel();
         var cluster = route.Cluster!;
 
+
+        var activityForTracing = ((ReverseProxyFeature)reverseProxyFeature).ActivityForTracing;
```
Tratcher, line 46: unsafe downcast. Use an as cast in case the type isn't want you expect. Or find a more reliable way to flow this data, like a separate feature?

src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
```diff
@@ -46,10 +48,15 @@ public Task Invoke(HttpContext context)
...
+            AvailableDestinations = destinationsState.AvailableDestinations,
+            ActivityForTracing = Observability.YarpActivitySource.StartActivity("Proxy Forwarder", ActivityKind.Server)
         });
+       
 
-        return _next(context);
+        await _next(context);
+
+        // Re-fetch the feature as it could have been changed during processing 
+        (context.GetReverseProxyFeature() as ReverseProxyFeature)?.ActivityForTracing?.Stop();
```
Tratcher, line 59: Only stop your own activity, not someone else'.

Start the activity, store it in a local field with a using, and let dispose call stop.
```
using var activity = Observability.YarpActivitySource.StartActivity("Proxy Forwarder", ActivityKind.Server);
...
ActivityForTracing = activity,
...
await _next(context);
```

src/ReverseProxy/Model/ReverseProxyFeature.cs
MihaZupan: I don't think our implementation of `IReverseProxyFeature` should be special here.
You can do so as an optimization, but the tracing shouldn't stop working if you create your own.

E.g. you could do
```c#
public static Activity? GetCurrentYarpActivity(IReverseProxyFeature feature)
{
    if (feature is ReverseProxyFeature knownFeature)
    {
        return knownFeature.Activity;
    }

    Activity? activity = Activity.Current;
    while (activity is not null && activity.Source != Observability.ActivitySource)
    {
        activity = activity.Parent;
    }
    return activity;
}
```

src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
MihaZupan: This will now allocate an extra async state machine even if the request isn't being sampled.
Can you please change the logic to something along the lines of what I shared here: https://github.com/microsoft/reverse-proxy/pull/2098#discussion_r1165816252?

src/ReverseProxy/Model/ReverseProxyFeature.cs
```diff
@@ -31,4 +33,9 @@ public IReadOnlyList<DestinationState> AvailableDestinations
 
     /// <inheritdoc/>
     public DestinationState? ProxiedDestination { get; set; }
+
+    /// <summary>
+    /// Diagnostics Activity for distributed tracing. Will only be created if there is a listener for "Yarp.ReverseProxy"
+    /// </summary>
+    internal Activity? ActivityForTracing { get; init; }
```
Tratcher, line 40: Move this to its own internal feature so you don't have to cast all the time.

src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
samsp-msft: What I have is probably the most efficient. Its internal to the proxy feature so it can't get replaced by somebody else - but if you replace the cluster/route, we create a new feature, and copy across the state including this. A local may work.

src/ReverseProxy/Model/ReverseProxyFeature.cs
samsp-msft: That's more state that needs to be added to HttpContext. We already have this object, so why not use it?
I only made it private as Miha didn't want it exposed.

Tratcher: > We already have this object, so why not use it?

because using it like this requires problematic coding practices. It's also really common for us to separate features into different groups based on functionality, making them independently pluggable.

samsp-msft: I was not keen on this using this approach, but after implementing, it's actually made it much cleaner with a helper in the observability class.

src/ReverseProxy/Health/ActiveHealthCheckMonitor.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.Extensions.Options;
 using Yarp.ReverseProxy.Model;
 using Yarp.ReverseProxy.Utilities;
+using System.Diagnostics;
```
Tratcher, line 13: Sort

src/ReverseProxy/Model/HttpContextFeaturesExtensions.cs
```diff
@@ -76,7 +76,7 @@ public static void ReassignProxyRequest(this HttpContext context, RouteModel rou
             Cluster = cluster.Model,
             AllDestinations = destinations.AllDestinations,
             AvailableDestinations = destinations.AvailableDestinations,
-            ProxiedDestination = oldFeature.ProxiedDestination,
```
Tratcher, line 79: revert

src/ReverseProxy/Model/IReverseProxyFeature.cs
```diff
@@ -35,4 +35,5 @@ public interface IReverseProxyFeature
     /// The actual destination that the request was proxied to.
     /// </summary>
     DestinationState? ProxiedDestination { get; set; }
+
```
Tratcher, line 38: revert

src/ReverseProxy/Model/ReverseProxyFeature.cs
```diff
@@ -3,6 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using Yarp.ReverseProxy.Utilities;
```
Tratcher, line 7: revert

src/ReverseProxy/SessionAffinity/SessionAffinityMiddleware.cs
```diff
@@ -56,10 +56,14 @@ private async Task InvokeInternal(HttpContext context, IReverseProxyFeature prox
...
         var affinityResult = await policy.FindAffinitizedDestinationsAsync(context, cluster, config, destinations, context.RequestAborted);
 
+        // Used for Distributed Tracing as part of Open Telemetry, null if there are no listeners
+        var activity = context.GetYarpActivity();
+
         switch (affinityResult.Status)
         {
             case AffinityStatus.OK:
                 proxyFeature.AvailableDestinations = affinityResult.Destinations!;
+                activity?.SetTag("Session Affinity", $"Destination selected by affinity policy {policy.Name}");
```
Tratcher, line 66: Are the tag contents supposed to be messages like logs? That seems like it would take a lot of extra storage. You're also creating extra strings even if the activity doesn't exist.
```suggestion
                activity?.SetTag("SessionPolicySuccess", policy.Name);
                ...
                activity?.SetTag("SessionPolicyFailure", policy.Name);
```

src/ReverseProxy/Utilities/Observability.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Yarp.ReverseProxy.Utilities
+{
+    internal static class Observability
+    {
+        public static readonly ActivitySource YarpActivitySource = new ActivitySource("Yarp.ReverseProxy");
+
+        public static Activity? GetYarpActivity(this HttpContext context)
+        {
+            return context.Features.Get<ActivityWrapper>()?.YarpActivity;
```
Tratcher, line 18: Much cleaner, thanks!

Tratcher (commented):
Great, just needs some cleaup.

src/ReverseProxy/SessionAffinity/SessionAffinityMiddleware.cs
```diff
@@ -56,10 +56,14 @@ private async Task InvokeInternal(HttpContext context, IReverseProxyFeature prox
...
         var affinityResult = await policy.FindAffinitizedDestinationsAsync(context, cluster, config, destinations, context.RequestAborted);
 
+        // Used for Distributed Tracing as part of Open Telemetry, null if there are no listeners
+        var activity = context.GetYarpActivity();
+
         switch (affinityResult.Status)
         {
             case AffinityStatus.OK:
                 proxyFeature.AvailableDestinations = affinityResult.Destinations!;
+                activity?.SetTag("SessionAffinitySuccess", policy.Name);
```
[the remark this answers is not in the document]
rwkarg: Similarly consider naming for other tags such as `activity?.AddTag("ClusterId", cluster.ClusterId);` to `activity?.AddTag("yarp.cluster_id", cluster.ClusterId);` to differentiate (Is this an Orleans cluster? Kubernetes cluster? Yarp cluster?)

src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
```diff
@@ -46,10 +48,15 @@ public Task Invoke(HttpContext context)
...
             AllDestinations = destinationsState.AllDestinations,
-            AvailableDestinations = destinationsState.AvailableDestinations
+            AvailableDestinations = destinationsState.AvailableDestinations,
         });
 
-        return _next(context);
+        var activity = Observability.YarpActivitySource.StartActivity("Proxy Forwarder", ActivityKind.Server);
+        context.SetYarpActivity(activity);
+
+        await _next(context);
```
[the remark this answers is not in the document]
samsp-msft: > This will now allocate an extra async state machine even if the request isn't being sampled. Can you please change the logic to something along the lines of what I shared here: [#2098 (comment)](https://github.com/microsoft/reverse-proxy/pull/2098#discussion_r1165816252)?

It only adds it if the activity is created, which depends on the listener. If the activity is null, nothing gets set.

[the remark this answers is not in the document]
MihaZupan: The issue is that the method is now `async`, which means you get an extra state machine allocation even if you don't have the activity.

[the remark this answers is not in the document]
MihaZupan: If you want you can keep it as-is and I can clean it up in a follow up PR

src/ReverseProxy/Forwarder/ForwarderMiddleware.cs
```diff
@@ -3,7 +3,9 @@
 
 using System;
 using System.Threading.Tasks;
+using System.Diagnostics;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Connections;
```
Tratcher, line 8: ?

Tratcher (approved):

src/ReverseProxy/Utilities/Observability.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                { "error", message },
+                { "description", description }
+            };
+
+            activity.AddEvent(new ActivityEvent("Error", default, tagsCollection));
+        }
+    }
+
+    private class YarpActivity
```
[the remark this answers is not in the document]
samsp-msft: That was @Kahbazi's idea.

src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
```diff
@@ -51,12 +51,11 @@ public async Task Invoke(HttpContext context)
...
         });
 
-        var activity = Observability.YarpActivitySource.StartActivity("proxy.forwarder", ActivityKind.Server);
-        context.SetYarpActivity(activity);
-
-        await _next(context);
-
-        activity?.Stop();
+        using (var activity = Observability.YarpActivitySource.StartActivity("proxy.forwarder", ActivityKind.Server))
+        {
```
Tratcher, line 55: FYI the brackets are optional. This will imply brackets from here to the end of the method.
`using var activity = Observability.YarpActivitySource.StartActivity("proxy.forwarder", ActivityKind.Server);`

src/ReverseProxy/Utilities/Observability.cs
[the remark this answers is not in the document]
Kahbazi: We must hunt down all the allocations 😆

PR #2108, opened 2023-04-20
Using new TimeProvider 

### LINKED ISSUE #2084: Consider using new TimeProvider instead of ITimer and ITimerFactory
Now that TimeProvider APIs have been added to .NET 8.0, we should consider if we want to replace some internal types that are used by Yarp. For instance, we should be able to replace

https://github.com/microsoft/reverse-proxy/blob/b6f00851fc1032706c275b414244a95ef70d3a26/src/ReverseProxy/Utilities/ITimerFactory.cs#L8-L11

with The new `TimeProvider.CreateTimer` that was introduced:

https://github.com/dotnet/runtime/blob/ce4857970dff1e987a3991238c8b32e8eae3cab7/src/libraries/Common/src/System/TimeProvider.cs#L137

cc: @tarekgh @Tratcher

Tratcher:
Fixes #2084

IClock was the only public abstraction type, it's been marked obsolete. It also wasn't consumed publicly anywhere, only internally via DI. All other internal abstractions have been removed. The only people potentially broken by this change are those injecting IClock into their own code or trying to inject their own IClock into our components for some reason.

The K8 code was using ISystemClock from Microsoft.Extensions. There are breaking public API changes there to replace it.

The SDK update broke some asserts in the telemetry consumers that I've commented out. Miha will fix these later.


test/Tests.Common/TestTimeProvider.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        public TimerCallback Callback { get; private set; }
+
+        public object State { get; private set; }
+
+        public bool IsDisposed { get; private set; }
+
+        public bool Change(TimeSpan dueTime, TimeSpan period)
+        {
+            DueTime = dueTime;
```
sebastienros, line 128: I am not sure if there is anything calling it this way, but there are some rules associated with special values of dueTime (0, -1) to disable it. And that's important when calling Task.Delay because it uses these values (calling CreateTimer(-1, -1) then Change(value, value).

src/ReverseProxy/Forwarder/StreamCopier.cs
```diff
@@ -142,27 +142,27 @@ private static async ValueTask<(StreamCopyResult, Exception?)> CopyAsync(Stream
...
             _isRequest = isRequest;
-            _clock = clock ?? throw new ArgumentNullException(nameof(clock));
-            _firstReadTime = new TimeSpan(-1);
+            _timePovider = timeProvider ?? throw new ArgumentNullException(nameof(timeProvider));
+            _firstReadTime = -1;
 
             ForwarderTelemetry.Log.ForwarderStage(isRequest ? ForwarderStage.RequestContentTransferStart : ForwarderStage.ResponseContentTransferStart);
 
-            _lastTime = clock.GetStopwatchTime();
+            _lastTime = timeProvider.GetTimestamp();
```
sebastienros, line 165: I don't think that's correct. The previous implementation was returning a Time (difference between two timestamps), `_stopwatch.Elapsed` but `GetTimestamp()` return `Stopwatch.GetTimestamp()` which is a point in time.

Here I believe to you need to keep a `startTimestamp` and all `timeProvider.GetElapsed(startTimestamp, timeProvider.GetTimestamp()).Ticks`. 

There is a PR in runtime to make it a existing extension method on `TimeProvider`.

Tratcher: GetStopwatchTime was an elapsed time since the Clock was created at the start of the app. Really it is just a point in time too. What's important is the time between _firstReadTime, _lastTime, etc..
https://github.com/microsoft/reverse-proxy/pull/2108/files/a71025b6f50bf2e9491431baa535a5437f2ec77d#diff-c863c8859756ee31b5c113136522cb73710fe7964b88e6f2cd9811863a99e369L12

test/Tests.Common/TestTimeProvider.cs
Tratcher: No, this isn't currently used with Task.Delay.

src/ReverseProxy/Health/TransportFailureRateHealthPolicy.cs
```diff
@@ -61,8 +61,8 @@ private DestinationHealth EvaluateProxiedRequest(ClusterState cluster, Destinati
         lock (history)
         {
             var failureRate = history.AddNew(
-                _clock.TickCount,
-                (long)_policyOptions.DetectionWindowSize.TotalMilliseconds,
+                _timeProvider.GetTimestamp(),
```
MihaZupan, line 64: The `tickCount` here was the number of milliseconds before.
For example, [`RecordWindowSize`](https://github.com/microsoft/reverse-proxy/blob/a71025b6f50bf2e9491431baa535a5437f2ec77d/src/ReverseProxy/Health/TransportFailureRateHealthPolicy.cs#L103) is currently a constant 1000, while it should be `TimeProvider.TimestampFrequency`.

It may be easier to store the `TimeProvider` in `ProxiedRequestHistory` and do all the math there.

Tratcher:
Note to self, offline feedback from Miha.
> I think the issue is in the test time provider.
You're returning TimeSpan ticks from [GetTimestamp](https://github.com/microsoft/reverse-proxy/blob/c836687b877d92e04dd53fa61fecf02d173a622a/test/Tests.Common/TestTimeProvider.cs#L73), but the frequency is still set to Stopwatch.Frequency.
You can either override TimestampFrequency and return TimeSpan.TicksPerSecond, or return (long)(_currentTime.TotalSeconds * TimestampFrequency) from GetTimestamp.
Or use some random frequency number that's different from TimeSpan/Stopwatch so that our tests don't accidentally work.
The frequency of Stopwatch on Unix is in nanoseconds, while TimeSpan ticks are 100-nanoseconds

test/ReverseProxy.Tests/Health/EntityActionSchedulerTests.cs
```diff
@@ -12,46 +12,46 @@ namespace Yarp.ReverseProxy.Health.Tests;
...
         VerifyEntities(scheduler, entity0, entity1);
-        Assert.Equal(2, timerFactory.Count);
-        timerFactory.VerifyTimer(0, Period0);
-        timerFactory.VerifyTimer(1, Period1);
+        Assert.Equal(2, timeProvider.TimerCount);
+        timeProvider.VerifyTimer(0, Period0);
+        timeProvider.VerifyTimer(1, Period1);
 
-        timerFactory.FireTimer(1);
+        timeProvider.FireTimer(1);
```
sebastienros, line 39: This PR has nothing to do with this, but I have to say that I find it a weird way to handle fake time. I would just expect the timer to call `Advance` and have the callbacks be triggers automatically, not the test to explicitly invoke the callback.

sebastienros (approved):
Left two comments but I assume you will have an answer and nothing to change

test/Tests.Common/TestTimeProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+/// </remarks>
+public class TestTimeProvider : TimeProvider
+{
+    private readonly List<TestTimer> _timers = new();
+
+    private TimeSpan _currentTime;
+
+    public int TimerCount => _timers.Count;
+
+    public override long TimestampFrequency => TimeSpan.TicksPerSecond;
```
MihaZupan, line 32:
```suggestion
    public override long TimestampFrequency => TimeSpan.TicksPerSecond * 7;
```

Nit: Can we have this return a non-standard value, so that we can catch product errors that our tests would be happy with (e.g. we offset a timestamp by `timeSpan.Ticks` somewhere instead of going through `TimeProvider`)?
`TimestampFrequency` and `GetTimestamp` are the only places that should have to change.

test/Tests.Common/TestTimeProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        // We could use this to fire timers, but timers are currently fired manually by tests.
+
+        _currentTime = targetTime;
+    }
+
+    public override DateTimeOffset GetUtcNow() => new DateTime(_currentTime.Ticks, DateTimeKind.Utc);
+
+    public override long GetTimestamp() => _currentTime.Ticks;
```
MihaZupan, line 75:
```suggestion
    public override long GetTimestamp() => _currentTime.Ticks * 7;
```

MihaZupan (approved):
Looks good, thanks!

test/ReverseProxy.Tests/Health/EntityActionSchedulerTests.cs
MihaZupan: Do we know when the built-in test time provider will ship, such that we can remove this one?

src/ReverseProxy/Health/EntityActionScheduler.cs
```diff
@@ -183,7 +183,7 @@ private void SetTimer()
 
             try
             {
-                _timer.Change(_period, Timeout.Infinite);
+                _timer.Change(_period, Timeout.InfiniteTimeSpan);
```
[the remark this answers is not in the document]
Tratcher: It re-starts itself after running the current callback.
https://github.com/microsoft/reverse-proxy/blob/4868c5529d69ed55ef7a7cd101a3a7ae183583e4/src/ReverseProxy/Health/EntityActionScheduler.cs#L238

PR #2120, opened 2023-05-03
Fix docs on route metadata

jmezach:
I recently tried using the Kubernetes Ingress controller pieces from this repository. While trying out some things I noticed that the doc wasn't correct on how metadata should be configured using annotations on the Ingress. If I did it the way the docs said I got the following error:

```
warn: Yarp.Kubernetes.Controller.Services.Reconciler[0]
      Uncaught exception occured while reconciling ingress default/minimal-ingress
      (Line: 1, Col: 1, Idx: 0) - (Line: 1, Col: 1, Idx: 0): Expected 'MappingStart', got 'SequenceStart' (at Line: 1, Col: 1, Idx: 0).
```

After removing the dashes from the annotation it worked correctly.


MihaZupan (approved):
Thank you

Issue #2018, opened 2023-01-30
Documentation Maintenance For Kubernetes Controller


nquandt:
### What should we add or change to make your life better?

  I noticed an update to the Kubernetes.Controller by @witskeeper, I realize now that "reviewing" a completed PR will probably not be noticed, so I submitting this issue as reference.
  #1913 

### Why is this important to you?

Required configurations should be mentioned / documented somewhere when changed.

![image](https://user-images.githubusercontent.com/57432850/215622172-ea85358e-83a0-4ba7-ba5b-5342b621be31.png)

MihaZupan:
I agree, at a minimum, it would be worth commenting on the `YarpOptions` when things are deemed required and updating examples to match. If there are any required settings, we should also validate that and fail early.

Even if updating the status failed, we likely shouldn't prevent the reconciler loop from continuing.

specialforest:
Also need `get` on `ingresses/status`. Samples don't have it.

specialforest:
Could you please update the title to make the bug searchable? 

Nevertheless, reposting error stack trace:
```
warn: Yarp.Kubernetes.Controller.Services.Reconciler[0]
      Uncaught exception occured while reconciling
      System.ArgumentNullException: Value cannot be null. (Parameter 'name')
         at k8s.AbstractKubernetes.k8s.ICoreV1Operations.ReadNamespacedServiceStatusWithHttpMessagesAsync(String name, String namespaceParameter, Nullable`1 pretty, IReadOnlyDictionary`2 customHeaders, CancellationToken cancellationToken)
         at k8s.CoreV1OperationsExtensions.ReadNamespacedServiceStatusAsync(ICoreV1Operations operations, String name, String namespaceParameter, Nullable`1 pretty, CancellationToken cancellationToken)
         at Yarp.Kubernetes.Controller.Client.V1IngressResourceStatusUpdater.UpdateStatusAsync()
         at Yarp.Kubernetes.Controller.Services.Reconciler.ProcessAsync(CancellationToken cancellationToken) in src\Kubernetes.Controller\Services\Reconciler.cs:line 71
```
and keywords ControllerServiceName ControllerServiceNamespace.

specialforest:
Also it updates a status of all ingresses in the cluster without `IsYarpIngress` check :(

PR #2136, opened 2023-05-17
Ingress status update fixes

specialforest:
- Updates samples with required permissions and ingress controller options.
- Adds code documentation for required options.
- Fixes case when the ingress controller updates statuses of ingresses not belonging to the controller.

This should address #2018. Also refactors changes from #1953 to avoid future errors.

### DIFF
--- a/samples/KubernetesIngress.Sample/Combined/ingress-controller.yaml
+++ b/samples/KubernetesIngress.Sample/Combined/ingress-controller.yaml
@@ -14,7 +14,9 @@ data:
       "Yarp": {
         "ControllerClass": "microsoft.com/ingress-yarp",
         "ServerCertificates": false,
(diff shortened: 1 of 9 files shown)


specialforest:
Dockerfiles Build failed with unrelated error:
```
RUN curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --jsonfile global.json
#11 18.78 curl: (56) OpenSSL SSL_read: error:0A000126:SSL routines::unexpected eof while reading, errno 32
...
#21 0.080 /bin/sh: 1: /root/.dotnet/dotnet: not found
```

Is there a way to re-run checks without making a dummy commit?

MihaZupan (approved):
Thank you

PR #2144, opened 2023-05-23
Enable AOT Compatibility

eerhardt:
Add `IsAotCompatible=true` to Yarp.ReverseProxy, which enables trimming and AOT analyzers.

Address the current warnings.

cc @agocke @karelz @adityamandaleeka


agocke (approved):
Lgtm

MihaZupan:
We will probably want to do this for the [Yarp.Telemetry.Consumption](https://github.com/microsoft/reverse-proxy/blob/main/src/TelemetryConsumption/Yarp.Telemetry.Consumption.csproj) package as well.

Issue #2145, opened 2023-05-24
Enable AOT compatibility for Yarp.Kubernetes.Controller


eerhardt:
### What should we add or change to make your life better?

Address all trimming and AOT warnings in Yarp.Kubernetes.Controller.

The current warnings  are from 2 API usages:

1. Using System.Text.Json Serialization
2. Using ConfigurationBinder

Both of these APIs use unbounded runtime Reflection. There are 2 general approaches to resolving these warnings:

1. Use the Source Generators for each. Note that the ConfigurationBinder source generator is new in .NET 8 - https://github.com/dotnet/runtime/issues/44493.
2. Manually writing the code the source generator/Reflection is doing. For example, using Utf8JsonReader/Writer.

I would suggest using the Source Generators.

Note that using the JSON Source Generator right now isn't possible because of https://github.com/dotnet/runtime/issues/79311. We need this functionality to serialize a `Message`:

https://github.com/microsoft/reverse-proxy/blob/7ef816467bb5bd0e82f08898a800e546ca0f86b9/src/Kubernetes.Controller/Protocol/Message.cs#L18-L22

So we will have to wait until that bug is fixed.

Here are the current warnings from this project:

```
1>C:\git\reverse-proxy\src\Kubernetes.Controller\Services\Reconciler.cs(62,36,62,82): warning IL3050: Using member 'System.Text.Json.JsonSerializer.Serialize<TValue>(TValue, JsonSerializerOptions)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. JSON serialization and deserialization might require types that cannot be statically analyzed and might need runtime code generation. Use System.Text.Json source generation for native AOT applications.
1>C:\git\reverse-proxy\src\Kubernetes.Controller\Services\Reconciler.cs(63,36,63,70): warning IL3050: Using member 'System.Text.Json.JsonSerializer.Serialize<TValue>(TValue, JsonSerializerOptions)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. JSON serialization and deserialization might require types that cannot be statically analyzed and might need runtime code generation. Use System.Text.Json source generation for native AOT applications.
[...]

PR #2146, opened 2023-05-24
Enable AOT compatibility for Yarp.Telemetry.Consumption

eerhardt:
Also fix a simple warning in Yarp.Kubernetes.Controller. The rest of the warnings in this project will be addressed in https://github.com/microsoft/reverse-proxy/issues/2145.

Follow up to https://github.com/microsoft/reverse-proxy/pull/2144#issuecomment-1561269092

### DIFF
--- a/src/Kubernetes.Controller/Management/KubernetesReverseProxyServiceCollectionExtensions.cs
+++ b/src/Kubernetes.Controller/Management/KubernetesReverseProxyServiceCollectionExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Diagnostics.CodeAnalysis;
 using k8s;
 using k8s.Models;
 using Microsoft.Extensions.Configuration;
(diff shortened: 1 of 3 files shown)


MihaZupan (approved):
How come something like
https://github.com/microsoft/reverse-proxy/blob/d6fee86a9af9537ac8c7baad17460d1f0aa4d3fa/src/TelemetryConsumption/TelemetryConsumptionExtensions.cs#L89
does not raise a warning, but something like
https://github.com/microsoft/reverse-proxy/blob/7ef816467bb5bd0e82f08898a800e546ca0f86b9/src/ReverseProxy/Management/ReverseProxyServiceCollectionExtensions.cs#L122
does?

eerhardt:
Oh wow, nice catch! This is a bug in the analyzer. https://github.com/dotnet/runtime/issues/86720.

I've fixed the warning and ensured when I publish an app rooting both Yarp assemblies there aren't any warnings from the trimming tool (which is separate from the Roslyn analyzer).

MihaZupan:
Thanks!

PR #2166, opened 2023-06-23
created configuration-driven proxy filter, which uses a handlebar-lik…

### LINKED ISSUE #2165: Feature request: Configuration-driven proxy filtering
In most of the scenarios where I've used YARP, I've found it useful to copy in a class I borrowed from another sample created during the `Project Tye` era that would read in values from `IConfiguration` to be pumped into the route and cluster endpoint values. In an example by Shayne Boyer, which I've mimicked in the `CustomConfigFilter` [class I created for the Build 2023 cloud native .NET demo code](https://github.com/bradygaster/dotnet-cloud-native-build-2023/blob/main/ContosoOnline/Proxy/CustomConfigFilter.cs), a configuration-driven filter would give developers a conventional way of mapping environment variables to placeholders in their YARP configuration. 

Take the `appsettings.json` code below, from the same Build 2023 demo repo:

```json
{
  "Logging": {
    "LogLevel": {
      "Default": "Information",
      "Microsoft.AspNetCore": "Warning",
      "System.Net.Http.HttpClient": "Warning",
      "Yarp": "Warning"
    }
  },
  "ORDERS_API": "http://localhost:5004",
  "STORE_UI": "http://localhost:5176",
  "ReverseProxy": {
    "Clusters": {
      "Frontend": {
        "Destinations": {
          "Store": {
            "Address": "{{STORE_UI}}"
          }
        }
      },
      "OrdersApi": {
        "Destinations": {
          "Store": {
            "Address": "{{ORDERS_API}}"
          }
        }
      }
    },
    "Routes": {
      "ClientRoute": {
        "ClusterId": "Frontend",
        "Match": {
          "Path": "{**catch-all}"
        }
      },
      "OrdersApiRoute": {
        "ClusterId": "OrdersApi",
        "Match": {
          "Path": "/api/orders/{**catch-all}"
        },
        "Transforms": [
          { "PathPattern": "/api/orders/{**remainder}" },
          { "PathRemovePrefix": "/api" }
        ]
      }
    }
  }
}
```

In this code, the `ORDERS_API` and `STORE_UI` values are hard-coded in the file as settings, so they'll be read in from here by default or from environment variables. The placeholders `{{ORDERS_API}}` and `{{STORE_UI}}` are replaced with the values in these settings or environment variables at runtime, making it easy for developers to use environment variables to route traffic between services. It also gives developers who want to integrate things like Tye with YARP to create new service discovery approaches. 

I'll follow this issue up with a pull request to add the filter to the product's code.

bradygaster:
created configuration-driven proxy filter, which uses a handlebar-like approach to wiring up configuration values - like routes and endpoints - based on values coming out of configuration.

resolves #2165

### DIFF
--- a/src/ReverseProxy/Configuration/ConfigurationDrivenFilter/ConfigurationDrivenFilter.cs
+++ b/src/ReverseProxy/Configuration/ConfigurationDrivenFilter/ConfigurationDrivenFilter.cs
@@ -1,0 +1,65 @@
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
+
+namespace Yarp.ReverseProxy.Configuration;
+
+internal class ConfigurationDrivenFilter : IProxyConfigFilter
+{
+    // Matches {{env_var_name}} or {{my-name}} or {{123name}} etc.
+    private readonly Regex _exp = new("\\{\\{(\\w+\\-?\\w+?)\\}\\}");
+    private readonly IConfiguration _configuration;
+
+    public ConfigurationDrivenFilter(IConfiguration configuration)
+    {
+        _configuration = configuration;
+    }
+
+    public ValueTask<ClusterConfig> ConfigureClusterAsync(ClusterConfig cluster, CancellationToken cancel)
+    {
+        if (cluster.Destinations is null)
+        {
+            return new ValueTask<ClusterConfig>(cluster);
+        }
+
+        // Each cluster has a dictionary of destinations, which is read-only, so we'll create a new one with our updates 
+        var newDests = new Dictionary<string, DestinationConfig>(StringComparer.OrdinalIgnoreCase);
+
+        foreach (var d in cluster.Destinations)
+        {
+            var origAddress = d.Value.Address;
+            if (_exp.IsMatch(origAddress))
+            {
+                var lookup = _exp.Matches(origAddress)[0].Groups[1].Value;
+                var newAddress = _configuration.GetValue<string>(lookup);
+
+                if (string.IsNullOrWhiteSpace(newAddress))
+                {
+                    throw new System.ArgumentException($"Configuration Filter Error: Substitution for '{lookup}' in cluster '{d.Key}' not found in configuration.");
+                }
+
+                var modifiedDest = d.Value with { Address = newAddress };
+                newDests.Add(d.Key, modifiedDest);
+            }
+            else
+            {
+                newDests.Add(d.Key, d.Value);
+            }
+        }
+
+        return new ValueTask<ClusterConfig>(cluster with { Destinations = newDests });
+    }
+
+    public ValueTask<RouteConfig> ConfigureRouteAsync(RouteConfig route, ClusterConfig? cluster, CancellationToken cancel)
+    {
+        if (route.Order.HasValue && route.Order.Value < 1)
+        {
+            return new ValueTask<RouteConfig>(route with { Order = 1 });
+        }
+
+        return new ValueTask<RouteConfig>(route);
+    }
+}
--- a/src/ReverseProxy/Configuration/ConfigurationDrivenFilter/ConfigurationDrivenFilterReverseProxyBuilderExtensions.cs
+++ b/src/ReverseProxy/Configuration/ConfigurationDrivenFilter/ConfigurationDrivenFilterReverseProxyBuilderExtensions.cs
@@ -1,0 +1,9 @@
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Yarp.ReverseProxy.Configuration;
+
+public static class ConfigurationDrivenFilterReverseProxyBuilderExtensions
+{
+    public static IReverseProxyBuilder AddConfigurationDrivenProxyFilter(this IReverseProxyBuilder builder)
+        => builder.AddConfigFilter<ConfigurationDrivenFilter>();
+}
--- a/src/ReverseProxy/Yarp.ReverseProxy.csproj
+++ b/src/ReverseProxy/Yarp.ReverseProxy.csproj
@@ -23,7 +23,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <None Include="README.md" Pack="true" PackagePath="\"/>
+    <None Include="README.md" Pack="true" PackagePath="\" />
   </ItemGroup>
 
   <ItemGroup>


src/ReverseProxy/Configuration/ConfigurationDrivenFilter/ConfigurationDrivenFilter.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        return new ValueTask<ClusterConfig>(cluster with { Destinations = newDests });
+    }
+
+    public ValueTask<RouteConfig> ConfigureRouteAsync(RouteConfig route, ClusterConfig? cluster, CancellationToken cancel)
+    {
+        if (route.Order.HasValue && route.Order.Value < 1)
+        {
+            return new ValueTask<RouteConfig>(route with { Order = 1 });
+        }
```
Tratcher, line 61: This is unrelated, it's copy-pasted from a different sample, right? Remove.

Tratcher (changes requested):
I don't think we want to do it this way. MapForwarder seems like a better fit.
https://github.com/microsoft/reverse-proxy/issues/2165#issuecomment-1605062293

Also, config filters have a weakness where they're not able to respond to configuration change notifications. We'll want to solve that before building something like this on them.

samsp-msft:
> Also, config filters have a weakness where they're not able to respond to configuration change notifications. We'll want to solve that before building something like this on them.

I agree on the problem of config filters adapting to changed, but I think the primary source for these values is environment variables that should be considered immutable during a process/container's lifetime.

samsp-msft:
> I don't think we want to do it this way. MapForwarder seems like a better fit. [#2165 (comment)](https://github.com/microsoft/reverse-proxy/issues/2165#issuecomment-1605062293)
> 
> Also, config filters have a weakness where they're not able to respond to configuration change notifications. We'll want to solve that before building something like this on them.

Nope. We want full goodness of YARP which you don't get with the forwarder. This needs name substitution to be able to deal with running locally and then in the cloud - and wanting to map service names for each of them based on environment.

src/ReverseProxy/Configuration/ConfigurationDrivenFilter/ConfigurationDrivenFilter.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
+
+namespace Yarp.ReverseProxy.Configuration;
+
+internal class ConfigurationDrivenFilter : IProxyConfigFilter
+{
+    // Matches {{env_var_name}} or {{my-name}} or {{123name}} etc.
+    private readonly Regex _exp = new("\\{\\{(\\w+\\-?\\w+?)\\}\\}");
```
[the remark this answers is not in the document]
bradygaster: i don't think it's a bad thing; i think the handlebars are something most web folks are accustomed to using.

bradygaster:
ping to @Tratcher and @samsp-msft - all changes you both suggested have been made. Let me know if there's anything additional needed here, and appreciate the reviews!

src/ReverseProxy/Configuration/Filters/ConfigSubstitutionFilter.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
+
+namespace Yarp.ReverseProxy.Configuration;
+
+public class ConfigSubstitutionFilter : IProxyConfigFilter
+{
+    // Matches {{env_var_name}} or {{my-name}} or {{123name}} etc.
+    private readonly Regex _exp = new("\\{\\{(\\w+\\-?\\w+?)\\}\\}");
```
davidfowl, line 13: Can we get a review on this regex? Also this should probably use the regex source generator.

src/ReverseProxy/Configuration/Filters/ConfigSubstitutionFilterProxyBuilderExtensions.cs
```diff
@@ -0,0 +1,9 @@
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Yarp.ReverseProxy.Configuration;
+
+public static class ConfigSubstitutionFilterProxyBuilderExtensions
+{
+    public static IReverseProxyBuilder AddConfigurationDrivenProxyFilter(this IReverseProxyBuilder builder)
```
davidfowl, line 7: API review?

src/ReverseProxy/Configuration/Filters/ConfigSubstitutionFilter.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        // Each cluster has a dictionary of destinations, which is read-only, so we'll create a new one with our updates 
+        var newDests = new Dictionary<string, DestinationConfig>(StringComparer.OrdinalIgnoreCase);
+
+        foreach (var d in cluster.Destinations)
+        {
+            var origAddress = d.Value.Address;
+            if (_exp.IsMatch(origAddress))
+            {
+                var lookup = _exp.Matches(origAddress)[0].Groups[1].Value;
```
stephentoub, line 36: Why IsMatch and then Matches(..)[0] rather than just using Match?

src/ReverseProxy/Configuration/Filters/ConfigSubstitutionFilter.cs
stephentoub: It'd be easier to read if the expression had fewer slashes:
```C#
@"{{(\w+-?\w+?)}}"
```

Is the intention that this can also match `{{my-}}` with no word after a dash?  If not, it'd be better as:
```C#
@"{{(\w+(?:-\w+)?)}}"
```

src/ReverseProxy/Configuration/Filters/ConfigSubstitutionFilter.cs
bradygaster: I presume they'd be equivalent, so that's a good call-out. @samsp-msft do you think this would enable consistent behavior?

stephentoub: IsMatch followed by Matches will run the regex twice rather than once, and Matches will result in allocating an additional collection object.  Just using Match, e.g.
```C#
Match m = _exp.Match(origAddress);
if (m.Success)
{
    var lookup = match.Groups[1].Value;
    ...
}
```
will have exactly the same functional semantics but be a heck of a lot cheaper.

bradygaster: Thanks! I'll update the PR with that.

karelz:
Triage: Looks like plans have changed, closing.

PR #2184, opened 2023-07-10
add cluster destination address validation

### LINKED ISSUE #2175: Cluster config validation does not error on a null destination address
### Describe the bug
I'm trying to make sure my config is valid on startup instead of later when we try to use it. To that end I setup a simple hosted service that will get the config on start and run the endpoints and clusters through the validator and throw if there's an error. However even with a null destination address I don't get any warnings.

### To Reproduce
setup a yarp with a basic route config and this as your cluster config:
```json
"Clusters": {
  "hgWeb": {
    "Destinations": {
      "hgWebServer": {
            "Address": null
      }
    }
  },
}
```
validation doesn't throw an error, however when you try to use the route you obviously get an exception:
```
      System.ArgumentNullException: Value cannot be null. (Parameter 'destinationPrefix')
         at Yarp.ReverseProxy.Forwarder.HttpForwarder.SendAsync(HttpContext context, String destinationPrefix, HttpMessageInvoker httpClient, ForwarderReq
uestConfig requestConfig, HttpTransformer transformer, CancellationToken cancellationToken)
         at Yarp.ReverseProxy.Forwarder.ForwarderMiddleware.Invoke(HttpContext context)
         at LexSyncReverseProxy.ProxyKernel.<>c.<<MapSyncProxy>b__1_1>d.MoveNext() in C:\dev\LexBox\backend\SyncReverseProxy\ProxyKernel.cs:line 74
      --- End of stack trace from previous location ---
         at Microsoft.AspNetCore.Routing.EndpointMiddleware.<Invoke>g__AwaitRequestTask|6_0(Endpoint endpoint, Task requestTask, ILogger logger)
         at Microsoft.AspNetCore.Authorization.AuthorizationMiddleware.Invoke(HttpContext context)
         at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
```

### Further technical details

- version: 2.0.1
- windows 11

hahn-kev:
closes #2175

This PR adds validation that the destination address in a cluster config is not null or empty. It's possible that the destination property of the cluster is null or empty, I decided not to validate that because it was not clear if that should be considered valid. I chose to not report that as an error. If it should be an error, then all the Validation unit tests would need to be updated as they currently expect null destinations to not produce an error.

### DIFF
--- a/src/ReverseProxy/Configuration/ConfigValidator.cs
+++ b/src/ReverseProxy/Configuration/ConfigValidator.cs
@@ -112,6 +112,7 @@ internal sealed class ConfigValidator : IConfigValidator
         }
 
         errors.AddRange(_transformBuilder.ValidateCluster(cluster));
+        ValidateDestinations(errors, cluster);
         ValidateLoadBalancing(errors, cluster);
         ValidateSessionAffinity(errors, cluster);
         ValidateProxyHttpClient(errors, cluster);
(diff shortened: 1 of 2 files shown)


adityamandaleeka (approved):

adityamandaleeka:
Thanks @hahn-kev!

PR #2210, opened 2023-08-01
Add `IDestinationResolver` for resolving cluster destination addresses

ReubenBond:
This PR adds `IDestinationResolver`, an abstraction for expanding the `ClusterConfig.Destinations` property, resolving the destination addresses into one or more addresses. The intention is to better support service discovery in YARP.

```csharp
/// <summary>
/// Resolves destination addresses.
/// </summary>
public interface IDestinationResolver
{
    /// <summary>
    /// Resolves the provided destinations and returns resolved destinations.
    /// </summary>
    /// <param name="destinations">The destinations to resolve.</param>
    /// <param name="cancellationToken">The cancellation token.</param>
    /// <returns>
    /// The resolved destinations and a change token used to indicate when resolution should be performed again.
    /// </returns>
    ValueTask<ResolvedDestinationCollection> ResolveDestinationsAsync(
        IReadOnlyDictionary<string, DestinationConfig> destinations,
        CancellationToken cancellationToken);
}
```

The PR also currently adds a `Host` property to the `DestinationConfig`, which is indented to act as the HTTP host name (for virtual host & SNI support).


ReubenBond:
Is someone able to advise me on the best way to implement the abovementioned `Host` feature?

ReubenBond:
I spoke with @Tratcher about how to flow the `Host` property through, and have implemented that in the current version of the PR.

ReubenBond:
Alright, I consider this ready for review.

src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -158,11 +160,19 @@ internal async Task<EndpointDataSource> InitialLoadAsync()
             var routes = new List<RouteConfig>();
             var clusters = new List<ClusterConfig>();
 
+            // Begin resolving config providers concurrently.
+            var resolvedConfigs = new List<(int Index, IProxyConfigProvider Provider, ValueTask<IProxyConfig> Config)>();
```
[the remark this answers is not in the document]
ReubenBond: Thanks, @Kahbazi!

ReubenBond:
@Tratcher @MihaZupan PTAL

src/ReverseProxy/Transforms/RequestHeaderOriginalHostTransform.cs
```diff
@@ -31,24 +32,26 @@ private RequestHeaderOriginalHostTransform(bool useOriginalHost)
...
+                // If there was no original host specified, allow the destination config host to flow through.
+                context.ProxyRequest.Headers.TryAddWithoutValidation(HeaderNames.Host, originalHost ?? destinationConfigHost);
             }
         }
-        else if (context.HeadersCopied
-            // Don't remove a custom host, only the original
-            && RequestUtilities.TryGetValues(context.ProxyRequest.Headers, HeaderNames.Host, out var existingHost)
-            && string.Equals(context.HttpContext.Request.Host.Value, existingHost.ToString(), StringComparison.Ordinal))
+        else if (((context.HeadersCopied && existingHost is not null && string.Equals(originalHost, existingHost, StringComparison.Ordinal))
+            || (!context.HeadersCopied && existingHost is null)))
```
MihaZupan, line 49: Should this be something like `else if (existingHost is null || existingHost == originalHost)` instead?

As written, we wouldn't make use of `destinationConfigHost` if `HeadersCopied && originalHost is null`, or if `HeadersCopied && existingHost is null`. Both of these feel like they should fall back to `destinationConfigHost`.

test/ReverseProxy.Tests/Management/ProxyConfigManagerTests.cs
```diff
@@ -1264,4 +1272,336 @@ public async Task LoadAsync_ConfigFilterRouteActionThrows_Throws()
...
+        var ex = configLoadException.Exception;
+        Assert.Equal("Throwing!", ex.Message);
+    }
+
+    [Fact]
+    public async Task LoadAsync_DestinationResolver_Reload_ThrowsAsync()
+    {
+        var configListener = new TestConfigChangeListener(includeApply: false);
+        var cts = new CancellationTokenSource();
+        var syncThrowResolver = new FakeDestinationResolver(async (destinations, cancellation) =>
```
MihaZupan, line 1567: Nit: `asyncThrowResolver` / `throwResolver`

test/ReverseProxy.Tests/Transforms/Builder/TransformBuilderTests.cs
```diff
@@ -54,8 +56,6 @@ private void NullOrEmptyTransforms_AddsDefaults(IReadOnlyList<IReadOnlyDictionar
         Assert.Empty(results.ResponseTrailerTransforms);
 
         Assert.Equal(6, results.RequestTransforms.Length);
-        var hostTransform = Assert.Single(results.RequestTransforms.OfType<RequestHeaderOriginalHostTransform>());
-        Assert.False(hostTransform.UseOriginalHost);
```
MihaZupan, line 58: Are these not applicable anymore?

test/ReverseProxy.Tests/Transforms/Builder/TransformBuilderTests.cs
```diff
@@ -283,41 +291,53 @@ public async Task UseOriginalHost(bool? useOriginalHost, bool? copyHeaders)
...
-        else if (copyHeaders.GetValueOrDefault(true))
-        {
-            var transform = Assert.Single(results.RequestTransforms);
-            var hostTransform = Assert.IsType<RequestHeaderOriginalHostTransform>(transform);
-            Assert.False(hostTransform.UseOriginalHost);
-        }
-        else
-        {
-            Assert.Empty(results.RequestTransforms);
+            //Assert.Equal(useOriginalHost.Value, results.ShouldCopyHostHeader.Value);
```
MihaZupan, line 315: Nit: There's a few commented-out asserts like this one - should they be removed?

src/ReverseProxy/Transforms/RequestHeaderOriginalHostTransform.cs
MihaZupan: (I stared at this for way too long so I may be missing some edge case)

MihaZupan (approved):
A few minor suggestions, otherwise this LGTM, thank you.

Might be worth adding a mention in the docs somewhere that this is an available extension point.

src/ReverseProxy/Transforms/RequestHeaderOriginalHostTransform.cs
ReubenBond: Agreed & good point

ReubenBond:
Feedback addressed, thank you @MihaZupan

ReubenBond:
Added a short doc on destination resolvers

src/ReverseProxy/Configuration/NoOpDestinationResolver.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Yarp.ReverseProxy.Configuration;
+
+/// <summary>
+/// An <see cref="IDestinationResolver"/> which performs no action.
+/// </summary>
+internal sealed class NoOpDestinationResolver : IDestinationResolver
+{
+    public ValueTask<ResolvedDestinationCollection> ResolveDestinationsAsync(IReadOnlyDictionary<string, DestinationConfig> destinations, CancellationToken cancellationToken)
+        => new(new ResolvedDestinationCollection(destinations, ChangeToken: null));
```
davidfowl, line 16: Unforutnate this needs to allocate.

davidfowl: Got it

ReubenBond:
`DnsDestinationResolver` defaults to adding a destination per IP, but that means we will get 2 destinations if DNS gives back an IPv4 & IPv6 address for the same host. You can specify the `AddressFamily` to query in the options.

Alternatively, we could scan over the query results and choose to prefer only IPv4 addresses or only IPv6 addresses if there is a mix of IPv4 and IPv6 addresses. I'm open to feedback.

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
```diff
@@ -0,0 +1,147 @@
...
+            uriBuilder.Host = addressString;
+            var resolvedAddress = uriBuilder.ToString();
+            var healthAddress = originalConfig.Health;
+            if (healthUriBuilder is not null)
+            {
+                healthUriBuilder.Host = addressString;
+                healthAddress = healthUriBuilder.ToString();
+            }
+
+            var name = $"{originalName}[{results.Count}]";
```
Tratcher, line 102: This naming is unstable if entries are added or removed over time. Use `{originalName}[{addressString}]` instead. We need stable names for session affinity.

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
```diff
@@ -0,0 +1,147 @@
...
+        var result = new UriBuilder()
+        {
+            Scheme = uri.Scheme,
+            Host = uri.Host,
+            Path = uri.AbsolutePath,
+            Query = uri.Query,
+            Fragment = uri.Fragment,
+        };
+
+        var userInfo = uri.UserInfo;
```
Tratcher, line 124: Drop UserInfo, it should _never_ be used.

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
```diff
@@ -0,0 +1,147 @@
...
+            results.Add((name, config));
+        }
+
+        return results;
+    }
+
+    /// <summary>
+    /// Creates a <see cref="UriBuilder"/> from the provided address, only setting the Port property if it is not the default value.
+    /// </summary>
+    private static UriBuilder CreateUriBuilder(Uri uri)
```
Tratcher, line 113: How is this different from `new UriBuilder(uri)`?

ReubenBond:
BTW note that I placed `IDestinationResolver` and friends into a new namespace, _Yarp.ReverseProxy.ServiceDiscovery_.

test/ReverseProxy.Tests/Transforms/Builder/TransformBuilderTests.cs
```diff
@@ -240,22 +242,31 @@ public void DefaultsCanBeDisabled()
         var results = transformBuilder.BuildInternal(route, new ClusterConfig());
         Assert.NotNull(results);
         Assert.False(results.ShouldCopyRequestHeaders);
-        Assert.Empty(results.RequestTransforms);
+        Assert.Single(results.RequestTransforms);
```
Tratcher, line 245: It's unfortunate that this can't be turned off completely like before. I don't want us to end up with default behavior that the user can't override. They can still replace or remove the host themselves, right?

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolverOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+/// </summary>
+public class DnsDestinationResolverOptions
+{
+    /// <summary>
+    /// The period between requesting a refresh of a resolved name.
+    /// </summary>
+    /// <remarks>
+    /// Defaults to 5 minutes.
+    /// </remarks>
+    public TimeSpan? RefreshPeriod { get; set; } = TimeSpan.FromMinutes(5);
```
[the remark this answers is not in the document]
davidfowl: Yea, but this is what gRPC ships with too (we should check their impl). Anything else requires external dependencies.

cc @jamesnk

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
ReubenBond: It's the `Port`. When I just use `new UriBuilder(uri)`, it will include ":80" in the Host, which is ugly (it's probably not a showstopper). It should only include the port if it's a non-default port for the scheme. The doc comment tries to indicate this: "only setting the Port property if it is not the default value."

test/ReverseProxy.Tests/Transforms/Builder/TransformBuilderTests.cs
ReubenBond: I agree. I had a version of this which moved the behavior to the front of the transform pipeline inside `StructuredTransformer` itself, but after speaking with Miha about it, I moved it to the existing transform at the end of the pipeline instead.

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
```diff
@@ -0,0 +1,147 @@
...
+                { } refreshPeriod when refreshPeriod > TimeSpan.Zero => new CancellationChangeToken(new CancellationTokenSource(refreshPeriod).Token),
+                _ => null,
+            };
+
+            return new ResolvedDestinationCollection(results, changeToken);
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Error resolving destinations");
+            throw;
```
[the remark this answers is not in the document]
ReubenBond: The caller will rethrow on first load, with "Unable to load or apply the proxy configuration."
On reload, it will log "Failed to reload config. Unable to register for change notifications, polling for changes until successful."

i.e, it will be logged the same as an `IProxyConfig.GetConfig()` exception. Is that fine?

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    private static async Task<List<(string Name, DestinationConfig Config)>> ResolveHostAsync(
+        DnsDestinationResolverOptions options,
+        string originalName,
+        DestinationConfig originalConfig,
+        CancellationToken cancellationToken)
+    {
+        var originalUri = new Uri(originalConfig.Address);
```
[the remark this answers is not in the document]
ReubenBond: What should it do if the TryParse fails?

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
```diff
@@ -0,0 +1,147 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+using Yarp.ReverseProxy.Configuration;
+
+namespace Yarp.ReverseProxy.ServiceDiscovery;
+
+/// <summary>
+/// Implementation of <see cref="IDestinationResolver"/> which resolves host names to IP addresses using DNS.
+/// </summary>
+public class DnsDestinationResolver : IDestinationResolver
```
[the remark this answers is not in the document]
ReubenBond: Not anymore, will change.

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolverOptions.cs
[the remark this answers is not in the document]
JamesNK: gRPC keeps open connections with resolved endpoints. This allows it to track the health of endpoints in realtime. If one of those connections goes down, such as because a k8s pod instance is restarted, then gRPC immediately asks the resolver to refresh itself. k8s will then return the new set of IP addresses, with the old pod instance removed and the new instance added, and gRPC attempts to connect to the new address.

The gRPC DNS resolve also has the option to configure an explicit refresh interval. This is mostly useful in situations where k8s scales up. Periodically checking DNS lets the client notice that there are new endpoints available.

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
Tratcher: And there's no way to make it clear the default port later?

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolverOptions.cs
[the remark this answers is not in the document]
Tratcher: Does DnsDestionationResolver need some kind of health check integration to trigger a refresh?

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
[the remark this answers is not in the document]
Tratcher: Log and return the original destination?

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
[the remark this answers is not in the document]
Tratcher: Yes, that's fine. If you want a more specific log, log without the full exception, maybe just the message?

Tratcher (approved):

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolverOptions.cs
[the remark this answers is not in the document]
ReubenBond: We could integrate YARP's health checks here, potentially to reload the `IProxyConfig` which the faulting destination belongs to. With DNS, we have no notification to tell us to perform a reload, but a Kubernetes-native solution could subscribe to those events by watching the relevant k8s Service resource for changes.

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
ReubenBond: Not from what I could tell, and not from what StackOverflow seemed to indicate 😁

src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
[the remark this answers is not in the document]
ReubenBond: Is that valid behavior? Eg, a Warning level log saying `Destination {DestinationId} has an unsupported Address. Addresses must be valid URIs.`?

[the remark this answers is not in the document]
Tratcher: We're careful not to over-validate/restrict destination addresses because people can plug in alternative transports like named pipes and use custom address formats. I think we only validate that it's not null or empty. A warning log seems appropriate in this case.

adityamandaleeka:
Woohoo, looks like this is green now. Ready to merge?

Issue #2163, opened 2023-06-19
Flow the status code from BadHttpRequestException 


Tratcher:
The AspNetCore servers (especially Kestrel) will throw BadHttpRequestException from the request body when something goes wrong and include a HTTP status code that can be forwarded to the client. E.g. 413 when the request body exceeds size limits.

Today YARP captures the request body exceptions and returns a generic 400. It could instead check for this exception and flow a more specific status code back to the client.

Hartaxfr:
Up !

PR #2227, opened 2023-08-27
Request body exception status code granularity

antonybstack:
This addresses the open issue, https://github.com/microsoft/reverse-proxy/issues/2163, which is related to the lack of status code granularity when returning a request body client error. The non-generic status codes of the bad request exceptions returned from aspnet net servers such as Kestrel or HttpSys were not being forwarded to the client.

### DIFF
--- a/test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
+++ b/test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
@@ -15,6 +15,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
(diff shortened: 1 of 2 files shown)


test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
```diff
@@ -1689,6 +1690,50 @@ public async Task RequestBodyClientErrorBeforeResponseError_Returns400()
         });
     }
 
+    [Theory]
+    [InlineData(StatusCodes.Status405MethodNotAllowed)]
+    [InlineData(StatusCodes.Status406NotAcceptable)]
+    [InlineData(StatusCodes.Status411LengthRequired)]
+    [InlineData(StatusCodes.Status412PreconditionFailed)]
+    [InlineData(StatusCodes.Status413PayloadTooLarge)]
```
Tratcher, line 1698: No need to test all of these combinations, most won't ever show up in this code path and the behavior doesn't depend on the specific status code. Just do 413 for now.

Tratcher (approved):

test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
antonybstack: Got it thanks. I made that change, but the build now has test failures. Reverting my changes did not fix the test failures locally, so from a quick glance, the failures seem to not be unrelated to this PR.

Tratcher:
Thanks, I'll deal with the unrelated test.

PR #2241, opened 2023-09-05
Improve diagnostics for IDestinationResolver exceptions

ReubenBond:
When an exception is thrown from `IDestinationResolver.ResolveDestinationsAsync`, we should catch the exception and rethrow it with context, such as which cluster failed resolution and, if possible, which destination failed to be resolved.

### DIFF
--- a/src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
+++ b/src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
@@ -66,11 +66,21 @@ internal class DnsDestinationResolver : IDestinationResolver
     {
         var originalUri = new Uri(originalConfig.Address);
         var originalHost = originalConfig.Host is { Length: > 0 } host ? host : originalUri.Authority;
-        var addresses = options.AddressFamily switch
+        var hostName = originalUri.DnsSafeHost;
+        IPAddress[] addresses;
+        try
         {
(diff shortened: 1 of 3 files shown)


src/ReverseProxy/ServiceDiscovery/DnsDestinationResolver.cs
```diff
@@ -66,11 +66,21 @@ public async ValueTask<ResolvedDestinationCollection> ResolveDestinationsAsync(I
...
-        };
+            addresses = options.AddressFamily switch
+            {
+                { } addressFamily => await Dns.GetHostAddressesAsync(hostName, addressFamily, cancellationToken).ConfigureAwait(false),
+                null => await Dns.GetHostAddressesAsync(hostName, cancellationToken).ConfigureAwait(false)
+            };
+        }
+        catch (Exception exception)
+        {
+            throw new InvalidOperationException($"Failed to resolve host {hostName}. See {nameof(Exception.InnerException)} for details.", exception);
```
Tratcher, line 81:
```suggestion
            throw new InvalidOperationException($"Failed to resolve host '{hostName}'. See {nameof(Exception.InnerException)} for details.", exception);
```

Tratcher (approved):

ReubenBond:
Thanks for the review. Will merge when CI completes

PR #2243, opened 2023-09-06
Update DailyBuilds.md

danmoseley:
fix broken link


MihaZupan (approved):
Thanks

MihaZupan:
@danmoseley this permission seems to be blocked on the organization level.

image

is there a way to add the backport action to some sort of an allow list?

PR #2244, opened 2023-09-06
add markdownlint

danmoseley:
this wouldn't have found https://github.com/microsoft/reverse-proxy/pull/2243, but it will at least find broken links within the repo.


danmoseley:
actually markdown link doesn't even find links between markdowns that are broken. it only validates links to anchors in the same file. still, it has value in that and its other rules.

MihaZupan (approved):
Thanks

Issue #1346, opened 2021-11-02
Document Kestrel response header encoding


MihaZupan:
From #1055:

Document Kestrel response header encoding (dotnet/aspnetcore#33776) [here](https://github.com/microsoft/reverse-proxy/blob/16ab855a5be980a0cf5fd21e9a7028190b6e7dda/docs/docfx/articles/http-client-config.md#httpclient).

ChintanRaval:
hi Chris @Tratcher / @MihaZupan / @karelz .. i'm trying to use yarp and facing [this encoding issue](https://github.com/microsoft/reverse-proxy/issues/440) due to a downstream system returning non-ASCII chars in the response header. since I don't have control over the downstream system, I am trying to see how we can be more tolerant at yarp's end. 

now that aspnetcore/kerner [is updated](https://github.com/dotnet/aspnetcore/pull/33776) (thanks Chris!) to optionally be more tolerant to non-ASCII characters, how can we tell yarp to leverage the tolerance. is there something similar to RequestHeaderEncoding coming up like ResponseHeaderEncoding or so? 

I was hoping this documentation update issue might help make that clear. even if someone can give a TL;DR on how to achieve this without this documentation-update-issue being resolved, it'd be great.

Tratcher:
You don't need to do anything YARP specific, you configure the Kestrel options directly in your proxy app.

ChintanRaval:
ah sorry. thanks @Tratcher. i'll try something via appSettings.. or even code like the below. 

```c#
builder.WebHost.ConfigureKestrel(serverOptions =>
{
    serverOptions.ResponseHeaderEncodingSelector = _ => System.Text.Encoding.Unicode;
});
```

Issue #2076, opened 2023-03-24
502 Errors for content disposition encoded in Base64


alkampfergit:
I have intermittent error calling an api that download a document and I find that the error is the following one.

```
"ResponseHeaders": The destination returned a response that cannot be proxied back to the client.
System.InvalidOperationException: Invalid non-ASCII or control character in header: 0x00C2
```

Actually I have this exact response-header in original response.
```
Content-disposition: attachment; filename="=?utf-8?B?T0YwMTMxNjEgLSAxsCBMaW5lYQ==?="
```

It seems that YARP does not like RFC2045 encoding, actually decoding from Base-64 cyber chef reports valid UTF-8 base64 encoded 

### To Reproduce
Just proxy a request that returned the exact content-disposition header.

I'm using version 2.0.0

MihaZupan:
YARP won't try decoding the filename, it'll copy the header back to the client as-is.

> Invalid non-ASCII or control character in header: 0x00C2

It looks like the header value actually contained a non-ASCII character (that wasn't encoded in base64).
I can't repro the issue by using `Content-disposition: attachment; filename="=?utf-8?B?T0YwMTMxNjEgLSAxsCBMaW5lYQ==?="` exactly, I suspect that the server response contained other characters.

You can try looking at the actual bytes the server responded with on the network, or a test transform like
```c#
.AddTransforms(builder =>
{
    builder.CopyResponseHeaders = false;

    builder.AddResponseTransform(context =>
    {
        if (context.ProxyResponse is not null)
        {
            if (context.ProxyResponse.Content.Headers.NonValidated.TryGetValues("Content-disposition", out var values))
            {
                Console.WriteLine($"Server sent 'Content-disposition: {values}' ({string.Join(", ", values.ToString().Select(c => (int)c))})");
            }
        }

        return default;
    });
});
```

Tratcher:
0xC@ is `Á`. Was that in a different header?

alkampfergit:
Actually it contains a file name that has some customer data so after my first post I've edited removing the sensitive part.

MihaZupan:
Do you have any custom logic in your YARP app that processes the headers? Can you setup a minimal repro?

If it is the backend server generating this header, I can't reproduce any issues with proxying such a response through YARP.

alkampfergit:
Managed to reproduce the issue https://github.com/alkampfergit/yarp-bug-reproduction 

[...]

PR #2254, opened 2023-09-14
💡feat: allow for simpler override of Response header encoding of Forwarded Requests

ChintanRaval:
## Overview

This change intends to allow consumers to be able to choose tolerance for non-default (non-ASCII) encoding of Response headers for forwarded requests by leveraging [`SocketsHttpHandler.ResponseHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler.responseheaderencodingselector). The idea is to provide something similar to the existing [`HttpClientConfig.RequestHeaderEncoding`](https://microsoft.github.io/reverse-proxy/api/Yarp.ReverseProxy.Configuration.HttpClientConfig.html#Yarp_ReverseProxy_Configuration_HttpClientConfig_RequestHeaderEncoding) property via a new `HttpClientConfig.ResponseHeaderEncoding` property.

Hopefully, this helps close https://github.com/microsoft/reverse-proxy/issues/1346 and helps simplify solution for https://github.com/microsoft/reverse-proxy/issues/2076.

## Changes

1. add support for [`SocketsHttpHandler.ResponseHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler.responseheaderencodingselector) via new `ResponseHeaderEncoding` member in [`HttpClientConfig`](https://microsoft.github.io/reverse-proxy/api/Yarp.ReverseProxy.Configuration.HttpClientConfig.html)
1. honour the same in [`ForwarderHttpClientFactory.ConfigureHandler`](https://microsoft.github.io/reverse-proxy/api/Yarp.ReverseProxy.Forwarder.ForwarderHttpClientFactory.html#Yarp_ReverseProxy_Forwarder_ForwarderHttpClientFactory_ConfigureHandler_Yarp_ReverseProxy_Forwarder_ForwarderHttpClientContext_SocketsHttpHandler_)
1. update existing & add new relevant tests
1. update relevant documentation for API, usage & samples

## Testing

No additional testing has been performed other than relying on CI to validate updated & newly added tests in code. Happy to discuss more testing if required.

### DIFF
--- a/docs/docfx/articles/http-client-config.md
+++ b/docs/docfx/articles/http-client-config.md
@@ -12,7 +12,7 @@ The configuration is represented differently if you're using the [IConfiguration
 These types are focused on defining serializable configuration. The code based configuration model is described below in the "Code Configuration" section.
 
 ### HttpClient
-HTTP client configuration is based on [HttpClientConfig](xref:Yarp.ReverseProxy.Configuration.HttpClientConfig) and represented by the following configuration schema.
+HTTP client configuration is based on [HttpClientConfig](xref:Yarp.ReverseProxy.Configuration.HttpClientConfig) and represented by the following configuration schema. If you need a more granular approach, please use a [custom implementation](https://microsoft.github.io/reverse-proxy/articles/http-client-config.html#custom-iforwarderhttpclientfactory) of `IForwarderHttpClientFactory`.
 ```JSON
 "HttpClient": {
     "SslProtocols": [ "<protocol-names>" ],
@@ -52,21 +52,6 @@ Configuration settings:
 - ResponseHeaderEncoding - enables other than ASCII encoding for incoming response headers (from requests that the proxy would forward out). Setting this value will leverage [`SocketsHttpHandler.ResponseHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler.responseheaderencodingselector) and use the selected encoding for all headers. The value is then parsed by [`Encoding.GetEncoding`](https://docs.microsoft.com/dotnet/api/system.text.encoding.getencoding#System_Text_Encoding_GetEncoding_System_String_), use values like: "utf-8", "iso-8859-1", etc.
 ```JSON
 "ResponseHeaderEncoding": "utf-8"
-```
-If you need more granular approach, please use custom `IForwarderHttpClientFactory`. However, if you're using an encoding other than ASCII (or UTF-8 for Kestrel) you also need to set your server to accept requests and/or send responses with such headers. For example, use [`KestrelServerOptions.RequestHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector) / [`.ResponseHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.ResponseHeaderEncodingSelector) to set up Kestrel to accept Latin1 ("iso-8859-1") headers:
-```C#
-private static IHostBuilder CreateHostBuilder(string[] args) =>
-    Host.CreateDefaultBuilder(args)
-        .ConfigureWebHostDefaults(webBuilder =>
-        {
-            webBuilder.UseStartup<Startup>()
-                      .ConfigureKestrel(kestrel =>
-                      {
-                          kestrel.RequestHeaderEncodingSelector = _ => Encoding.Latin1;
-                          // and/or
-                          kestrel.ResponseHeaderEncodingSelector = _ => Encoding.Latin1;
-                      });
-        });
 ```
 - EnableMultipleHttp2Connections - enables opening additional HTTP/2 connections to the same server when the maximum number of concurrent streams is reached on all existing connections. The default is `true`. See [SocketsHttpHandler.EnableMultipleHttp2Connections](https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler.enablemultiplehttp2connections)
 ```JSON
--- a/docs/docfx/articles/http-client-config.md
+++ b/docs/docfx/articles/http-client-config.md
@@ -53,7 +53,7 @@ Configuration settings:
 ```JSON
 "ResponseHeaderEncoding": "utf-8"
 ```
-Note that if you're using an encoding other than ASCII, you also need to set your server to accept-requests and/or send-responses with such headers. For example, when using Kestrel as server, use [`KestrelServerOptions.RequestHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector) / [`.ResponseHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.ResponseHeaderEncodingSelector) to configure Kestrel to allow `Latin1` ("`iso-8859-1`") headers:
+Note that if you're using an encoding other than ASCII, you also need to set your server to accept requests and/or send responses with such headers. For example, when using Kestrel as the server, use [`KestrelServerOptions.RequestHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector) / [`.ResponseHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.ResponseHeaderEncodingSelector) to configure Kestrel to allow `Latin1` ("`iso-8859-1`") headers:
 ```C#
 private static IHostBuilder CreateHostBuilder(string[] args) =>
     Host.CreateDefaultBuilder(args)
--- a/docs/docfx/articles/http-client-config.md
+++ b/docs/docfx/articles/http-client-config.md
@@ -19,6 +19,7 @@ HTTP client configuration is based on [HttpClientConfig](xref:Yarp.ReverseProxy.
     "MaxConnectionsPerServer": "<int>",
     "DangerousAcceptAnyServerCertificate": "<bool>",
     "RequestHeaderEncoding": "<encoding-name>",
+    "ResponseHeaderEncoding": "<encoding-name>",
     "EnableMultipleHttp2Connections": "<bool>"
     "WebProxy": {
         "Address": "<url>",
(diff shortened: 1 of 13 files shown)


ChintanRaval:
resolves https://github.com/microsoft/reverse-proxy/issues/1346

docs/docfx/articles/http-client-config.md
```diff
@@ -44,11 +45,15 @@ Configuration settings:
...
+```JSON
+"ResponseHeaderEncoding": "utf-8"
+```
+If you need more granular approach, please use custom `IForwarderHttpClientFactory`. However, if you're using an encoding other than ASCII (or UTF-8 for Kestrel) you also need to set your server to accept requests and/or send responses with such headers. For example, use [`KestrelServerOptions.RequestHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector) / [`.ResponseHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.ResponseHeaderEncodingSelector) to set up Kestrel to accept Latin1 ("iso-8859-1") headers:
```
ChintanRaval, line 56: Initially, I attempted to combine both `RequestHeaderEncoding` and `ResponseHeaderEncoding` secitons together to DRY things up but reverted to only DRY-ing up the common info below the two sections.

ChintanRaval:
I'll suss out the license agreement soon once I figure out what exact company name to use wiht my employer.

MihaZupan (approved):
The change looks good, thanks.

I don't really mind exposing the option given we have a corresponding setting for request headers already, but I question the usefulness of having it in the config given that Kestrel uses a different default for response headers, forcing you into code anyway.

docs/docfx/articles/http-client-config.md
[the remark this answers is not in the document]
ChintanRaval: > Kestrel doesn't allow you to send non-ASCII response headers by default. This comment should reflect that you must set that option even if you're proxying back UTF8

correct @MihaZupan. I _think_ (one) problem here is that the original commentary `or UTF-8 for Kestrel` is misleading. This change is merely consolidating some of that commentary towards the bottom while expanding it to include response terminology. I think it'd read more accurately if updated.

Moreover, i do like the suggestion of making the comment starting with `However..` redundant and move a sensible line to decorate the `RequestHeaderEncoding` & `ResponseHeaderEncoding` properties instead. 

> @Tratcher there's no way to set Kestrel's header encoding selectors via config, right?

@MihaZupan I believe that [this is possible indeed](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/options) - which makes it all config-drive now that we have both Request & Response header encoding properties in HttpClientConfig in YARP.

[Here's a commit](https://github.com/ChintanRaval/reverse-proxy/commit/463e02384847fc1f09aa4bdc1368bfc55bba74c3) that attempts to address all of this. _The code XML comments might be harder to read in a PR but look less-verbose and hopefully more-helpful or helpful enough to no longer require the additional commentary in the markdown documents. 🤞

ChintanRaval:
@microsoft-github-policy-service agree company="PageUpPeople"

docs/docfx/articles/http-client-config.md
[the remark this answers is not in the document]
MihaZupan: > I believe that [this is possible indeed](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/options)

I know it's possible to set some options, but I don't think the header encoding selectors work (they're delegates and not simple POCOs), have you tried that?

> I think (one) problem here is that the original commentary `or UTF-8 for Kestrel` is misleading.

It made sense when the doc was only talking about request encoding, as that means you often don't actually have to touch the Kestrel config to make the scenario work.

docs/docfx/articles/http-client-config.md
```diff
@@ -44,22 +45,13 @@ Configuration settings:
...
+- RequestHeaderEncoding - enables other than ASCII encoding for outgoing request headers. Setting this value will leverage [`SocketsHttpHandler.RequestHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler.requestheaderencodingselector) and use the selected encoding for all headers. The value is then parsed by [`Encoding.GetEncoding`](https://docs.microsoft.com/dotnet/api/system.text.encoding.getencoding#System_Text_Encoding_GetEncoding_System_String_), use values like: "utf-8", "iso-8859-1", etc.
 ```JSON
 "RequestHeaderEncoding": "utf-8"
 ```
-If you're using an encoding other than ASCII (or UTF-8 for Kestrel) you also need to set your server to accept requests with such headers. For example, use [`KestrelServerOptions.RequestHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector) to set up Kestrel to accept Latin1 ("iso-8859-1") headers:
```
MihaZupan, line 51: We should keep a section talking about how to set the Kestrel options.
For example, if you want to proxy UTF8 both ways, you'll have to set the `kestrel.ResponseHeaderEncodingSelector`.

Most people will set the client `ResponseHeaderEncoding` if they want to pass non-ASCII headers all the way through the proxy. We should warn them that they must also set the corresponding Kestrel option.

docs/docfx/articles/http-client-config.md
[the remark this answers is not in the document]
ChintanRaval: > I don't think the header encoding selectors work (they're delegates and not simple POCOs), have you tried that?

good point. I din't think of that. and neither have I tried that since i've achieved success by using code based options like advertised before in `http-client-config.md`. i'll add the code sample back to mitigate the risk.

docs/docfx/articles/http-client-config.md
```diff
@@ -44,11 +45,15 @@ Configuration settings:
...
+```JSON
+"ResponseHeaderEncoding": "utf-8"
+```
+Note that if you're using an encoding other than ASCII, you also need to set your server to accept-requests and/or send-responses with such headers. For example, when using Kestrel as server, use [`KestrelServerOptions.RequestHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector) / [`.ResponseHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.ResponseHeaderEncodingSelector) to configure Kestrel to allow `Latin1` ("`iso-8859-1`") headers:
```
MihaZupan, line 56:
```suggestion
Note that if you're using an encoding other than ASCII, you also need to set your server to accept requests and/or send responses with such headers. For example, when using Kestrel as the server, use [`KestrelServerOptions.RequestHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector) / [`.ResponseHeaderEncodingSelector`](https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.ResponseHeaderEncodingSelector) to configure Kestrel to allow `Latin1` ("`iso-8859-1`") headers:
```

src/ReverseProxy/Configuration/HttpClientConfig.cs
```diff
@@ -45,10 +48,33 @@ public sealed record HttpClientConfig
...
 
     /// <summary>
-    /// Enables non-ASCII header encoding for outgoing requests.
+    /// Allows overriding the default (ASCII) encoding for outgoing request headers.
+    /// <para>
+    /// Setting this value will in turn set <see href="https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler.requestheaderencodingselector">SocketsHttpHandler.RequestHeaderEncodingSelector</see> and use the selected encoding for all request headers. The value is then parsed by <see href="https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getencoding#system-text-encoding-getencoding(system-string)">Encoding.GetEncoding</see>, so use values like: "utf-8", "iso-8859-1", etc.
+    /// </para>
     /// </summary>
+    /// <remarks>
+    /// Note: If you're using an encoding other than ASCII here, then you may also need to set your server to accept request headers with such encoding via the corresponding options for the server.
```
MihaZupan, line 57:
```suggestion
    /// Note: If you're using an encoding other than UTF-8 here, then you may also need to configure your server to accept request headers with such an encoding via the corresponding options for the server.
```

src/ReverseProxy/Configuration/HttpClientConfig.cs
```diff
@@ -45,10 +48,33 @@ public sealed record HttpClientConfig
...
+    /// <para>
+    /// For example, when using Kestrel as server, use <see href="https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector">KestrelServerOptions.RequestHeaderEncodingSelector</see> to <see href="https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/options">configure Kestrel</see> to use the same encoding.
+    /// </para>
+    /// </remarks>
     public string? RequestHeaderEncoding { get; init; }
 
+    /// <summary>
+    /// Allows overriding the default (ASCII) encoding for incoming request headers.
```
MihaZupan, line 65:
```suggestion
    /// Allows overriding the default (Latin1) encoding for incoming request headers.
```

src/ReverseProxy/Configuration/HttpClientConfig.cs
```diff
@@ -45,10 +48,33 @@ public sealed record HttpClientConfig
...
+    /// <para>
+    /// Setting this value will in turn set <see href="https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler.responseheaderencodingselector">SocketsHttpHandler.ResponseHeaderEncodingSelector</see> and use the selected encoding for all response headers. The value is then parsed by <see href="https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getencoding#system-text-encoding-getencoding(system-string)">Encoding.GetEncoding</see>, so use values like: "utf-8", "iso-8859-1", etc.
+    /// </para>
+    /// </summary>
+    /// <remarks>
+    /// Note: If you're using an encoding other than ASCII here, then you may also need to set your server to send response headers with such encoding via the corresponding options for the server.
```
MihaZupan, line 71:
```suggestion
    /// Note: If you're using an encoding other than ASCII here, then you may also need to configure your server to send response headers with such an encoding via the corresponding options for the server.
```

src/ReverseProxy/Configuration/HttpClientConfig.cs
```diff
@@ -45,10 +48,33 @@ public sealed record HttpClientConfig
...
+    /// Setting this value will in turn set <see href="https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler.requestheaderencodingselector">SocketsHttpHandler.RequestHeaderEncodingSelector</see> and use the selected encoding for all request headers. The value is then parsed by <see href="https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getencoding#system-text-encoding-getencoding(system-string)">Encoding.GetEncoding</see>, so use values like: "utf-8", "iso-8859-1", etc.
+    /// </para>
     /// </summary>
+    /// <remarks>
+    /// Note: If you're using an encoding other than ASCII here, then you may also need to set your server to accept request headers with such encoding via the corresponding options for the server.
+    /// <para>
+    /// For example, when using Kestrel as server, use <see href="https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector">KestrelServerOptions.RequestHeaderEncodingSelector</see> to <see href="https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/options">configure Kestrel</see> to use the same encoding.
```
MihaZupan, line 59:
```suggestion
    /// For example, when using Kestrel as the server, use <see href="https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector">KestrelServerOptions.RequestHeaderEncodingSelector</see> to <see href="https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/options">configure Kestrel</see> to use the same encoding.
```

src/ReverseProxy/Configuration/HttpClientConfig.cs
```diff
@@ -45,10 +48,33 @@ public sealed record HttpClientConfig
...
+    /// Setting this value will in turn set <see href="https://docs.microsoft.com/dotnet/api/system.net.http.socketshttphandler.responseheaderencodingselector">SocketsHttpHandler.ResponseHeaderEncodingSelector</see> and use the selected encoding for all response headers. The value is then parsed by <see href="https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getencoding#system-text-encoding-getencoding(system-string)">Encoding.GetEncoding</see>, so use values like: "utf-8", "iso-8859-1", etc.
+    /// </para>
+    /// </summary>
+    /// <remarks>
+    /// Note: If you're using an encoding other than ASCII here, then you may also need to set your server to send response headers with such encoding via the corresponding options for the server.
+    /// <para>
+    /// For example, when using Kestrel as server, use <see href="https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector">KestrelServerOptions.RequestHeaderEncodingSelector</see> to <see href="https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/options">configure Kestrel</see> to use the same encoding.
```
MihaZupan, line 73:
```suggestion
    /// For example, when using Kestrel as the server, use <see href="https://docs.microsoft.com/dotnet/api/Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions.RequestHeaderEncodingSelector">KestrelServerOptions.RequestHeaderEncodingSelector</see> to <see href="https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/options">configure Kestrel</see> to use the same encoding.
```

src/ReverseProxy/Configuration/HttpClientConfig.cs
```diff
@@ -9,6 +9,9 @@ namespace Yarp.ReverseProxy.Configuration;
 /// <summary>
 /// Options used for communicating with the destination servers.
 /// </summary>
+/// <remarks>
+/// If you need a more granular approach, please use a <see href="https://microsoft.github.io/reverse-proxy/articles/http-client-config.html#custom-iforwarderhttpclientfactory">custom implementation of <c>IForwarderHttpClientFactory</c></see>
```
MihaZupan, line 13:
```suggestion
/// If you need a more granular approach, please use a <see href="https://microsoft.github.io/reverse-proxy/articles/http-client-config.html#custom-iforwarderhttpclientfactory">custom implementation of <c>IForwarderHttpClientFactory</c></see>.
```

docs/docfx/articles/http-client-config.md
MihaZupan: Thank you for the PR :)
I just tweaked the comments.

PR #2258, opened 2023-09-22
Update NotContains match mode for Header and QueryParameters to support missing/empty values

### LINKED ISSUE #2260: NotContains match modes for Header and QueryParameters incorrect for missing/empty values
Currently the `NotContains` mode for both `Header` and `QueryParameter` match types do not match correctly when the value of the header/query is empty or if the key does not exist entirely.

The "CurrentMatch" column below shows YARP's current match result while the "ExpectedMatch" column shows the expected match result.

| Type | Mode | RequestKey | RequestValue | CurrentMatch | ExpectedMatch |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Header/QueryParam | NotContains | None | - | No | **Yes** |
| Header/QueryParam | NotContains | Exists | Empty | No | **Yes** |
| Header/QueryParam | NotContains | Exists | NonEmpty | No | No |

evmurphy:
fixes #2260

Currently the NotContains modes for both Header and QueryParameter matches do not match correctly when the value of the header/query is empty or if the key does not exist entirely. 

Updated the match logic to the below (assuming any non-empty request values match the value specified by the route):

| Type | Mode | RequestKey | RequestValue | CurrentMatch | ExpectedMatch |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Header/QueryParam | NotContains | None | - | No | **Yes** |
| Header/QueryParam | NotContains | Exists | Empty | No | **Yes** |
| Header/QueryParam | NotContains | Exists | NonEmpty | No | No |

### DIFF
--- a/test/ReverseProxy.Tests/Routing/HeaderMatcherPolicyTests.cs
+++ b/test/ReverseProxy.Tests/Routing/HeaderMatcherPolicyTests.cs
@@ -154,7 +154,7 @@ public class HeaderMatcherPolicyTests
     [InlineData("", HeaderMatchMode.Exists, false)]
     [InlineData("abc", HeaderMatchMode.Exists, true)]
     [InlineData(null, HeaderMatchMode.NotExists, true)]
-    [InlineData("", HeaderMatchMode.NotExists, true)]
+    [InlineData("", HeaderMatchMode.NotExists, false)]
     [InlineData("abc", HeaderMatchMode.NotExists, false)]
     public async Task ApplyAsync_MatchingScenarios_AnyHeaderValue(string incomingHeaderValue, HeaderMatchMode mode, bool shouldMatch)
     {
--- a/test/ReverseProxy.Tests/Routing/HeaderMatcherPolicyTests.cs
+++ b/test/ReverseProxy.Tests/Routing/HeaderMatcherPolicyTests.cs
@@ -154,7 +154,7 @@ public class HeaderMatcherPolicyTests
     [InlineData("", HeaderMatchMode.Exists, false)]
     [InlineData("abc", HeaderMatchMode.Exists, true)]
     [InlineData(null, HeaderMatchMode.NotExists, true)]
-    [InlineData("", HeaderMatchMode.NotExists, false)]
+    [InlineData("", HeaderMatchMode.NotExists, true)]
     [InlineData("abc", HeaderMatchMode.NotExists, false)]
     public async Task ApplyAsync_MatchingScenarios_AnyHeaderValue(string incomingHeaderValue, HeaderMatchMode mode, bool shouldMatch)
     {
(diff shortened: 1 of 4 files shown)


evmurphy:
> @evmurphy please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

test/ReverseProxy.Tests/Routing/HeaderMatcherPolicyTests.cs
```diff
@@ -154,7 +154,7 @@ public void AppliesToEndpoints_NoMetadata_DoesNotApply()
     [InlineData("", HeaderMatchMode.Exists, false)]
     [InlineData("abc", HeaderMatchMode.Exists, true)]
     [InlineData(null, HeaderMatchMode.NotExists, true)]
-    [InlineData("", HeaderMatchMode.NotExists, false)]
+    [InlineData("", HeaderMatchMode.NotExists, true)]
```
Tratcher, line 157: I'm suspicious that we already had test coverage for this, meaning this was likely by design, not a bug. That makes the bar higher for changing the behavior. We need to talk to the partner team that originally submitted this feature.

MihaZupan: The `NotContains` change here looks good, an empty string truly does not contain other non-empty values.

I find the `Exists`/`NotExists` odd though. I'd expect that `NotExists` would only match if a header actually doesn't exist. If it's empty, the header does exist, it just had an empty value, but maybe that's just me?
On the other hand, it's odd to have cases where it's possible for `Exists` and `NotExists` to be the same. I'd personally rather see `Exists` return  `true` for empty values instead, but that goes against how we've documented this in the past.

evmurphy: Right so the issue is that currently `Exists` takes both the key and value into account whereas `NotExists` takes only the key into account. 

@MihaZupan I decided to update `NotExists` to take the value into account as well to align the two in this PR. But I agree and think both `Exists` and `NotExists` should both take only the presence of the key into account. For our use case of YARP we only need the `NotContains` logic to work correctly in the short term. Should we align `Exists` and `NotExists` in a separate PR?

Tratcher: Headers without values have no semantic meaning in HTTP and should be considered as non-existent.

evmurphy: I have reduced the scope of this PR to only fix NotContains match mode for now

MihaZupan (approved):
Thanks, I just changed the style a little bit

PR #2283, opened 2023-10-19
failing tests for rate limiter policy validation when 'default' or 'disable' used

jakebanks:
Debugging during a test run I can see that `_rateLimiterPolicyProvider.GetPolicyAsync(rateLimiterPolicyName)` always returns null for the values `default` or `disable`, and the reason seems to be that the collection the method is reading from is empty - should it contain some default policies?

### DIFF
--- a/src/ReverseProxy/Configuration/ConfigValidator.cs
+++ b/src/ReverseProxy/Configuration/ConfigValidator.cs
@@ -312,6 +312,7 @@ internal sealed class ConfigValidator : IConfigValidator
                 return;
             }
 
+            //TODO: this only gets invoked if a custom RateLimiter policy is named one of the following, AND it gets used. Is there a better way to check this at startup? or is it unlikely that the policy will be registered but not used?
             if (string.Equals(RateLimitingConstants.Default, rateLimiterPolicyName, StringComparison.OrdinalIgnoreCase)
                 || string.Equals(RateLimitingConstants.Disable, rateLimiterPolicyName, StringComparison.OrdinalIgnoreCase))
             {
--- a/src/ReverseProxy/Configuration/ConfigValidator.cs
+++ b/src/ReverseProxy/Configuration/ConfigValidator.cs
@@ -308,6 +308,13 @@ internal sealed class ConfigValidator : IConfigValidator
 
             if (policy is null)
             {
+                if (string.Equals(RateLimitingConstants.Default, rateLimiterPolicyName, StringComparison.OrdinalIgnoreCase)
+                    || string.Equals(RateLimitingConstants.Disable, rateLimiterPolicyName, StringComparison.OrdinalIgnoreCase))
+                {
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Configuration/ConfigValidator.cs
```diff
@@ -312,6 +312,7 @@ private async ValueTask ValidateRateLimiterPolicyAsync(IList<Exception> errors,
                 return;
             }
 
+            //TODO: this only gets invoked if a custom RateLimiter policy is named one of the following, AND it gets used. Is there a better way to check this at startup? or is it unlikely that the policy will be registered but not used?
             if (string.Equals(RateLimitingConstants.Default, rateLimiterPolicyName, StringComparison.OrdinalIgnoreCase)
                 || string.Equals(RateLimitingConstants.Disable, rateLimiterPolicyName, StringComparison.OrdinalIgnoreCase))
```
jakebanks, line 317: if the policy is not null then this line will throw, even if I haven't defined a custom RateLimiter policy with a conflicting name

MihaZupan: The current logic looks like it will report an error for a policy with such a name no matter if that policy exists or not, which is unfortunate.

I think this check (`name is Default or Disable`) should be moved up to before we call `GetPolicyAsync`, and just early-exit without reporting an error.

The logic that is acting on these policy names looks correct though.
https://github.com/mburumaxwell/reverse-proxy/blob/1be98d88cd42340d67e40fb352bd3431677e6b71/src/ReverseProxy/Routing/ProxyEndpointFactory.cs#L121-L132

Tratcher: > I think this check (`name is Default or Disable`) should be moved up to before we call `GetPolicyAsync`, and just early-exit without reporting an error.

We do want to report a name conflict though. This check is correct for doing that.

What's missing is that if the policy is null, we should not produce an error for the names Default or Disable. That goes on line 311.

Tratcher:
There should not be policies in the collection with these names, ever, that's what the check is for. However, we need to allow for a null result in that case.

src/ReverseProxy/Configuration/ConfigValidator.cs
Tratcher: Compare with ValidateCorsPolicyAsync

Tratcher: I adjusted this to match how we validate the auth policy in the prior method.

Tratcher:
@jakebanks please ack the CLA.

jakebanks:
@microsoft-github-policy-service agree [company="PageUpPeople"]

jakebanks:
> @jakebanks please ack the CLA.

sorry @Tratcher I only just caught up with your changes, done now

Tratcher (approved):

PR #2286, opened 2023-10-20
Set IsPackable to true for the k8s project on non rtw builds

### LINKED ISSUE #2134: Daily build broken Yarp.Kubernetes.Controller
### Some details

**How many backends are in your application?**

- [ ] 1-2
- [ ] 3-5
- [X] 6-10
- [ ] 10+

**How do you host your application?**

- [X] Kubernetes (AKS)
- [ ] Azure App Service
- [ ] Azure VMs
- [ ] Other Cloud Provider (please include details below)
- [ ] Other Hosting model (please include details below)

### What did you think of YARP?

Daily build for **only** the kubernetes controller is broken
it's stuck on
- 2.0.0-rc.2.23068.2
- Wednesday, January 18, 2023 (1/18/2023)

while the main yarp nuget is on
- 2.1.0-preview.1.23259.1
- Tuesday, May 9, 2023 (5/9/2023)

benjaminpetit:
Fix #2134

This will enable the package creation only for non rtw build, meaning that the NuGet will be created for nightly and preview builds.

### DIFF
--- a/src/Kubernetes.Controller/Yarp.Kubernetes.Controller.csproj
+++ b/src/Kubernetes.Controller/Yarp.Kubernetes.Controller.csproj
@@ -5,7 +5,8 @@
     <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
     <OutputType>Library</OutputType>
     <NoWarn>$(NoWarn);CS8002</NoWarn>
-    <IsPackable Condition="'$(PreReleaseVersionLabel)' == 'rtw'">false</IsPackable>
+    <IsPackable>true</IsPackable>
+    <IsShipping>false</IsShipping>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


MihaZupan:
I believe we wanted to re-enable only the nightly-build part that goes to the separate feed.
If we're shipping a preview version of YARP to NuGet, we don't want to also ship a preview of the k8s package. I don't know how we would go about disabling the publishing of this package only for the NuGet publishing though.

Ideally, we would find a way to do so without having to re-disable and re-enable this flag every time we publish a preview, but maybe that's an okay compromise given we don't release all that often.

benjaminpetit:
Then the only solution is to modify the powershell script that publishes the nuget to explicitly ignore this package (or even better, only publish a certain list of packages).

Any reason to not ship this package as a preview in the official nuget feed?

Tratcher:
> Any reason to not ship this package as a preview in the official nuget feed?

We wouldn't want to ship an RC version for example.

Suggestion:
- Make it always packable
- Mark it as non-shipping
- Update the release script to pull from '_microsoft-reverse-proxy-official\artifacts\release\shipping\' (excluding symbol packages) instead of '_microsoft-reverse-proxy-official\PackageArtifacts\'

Tratcher (approved):

PR #2307, opened 2023-11-09
Timeout integration

### LINKED ISSUE #2074: Integrate with or disable the AspNetCore 8 request timeouts
See https://github.com/dotnet/aspnetcore/issues/46818

AspNetCore 8.0 adds a new middleware for applying request timeouts. This augments HttpContext.RequestAborted. Timeouts can be configured globally, per endpoint, and/or by policy.

Problem 1) global timeouts can negatively impact long-running requests like WebSockets, gRPC, SignalR, etc.. Since YARP manages its own timeouts it should consider disabling the global default when proxying.

Problem 2) should YARP allow specifying timeout policies per registered route, similar to CORS, Auth, etc..? Or is that redundant with YARP's timeout settings?

Tratcher:
Fixes #2074 by exposing config that integrates with the new timeout middleware. This is a nearly identical clone of the Cors feature.

### DIFF
--- a/docs/docfx/articles/timeouts.md
+++ b/docs/docfx/articles/timeouts.md
@@ -45,37 +45,26 @@ Example:
 }
 ```
 
-Timeout policies can be configured in Startup.ConfigureServices as follows:
-```
-public void ConfigureServices(IServiceCollection services)
+Timeout policies and the default policy can be configured in the service collection and the middleware can be added as follows:
+```csharp
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddReverseProxy()
+    .LoadFromConfig(builder.Configuration.GetSection("ReverseProxy"));
+
+builder.Services.AddRequestTimeouts(options =>
 {
-    services.AddRequestTimeouts(options =>
-    {
-        options.AddPolicy("customPolicy", TimeSpan.FromSeconds(20));
-    });
-}
+    options.AddPolicy("customPolicy", TimeSpan.FromSeconds(20));
+});
+
+var app = builder.Build();
+
+app.UseRequestTimeouts();
+
+app.MapReverseProxy();
+
+app.Run();
 ```
-
-In Startup.Configure add the timeout middleware between Routing and Endpoints.
-
-```
-public void Configure(IApplicationBuilder app)
-{
-    app.UseRouting();
-
-    app.UseRequestTimeouts();
-
-    app.UseEndpoints(endpoints =>
-    {
-        endpoints.MapReverseProxy();
-    });
-}
-```
-
-
-### DefaultPolicy
-
-Specifying the value `default` in a route's `TimeoutPolicy` parameter means that route will use the policy defined in [RequestTimeoutOptions.DefaultPolicy](https://learn.microsoft.com/dotnet/api/microsoft.aspnetcore.http.timeouts.requesttimeoutoptions.defaultpolicy#microsoft-aspnetcore-http-timeouts-requesttimeoutoptions-defaultpolicy).
 
 ### Disable timeouts
 
--- a/docs/docfx/articles/timeouts.md
+++ b/docs/docfx/articles/timeouts.md
@@ -1,0 +1,86 @@
+# Request Timeouts
+
+## Introduction
+
+.NET 8 introduced the [Request Timeouts Middleware](https://learn.microsoft.com/aspnet/core/performance/timeouts) to enable configuring request timeouts globally as well as per endpoint. This functionality is also available in YARP 2.1 when running on .NET 8.
+
+## Defaults
+Requests do not have any timeouts by default, other than the [Activity Timeout](http-client-config.md#HttpRequest) used to clean up idle requests. A default policy specified in [RequestTimeoutOptions](https://learn.microsoft.com/dotnet/api/microsoft.aspnetcore.http.timeouts.requesttimeoutoptions) will apply to proxied requests as well.
+
+## Configuration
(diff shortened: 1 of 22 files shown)


src/ReverseProxy/Routing/ProxyEndpointFactory.cs
```diff
@@ -131,6 +138,25 @@ public Endpoint CreateEndpoint(RouteModel route, IReadOnlyList<Action<EndpointBu
             endpointBuilder.Metadata.Add(new EnableRateLimitingAttribute(config.RateLimiterPolicy));
         }
 #endif
+#if NET8_0_OR_GREATER
+        if (string.Equals(TimeoutPolicyConstants.Default, config.TimeoutPolicy, StringComparison.OrdinalIgnoreCase))
```
Kahbazi, line 142: Is the `TimeoutPolicyConstants.Default` really necessary? AuthZ and CORS needed a default one in order to add the default metadata. But here (and also in rate limiting) it seems it has no value and also limits the application of using `default` as a policy name.

src/ReverseProxy/Configuration/ConfigValidator.cs
```diff
@@ -294,7 +311,38 @@ private async ValueTask ValidateAuthorizationPolicyAsync(IList<Exception> errors
...
+                errors.Add(new ArgumentException($"Timeout policy '{timeoutPolicyName}' not found for route '{routeId}'."));
+            }
 
+            if (timeout.HasValue)
+            {
+                errors.Add(new ArgumentException($"Route '{routeId}' has both a Timeout '{timeout}' and TimeoutPolicy '{timeoutPolicyName}'."));
+            }
+        }
+
+        if (timeout.HasValue && timeout.Value.TotalMicroseconds <= 0)
```
Kahbazi, line 340: `timeout.Value.TotalMicroseconds <= 0` I guess this won't support the `Timeout.InfiniteTimeSpan` which has the -1 value for `TotalMilliseconds`

Kahbazi: Also the error message said that timeout should be greater than zero milliseconds but the condition is checking the microseconds.

src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
```diff
@@ -41,7 +47,17 @@ public Task Invoke(HttpContext context)
             context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
             return Task.CompletedTask;
         }
-
+#if NET8_0_OR_GREATER
+        // There's no way to detect the presence of the timeout middleware before this, only the options.
+        if (endpoint.Metadata.GetMetadata<RequestTimeoutAttribute>() != null
```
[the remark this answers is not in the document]
Tratcher: > 
```csharp
        if (timeSpan is null || timeSpan == Timeout.InfiniteTimeSpan)
        {
            return _next(context);
        }
```

This case is ok, RequestTimeoutAttribute shouldn't be set without a timeout or policy, and we don't allow InfiniteTimeSpan.

> 
```
        if (context.RequestAborted.IsCancellationRequested)
        {
            return _next(context);
        }
```

This case is harder... we could just return early in that case, but it might mess with the logging.

src/ReverseProxy/Routing/ProxyEndpointFactory.cs
Tratcher: I think you're right, we don't need an opt-in, that's handled centrally, but we do need an opt-out.

docs/docfx/articles/timeouts.md
```diff
@@ -0,0 +1,86 @@
...
+            "Address": "https://localhost:10001/"
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+Timeout policies can be configured in Startup.ConfigureServices as follows:
```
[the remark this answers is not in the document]
Tratcher: Most of our docs do this since we used to support 3.1. We can update them now that we require .NET 6. Filed https://github.com/microsoft/reverse-proxy/issues/2324. I'll only update this one for now.

src/ReverseProxy/Configuration/ConfigValidator.cs
Tratcher: > `timeout.Value.TotalMicroseconds <= 0` I guess this won't support the `Timeout.InfiniteTimeSpan` which has the -1 value for `TotalMilliseconds`

Right, there's no need when it's nullable.

src/ReverseProxy/Routing/ProxyEndpointFactory.cs
Kahbazi: Is the support for rate limiting has been released? Is it too late to remove the `Default` from rate limiting?

src/ReverseProxy/Configuration/ConfigValidator.cs
```diff
@@ -294,7 +311,37 @@ private async ValueTask ValidateAuthorizationPolicyAsync(IList<Exception> errors
...
+                errors.Add(new ArgumentException($"Timeout policy '{timeoutPolicyName}' not found for route '{routeId}'."));
+            }
 
+            if (timeout.HasValue)
+            {
+                errors.Add(new ArgumentException($"Route '{routeId}' has both a Timeout '{timeout}' and TimeoutPolicy '{timeoutPolicyName}'."));
+            }
+        }
+
+        if (timeout.HasValue && timeout.Value.TotalMilliseconds <= 0)
```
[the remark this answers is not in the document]
Tratcher: Correct. It wouldn't make much sense in the config format anyways "-00:00:00.01".

I think this is already covered by the current comments that the value must be > 0.

src/ReverseProxy/Routing/ProxyEndpointFactory.cs
Tratcher: Unfortunately yes, rate limiting shipped in 2.0. Luckily it's harmless.

benjaminpetit (approved):

PR #2311, opened 2023-11-10
Make RequestTransformContext.DestinationPrefix read-write

### LINKED ISSUE #2150: RequestTransformContext.DestinationPrefix should be read-write
### Describe the bug
commit:  527fafedb8bc049d89d864311e8012c743386672

Comments in [RequestTransformContext](https://github.com/microsoft/reverse-proxy/blob/main/src/ReverseProxy/Transforms/RequestTransformContext.cs) suggest that `DestinationPrefix` is able to be updated:

```csharp
     /// <summary>
    /// The outgoing proxy request. All field are initialized except for the 'RequestUri' and optionally headers.
    /// If no value is provided then the 'RequestUri' will be initialized using the updated 'DestinationPrefix',
    /// 'Path', and 'Query' properties after the transforms have run. The headers will be copied later when
    /// applying header transforms.
    /// </summary>
    public HttpRequestMessage ProxyRequest { get; init; } = default!;
```

but `DestinationPrefix` cannot be changed:

```csharp
    /// <summary>
    /// The URI prefix for the proxy request. This includes the scheme and host and can optionally include a
    /// port and path base. The 'Path' and 'Query' properties will be appended to this after the transforms have run.
    /// </summary>
    public string DestinationPrefix { get; init; } = default!;
```

**Question**
Is this a bug in the code or the comment?

benjaminpetit:
Fix #2150

### DIFF
--- a/test/ReverseProxy.Tests/Transforms/DestinationPrefixTransformTests.cs
+++ b/test/ReverseProxy.Tests/Transforms/DestinationPrefixTransformTests.cs
@@ -1,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Yarp.ReverseProxy.Transforms.Tests;
+
+public class DestinationPrefixTransformTests
+{
+    [Fact]
+    public async Task UpdateDestinationPrefix()
+    {
+        const string newDestinationPrefix = "http://localhost:8080";
+        var context = new RequestTransformContext()
+        {
+            DestinationPrefix = "http://contoso.com:5000"
+        };
+        var transform = new DestinationPrefixTransform(newDestinationPrefix);
+        await transform.ApplyAsync(context);
+    }
+
+    private class DestinationPrefixTransform(string newDestinationPrefix) : RequestTransform
+    {
+        public override ValueTask ApplyAsync(RequestTransformContext context)
+        {
+            context.DestinationPrefix = newDestinationPrefix;
+            return ValueTask.CompletedTask;
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


test/ReverseProxy.Tests/Transforms/DestinationPrefixTransformTests.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Yarp.ReverseProxy.Transforms.Tests;
+
+public class DestinationPrefixTransformTests
+{
+    [Fact]
+    public async Task UpdateDestinationPrefix()
```
Tratcher, line 12: A more meaningful test would be if the HttpForwarder correctly consumes the new value after the transforms run.

https://github.com/microsoft/reverse-proxy/blob/3a5a68fac3bccbcf19e785a82497abf4b1c30053/src/ReverseProxy/Forwarder/HttpForwarder.cs#L423-L438

Tratcher: Wait, maybe it does...
https://github.com/microsoft/reverse-proxy/blob/3a5a68fac3bccbcf19e785a82497abf4b1c30053/src/ReverseProxy/Transforms/Builder/StructuredTransformer.cs#L119-L125

benjaminpetit: Yeah the `HttpForwarder` has no notion of `RequestTransformContext`, it's only used in the `StructuredTransformer`. We don't have any test for this class, I can add a test case if you want.

Tratcher: Is the primary implementation, most of it's coverage is in HttpTransformerTests.
https://github.com/microsoft/reverse-proxy/blob/main/test/ReverseProxy.Tests/Forwarder/HttpTransformerTests.cs

It would be good to add a few tests for StructuredTransform specific scenarios though.

Tratcher (approved):

PR #2325, opened 2023-11-14
Clarify websocket activity failure

Tratcher:
This addresses outstanding feedback from @mihazupan about the error reporting changes in #2167. The intent was to more correctly attribute blame for disconnects to something caused by the client or destination. For ActivityTimeouts the blame would go to whoever we were currently expecting data from. However, in the WebSocket case we may be expecting data in both ways all the time, and an ActivityTimeout can be considered a no-blame situation that deserves separate consideration.

I'm not sure if the same consideration should apply to gRPC/streaming.

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -745,9 +745,10 @@ internal sealed class HttpForwarder : IHttpForwarder
         ForwarderError error;
 
         var (firstResult, firstException) = await firstTask;
+        var activityTimedOut = activityCancellationSource.IsCancellationRequested && !activityCancellationSource.CancelledByLinkedToken;
         if (firstResult != StreamCopyResult.Success)
         {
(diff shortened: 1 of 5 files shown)


src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -745,9 +745,10 @@ private static ValueTask<bool> CopyResponseStatusAndHeadersAsync(HttpResponseMes
         ForwarderError error;
 
         var (firstResult, firstException) = await firstTask;
+        var activityTimedOut = activityCancellationSource.IsCancellationRequested && !activityCancellationSource.CancelledByLinkedToken;
```
benjaminpetit, line 748: Probably NIT: could we move this check into `ReportResult` and pass the `activityCancellationSource` in parameter instead?

It seems unnecessary to check this every time in case of a sucess.

benjaminpetit (approved):

PR #2353, opened 2023-12-13
add weight cluster feature

endless001:
Add weighted clustering feature to yarp

### DIFF
--- a/src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
+++ b/src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 #if NET8_0_OR_GREATER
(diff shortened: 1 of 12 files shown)


Tratcher:
Please don't submit PRs without first filing an issue to discuss the requirements and design. There are many variations of this scenario and many ways to implement it.

src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
```diff
@@ -39,6 +40,12 @@ public Task Invoke(HttpContext context)
         var route = endpoint.Metadata.GetMetadata<RouteModel>()
             ?? throw new InvalidOperationException($"Routing Endpoint is missing {typeof(RouteModel).FullName} metadata.");
 
+        var weightClusters = route.WeightClusters;
+
+        if (weightClusters.Any())
+        {
+            route.SetCluster(PickCluster(weightClusters));
```
Tratcher, line 47: The route object is shared across all requests, you can't change it per request.

src/ReverseProxy/Model/WeightedList.cs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    public void Clear()
+    {
+        _list.Clear();
+        _weights.Clear();
+        Recalculate();
+    }
+
+    public T Next()
```
Tratcher, line 46: This would need to be thread safe.

endless001:
Sorry, I submitted the "PR" without any discussion. I have seen that many reverse proxy tools have built-in weight clustering function, so I want to add this function to Yarp as well.

karelz:
@endless001 as mentioned above, we should first discuss the feature and its design in an issue. Closing the PR for now. We can reopen it once we have feature agreement.

PR #2400, opened 2024-02-15
Fix X-Forwarded-For for IPv4 with dual stack sockets #2399

### LINKED ISSUE #2399: RequestHeaderXForwardedForTransform attaches IPv4-Mapped IPv6 Addresses to X-Forwarded-For in dual stack settings
### Describe the bug
TL;DR: When doing a request from a IPv4 IP `10.0.0.12` I noticed that YARP attached  `::ffff:10.0.0.12` to the X-Forwarded-For header.

When running with dual stack sockets in ASP.NET the remote IP is a [IPv4-Mapped IPv6 Address](https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.5.2) meaning a request from an IPv4 address ends up as an IPv6 IPAddress with [IsIPv4MappedToIPv6](https://learn.microsoft.com/en-us/dotnet/api/system.net.ipaddress.isipv4mappedtoipv6?view=net-8.0#system-net-ipaddress-isipv4mappedtoipv6) being true.

E.g. a connection from the IPv4 address `10.0.0.12` becomes the IPv6 address `::ffff:10.0.0.12` inside .NET.

To create the string to attach to the X-Forwarded-For header the transform uses `ToString()` on the remote address 
https://github.com/microsoft/reverse-proxy/blob/b12b34bae71ca4042160d73c8ede9149f11fcb3a/src/ReverseProxy/Transforms/RequestHeaderXForwardedForTransform.cs#L45

This results in the the IPv4-Mapped IPv6 address ending up in the X-Forwarded-For header.

As these addresses are [non-routable](https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml) and more an implementation quirk of dual stack sockets than "real" addresses it seems incorrect to me to place them in the X-Forwarded-For header.

The fix for this should be relatively simple. If the address is mapped, map it back to IPv4 and stringify that.

### To Reproduce
- Run YARP in on a dual stack setup. E.g. by using [the getting started example](https://microsoft.github.io/reverse-proxy/articles/getting-started.html) and running it with `dotnet run --urls "http://*:5020"` against a backend that allows inspecting headers.
- Execute a request from an IPv4 address (e.g. using `curl http://127.0.0.1:5020/`)
- Observe that the backend receives a X-Forwarded-For header with the non-routable IPv4-Mapped IPv6 Address (e.g. `::ffff:127.0.0.1`)

### Further technical details
- Mostly tried on Linux but I think it is the same on Windows

hacst:
Fixes #2399 . As discussed there checks whether the RemoteIpAddress is a IPv4-Mapped IPv6 Address and maps it back to a real IPv4 address before turning it into a string for the X-Forwarded-For header.


src/ReverseProxy/Transforms/RequestHeaderXForwardedForTransform.cs
```diff
@@ -42,7 +42,13 @@ public override ValueTask ApplyAsync(RequestTransformContext context)
             throw new ArgumentNullException(nameof(context));
         }
 
-        var remoteIp = context.HttpContext.Connection.RemoteIpAddress?.ToString();
+        string? remoteIp = null;
```
Tratcher, line 45: var remoteIpAddress = context.HttpContext.Connection.RemoteIPAddress;
....

Tratcher (approved):

src/ReverseProxy/Transforms/RequestHeaderXForwardedForTransform.cs
hacst: Changed as indicated.

MihaZupan (approved):
Thanks

PR #2412, opened 2024-02-20
Update notExists to true when value is empty

qianyuansun:
Currently **Exists** and **NotExists** are the same when value is empty. Headers with empty values have no semantic meaning in HTTP. So I updated NotExists with empty value to true. This will go against the current behavior by design. 
Feel free to post any comments or suggestions. Initial discussion is here https://github.com/microsoft/reverse-proxy/pull/2258#discussion_r1335320211


qianyuansun:

MihaZupan (approved):
Thanks.
This makes sense to me to get the behavior that `Exists` is always `!NotExists`.

adityamandaleeka (approved):
Triage: we're okay with this even though it is technically a breaking change.

PR #2415, opened 2024-02-22
Do not transition servers into the healthy state until at least one healthy probe is detected.

### LINKED ISSUE #2079: ConsecutiveFailuresHealthPolicy would wrongly mark a new faulty destination as healthy, causing proxy failure.
### Describe the bug

We noticed that there is a bug in `ConsecutiveFailuresHealthPolicy`: [link](https://github.com/microsoft/reverse-proxy/blob/ac99291bc8aba7b4a8aa49918be5ff2cae84d404/src/ReverseProxy/Health/ConsecutiveFailuresHealthPolicy.cs#L71)

`ConsecutiveFailuresHealthPolicy` doesn't factor in whether the current active `DestinationHealth` is Unknown before applying its regular failure count logic. This means that when a new faulty destinations are added but immediately return failures from active health probes, the policy would have a period where it mark destination as `DestinationHealth.Healthy` before it transition to` DestinationHealth.Unhealthy` based on the configured consecutive failure count.

Example if consecutive failures threshold is 2. 
When a new unreachable destination comes up:

| time | health state | Should be |
| ----------- | ----------- | ----------|
| t0 |  unknown | unknown |
| t1 |  healthy | unhealthy |
| t2 |  healthy | unhealthy |
| t3 |  unhealthy |  unhealthy |

The reason we encounter this problem is because we host service on Service Fabric Cluster. And we would do weekly deployment. Every time we have weekly deployment, we see proxy failure. Because, when new destination showing up, the service on this new destination are not reachable during the start up. But destination are marked as healthy in our gateway, due to the ConsecutiveFailuresHealthPolicy.

### Suggestion

Can we add a condition in ` EvaluateHealthState`?

If previous health state is unknown, and new health state is unhealthy, immediately return DestinationHealthState = unhealthy.

robbieknuth:
Fixes #2079 

~This fix is sort of a hack, but does not add any additional book keeping or change the health model while still yielding the desired behavior.~

~By simply initializing the failure count to the failure threshold, the first health probe will either keep the destination unhealthy, or it will swap to healthy just like before.~

Use the destination's previous state to inform the next state if we are below the failure theshold. This will prevent new destinations with slower server startup times from erroneously tranisitioning into healthy temporarily until the unhealthy threshold is eventually reached.

- Destinations with a state of `Unknown` will continue to be unknown until either the failure threshold is reached (=> Unhealthy) or a successful probe occurs (=> Healthy). This scenario is the patch.
- Unhealthy destinations are unaffected because the failure threshold is already reached.
- Healthy destinations maintain existing behavior and behave the same as bullet point 1.

### DIFF
--- a/src/ReverseProxy/Health/ConsecutiveFailuresHealthPolicy.cs
+++ b/src/ReverseProxy/Health/ConsecutiveFailuresHealthPolicy.cs
@@ -40,9 +40,10 @@ internal sealed class ConsecutiveFailuresHealthPolicy : IActiveHealthCheckPolicy
         for (var i = 0; i < probingResults.Count; i++)
         {
             var destination = probingResults[i].Destination;
+            var previousState = destination.Health.Active;
 
             var count = _failureCounters.GetOrCreateValue(destination);
-            var newHealth = EvaluateHealthState(threshold, probingResults[i].Response, count);
+            var newHealth = EvaluateHealthState(threshold, probingResults[i].Response, count, previousState);
             newHealthStates[i] = new NewActiveDestinationHealth(destination, newHealth);
         }
 
@@ -55,7 +56,7 @@ internal sealed class ConsecutiveFailuresHealthPolicy : IActiveHealthCheckPolicy
         return thresholdEntry.GetParsedOrDefault(_options.DefaultThreshold);
     }
 
-    private static DestinationHealth EvaluateHealthState(double threshold, HttpResponseMessage? response, AtomicCounter count)
+    private static DestinationHealth EvaluateHealthState(double threshold, HttpResponseMessage? response, AtomicCounter count, DestinationHealth previousState)
     {
         DestinationHealth newHealth;
         if (response is not null && response.IsSuccessStatusCode)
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Utilities/AtomicCounter.cs
```diff
@@ -9,6 +9,17 @@ internal sealed class AtomicCounter
...
+    /// <summary>
+    /// Creates an AtomicCounter with an initial value of 0.
+    /// </summary>
+    public AtomicCounter() : this(0) { }
+
+    /// <summary>
+    /// Creates an Atomic counter with an initial value of <paramref name="initialValue"/>
+    /// </summary>
+    /// <param name="initialValue">The initial value of the counter.</param>
+    public AtomicCounter(int initialValue) => _value = initialValue;
```
MihaZupan, line 21:
```suggestion
    /// <summary>
    /// Creates an Atomic counter with an initial value of <paramref name="initialValue"/>.
    /// </summary>
    /// <param name="initialValue">The initial value of the counter.</param>
    public AtomicCounter(int initialValue = 0) => _value = initialValue;
```

MihaZupan (commented):
Thank you for contributing here.
I think the behavior we want here is to keep the destination health as `Unknown` in this case until we see enough failures - see https://github.com/microsoft/reverse-proxy/issues/2079#issuecomment-1483666365.

We should have the necesarry bookkeeping in place already. We can pass the current destination health to `EvaluateHealthState` and then change the failure condition like so

```diff
-newHealth = currentFailureCount < threshold ? DestinationHealth.Healthy : DestinationHealth.Unhealthy;
+newHealth = currentFailureCount < threshold ? previousHealth : DestinationHealth.Unhealthy;
```

robbieknuth:
By simply using the previous state, the change can be smaller. The changes to AtomicCounter (and those tests) are no longer required.

MihaZupan (approved):
Thanks

PR #2421, opened 2024-02-27
Allow for using query strings in DefaultProbingRequestFactory.

### LINKED ISSUE #2420: Consider allowing query strings in health probes
Health probes are currently implemented strictly as paths. It is not possible to probe at at `/myhealth?key=value`. The resulting probe becomes `/myhealth%3Fkey=value`

It may or may not be desirable to include query strings in `DefaultProbingRequestFactory`, however I am opening this for consideration. 

Our current health probe includes a query string. It is simple enough to work around this by implementing a `IProbingRequestFactory`.

robbieknuth:
fixes #2420 

An attempt at allowing query strings in health probes.

If a query is specified both on the probe address and the probe path, the probe path wins.

### DIFF
--- a/src/ReverseProxy/Health/DefaultProbingRequestFactory.cs
+++ b/src/ReverseProxy/Health/DefaultProbingRequestFactory.cs
@@ -4,6 +4,7 @@
 using System.Net;
 using System.Net.Http;
 using System.Reflection;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.Net.Http.Headers;
 using Yarp.ReverseProxy.Model;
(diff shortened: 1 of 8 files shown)


src/ReverseProxy/Health/DefaultProbingRequestFactory.cs
```diff
@@ -18,9 +19,20 @@ internal sealed class DefaultProbingRequestFactory : IProbingRequestFactory
     public HttpRequestMessage CreateRequest(ClusterModel cluster, DestinationModel destination)
     {
         var probeAddress = !string.IsNullOrEmpty(destination.Config.Health) ? destination.Config.Health : destination.Config.Address;
+        UriHelper.FromAbsolute(probeAddress, out var destinationScheme, out var destinationHost, out var destinationPathBase, out var query, out _);
+
         var probePath = cluster.Config.HealthCheck?.Active?.Path;
```
Tratcher, line 24: The field is already named Path, including the query might be confusing. Would we want a separate field for the query instead?

robbieknuth: Sure. Probably better for compat too so that the way any existing `Path` is handled doesn't change.

robbieknuth:
Build issues: New property needs to be initialized in other places. Checking.

robbieknuth:
> Build issues: New property needs to be initialized in other places. Checking.

Issues have been addresssed. `Query` will now be properly supported via config files as well.

robbieknuth:
Close button is confusing.

src/ReverseProxy/Health/DefaultProbingRequestFactory.cs
```diff
@@ -20,7 +21,8 @@ public HttpRequestMessage CreateRequest(ClusterModel cluster, DestinationModel d
         var probeAddress = !string.IsNullOrEmpty(destination.Config.Health) ? destination.Config.Health : destination.Config.Address;
         var probePath = cluster.Config.HealthCheck?.Active?.Path;
         UriHelper.FromAbsolute(probeAddress, out var destinationScheme, out var destinationHost, out var destinationPathBase, out _, out _);
-        var probeUri = UriHelper.BuildAbsolute(destinationScheme, destinationHost, destinationPathBase, probePath, default);
+        var query = QueryString.FromUriComponent(cluster.Config.HealthCheck?.Active?.Query ?? "");
```
robbieknuth, line 24: FromUriComponent can actually handle a null string, but it isn't annotated as such.

MihaZupan (approved):
Thanks.
A bit unfortunate we didn't just make it `PathAndQuery` to begin with.

PR #2426, opened 2024-03-01
Update HttpSysDelegator to support detaching from and re-initializing queues when the http.sys queue no longer exists

### LINKED ISSUE #2425: Http.sys delegation should handle ERROR_OBJECT_NO_LONGER_EXISTS and try to re-attach to the queue
To fix the issue where receivers can't properly restart and re-create their queues due to the fact delegators still have an open handle, the Windows http.sys team had made a change to detach the handle the delegator has when the receiver detaches from the queue. When this happens, http.sys will return an ERROR_OBJECT_NO_LONGER_EXISTS error when trying to delegate. The expectation when this happens is for the delegator to detach from the no-longer valid queue and attempt to attach to the newly created queue (assuming it has been created already).

NGloreous:
This PR fixes #2425 by detaching and attempting to re-attach to destination queues when an ERROR_OBJECT_NO_LONGER_EXISTS  error is received when attempting to delegate requests.

### DIFF
--- a/src/ReverseProxy/Utilities/EventIds.cs
+++ b/src/ReverseProxy/Utilities/EventIds.cs
@@ -68,4 +68,5 @@ internal static class EventIds
     public static readonly EventId RetryingWebSocketDowngradeNoHttp2 = new EventId(62, "RetryingWebSocketDowngradeNoHttp2");
     public static readonly EventId InvalidSecWebSocketKeyHeader = new EventId(63, "InvalidSecWebSocketKeyHeader");
     public static readonly EventId TimeoutNotApplied = new(64, nameof(TimeoutNotApplied));
+    public static readonly EventId DelegationQueueNoLongerExists = new(64, nameof(DelegationQueueNoLongerExists));
 }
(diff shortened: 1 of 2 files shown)


src/ReverseProxy/Utilities/EventIds.cs
```diff
@@ -68,4 +68,5 @@ internal static class EventIds
     public static readonly EventId RetryingWebSocketDowngradeNoHttp2 = new EventId(62, "RetryingWebSocketDowngradeNoHttp2");
     public static readonly EventId InvalidSecWebSocketKeyHeader = new EventId(63, "InvalidSecWebSocketKeyHeader");
     public static readonly EventId TimeoutNotApplied = new(64, nameof(TimeoutNotApplied));
+    public static readonly EventId DelegationQueueNoLongerExists = new(64, nameof(DelegationQueueNoLongerExists));
```
MihaZupan, line 71:
```suggestion
    public static readonly EventId DelegationQueueNoLongerExists = new(65, nameof(DelegationQueueNoLongerExists));
```

MihaZupan (approved):
Thank you

PR #2433, opened 2024-03-08
Add icon and update nuget metadata

jamesmontemagno:
Now has an icon! 
image


src/ReverseProxy/Yarp.ReverseProxy.csproj
```diff
@@ -25,6 +27,7 @@
 
   <ItemGroup>
     <None Include="README.md" Pack="true" PackagePath="\" />
+    <None Include="..\..\assets\icon.png" PackagePath="icon.png" Pack="true" />
```
WeihanLi, line 30: Think maybe we could define it in `Directory.Build.props` to avoid duplicate code, as all packages would need this

https://github.com/microsoft/reverse-proxy/blob/main/src/Directory.Build.props

MihaZupan (approved):
Thanks.
Mind adding it to the `Yarp.Kubernetes.Controller.csproj` package as well (not shipping yet)? Or include it in `Directory.Build.props` if possible

PR #2436, opened 2024-03-13
Add yarp.ingress.kubernetes.io/route-queryparameters to allow ingress forking based on query parameters

jsedlak-microsoft:


MihaZupan (approved):
Thanks. This looks reasonable to me to add

PR #2444, opened 2024-03-20
Remove 1.x builds from weekly builds

benjaminpetit:


MihaZupan (approved):
Is this because we're not updating the pipelines on these branches?

benjaminpetit:
Yes, and it's not worth it running them every week if we don't moify these branches anyway. If we ever had to, we will have to first migrate them to the new template build.

PR #2446, opened 2024-03-20
Update KubernetesIngress docker builds

benjaminpetit:


tomap:
-c Release is now the default with DotNet publish in  .Net 8

benjaminpetit:
> -c Release is now the default with DotNet publish in .Net 8

Better to be explicit :)

samples/KubernetesIngress.Sample/Combined/Dockerfile
```diff
@@ -1,12 +1,13 @@
 #See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.
 
-FROM mcr.microsoft.com/dotnet/aspnet:6.0 AS base
+FROM mcr.microsoft.com/dotnet/aspnet:8.0 AS base
 WORKDIR /app
 EXPOSE 80
 EXPOSE 443
 
-FROM amd64/buildpack-deps:jammy-curl AS publish
+FROM mcr.microsoft.com/dotnet/sdk:8.0  AS publish
```
WeihanLi, line 8: Since we install dotnet sdk via dotnet-install script, do we still need to use sdk image as base image?

benjaminpetit: Good question, we still need to install the SDK because we target versions that are not published in docker.

The SDK docker image contains dependencies needed for the dotnet script to run properly.

PR #2448, opened 2024-03-24
feat: removed unnecessary to array allocation

AliKhalili:
This PR aims to remove some extra memory allocation due to unnecessary calling of `ToArray` method.


AliKhalili:

MihaZupan (commented):
Thank you for looking to contribute! We're happy to accept performance improvements, but it's not always obvious whether something will be an improvement just by reading the code.

Can you please revert the places I've mentioned where this would be a regression?

PR #2474, opened 2024-04-11
Turn off a bunch of optional features in the benchmark

MihaZupan:
Skips logic such as ASP.NET's auth middlewares, routing, .... Also skips some optional features like the `X-Forwarded` headers being added by default.

This way it more closely matches what we do in the proxy-httpclient scenario: https://github.com/aspnet/Benchmarks/blob/main/src/BenchmarksApps/HttpClient/Proxy/Program.cs

Net effect a is ~17 % RPS bump on the basic h1-h1 scenario.

### DIFF
--- a/testassets/BenchmarkApp/Properties/launchSettings.json
+++ b/testassets/BenchmarkApp/Properties/launchSettings.json
@@ -2,7 +2,7 @@
   "profiles": {
     "BenchmarkApp": {
       "commandName": "Project",
-      "commandLineArgs": "--urls http://localhost:5000 --clusterUrls http://localhost:10010",
+      "commandLineArgs": "--urls http://localhost:5000 --clusterUrls http://httpbin.org",
       "environmentVariables": {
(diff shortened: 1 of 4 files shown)


testassets/BenchmarkApp/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "BenchmarkApp": {
       "commandName": "Project",
-      "commandLineArgs": "--urls http://localhost:5000 --clusterUrls http://localhost:10010",
+      "commandLineArgs": "--urls http://localhost:5000 --clusterUrls http://httpbin.org",
```
benjaminpetit, line 5: Do we want to commit that?

MihaZupan: Do you see any other problems with the change?

benjaminpetit (approved):

PR #2487, opened 2024-05-03
Update queue get/create/update logic in HttpSysDelegator.AddOrUpdateRules to avoid GC timing bug

### LINKED ISSUE #2486: HttpSysDelegator - AddOrUpdateRules fails to add rule due to bug with WeakReference/GC
### Describe the bug
[This unexpected condition](https://github.com/microsoft/reverse-proxy/blob/b8cc3f3df202ca7aa7c1bc03f15013555ba2d9e9/src/ReverseProxy/Delegation/HttpSysDelegator.cs#L200) can be hit if a GC occurs after a [new DelegationQueue is initialized](https://github.com/microsoft/reverse-proxy/blob/b8cc3f3df202ca7aa7c1bc03f15013555ba2d9e9/src/ReverseProxy/Delegation/HttpSysDelegator.cs#L171) but before it's [pulled back out of the WeakReference](https://github.com/microsoft/reverse-proxy/blob/b8cc3f3df202ca7aa7c1bc03f15013555ba2d9e9/src/ReverseProxy/Delegation/HttpSysDelegator.cs#L178). 

That is because even though the queue variable is still used after [here](https://github.com/microsoft/reverse-proxy/blob/b8cc3f3df202ca7aa7c1bc03f15013555ba2d9e9/src/ReverseProxy/Delegation/HttpSysDelegator.cs#L172) the actual object isn't necessarily referenced, so the JIT/GC considers it fine to clean up the object.

NGloreous:
Fixes #2486

### DIFF
--- a/src/ReverseProxy/Delegation/HttpSysDelegator.cs
+++ b/src/ReverseProxy/Delegation/HttpSysDelegator.cs
@@ -164,48 +164,32 @@ internal sealed class HttpSysDelegator : IHttpSysDelegator, IClusterChangeListen
                 var queueKey = new QueueKey(queueName, urlPrefix);
                 if (!_queuesPerDestination.TryGetValue(destination, out var queue) || !queue.Equals(queueKey))
                 {
-                    if (!_queues.TryGetValue(queueKey, out var queueWeakRef) || !queueWeakRef.TryGetTarget(out queue))
+                    var queueWeakRef = _queues.GetOrAdd(queueKey, key => new WeakReference<DelegationQueue>(new DelegationQueue(key.QueueName, key.UrlPrefix)));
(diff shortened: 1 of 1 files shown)


MihaZupan (approved):
Thanks

NGloreous:
@MihaZupan can I get this back ported to a current release so I we can pick up the next right away? I can create the PR but what should I target?

MihaZupan:
Are you able to use [daily builds](https://github.com/microsoft/reverse-proxy/blob/main/docs/DailyBuilds.md) until we release the next version?

NGloreous:
No, we can only consume from nuget.org since this will go to our production services.

PR #2490, opened 2024-05-06
Stop using -pt images

MihaZupan:
Internal pipelines are broken since Thursday because of using these images.


MihaZupan:
Tested on an internal build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2445812&view=results

Merging to unblock the daily build feed

PR #2491, opened 2024-05-06
[release/2.1] Stop using -pt images

MihaZupan:
Backport of #2490


MihaZupan:
Looks like we may need to pick up more recent arcade as well for this branch.
We should consider re-enabling automated darc PRs for release/2.1

MihaZupan:
I added a darc subscription to keep this branch updated with arcade changes.
After #2497 this now builds again

benjaminpetit (approved):

PR #2503, opened 2024-05-15
Move request cancellations to a separate log

### LINKED ISSUE #2195: Control the handling of TaskCanceledException and other exceptions
#### Our setup
We use YARP as a gateway from our frontend client to the backend services, with communication via gRPC-web.

#### The suggestion
We encounter an issue when the client leaves a page while streaming data from a backend service through the YARP proxy. YARP throws a `TaskCanceledException` in this scenario. No error is thrown or logged by the backend service. We expect this exception to be thrown, as it usually is when a task is canceled. However, it is currently not possible to catch and handle the exception with YARP. This is problematic for us because YARP logs the exception, which then gets caught by our analytics.

We have attempted to add middleware to address this issue, but we have been unable to stop the logging of the exception.

Changing the log level to not log errors is not a viable solution as we may miss other important errors.

We suggest one of the following solutions:
1. Add a configuration flag to toggle the logging of `TaskCanceledException` since it is an expected exception.
2. Make it possible for the consumer of YARP to decide how to handle exceptions. This may be achievable if we can catch it in middleware.

This is important to us as we don't want to log expected exceptions, which then gets reported by our analytics.

MihaZupan:
Fixes #2195

This moves `RequestCanceled`, `RequestBodyCanceled` and `UpgradeRequestCanceled` to a debug-level log.
The only thing I'm not sure about is whether `ForwardingRequestCancelled` is the right name for it in case we decide to switch more errors to it in the future.

Most of the change is switching error tests to use a shared helper that now also asserts that the error was logged.

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -982,6 +982,11 @@ internal sealed class HttpForwarder : IHttpForwarder
             EventIds.ForwardingError,
             "{error}: {message}");
 
+        private static readonly Action<ILogger, ForwarderError, string, Exception> _proxyRequestCancelled = LoggerMessage.Define<ForwarderError, string>(
+            LogLevel.Debug,
+            EventIds.ForwardingRequestCancelled,
+            "{error}: {message}");
+
         private static readonly Action<ILogger, int, Exception?> _notProxying = LoggerMessage.Define<int>(
             LogLevel.Information,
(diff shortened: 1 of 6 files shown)


benjaminpetit (approved):

hughbidone:
Hi, thanks for this but has this been released yet i.e. does the latest version `2.2.0` have this enhancement?

I am using `2.1.0` and it is just spamming our exceptions log with these cancelled requests.

PR #2509, opened 2024-05-17
Fix Forwarded header for IPv4 with dual stack sockets

rkargMsft:
Similar to #2400  but for the `Forwarded` header


rkargMsft:

MihaZupan (commented):
Thank you.
Could you move this logic into `AppendNode` such that it also applies to `by=`'s `LocalIpAddress`?

rkargMsft:
Pulled out handling for both IP addresses

src/ReverseProxy/Transforms/RequestHeaderForwardedTransform.cs
```diff
@@ -141,6 +141,12 @@ private void AppendBy(HttpContext context, ref ValueStringBuilder builder)
     // https://tools.ietf.org/html/rfc7239#section-6
     private void AppendNode(IPAddress? ipAddress, int port, NodeFormat format, ref ValueStringBuilder builder)
     {
+        // Prefer IPv4 formatting
+        if (ipAddress is {IsIPv4MappedToIPv6:true})
```
MihaZupan, line 145:
```suggestion
        if (ipAddress is { IsIPv4MappedToIPv6: true })
```

MihaZupan (approved):
Thanks

PR #2530, opened 2024-06-24
Forwarded Header Transforms Doc Comments

mroberts91:
Add contextual doc comments indicating that using `AddForwarded` or `AddXForwarded` header transforms will remove the other from the `RequestTransforms` when either is enabled. Without the comments, this may seem to be an unexpected behavior without reading the source code of the extension methods to understand the effect each method has on the other at runtime.

### DIFF
--- a/src/ReverseProxy/Transforms/ForwardedTransformExtensions.cs
+++ b/src/ReverseProxy/Transforms/ForwardedTransformExtensions.cs
@@ -113,6 +113,9 @@ public static class ForwardedTransformExtensions
     /// <summary>
     /// Adds the transform which will add X-Forwarded-* request headers.
     /// </summary>
+    /// <remarks>
+    /// Removes the Forwarded header when an X-Forwarded transform is enabled
+    /// </remarks>
     public static TransformBuilderContext AddXForwarded(this TransformBuilderContext context, ForwardedTransformActions action = ForwardedTransformActions.Set)
     {
         context.AddXForwardedFor(action: action);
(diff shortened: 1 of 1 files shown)


src/ReverseProxy/Transforms/ForwardedTransformExtensions.cs
```diff
@@ -113,6 +113,9 @@ public static TransformBuilderContext AddXForwardedPrefix(this TransformBuilderC
     /// <summary>
     /// Adds the transform which will add X-Forwarded-* request headers.
     /// </summary>
+    /// <remarks>
+    /// Removes the Forwarded header when an X-Forwarded transform is enabled
```
MihaZupan, line 117:
```suggestion
    /// Also removes the <c>Forwarded</c> header when enabled.
```

src/ReverseProxy/Transforms/ForwardedTransformExtensions.cs
```diff
@@ -173,6 +176,9 @@ public static TransformBuilderContext AddXForwarded(this TransformBuilderContext
     /// <summary>
     /// Adds the transform which will add the Forwarded header as defined by [RFC 7239](https://tools.ietf.org/html/rfc7239).
     /// </summary>
+    /// <remarks>
+    /// Removes the X-Forwarded headers when an Forwarded transform is enabled
```
MihaZupan, line 180:
```suggestion
    /// Also removes the <c>X-Forwarded</c> headers when enabled.
```

MihaZupan (approved):
Thank you

mroberts91:
Thank you

PR #2548, opened 2024-07-10
Allow ProxyConfigManager to be resolved earlier

FlorianGrimm:
This is for issue 1618 tunnel.
Allowing to be configured, I need to resolve the ProxyConfigManager within a ConfigureKestrel callback.
If you prefer a bigger PR please let me know.


FlorianGrimm:

Issue #85, opened 2020-04-23
Support FastCGI proxying


benaadams:
Support the FastCGI protocol

This would enable proxying for a wider ecosystem e.g. PHP, Python etc

anurse:
I think this is pretty beyond our 1.0 radar to be honest. It may not even be a major goal right now. We're not really looking at process launching at all (let alone FastCGI) as a main goal of this. It's definitely expected that this is something you could extend YARP to do though, and might fit well into an ecosystem beyond the core of YARP.

Certainly worth keeping this around as a backlog item to collect more feedback on though.

samsp-msft:
Scenario question: How would you weigh this against being able to have fast loopback or domain socket support, so the proxy could be hosted on box, and route requests to the other services?

xrkolovos:
Is there a new road map for yarp for the next year? Would this fix inside?

Tratcher:
This is currently in our backlog milestone, out of scope for our initial release.

The items we're actively working on are in the 1.0.0 [milestones](https://github.com/microsoft/reverse-proxy/milestones).

samsp-msft:
We should talk about this

karelz:
Triage: Outbound protocol, interesting for interop and likely perf.

posledam:
Extremely interested in this functionality.

mike7ang1rdz:
is it time to add fastCGI  support?  I'd like to run php along with my net app, i don't want to have another proxy/server like iis or apache

xrkolovos:
Hello
any plans for this feature?

kasenhoo:
any update.....we have to replace nginx ....

PR #2572, opened 2024-08-15
add FastCgiHttpMessageHandler with basic tests

funcmike:
This allows to send normal HTTP requests using FastCGI protocol.
Implementation is based on Caddy (https://github.com/caddyserver/caddy/tree/master/modules/caddyhttp/reverseproxy/fastcgi). It's a working draft - basic test performed with PHP-FPM & Go.
I need feedback / review - I'm  open to discuss how to progress with it and make it mergeable.

Solves: #85

### DIFF
--- a/src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
+++ b/src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
@@ -145,9 +145,7 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
         }
         catch (BadFastCgiResponseException ex)
         {
-            response.StatusCode = HttpStatusCode.InternalServerError;
-            response.Content = new StringContent(ex.Message);
-            return response;
+            throw new HttpRequestException("Could not process FastCGI Response", ex, HttpStatusCode.InternalServerError);
         }
     }
 
--- a/src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
+++ b/src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
@@ -133,7 +133,7 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
             connection, cancellationToken);
 
         var response = new HttpResponseMessage() { RequestMessage = request, StatusCode = HttpStatusCode.OK };
-        var httpResponseReader = new HttpResponseReaderFastcgiRecordHandler(response, logger);
+        var httpResponseReader = new HttpResponseReaderFastCgiRecordHandler(response, logger);
 
         try
         {
@@ -219,20 +219,20 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
             documentRoot = string.Empty;
         }
 
+        if (!request.Options.TryGetValue(FastCgiHttpOptions.SCRIPT_FILENAME, out var scriptFilename))
+        {
+            scriptFilename = string.Empty;
+        }
+
         var fpath = request.RequestUri?.LocalPath ?? string.Empty;
 
-        var pathInfo = fpath;
-        var documentURI = fpath;
         var scriptName = fpath;
-
         // Ensure the SCRIPT_NAME has a leading slash for compliance with RFC3875
         // Info: https://tools.ietf.org/html/rfc3875#section-4.1.13
         if (scriptName.Length > 0 && !scriptName.StartsWith('/'))
         {
             scriptName = $"/{scriptName}";
         }
-
-        var scriptFilename = SanitizedPathJoin(documentRoot, pathInfo);
 
         var serverName = request.Headers.Host ?? string.Empty;
         if (serverName.Length > 0)
@@ -252,20 +252,15 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
         yield return new(FastCgiCoreParams.CONTENT_TYPE, request.Content?.Headers.ContentType?.ToString() ?? (request.Method == System.Net.Http.HttpMethod.Post ? "application/x-www-form-urlencoded" : string.Empty));
 
         yield return new(FastCgiCoreParams.DOCUMENT_ROOT, documentRoot);
-        yield return new(FastCgiCoreParams.DOCUMENT_URI, documentURI);
+        yield return new(FastCgiCoreParams.DOCUMENT_URI, fpath);
 
         yield return new(FastCgiCoreParams.GATEWAY_INTERFACE, FastCgiCoreParamValues.GATEWAY_INTERFACE_CGI11);
 
         yield return new(FastCgiCoreParams.HTTPS, isHttps ? FastCgiCoreParamValues.HTTPS_ON : FastCgiCoreParamValues.HTTPS_OFF);
-        //new(FastCgiCoreParams.HTTP_HOST, request.Headers.Host ?? string.Empty), // will be added later
-
-        yield return new(FastCgiCoreParams.PATH_INFO, pathInfo);
-        yield return new(FastCgiCoreParams.PATH_TRANSLATED, pathInfo.Length > 0 ? SanitizedPathJoin(documentRoot, pathInfo) : string.Empty);
 
         yield return new(FastCgiCoreParams.QUERY_STRING, request.RequestUri?.Query ?? string.Empty);
 
         yield return new(FastCgiCoreParams.REMOTE_ADDR, remoteEndpoint);
-        yield return new(FastCgiCoreParams.REMOTE_HOST, remoteEndpoint); // For performance, remote host lookup is disabled;
         yield return new(FastCgiCoreParams.REMOTE_PORT, remotePort);
 
         yield return new(FastCgiCoreParams.REQUEST_METHOD, request.Method.Method);
@@ -280,11 +275,6 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
         yield return new(FastCgiCoreParams.SERVER_PROTOCOL, HttpProtocol.GetHttpProtocol(request.Version));
         yield return new(FastCgiCoreParams.SERVER_SOFTWARE, FastCgiCoreParamValues.SERVER_SOFTWARE_YARP_2);
 
-        if (pathInfo.Length > 0)
-        {
-            yield return new(FastCgiCoreParams.PATH_TRANSLATED, SanitizedPathJoin(documentRoot, pathInfo));
-        }
-
         foreach (var header in request.Headers)
         {
             yield return new(header.Key.ReplaceToUpperAscii('-', '_'), string.Join(", ", header.Value), FastCgiCoreParams.PREFIX);
@@ -296,156 +286,6 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
             {
                 yield return new(header.Key.ReplaceToUpperAscii('-', '_'), string.Join(", ", header.Value, FastCgiCoreParams.PREFIX));
             }
-        }
-    }
-
-    private static string SanitizedPathJoin(string root, string reqPath)
-    {
-        if (root.Length == 0)
-        {
-            root = ".";
-        }
-
-        // TODO: its good enough for linux / unix but for windows it also needs finding of special devices names
-        // https://github.com/golang/go/issues/56336#issuecomment-1416214885
-        var relPath = PathClean("/" + reqPath);
-
-        return Path.Join(root, relPath);
-    }
-
-    // Copied from: https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/path/path.go;l=72
-    private static string PathClean(string path)
-    {
-        if (path.Length == 0)
-        {
-            return ".";
-        }
-
-        var pathChars = path.AsSpan();
-
-        var lb = new LazyBuf(path);
-
-        var (r, dotdot, n, rooted) = (0, 0, path.Length, pathChars[0] == '/');
-
-        if (rooted)
-        {
-            lb.Append('/');
-            (r, dotdot) = (1, 1);
-        }
-
-        while (r < n)
-        {
-            switch (pathChars[r])
-            {
-                case '/':
-                    // empty path element
-                    r++;
-                    break;
-                case '.' when (r + 1 == n || pathChars[r + 1] == '/'):
-                    // . element
-                    r++;
-                    break;
-                case '.' when (pathChars[r + 1] == '.' && (r + 2 == n || pathChars[r + 2] == '/')):
-                    // .. element: remove to last /
-                    r += 2;
-
-                    if (lb.W > dotdot)
-                    {
-                        // can backtrack
-                        lb.W--;
-                        while (lb.W > dotdot && lb.Index(lb.W) != '/')
-                        {
-                            lb.W--;
-                        }
-                    }
-                    else if (!rooted)
-                    {
-                        // cannot backtrack, but not rooted, so append .. element.
-                        if (lb.W > 0)
-                        {
-                            lb.Append('/');
-                        }
-                        lb.Append('.');
-                        lb.Append('.');
-                        dotdot = lb.W;
-                    }
-                    break;
-                default:
-                    // real path element.
-                    // add slash if needed
-                    if (rooted && lb.W != 1 || !rooted && lb.W != 0)
-                    {
-                        lb.Append('/');
-                    }
-                    // copy element
-                    while (r < n && pathChars[r] != '/')
-                    {
-                        lb.Append(pathChars[r]);
-                        r++;
-                    }
-                    break;
-            }
-        }
-
-        if (lb.W == 0)
-        {
-            return ".";
-        }
-
-        return lb.ToString();
-    }
-
-    private ref struct LazyBuf(string s)
-    {
-        private char[]? _rented = default;
-        public int W { get; set; } = 0;
-        private readonly string _s = s;
-
-        public new readonly string ToString()
-        {
-            if (_rented == null)
-            {
-                return _s;
-            }
-            var v = new string(_rented[..W]);
-            ArrayPool<char>.Shared.Return(_rented);
-            return v;
-        }
-
-        public void Append(char c)
-        {
-            if (_rented == null)
-            {
-
-                if (W < _s.Length && _s[W] == c)
-                {
-                    W++;
-                    return;
-                }
-
-                _rented = ArrayPool<char>.Shared.Rent(_s.Length);
-                try
-                {
-                    _s.CopyTo(0, _rented, 0, W);
-                }
-                catch
-                {
-                    ArrayPool<char>.Shared.Return(_rented);
-                    throw;
-                }
-            }
-
-            _rented[W] = c;
-            W++;
-        }
-
-        public readonly char Index(int i)
-        {
-            if (_rented != null)
-            {
-                return _rented[W];
-            }
-            return _s[i];
         }
     }
 
@@ -471,7 +311,7 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
                     {
                         try
                         {
-                            if (!handler.TryOnFastcgiRecord(ref record))
+                            if (!handler.TryOnFastCgiRecord(ref record))
                             {
                                 return;
                             }
@@ -586,7 +426,7 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
 
     private interface FastCgiRecordHandler
     {
-        bool TryOnFastcgiRecord(ref FastCgiRecord fastcgiRecord);
+        bool TryOnFastCgiRecord(ref FastCgiRecord fastCgiRecord);
     }
 
     private interface IFastCgiContentDataWriter
@@ -594,15 +434,15 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
         void WriteTo(IBufferWriter<byte> buffer);
     }
 
-    private sealed class HttpResponseReaderFastcgiRecordHandler(HttpResponseMessage result, ILogger logger) : FastCgiRecordHandler
+    private sealed class HttpResponseReaderFastCgiRecordHandler(HttpResponseMessage result, ILogger logger) : FastCgiRecordHandler
     {
         private readonly HttpResponseReader _reader = new(result);
         private bool _headersDone;
         private RentedMemorySegment<byte>? _start;
         private RentedMemorySegment<byte>? _end;
-        public bool TryOnFastcgiRecord(ref FastCgiRecord fastcgiRecord)
+        public bool TryOnFastCgiRecord(ref FastCgiRecord fastCgiRecord)
         {
-            switch (fastcgiRecord.Header.Type)
+            switch (fastCgiRecord.Header.Type)
             {
                 case FastCgiRecordHeader.RecordType.Stdout when _headersDone:
                     {
@@ -610,51 +450,51 @@ public sealed class FastCgiHttpMessageHandler(IOptions<SocketConnectionFactoryOp
                         // After headers are done it could act as stream and lazy parse rest of the data on the fly.
                         // This would lower memory footprint to only 1 record (max 65535).
                         // Stream would stop at EndRequest Record.
-                        if (fastcgiRecord.ContentData.Memory.Length == 0)
+                        if (fastCgiRecord.ContentData.Memory.Length == 0)
                         {
-                            fastcgiRecord.Dispose();
+                            fastCgiRecord.Dispose();
                             return true;
                         }
 
                         if (_start is null)
                         {
-                            _start = new RentedMemorySegment<byte>(fastcgiRecord.ContentData);
+                            _start = new RentedMemorySegment<byte>(fastCgiRecord.ContentData);
                         }
                         else if (_end is null)
                         {
-                            _end = _start.Append(fastcgiRecord.ContentData);
+                            _end = _start.Append(fastCgiRecord.ContentData);
                         }
                         else
                         {
-                            _end.Append(fastcgiRecord.ContentData);
+                            _end.Append(fastCgiRecord.ContentData);
                         }
 
                         return true;
                     }
                 case FastCgiRecordHeader.RecordType.Stdout:
                     {
-                        var reader = new SequenceReader<byte>(new ReadOnlySequence<byte>(fastcgiRecord.ContentData.Memory));
+                        var reader = new SequenceReader<byte>(new ReadOnlySequence<byte>(fastCgiRecord.ContentData.Memory));
                         if (_reader.ParseHttpHeaders(ref reader))
                         {
                             _headersDone = true;
 
-                            var left = fastcgiRecord.ContentData.Memory.Slice((int)reader.Consumed);
+                            var left = fastCgiRecord.ContentData.Memory.Slice((int)reader.Consumed);
                             if (left.Length > 0)
                             {
-                                _start = new RentedMemorySegment<byte>(new RentedReadOnlyMemory<byte>(left, fastcgiRecord.ContentData.Rented));
+                                _start = new RentedMemorySegment<byte>(new RentedReadOnlyMemory<byte>(left, fastCgiRecord.ContentData.Rented));
                                 return true;
                             }
                         }
 
-                        fastcgiRecord.Dispose();
+                        fastCgiRecord.Dispose();
                         return true;
                     }
                 // TODO: how to treat errors - caddy & nginx are just logging them
                 // stderr can interleave with stdout records so they do not interrupt connection
                 case FastCgiRecordHeader.RecordType.Stderr:
                     {
-                        logger.LogError(message: "stdErr {contentData}", Encoding.ASCII.GetString(fastcgiRecord.ContentData.Memory.Span));
-                        fastcgiRecord.Dispose();
+                        logger.LogError(message: "stdErr {contentData}", Encoding.ASCII.GetString(fastCgiRecord.ContentData.Memory.Span));
+                        fastCgiRecord.Dispose();
                         return true;
                     }
                 case FastCgiRecordHeader.RecordType.EndRequest:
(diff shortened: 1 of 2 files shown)


Tratcher:
Interesting the whole thing is just a HttpMessageHandler. Does it have uses outside of yarp that would warrant it being its own project?

funcmike:
I tried to make it as HttpMessageHandler to see how far I can go with it. It turn out doable with little tricks like DOCUMENT_ROOT is passed as HttpOption in request - its static option so it could be also passed in constructor.
I don’t know if it’s useful for others, nowadays FastCGI is used mainly by PHP but let others give opinion on that.

funcmike:
@Tratcher Tests are still just for me to debug responses (I will add more & better at some point), but code of  FastCgiHttpMessageHandler can be reviewed. I need some feedback to known if I'm going in a right direction with it.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+            {
+                var result = _httpParser.ParseHeaders(this, ref reader);
+                return result;
+            }
+
+            public void OnHeader(ReadOnlySpan<byte> name, ReadOnlySpan<byte> value)
+            {
+
+                bool added;
+                var valueValue = Encoding.ASCII.GetString(value);
```
Tratcher, line 442: Header decoding options.
https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.responseheaderencodingselector?view=net-8.0#system-net-http-socketshttphandler-responseheaderencodingselector

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+                {
+                    if (value.Length > 2)
+                    {
+                        var codePart = value.Slice(0, 3);
+                        if (int.TryParse(Encoding.ASCII.GetString(codePart), out var code))
+                        {
+                            response.StatusCode = (HttpStatusCode)code;
+                        }
+                    }
+                    added = response.Headers.TryAddWithoutValidation(status!, valueValue);
```
Tratcher, line 454: Shouldn't an unparsable status be an error?

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+        public bool TryOnFastCgiRecord(ref FastCgiRecord fastCgiRecord)
+        {
+            switch (fastCgiRecord.Header.Type)
+            {
+                case FastCgiRecordHeader.RecordType.Stdout when _headersDone:
+                    {
+                        // TODO: Idea to optimize it further.
+                        // After headers are done it could act as stream and lazy parse rest of the data on the fly.
+                        // This would lower memory footprint to only 1 record (max 65535).
+                        // Stream would stop at EndRequest Record.
```
Tratcher, line 341: Yes please. YARP wants the HttpResponseMessage back as soon as possible, it will stream the response body later.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+                    }
+                // TODO: how to treat errors - caddy & nginx are just logging them
+                // stderr can interleave with stdout records so they do not interrupt connection
+                case FastCgiRecordHeader.RecordType.Stderr:
+                    {
+                        logger.LogError(message: "stdErr {contentData}", Encoding.ASCII.GetString(fastCgiRecord.ContentData.Memory.Span));
+                        fastCgiRecord.Dispose();
+                        return true;
+                    }
+                case FastCgiRecordHeader.RecordType.EndRequest:
```
Tratcher, line 389: Are trailers supported by the protocol?

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+            return new HttpResponseMessage(HttpStatusCode.BadRequest) { Content = new StringContent("invalid request path") };
+        }
+
+        //TODO: maybe implement static connect to endpoint specified by options when creating FastCgiHttpMessageHandler.
+        var endPoint = new DnsEndPoint(request.RequestUri.Host, request.RequestUri.Port);
+
+        //TODO: implement connection pooling
+        await using var connection = await ConnectAsync(endPoint, noDelay: true, cancellationToken);
+
+        await SendAsync(
```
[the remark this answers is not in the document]
funcmike: ok, done

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+            if (Uri.TryCreate($"tcp://{serverName}", default, out var u))
+            {
+                serverName = u.Host;
+            }
+        }
+
+        //TODO: probably better https detection will be needed at some point
+        var isHttps = request.RequestUri?.Scheme == "https";
+
+        yield return new(FastCgiCoreParams.CONTENT_LENGTH, request.Content?.Headers.ContentLength?.ToString() ?? string.Empty);
```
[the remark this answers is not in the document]
funcmike: yes, I could skip it -Check added.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+        yield return new(FastCgiCoreParams.CONTENT_TYPE, request.Content?.Headers.ContentType?.ToString() ?? (request.Method == System.Net.Http.HttpMethod.Post ? "application/x-www-form-urlencoded" : string.Empty));
+
+        yield return new(FastCgiCoreParams.DOCUMENT_ROOT, documentRoot);
+        yield return new(FastCgiCoreParams.DOCUMENT_URI, fpath);
+
+        yield return new(FastCgiCoreParams.GATEWAY_INTERFACE, FastCgiCoreParamValues.GATEWAY_INTERFACE_CGI11);
+
+        yield return new(FastCgiCoreParams.HTTPS, isHttps ? FastCgiCoreParamValues.HTTPS_ON : FastCgiCoreParamValues.HTTPS_OFF);
+
+        yield return new(FastCgiCoreParams.QUERY_STRING, request.RequestUri?.Query ?? string.Empty);
```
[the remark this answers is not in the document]
funcmike: I thinks std http param encoding.
Caddy is using URL.RawQuery:
`RawQuery    [string](https://pkg.go.dev/builtin#string)    // encoded query values, without '`

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+
+        yield return new(FastCgiCoreParams.REMOTE_ADDR, remoteEndpoint);
+        yield return new(FastCgiCoreParams.REMOTE_PORT, remotePort);
+
+        yield return new(FastCgiCoreParams.REQUEST_METHOD, request.Method.Method);
+        yield return new(FastCgiCoreParams.REQUEST_SCHEME, request.RequestUri?.Scheme ?? FastCgiCoreParamValues.REQUEST_SCHEME_HTTP);
+        yield return new(FastCgiCoreParams.REQUEST_URI, request.RequestUri?.ToString() ?? string.Empty);
```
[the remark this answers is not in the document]
funcmike: yeah it should. I tested with nginx (+phpinfo())  and it does include query string in REQUEST_URI - corrected.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+        foreach (var header in request.Headers)
+        {
+            yield return new(header.Key.ReplaceToUpperAscii('-', '_'), string.Join(", ", header.Value), FastCgiCoreParams.PREFIX);
+        }
+
+        if (request.Content != null)
+        {
+            foreach (var header in request.Content.Headers)
+            {
+                yield return new(header.Key.ReplaceToUpperAscii('-', '_'), string.Join(", ", header.Value, FastCgiCoreParams.PREFIX));
```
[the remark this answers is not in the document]
funcmike: yeah, but that's not causing any problems, testing for it will be perhaps more costly then just duplicating it? If You want I can add compare.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+        public static async Task ProcessAsync(FastCgiRecordHandler handler, ConnectionContext connectionContext, CancellationToken cancellationToken)
+        {
+            var input = connectionContext.Transport.Input;
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                var result = await input.ReadAsync(cancellationToken);
+                var buffer = result.Buffer;
+
+                if (result.IsCompleted && buffer.Length == 0)
```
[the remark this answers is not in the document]
funcmike: Yeah, it would, good catch, thanks. I added additional check please see if this is correct? I'm just checking that when pipe is completed on every turn next buffer supposed to have different (smaller length) until is 0.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
funcmike: yes, Caddy is also returning error for that - I added exception.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
funcmike: Whole HTTP response in FastCgi Response is just "ContendData" in STDOUT record.
There is no special support for Header, Body, or Trailers.  FastCgiParams are only  for Request. Response is just bytes that need to be parsed. So Trailers are "supported". In need to add tests for it but I think in current implementation I'm sending everything after headers so trailers should be sent also but maybe I need to do something special in HttpResponseMessage?

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1313 @@
...
+
+            long? prevBuffLenOnCompleted = null;
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                var result = await input.ReadAsync(cancellationToken);
+                var buffer = result.Buffer;
+
+                if (result.IsCompleted)
+                {
+                    if (buffer.Length == 0)
```
Tratcher, line 229: If EndRequest wasn't received then this should be an error?

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+
+        yield return new(FastCgiCoreParams.SERVER_NAME, serverName);
+        yield return new(FastCgiCoreParams.SERVER_PORT, isHttps ? FastCgiCoreParamValues.SERVER_PORT_443 : FastCgiCoreParamValues.SERVER_PORT_80);
+        yield return new(FastCgiCoreParams.SERVER_PROTOCOL, HttpProtocol.GetHttpProtocol(request.Version));
+        yield return new(FastCgiCoreParams.SERVER_SOFTWARE, FastCgiCoreParamValues.SERVER_SOFTWARE_YARP_2);
+
+        foreach (var header in request.Headers)
+        {
+            yield return new(header.Key.ReplaceToUpperAscii('-', '_'), string.Join(", ", header.Value), FastCgiCoreParams.PREFIX);
+        }
```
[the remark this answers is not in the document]
funcmike: I added the same join as caddy:
https://github.com/caddyserver/caddy/blob/2028da4e74cd41f0f7f94222c6599da1a371d4b8/modules/caddyhttp/reverseproxy/fastcgi/fastcgi.go#L366

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+        try
+        {
+            await FastCgiRecordReader.ProcessAsync(httpResponseReader, connection, cancellationToken);
+            return response;
+        }
+        catch (BadFastCgiResponseException ex)
+        {
+            response.StatusCode = HttpStatusCode.InternalServerError;
+            response.Content = new StringContent(ex.Message);
+            return response;
```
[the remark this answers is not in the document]
funcmike: ok, done

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+            }
+        }
+
+        await SendAsync(
+            new FastCgiRecord(
+                Header: new FastCgiRecordHeader(Type: FastCgiRecordHeader.RecordType.Stdin),
+                ContentData: RentedReadOnlyMemory<byte>.Empty),
+            connection, cancellationToken);
+
+        var response = new HttpResponseMessage() { RequestMessage = request, StatusCode = HttpStatusCode.OK };
```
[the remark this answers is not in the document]
funcmike: ok, done

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+
+
+        await SendAsync(
+            new FastCgiRecord(
+                Header: new FastCgiRecordHeader(Type: FastCgiRecordHeader.RecordType.Params),
+                ContentData: RentedReadOnlyMemory<byte>.Empty),
+            connection, cancellationToken);
+
+
+        if (request.Content != null)
```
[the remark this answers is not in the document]
funcmike: After Param fastcgi record  http Body should be send as STDOUT, there is no "100-continue" support afaik in FastCGI Protocol.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+    {
+        bool TryOnFastCgiRecord(ref FastCgiRecord fastCgiRecord);
+    }
+
+    private interface IFastCgiContentDataWriter
+    {
+        void WriteTo(IBufferWriter<byte> buffer);
+    }
+
+    private static class FastCgiRecordReader
```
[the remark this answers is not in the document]
funcmike: Only FastCgiRecordReader or everything related to it also?

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
funcmike: yes, I think now it's correct, I added state to fastcgi record handler and callback when end is reached, please check it

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
funcmike: Ok but how to use it? If I just create empty SocketsHttpHandler the property of ResponseHeaderEncodingSelector is null.

Tratcher: I just mean that this is functionality you might need to replicate eventually.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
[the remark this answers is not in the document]
Tratcher: That doesn't mean it's right 😆, it's a common mistake in HTTP implementations.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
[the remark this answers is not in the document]
Tratcher: At least FastCgiRecordReader and it's accessories, it's too large and complex to inline. Break out any other large types too.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1342 @@
...
+            new FastCgiRecordBeginRequestBody(Role: FastCgiRecordBeginRequestBody.RoleType.Responder, KeepConn: false),
+            connection);
+
+        using var paramBuffer = new RentedArrayBufferWriter<byte>(FastCgiRecordHeader.MAX_CONTENT_SIZE);
+        {
+            foreach (var fcgiParam in BuildFastCgiParams(connection.RemoteEndPoint, request))
+            {
+                if (fcgiParam.ByteCount > paramBuffer.Capacity)
+                {
+                    throw new ArgumentOutOfRangeException(nameof(fcgiParam));
```
Tratcher, line 64: This seems like an assert? It shouldn't be affected by user input, correct?

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1342 @@
...
+        //TODO: probably better https detection will be needed at some point
+        var isHttps = request.RequestUri?.Scheme == "https";
+
+        if (request.Content?.Headers.ContentLength is { } cl)
+        {
+            yield return new(FastCgiCoreParams.CONTENT_LENGTH, cl.ToString());
+        }
+
+        // based on caddy implementation https://github.com/caddyserver/caddy/blob/9ddb78fadcdbec89a609127918604174121dcf42/modules/caddyhttp/reverseproxy/fastcgi/client.go#L289
+        yield return new(FastCgiCoreParams.CONTENT_TYPE, request.Content?.Headers.ContentType?.ToString() ?? (request.Method == System.Net.Http.HttpMethod.Post ? "application/x-www-form-urlencoded" : string.Empty));
```
Tratcher, line 586: That's suspect. I don't think we have any reason to add this. The content-type _should_ already be there, and if it's not, it's not our place to guess, let the server do that.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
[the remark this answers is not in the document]
Tratcher: Avoid duplicating headers, it can confuse the server. I also see Content-Type being duplicated.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
funcmike: yes, You are right, changed to debug assert

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
funcmike: ok removed

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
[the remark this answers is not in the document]
funcmike: ok done, I now send them separately.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1290 @@
...
+        await SendAsync(
+            new FastCgiRecord(
+                Header: new FastCgiRecordHeader(Type: FastCgiRecordHeader.RecordType.Params),
+                ContentData: RentedReadOnlyMemory<byte>.Empty),
+            connection, cancellationToken);
+
+
+        if (request.Content != null)
+        {
+            var contentStream = await request.Content.ReadAsStreamAsync(cancellationToken);
```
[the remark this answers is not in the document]
funcmike: ok, implemented, please check.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1377 @@
...
+        }
+
+        public override void Write(ReadOnlySpan<byte> buffer) => throw new NotSupportedException();
+
+        public override ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken = default) => throw new NotSupportedException();
+    }
+
+    private sealed class FastCgiStdinStream(ConnectionContext connection): Stream
+    {
+        public override void Write(byte[] buffer, int offset, int count)
```
Tratcher, line 1143: The sync Write overload isn't used, only WriteAsync(Memory, CancellationToken) and FushAsync(CancellationToken)

https://github.com/microsoft/reverse-proxy/blob/f2a8594489a0ecfa1c18b2294b2f84c75654e161/src/ReverseProxy/Forwarder/StreamCopier.cs#L102

funcmike: ok, I've implemented also the async write & flush, for now I left sync write to simplify testing, at some point I'll add test with YARP's forwarder.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
funcmike: Ok, first attempt ready for critique ;) it's not pretty but it works

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+                {
+                    if (buffer.Length == 0)
+                    {
+                        try
+                        {
+                            handler.OnEndOfData();
+                        }
+                        finally
+                        {
+                            input.AdvanceTo(buffer.Start, buffer.End);
```
Tratcher, line 208: The AdvanceTo can happen first, without a try/finally, because you're not using the content.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+                        return true;
+                    }
+
+                    if (prevBuffLenOnCompleted is { } prevLen && prevLen == buffer.Length)
+                    {
+                        input.AdvanceTo(buffer.Start, buffer.End);
+                        throw new BadFastCgiResponseException(FastCgiCoreExpStrings.BadResponse_IncompleteRecord, FastCgiResponseRejectionReason.IncompleteRecord);
+                    }
+
+                    prevBuffLenOnCompleted = buffer.Length;
```
Tratcher, line 219: I think it's better to check for this around line 228 and 230. If it can't read the record, and there's still data left, but the pipe is completed, then error.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+                return true;
+            }
+
+            fastCgiRecord = new FastCgiRecord(
+                Header: header,
+                ContentData: buffer.Slice(FastCgiRecordHeader.FCGI_HEADER_LEN, header.ContentLength));
+
+            // Advance the buffer
+            buffer = buffer.Slice(recordByteCount);
+            return true;
```
Tratcher, line 284: Remove duplication
```suggestion
            fastCgiRecord = new FastCgiRecord(
                Header: header,
                ContentData: header.ContentLength == 0 ?
                    ReadOnlySequence<byte>.Empty :
                    buffer.Slice(FastCgiRecordHeader.FCGI_HEADER_LEN, header.ContentLength));
            
            // Advance the buffer
            buffer = buffer.Slice(recordByteCount);
            return true;
```

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+        public HttpResponseReaderFastCgiRecordHandler(HttpRequestMessage request, ILogger logger, StateType interruptOnState = StateType.Body)
+        {
+            Result = new HttpResponseMessage { RequestMessage = request, StatusCode = HttpStatusCode.OK };
+            _reader = new HttpResponseReader(Result);
+            _pipe = new Pipe(new PipeOptions(pauseWriterThreshold: 0));
+            _logger = logger;
+            _interruptOnState = interruptOnState;
+        }
+
+        public ValueTask<FlushResult> FlushAsync(CancellationToken cancellationToken)
```
Tratcher, line 335: Why does a reader have a flush method?

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+        private bool _disposed;
+        private bool _readerFinished;
+
+        public override async ValueTask<int> ReadAsync(
+            Memory<byte> buffer,
+            CancellationToken cancellationToken = default)
+        {
+            if (!_readerFinished)
+            {
+                if (await FastCgiRecordReader.ProcessAsync(connection, handler, cancellationToken))
```
Tratcher, line 846: ProcessAsync has a while loop in it that could cause it to read to the end of the response, no? 

This Stream.ReadAsync API should return as soon as the given buffer is full. I'd expect a sequence more like
- do I have part of a prior message left? Copy that to the buffer
- while I have more space in the buffer, try reading the next message and copying that to the buffer
- store any partial message for future reads

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+            await stream.DisposeAsync();
+            await base.DisposeAsync();
+        }
+
+        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) => await ReadAsync(buffer.AsMemory(offset, count), cancellationToken);
+        public override void Flush() => stream.Flush();
+        public override int Read(byte[] buffer, int offset, int count) => throw new NotSupportedException();
+        public override long Seek(long offset, SeekOrigin origin) => stream.Seek(offset, origin);
+        public override void SetLength(long value) => stream.SetLength(value);
+        public override void Write(byte[] buffer, int offset, int count) => stream.Write(buffer, offset, count);
```
Tratcher, line 869: NotSupportedException

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+        public override async ValueTask DisposeAsync()
+        {
+            _disposed = true;
+            await connection.DisposeAsync();
+            await stream.DisposeAsync();
+            await base.DisposeAsync();
+        }
+
+        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) => await ReadAsync(buffer.AsMemory(offset, count), cancellationToken);
+        public override void Flush() => stream.Flush();
```
Tratcher, line 865: NotSupportedException

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+        public override int Read(byte[] buffer, int offset, int count) => throw new NotSupportedException();
+        public override long Seek(long offset, SeekOrigin origin) => stream.Seek(offset, origin);
+        public override void SetLength(long value) => stream.SetLength(value);
+        public override void Write(byte[] buffer, int offset, int count) => stream.Write(buffer, offset, count);
+
+        public override bool CanRead => stream.CanRead && !_disposed;
+        public override bool CanSeek => stream.CanSeek;
+        public override bool CanWrite => stream.CanWrite;
```
Tratcher, line 873: false

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+        public override long Length => stream.Length;
```
Tratcher, line 874: NotSupportedException

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+                offset += written;
+            }
+        }
+
+        public override bool CanRead { get; } = false;
+        public override bool CanSeek { get; } = false;
+        public override bool CanWrite { get; } = true;
+        public override long Length { get; } = 0;
+
+        public override long Position { get; set; }
```
Tratcher, line 922: NotSupportedException

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
```diff
@@ -0,0 +1,1131 @@
...
+            await FastCgiHttpMessageHandler.FlushAsync(connection, cancellationToken);
+        }
+
+        public override void Flush() => throw new NotSupportedException();
+        public override int Read(byte[] buffer, int offset, int count) => throw new NotSupportedException();
+        public override long Seek(long offset, SeekOrigin origin) => throw new NotSupportedException();
+        public override void SetLength(long value) => throw new NotSupportedException();
+    }
+
+    private static class HttpResponseHeaders
```
Tratcher, line 935: HttpResponseHeaders already exists
```suggestion
    private static class FastCgiResponseHeaders
```

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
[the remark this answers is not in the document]
Tratcher: Please split this file up across multiple files by type, reviewing one giant file makes it hard to find things.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
funcmike: Because current stream design  is like that

IO Pipe FastCgi Server Connection Reader  ->  Process FastCGI Record (only 1 STDOUT Record and interrupt) ->  Unpack content -> IO Pipe Internal Writer Stream -> Flush  (make it available for reader) -> IO Pipe Reader internal Stream    ... so reading stream is causing blocking and waiting for the  1 next frame / record to be buffered in IO Pipe Writer.

Internal Stream Pipe is in FastCgiMessageHandler so that's why it has Flush method. 

So its kinda a Extract Transform Load when Reader becomes also feeder/writer to second pipe that is consumed by final client.

src/ReverseProxy/Forwarder/FastCgiHttpMessageHandler.cs
funcmike: No, while loop is always blocking stream( after http Headers) only for just 1 Record/Frame. So it's basically works like You described, partial responses are buffered in IO PipeWriter and waits for next stream read.

povilas-st:
🙌

MihaZupan:
Thank you for looking into this.

Given the relatively low demand for fastcgi in YARP and the high complexity of the implementation (new protocol, connection pool, security considerations), and the lack of familiarity with it on the team, we don't believe this repository is the right place for this logic at this time.
Implementing the logic as a message handler is an interesting idea, with possible use outside of YARP. Since there doesn't seem to be anything YARP specific to the implementation (is there?), would you be interested in publishing the work in a separate repository / NuGet package?

funcmike:
Thanks for reply. I was planning to continue work on this next month (finally I have time). I understand Your decision. I don’t plan to release it as a separate package - too much work for one person to maintain it - ensuring future compatibility with YARP etc. I’m not sponsored - I was doing it voluntary in my spare time.

PR #2637, opened 2024-10-22
Use the correct Assert.Single overload

MihaZupan:
Unblocks #2634 and #2635 that include a new analyzer warning for this case


MihaZupan:
@MihuBot backport to release/2.1

MihaZupan:
@MihuBot backport to release/2.2

PR #2710, opened 2025-01-07
Skeleton app for containerization

benjaminpetit:
Basic application to be used in the future container.

It relies on yarp nuget packages for now because of the dependency to aspire service discovery.


eng/Publishing.props
```diff
@@ -3,4 +3,44 @@
    <PropertyGroup>
       <PublishingVersion>3</PublishingVersion>
    </PropertyGroup>
+
+    <PropertyGroup>
```
MihaZupan, line 7: Nit: indentation seems to be off in a couple places throughout this file

eng/yarpapppack/Common.projitems
MihaZupan, line 1: So the `Yarp.Application` is just a regular web app that happens to use service discovery, and all of the `eng/*` stuff is for container publishing?
And the yarpapppack projects don't reference the source, but just the binaries? If so, does the TFM set on these do anything?

src/Application/Yarp.Application.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <RuntimeIdentifiers>win-x64;win-arm64;linux-x64;linux-arm64;</RuntimeIdentifiers>
+    <TargetFramework>net8.0</TargetFramework>
```
MihaZupan, line 6:
```suggestion
    <TargetFramework>net9.0</TargetFramework>
```
If we're the only ones expected to build this, can we just always use the latest?

src/Application/Yarp.Application.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <Nullable>enable</Nullable>
+    <AssemblyName>yarp</AssemblyName>
+  </PropertyGroup>
+
+  <!-- Used by publishing infrastructure to get the version to use for blob publishing -->
+  <Target Name="ReturnPackageVersion" Returns="$(PackageVersion)" />
+
+  <ItemGroup>
+    <PackageReference Include="Yarp.ReverseProxy" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="8.6.0" />
```
MihaZupan, line 17: Do we need this ref?

src/Application/Yarp.Application.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <PackageReference Include="Microsoft.Extensions.ServiceDiscovery" Version="8.2.2" />
+    <PackageReference Include="Microsoft.Extensions.ServiceDiscovery.Yarp" Version="8.2.2" />
```
MihaZupan, line 19: Nit: please move the versions for these into https://github.com/microsoft/reverse-proxy/blob/main/eng/Versions.props

MihaZupan (approved):

eng/Publishing.props
benjaminpetit: Fixed

src/Application/Yarp.Application.csproj
benjaminpetit: We are the only one, but since it will be used in a Dockerfile in another repo, I prefer keeping the target framework explicitely fixed

src/Application/Yarp.Application.csproj
benjaminpetit: Not anymore

eng/yarpapppack/Common.projitems
benjaminpetit: > So the Yarp.Application is just a regular web app that happens to use service discovery, and all of the eng/* stuff is for container publishing?

Yes

> And the yarpapppack projects don't reference the source, but just the binaries? If so, does the TFM set on these do anything?

They do reference the csproj and do a "publish" instead of a smple "build" and then create a zip file. I followed what was done for the Aspire dashboard. I think it's simpler to have separate csproj files when we multiply the target runtime identifiers.

Issue #2727, opened 2025-01-27
Forward transform registration actively removes x-forwarded headers


rkargMsft:
### Describe the bug
Adding the Forwarded request transform doesn't just remove any x-forwarded transforms as the documentation says. It adds new x-forwarded transforms to actively remove any x-forwarded headers on the request.

### To Reproduce
```
            var proxyBuilder = services.AddReverseProxy()
                    .LoadFromConfig(Configuration)
                    .AddTransforms(context =>
                    {
                        context
                            .AddForwarded(forFormat: NodeFormat.IpAndPort, byFormat: NodeFormat.IpAndPort, action: ForwardedTransformActions.Append)
                            ;
                        // at this point, there is the forwarded transform as well as 4 x-forwarded-* transforms with action Remove
                        context
                            .AddXForwardedFor(action: ForwardedTransformActions.Append)
                            .AddXForwardedPrefix(action: ForwardedTransformActions.Append)
                            .AddXForwardedHost(action: ForwardedTransformActions.Append)
                            .AddXForwardedProto(action: ForwardedTransformActions.Append)
                            ;
                    })
```

### Further technical details

This behavior means that if there are proxies before this one that add x-forwarded headers, those will be cleared first, then the explicitly added `.Append` transforms will not see the prior ones and only have the current hop's information included instead of both the prior and current. All prior proxy information is lost in this scenario.

We include both the forwarded and x-forwarded because there are a variety of different backends behind this load balancer and some only support x-forwarded and others only the forwarded header.

Our workaround is to explicitly search for and remove any existing x-forwarded request transforms before we add them back in.

The [documentation](https://microsoft.github.io/reverse-proxy/articles/transforms.html#forwarded) states
| Enabling this transform will disable the default X-Forwarded transforms as they carry similar information in another format. The X-Forwarded transforms can still be explicitly enabled.

but it doesn't just disable any x-forwarded transforms as stated. It actively adds new x-forwarded transforms that remove any existing headers, making them invisible to any explicitly added x-forwarded transforms after the fact.

rkargMsft:
I agree that the default behavior of wanting to use only the `forwarded` header is likely the more common one, but it would be nice to have an optional parameter to suppress registering of the `x-forwarded` transforms with action `Remove` when calling `.AddForwarded`

MihaZupan:
I'm guessing that the workaround you could use in this case is swapping out
```c#
[...]

PR #2728, opened 2025-01-27
Allow bypassing of x-forwarded header removal

rkargMsft:
Addresses #2727


rkargMsft:
Looks like AddXForwarded similarly removes the Forwarded headers and needs similar treatment.

MihaZupan:
I agree that having the extra argument be `removeAllXForwardedHeaders` instead of `suppressRemoveAllXForwardedHeaders` is better.

The annoying part is that the current logic doesn't always remove the other header. `AddForwarded` does it _conditionally_.
https://github.com/dotnet/yarp/blob/41e563cced47925142bd5c6ddeb1e2958e0ebabd/src/ReverseProxy/Transforms/ForwardedTransformExtensions.cs#L188-L201

I'm not actually sure why someone would want to add a no-op `Forwarded` transform like that.
I'd be okay with making a small breaking change here and have the new default be that the other header gets removed unconditionally (well, conditioned on the new argument only).

rkargMsft:
I can swap the parameter naming.

I'm not sure what the best way to handle `action == ForwardedTransformActions.Off` is because, as you say, it doesn't make sense to call `AddForwarded` with that action. If it were a new method then I'd propose validation that would throw if it was Off because, just don't call `AddForwarded`. I think that since it exists with the current behavior (do nothing if `.Off`) that it be kept as is. I think it would be odd to have the desired behavior be that `.AddForwarded(..., action: ForwardedTransformActions.Off)` be to do nothing with Forwarded headers but add transforms to remove the XForwarded headers.

MihaZupan (approved):
Thanks.

Mind adding a few tests for these?

rkargMsft:
Minor pain point is that targeting the alpha SDK does add some friction to contributing

MihaZupan:
Thank you.

Did you run into SDK issues even after running `build.cmd` from the root?
(it's not that common of a step in general, but is a thing for most dotnet/* repos)

rkargMsft:
I was not aware of that. Running `build.bat` and `startvs.bat` worked great.

Someone should put that in the Readme, maybe somewhere near the top in a Build section or something. Wait a minute...

PR #2729, opened 2025-01-28
Always enforce HttpVersionPolicy on WebSocket version negotiation

MihaZupan:
When we added HTTP/2 WebSocket support, the intention behind version selection was that we would match `SocketsHttpHandler`/`ClientWebSocket` behavior.

When the user specified `Version=2.0, Policy=RequestVersionOrHigher`, we should be attempting HTTP/2 WebSockets, regardless of whether the destination is using https or not.
Currently, we'll use HTTP/1.1 instead, which goes against the specified policy.
Aside from violating the configured policy, the current behavior is problematic when the backend service only supports HTTP/2. Instead of using H2 as requested, we'll downgrade the H1, and all requests will fail.

This change makes YARP strictly follow the version policy. If the user told us not to use HTTP/1.X, we simply shouldn't.
If the backend server happens to only support HTTP/1.X, then the request will fail, and that's a proxy misconfiguration.

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -370,9 +370,10 @@ internal sealed class HttpForwarder : IHttpForwarder
         var outgoingUpgrade = false;
         var outgoingConnect = false;
         var tryDowngradingH2WsOnFailure = false;
+
         if (isSpdyRequest)
         {
-            // Can only be done on HTTP/1.1, force regardless of options.
+            // Can only be done on HTTP/1.1.
             outgoingUpgrade = true;
         }
         else if (isH1WsRequest || isH2WsRequest)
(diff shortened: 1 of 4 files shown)


src/ReverseProxy/Forwarder/HttpForwarder.cs
```diff
@@ -398,9 +401,16 @@ public async ValueTask<ForwarderError> SendAsync(
             }
         }
 
+        bool http1IsAllowed = outgoingPolicy == HttpVersionPolicy.RequestVersionOrLower || outgoingVersion.Major == 1;
+
         if (outgoingUpgrade)
         {
-            // Can only be done on HTTP/1.1, force regardless of options.
+            // Can only be done on HTTP/1.1, throw if disallowed by options.
```
BrennanConroy, line 408: Line 370 has:
```csharp
if (isSpdyRequest)
{
    // Can only be done on HTTP/1.1, force regardless of options.
    outgoingUpgrade = true;
}
```

This may now throw, not saying that's wrong, but just want to make sure it's known.
e.g. Maybe it should have a more specific error message?

BrennanConroy (approved):

benjaminpetit (approved):

adityamandaleeka:
@MihaZupan Can this be merged?

MihaZupan:
Yep, I'll add another test and merge it once CI is back up

PR #2741, opened 2025-02-06
Fix the link to the getting started page

Tragetaschen:


MihaZupan:
Thank you for the contribution.
We've set up redirects for all previous articles now, so this change is no longer required.

In the (near) future we'll be moving docs to a new location, so we'll be updating all existing links at that point.

PR #2750, opened 2025-02-13
Add OpenTelemetry support on yarp container

benjaminpetit:
Add opentelemetry support for the yarp container.

The main issue with the current solution is that when running in Aspire, the container will try to export telemetry to ` host.docker.internal`, which will trigger a SSL error, since the dev cert is only made for `localhost` .

For the moment, I propose that we ignore this error when running in development.

Other option:
- require the certificate to be present in the docker at runtime, and ignore only the name missmatch
- require the cert to be valid for `host.docker.internal`, and manually trust the cert at runtime (requiring switching to a non `distroless` base image)


src/Application/Extensions.cs
```diff
@@ -0,0 +1,148 @@
...
+                {
+                    if (cert is not null)
+                    {
+                        // We are in dev environment,
+                        // check if the cert matches the dev cert
+                        return cert.Extensions
+                            .OfType<X509Extension>()
+                            .Any(e => string.Equals(aspNetHttpsOid, e.Oid?.Value, StringComparison.Ordinal)); ;
+                    }
+                    return errors == System.Net.Security.SslPolicyErrors.None;
```
MihaZupan, line 111: I don't think this is any better than a `return true`.
If the cert is malicious, it can trivially fake the extension too.

benjaminpetit: I removed it and now require a special environment variable to be set

src/Application/Extensions.cs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private static TBuilder AddOpenTelemetryExporters<TBuilder>(this TBuilder builder) where TBuilder : IHostApplicationBuilder
+    {
+        var useOtlpExporter = !string.IsNullOrWhiteSpace(builder.Configuration["OTEL_EXPORTER_OTLP_ENDPOINT"]);
+
+        if (useOtlpExporter)
+        {
+
+            if (string.Equals(Environment.GetEnvironmentVariable("YARP_UNSAFE_SKIP_OLTP_CERT_VALIDATION"), "true", StringComparison.InvariantCultureIgnoreCase))
```
MihaZupan, line 79: Nit: If this is otherwise equivalent, can we change this to something like?
```c#

builder.Services.Configure<OtlpExporterOptions>(options =>
{
    options.HttpClientFactory = () =>
    {
        var handler = new HttpClientHandler();
    
        if (string.Equals(Environment.GetEnvironmentVariable("YARP_UNSAFE_SKIP_OLTP_CERT_VALIDATION"), "true", StringComparison.InvariantCultureIgnoreCase))
        {
            handler.ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator;
        }
    
        return new HttpClient(handler);
    }
});

builder.Services.AddOpenTelemetry()
    .WithLogging(logging => logging.AddOtlpExporter())
    .WithMetrics(metrics => metrics.AddOtlpExporter())
    .WithTracing(tracing => tracing.AddOtlpExporter());

```

eng/Versions.props
```diff
@@ -33,7 +33,15 @@
     <JsonSchemaNetVersion>7.0.2</JsonSchemaNetVersion>
     <!-- Container app dependencies -->
     <YarpNugetVersion>2.2.0</YarpNugetVersion>
-    <MicrosoftExtensionsServiceDiscovery>8.2.2</MicrosoftExtensionsServiceDiscovery>
-    <MicrosoftExtensionsServiceDiscoveryYarp>8.2.2</MicrosoftExtensionsServiceDiscoveryYarp>
+    <MicrosoftExtensionsServiceDiscovery>9.0.0</MicrosoftExtensionsServiceDiscovery>
+    <MicrosoftExtensionsServiceDiscoveryYarp>9.0.0</MicrosoftExtensionsServiceDiscoveryYarp>
+    <OpenTelemetryExporterOpenTelemetryProtocol>1.9.0 </OpenTelemetryExporterOpenTelemetryProtocol>
```
WeihanLi, line 38: Should we try to use the latest version here

src/Application/Extensions.cs
```diff
@@ -0,0 +1,117 @@
...
+        builder.ConfigureOpenTelemetry();
+
+        builder.AddDefaultHealthChecks();
+
+        builder.Services.AddServiceDiscovery();
+
+        builder.Services.ConfigureHttpClientDefaults(http =>
+        {
+            // Turn on resilience by default
+            http.AddStandardResilienceHandler();
```
MihaZupan, line 34: Does this do anything for us given that YARP doesn't use IHttpClientFactory?

src/Application/Extensions.cs
MihaZupan: My comment here was more so about potential subtle differenes in behavior between the two (is the handler configured the same?), not just `DangerousAcceptAnyServerCertificateValidator`.

Right now it's not obvious that the environment variable will only change the cert validation.

src/Application/Extensions.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                metrics.AddAspNetCoreInstrumentation()
+                    .AddHttpClientInstrumentation()
+                    .AddRuntimeInstrumentation()
+                    .SetExemplarFilter(ExemplarFilterType.TraceBased);
+            })
+            .WithTracing(tracing =>
+            {
+                tracing.AddAspNetCoreInstrumentation()
+                    .AddGrpcClientInstrumentation()
```
MihaZupan, line 62: Are we doing gRPC anywhere?

If not, my preference would be to remove the beta dependency.

MihaZupan (approved):

src/Application/Extensions.cs
```diff
@@ -0,0 +1,117 @@
...
+// To learn more about using this project, see https://aka.ms/dotnet/aspire/service-defaults
+public static class Extensions
+{
+    public static TBuilder AddServiceDefaults<TBuilder>(this TBuilder builder) where TBuilder : IHostApplicationBuilder
+    {
+        builder.ConfigureOpenTelemetry();
+
+        builder.AddDefaultHealthChecks();
+
+        builder.Services.AddServiceDiscovery();
```
MihaZupan, line 29: Similar question here - does this affect anything?

benjaminpetit: It's used by aspire

PR #2765, opened 2025-02-24
Move ResponseBodyCanceled and UpgradeResponseCanceled to debug-level log event

### LINKED ISSUE #2631: Don't log TaskCanceledException with warning level
Hi, after log level for some entries was raised to `Warning` in `v2.2.0` I am experiencing a lot (like 5k for one day on dev, inlcuding also cancelled SignalR connections) of entries where underlying exception is `TaskCanceledException`, for example:

`UpgradeResponseCanceled: Copying the upgraded response body was canceled.`
`ResponseBodyCanceled: Copying the response body was canceled.`

In both cases:
```
System.Threading.Tasks.TaskCanceledException: The operation was canceled.\n ---> System.IO.IOException: Unable to read data from the transport connection: Operation canceled.\n ---> System.Net.Sockets.SocketException (125): Operation canceled
```

Request being cancelled is a common situation when user is interacting with web application or automated tests based on Selenium are being run. Because of that there is a ton of such log entries. 

I would like to recommend to not log `TaskCanceledException` with `Warning` level or at least expose some configuration that would enable to exclude it on demand.

Like I wrote above, those cancelled requests are valid cases in my application because users or automated tests often closes browser tabs. In a current state those logs could hide much more critical warnings because of their sheer number. I don't want to exclude those logs since they are produced from `Yarp.ReverseProxy.Forwarder.HttpForwarder` and I believe there are other valuable logs that could be excluded in the process. BTW what is the benefit of logging it with this log level? Why have you decided it is worth marking it as `Warning`?

MihaZupan:
Closes #2631

Same idea as #2503, but for `ResponseBodyCanceled` and `UpgradeResponseCanceled`.

`UpgradeResponseCanceled` could be an indication that the backend became unresponsive, but it's ultimately triggered by the client disconnecting. If this is a concern, the activity timeout should be lowered enough to catch such cases instead.

### DIFF
--- a/test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
+++ b/test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
@@ -2871,7 +2871,14 @@ public class HttpForwarderTests
         Assert.Equal(error, errorFeature.Error);
         Assert.IsAssignableFrom<TException>(errorFeature.Exception);
 
-        var expectedId = error is ForwarderError.RequestCanceled or ForwarderError.RequestBodyCanceled or ForwarderError.UpgradeRequestCanceled
+        var errorIsDebugLevel = error is
+            ForwarderError.RequestCanceled or
+            ForwarderError.RequestBodyCanceled or
+            ForwarderError.ResponseBodyCanceled or
+            ForwarderError.UpgradeRequestCanceled or
+            ForwarderError.UpgradeResponseCanceled;
+
+        var expectedId = errorIsDebugLevel
             ? EventIds.ForwardingRequestCancelled
             : EventIds.ForwardingError;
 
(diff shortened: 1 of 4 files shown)


test/ReverseProxy.Tests/Forwarder/HttpForwarderTests.cs
```diff
@@ -2871,7 +2871,14 @@ private static TException AssertErrorInfo<TException>(
         Assert.Equal(error, errorFeature.Error);
         Assert.IsAssignableFrom<TException>(errorFeature.Exception);
 
-        var expectedId = error is ForwarderError.RequestCanceled or ForwarderError.RequestBodyCanceled or ForwarderError.UpgradeRequestCanceled
+        var errorIsDebugLevel = error is
```
adityamandaleeka, line 2874: Maybe rename this? I _think_ the intent is just to say that these are not severe.

adityamandaleeka (approved):

PR #2772, opened 2025-02-27
Fix WebProxy address format in sample

MihaZupan:


adityamandaleeka (approved):

adityamandaleeka:
Did this not work before this change, or is this cleanup due to schema validation hygiene?

MihaZupan:
I added the `WebProxy` part to this config while adding the schema support as it was one of the few features that weren't showcased (we could choose to remove the WebProxy part as well since it's a sample). In this case this config is valid by the schema, but results in an error at runtime due to the url not being valid.

The reason why I chose not to enforce uri formats on these settings as part of the schema is that we support rewriting strings at runtime with config filters, e.g.
https://github.com/dotnet/yarp/blob/f0d7e5df077b8b74040146248f1a2e4aed501f12/samples/ReverseProxy.ConfigFilter.Sample/appsettings.json#L40-L43

And we do have a test where all `appsettings.json` in the repo are validated against the schema.
https://github.com/dotnet/yarp/blob/f0d7e5df077b8b74040146248f1a2e4aed501f12/test/ReverseProxy.Tests/Configuration/ConfigProvider/ConfigurationConfigProviderTests.cs#L676

benjaminpetit (approved):

PR #2775, opened 2025-02-28
Drop unsupported TFMs

MihaZupan:
Drop .NET 6 and .NET 7


Tratcher:
Removing TFMs is a significant breaking change that should only happen in a major version.

MihaZupan:
I'm fine with bumping the major version for it.

I personally don't view that as super necessary though given we're removing things that are out of support already and can't silently break you.

benjaminpetit (approved):

PR #2806, opened 2025-03-24
Fix Sec-WebSocket-Key handling during H2 => H1 WebSocket downgrades

### LINKED ISSUE #2805: H2 => H1 WebSocket downgrade logic doesn't handle invalid requests with the `Sec-WebSocket-Key` header set
When doing an HTTP/2 => HTTP/1.1 WebSocket downgrade, YARP will validate the response `Sec-WebSocket-Accept` header

https://github.com/dotnet/yarp/blob/3f7edd5cf3efb86d781cd48682f0e94594baf9c8/src/ReverseProxy/Forwarder/HttpForwarder.cs#L832-L842

We're reading the key from `response.RequestMessage.Headers` because we know we were the ones to add it previously.
https://github.com/dotnet/yarp/blob/3f7edd5cf3efb86d781cd48682f0e94594baf9c8/src/ReverseProxy/Forwarder/HttpForwarder.cs#L469-L476

But what our logic fails to handle is the case where the `Sec-WebSocket-Key` header already exists in the client's request.
Since we're adding another value, the logic in `CreateSecWebSocketAccept` fails as the key is malformed, causing YARP to throw.

https://github.com/dotnet/yarp/blob/3f7edd5cf3efb86d781cd48682f0e94594baf9c8/src/ReverseProxy/Forwarder/ProtocolHelper.cs#L69-L71

This only affects cases where:
- The client is talking HTTP/2 to the proxy and we're downgrading to 1.1
- The client included a `Sec-WebSocket-Key` header, which it shouldn't do for H2 WebSockets
- The server behind YARP is too accepting (doesn't validate the header), allowing the upgrade instead of returning an error response

MihaZupan:
Closes #2805 (issue has more info)

### DIFF
--- a/src/ReverseProxy/Forwarder/HttpForwarder.cs
+++ b/src/ReverseProxy/Forwarder/HttpForwarder.cs
@@ -201,6 +201,13 @@ internal sealed class HttpForwarder : IHttpForwarder
                     (destinationRequest, requestContent, _) = await CreateRequestMessageAsync(
                         context, destinationPrefix, transformer, config, isStreamingRequest, activityCancellationSource);
 
+                    // Transforms generated a response, do not proxy.
+                    if (RequestUtilities.IsResponseSet(context.Response))
+                    {
+                        Log.NotProxying(_logger, context.Response.StatusCode);
+                        return ForwarderError.None;
+                    }
+
                     destinationResponse = await httpClient.SendAsync(destinationRequest, activityCancellationSource.Token);
                 }
             }
@@ -471,8 +478,22 @@ internal sealed class HttpForwarder : IHttpForwarder
             {
                 request.Headers.TryAddWithoutValidation(HeaderNames.Connection, HeaderNames.Upgrade);
                 request.Headers.TryAddWithoutValidation(HeaderNames.Upgrade, WebSocketName);
-                var key = ProtocolHelper.CreateSecWebSocketKey();
-                request.Headers.TryAddWithoutValidation(HeaderNames.SecWebSocketKey, key);
+
+                // The client shouldn't be sending a Sec-WebSocket-Key header with H2 WebSockets, but if it did, let's use it.
+                if (RequestUtilities.TryGetValues(request.Headers, HeaderNames.SecWebSocketKey, out var clientKey))
+                {
+                    if (!ProtocolHelper.CheckSecWebSocketKey(clientKey))
+                    {
+                        Log.InvalidSecWebSocketKeyHeader(_logger, clientKey);
+                        // The request will not be forwarded if we change the status code.
+                        context.Response.StatusCode = StatusCodes.Status400BadRequest;
+                    }
+                }
+                else
+                {
+                    var key = ProtocolHelper.CreateSecWebSocketKey();
+                    request.Headers.TryAddWithoutValidation(HeaderNames.SecWebSocketKey, key);
+                }
             }
             // H1->H1, re-add the original Connection, Upgrade headers.
             else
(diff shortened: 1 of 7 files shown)


benjaminpetit (approved):

jmezach:
@MihaZupan Any chance this could be backported to 2.3 or 3.0 could get released any time soon? We've just been hit by this issue and it is causing our WebSocket connections to no longer work. We recently had to migrate from Ingress NGINX to Envoy Gateway and it seems that Envoy Gateway is now using HTTP/2 WebSockets, but some of our underlying services are running on IIS which does not support that so we have to make this transition.

MihaZupan:
Hey @jmezach, would it be possible for you to workaround the issue for the time being by removing the header in this erroneous case?

Presumably it should be as simple as (untested):
```c#
app.Use((context, next) =>
{
    if (context.Features.Get<IHttpExtendedConnectFeature>() is { IsExtendedConnect: true } connectFeature &&
        string.Equals("websocket", connectFeature.Protocol, StringComparison.OrdinalIgnoreCase))
    {
        // For HTTP/2 WebSocket requests, remove the SecWebSocketKey since the client shouldn't have sent it in the first place
        context.Request.Headers.Remove(HeaderNames.SecWebSocketKey);
    }

    return next();
});
```

jmezach:
@MihaZupan Actually, we're kind of doing that already but having Envoy remove the header. Unfortunately that didn't work, since the client is sending a different key than the one YARP generates, so when the 101 response comes back to the client is has a Sec-WebSocket-Accept header with a value that doesn't match what the client is expecting.

MihaZupan:
In that case something like this?
```c#
app.Use((context, next) =>
{
    if (context.Features.Get<IHttpExtendedConnectFeature>() is { IsExtendedConnect: true } connectFeature &&
        string.Equals("websocket", connectFeature.Protocol, StringComparison.OrdinalIgnoreCase) &&
        context.Request.Headers.TryGetValue(HeaderNames.SecWebSocketKey, out StringValues clientKeyValues) &&
        clientKeyValues.Count == 1 && clientKeyValues.ToString() is { Length: 24 } clientKey)
    {
        // For HTTP/2 WebSocket requests, remove the SecWebSocketKey since the client shouldn't have sent it in the first place
        context.Request.Headers.Remove(HeaderNames.SecWebSocketKey);

        context.Response.OnStarting(() =>
        {
            if (context.Response.StatusCode == StatusCodes.Status200OK)
            {
                context.Response.Headers[HeaderNames.SecWebSocketAccept] = Convert.ToBase64String(SHA1.HashData([.. Encoding.ASCII.GetBytes(clientKey), .. "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"u8]));
            }

            return Task.CompletedTask;
        });
    }

    return next();
});
```
Edit: Now tested that the above workaround works fine

jmezach:
@MihaZupan I've tried this but it doesn't work. I believe this is because by the time we get to the OnStarting delegate registered in the middleware, [this code](https://github.com/dotnet/yarp/blob/aa4e6955f3ab86d7149515ebed184c2aab34a8b0/src/ReverseProxy/Forwarder/HttpForwarder.cs#L870) has already run. So essentially the status code is already set back to 200 OK and thus we do not set the Sec-WebSocket-Accept header. I guess I could change the `if` to check for StatusCode 200 OK, but I'm not sure if that might break things.

jmezach:
@MihaZupan Any updates on this? It looks like a 3.0 release isn't coming anytime soon, but we're still experiencing this issue

MihaZupan:
I've tested and fixed the workaround above. As you already saw the status code check should look for 200 instead of 101, and the GUID was missing a trailing '1'.

I don't see risk in using that workaround, it should also remain valid (just unnecessary) even once the updated YARP version ships.
I'll talk with the team about shipping new versions.

PR #2807, opened 2025-03-27
Allow multiple certificates in k8s Certificate Selector

arontsang:


src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,118 @@
...
+using System.Threading;
 using Microsoft.AspNetCore.Connections;
 
 namespace Yarp.Kubernetes.Controller.Certificates;
 
 internal class ServerCertificateSelector : IServerCertificateSelector
 {
-    private X509Certificate2 _defaultCertificate;
+    private readonly Dictionary<NamespacedName, CertificateCandidate> _certificates = new();
+    private readonly ReaderWriterLock _readerWriterLock = new();
```
benjaminpetit, line 18: I would rather not having to use any lock in case of read, or at least use `ReaderWriterLockSlim`

arontsang:
The alternative would be to rewrite the interface to allow us to atomically
replace the certificates.

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,118 @@
...
+
+        public X509Certificate2 Certificate { get; }
+
+        public CertificateCandidate(X509Certificate2 certificate)
+        {
+            Certificate = certificate;
+            Domains = ImmutableHashSet.CreateRange(StringComparer.InvariantCultureIgnoreCase, GetDomains(certificate));
+        }
+
+        private static IEnumerable<string> GetDomains(X509Certificate2 certificate)
```
MihaZupan, line 79: What about wildcard certs? That would further complicate the `GetCertificate` logic

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,118 @@
...
+        }
+        finally
+        {
+            _readerWriterLock.ReleaseWriterLock();
+        }
+    }
+
+    private record CertificateCandidate
+    {
+        public IReadOnlySet<string> Domains { get; }
```
MihaZupan, line 69: Since this is internal logic, there's no need to type this as the interface. It can be strongly typed.

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,118 @@
...
+
+        public X509Certificate2 Certificate { get; }
+
+        public CertificateCandidate(X509Certificate2 certificate)
+        {
+            Certificate = certificate;
+            Domains = ImmutableHashSet.CreateRange(StringComparer.InvariantCultureIgnoreCase, GetDomains(certificate));
```
MihaZupan, line 76:
```suggestion
            Domains = new HashSet(GetDomains(certificate), StringComparer.OrdinalIgnoreCase);
```
Invariant culture is rarely the right choice. `ImmutableHashSet` is also unnecessary overhead in this case since this is all internal logic.

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
MihaZupan: Ideally this would come down to a `ConcurrentDictionary`-based approach without explicit locking on reads.
We should also try not to walk all namespaces on every read.

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,118 @@
...
+                    return cert;
+                }
+            }
+        }
+        finally
+        {
+            _readerWriterLock.ReleaseReaderLock();
+        }
+
+        return _certificates.Values.FirstOrDefault()?.Certificate;
```
MihaZupan, line 51: This is not safe to access outside of the lock as written

arontsang:
Well aware of wildcards. But this is a PR in the right direction.

MihaZupan:
But as-is it breaks wildcard certs that are currently working, no?

arontsang:
Huh..true...

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
arontsang: I've just updated the PR to use ConcurrentDictionary.
If you are happy for me to bring in Rx, I could instead build a ImmutableDictionary for a fast path (for non-wild card case).
We can also compute a prefix trie for the wildcard case...
Again, I would want to bring in Rx to debounce the Trie/Dictionary creation.

arontsang:
Replacing this to implement a faster wildcard lookup

PR #2815, opened 2025-04-04
Implement a fast lookup for wildcard certificates

arontsang:


src/Kubernetes.Controller/Certificates/ImmutableCertificateCache.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        private static int Compare(ReadOnlySpan<char> x, ReadOnlySpan<char> y)
+        {
+
+            var length = Math.Min(x.Length, y.Length);
+
+            for (var i = 1; i <= length; i++)
+            {
+                var charA = x[^i] & 0x5F;
+                var charB = y[^i] & 0x5F;
```
gfoidl, line 108: You compare it backwards?
Is it better then, to write the loop backwards and have normal index access?

src/Kubernetes.Controller/Certificates/ImmutableCertificateCache.cs
```diff
@@ -0,0 +1,121 @@
...
+                _ => x.Domain.Length - y.Domain.Length
+            };
+        }
+
+        private static int Compare(ReadOnlySpan<char> x, ReadOnlySpan<char> y)
+        {
+
+            var length = Math.Min(x.Length, y.Length);
+
+            for (var i = 1; i <= length; i++)
```
gfoidl, line 105: So the JIT can't elide the bound-check for indexing into the spans.
I think it's better to slice the spans to the common (= min) length and use one span in the loop, in order to elide at least one bound check (the JIT should handle the downward loop (see other comment) in a recent version).

Roughly:
```suggestion
            if (x.Length < y.Length)
            {
                y = y.Slice(0, x.Length);
            }
            else if (y.Length < x.Length)
            {
                x = x.Slice(0, y.Length);
            }

            for (var i = x.Length; i >= 1; --i)
```

gfoidl (commented):
Nit: there are quite a few blank lines that could be removed too.

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,57 @@
...
+    private ImmutableX509CertificateCache _certificateStore = new(Array.Empty<X509Certificate2>());
 
     public void AddCertificate(NamespacedName certificateName, X509Certificate2 certificate)
     {
-        _defaultCertificate = certificate;
+        _certificates[certificateName] = certificate;
+        _hasBeenUpdated = true;
     }
 
     public X509Certificate2 GetCertificate(ConnectionContext connectionContext, string domainName)
```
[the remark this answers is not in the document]
KLuuKer: i would suggest skipping certificate lookups and immediately fallback to the default cert when `string.IsNullOrEmpty`
you WILL get this traffic anytime someone runs ssllabs server test, or some other kind of auditing tool
or for the handful of people running this on their own internal network and connecting with some kind of weird legacy system

arontsang:
> If this pull also needs to work for the Kubernetes.Controller the following needs to be modified also
> 
> https://github.com/dotnet/yarp/blob/785e5fc80540baf3f34db17686d90cad595000a3/src/Kubernetes.Controller/Caching/IngressCache.cs#L100C9-L103C10
> 
> note: i don't work for MS but i have this code running right now with those lines patched out. and from my limited testing (cloudflare strict certificate check) it seems to work.

Thanks for that

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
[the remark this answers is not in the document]
arontsang: Updated to handle non SNI TLS hand shake.

KLuuKer (approved):

src/Kubernetes.Controller/Caching/IngressCache.cs
```diff
@@ -99,10 +99,13 @@ public void Update(WatchEventType eventType, V1Secret secret)
 
         if (!string.Equals(namespacedName.ToString(), _options.DefaultSslCertificate, StringComparison.OrdinalIgnoreCase))
         {
-            return;
+            _logger.LogInformation("Found secret `{NamespacedName}` to use as default certificate for HTTPS traffic", namespacedName);
```
MihaZupan: Are these if/else logs inverted? (note the `!string.Equals`)

src/Kubernetes.Controller/Caching/IngressCache.cs
```diff
@@ -99,10 +99,13 @@ public void Update(WatchEventType eventType, V1Secret secret)
...
+        }
+        else
+        {
+            _logger.LogInformation("Found secret `{NamespacedName}` to use for HTTPS traffic", namespacedName);
```
MihaZupan: This might result in a bunch of misleading logs (including warnings) since we're assuming all secrets are certs

https://github.com/dotnet/yarp/blob/e04fc5ac7720742d35f7e97d0f7aa9c6cba40565/src/Kubernetes.Controller/Certificates/CertificateHelper.cs#L33-L37

We should probably check that it's a relevant secret before logging.

src/Kubernetes.Controller/Certificates/ImmutableCertificateCache.cs
```diff
@@ -0,0 +1,123 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+#nullable enable
+namespace Yarp.Kubernetes.Controller.Certificates;
+
+public abstract class ImmutableCertificateCache<TCert> where TCert : class
```
MihaZupan: Please make this internal.

Why does it need to be abstract and generic? We should be able to merge this and `ImmutableX509CertificateCache` into one type.

MihaZupan (commented):
Looks like there are some related build & test failures in CI

Xulei-NL:
@MihaZupan Our business requires the implementation of HTTPS termination in YARP Ingress Controller. May I work on certificate selector, perhaps via a different PR? This is a good opportunity to improve my C# skills.

arontsang:
> @MihaZupan Our business requires the implementation of HTTPS termination in YARP Ingress Controller. May I work on certificate selector, perhaps via a different PR? This is a good opportunity to improve my C# skills.

I don't own the issue. You are free to submit another PR if you like.

You are also free to take my PR and improve it!

My strategy is to binary search for a wildcard certificate matching right to left (that way supporting prefix search). I'm not sure it's the BEST way, but I figured it would have less string comparisons than the naive approach.

src/Kubernetes.Controller/Certificates/ImmutableCertificateCache.cs
arontsang: The idea of the abstract base class was simply to allow for Unit Testing with a simplified TValue.

src/Kubernetes.Controller/Certificates/ServerCertificateSelector.cs
```diff
@@ -1,27 +1,100 @@
...
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        // Poll every 10 seconds for updates to
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            await Task.Delay(TimeSpan.FromSeconds(10), stoppingToken);
+            if (_hasBeenUpdated)
+            {
+                _hasBeenUpdated = false;
```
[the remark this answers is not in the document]
arontsang: Instead of using locks we should Interlocked.Exchange for the check and a Volatile.Write for the set.

src/Kubernetes.Controller/Certificates/ImmutableCertificateCache.cs
```diff
@@ -0,0 +1,123 @@
...
+        certificate = null;
+        return false;
+    }
+
+    /// <summary>
+    /// Sorts domain names right to left.
+    /// This allows us to use a Binary Search to achieve a suffix
+    /// search.
+    /// </summary>
+    private class DomainNameComparer : IComparer<WildCardDomain>
```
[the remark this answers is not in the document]
arontsang: Yes. Do that!

PR #2883, opened 2025-06-23
[Container] Add OutputCache configuration from file

benjaminpetit:
When using the YARP container, configuring middleware via dependency injection (DI) is not an option. This limitation makes it difficult to customize middleware behavior without modifying and rebuilding the container image.

In this PR, I added the possibility to configure output caching in the container configuration file.

### DIFF
--- a/src/ReverseProxy/Configuration/Middlewares/OutputCacheConfig.cs
+++ b/src/ReverseProxy/Configuration/Middlewares/OutputCacheConfig.cs
@@ -47,7 +47,7 @@ public sealed record NamedCacheConfig
     public bool ExcludeDefaultPolicy { get; set; }
 
     /// <inheritdoc cref="OutputCachePolicyBuilder.Expire(TimeSpan)"/>
-    public TimeSpan? ExpirationTime { get; set; }
+    public TimeSpan? ExpirationTimeSpan { get; set; }
 
     /// <inheritdoc cref="OutputCachePolicyBuilder.NoCache"/>
     public bool NoCache { get; set; }
@@ -107,8 +107,8 @@ public static class OutputCacheConfigExtensions
 
     private static void PolicyBuilder(OutputCachePolicyBuilder builder, NamedCacheConfig policy)
     {
-        if (policy.Duration.HasValue)
-            builder.Expire(policy.Duration.Value);
+        if (policy.ExpirationTimeSpan.HasValue)
+            builder.Expire(policy.ExpirationTimeSpan.Value);
 
         if (policy.NoCache)
             builder.NoCache();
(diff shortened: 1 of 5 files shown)


src/ReverseProxy/Configuration/Middlewares/OutputCacheConfig.cs
```diff
@@ -0,0 +1,122 @@
...
+    public string[]? VaryByQueryKeys { get; set; }
+
+    /// <inheritdoc cref="OutputCachePolicyBuilder.SetVaryByHeader(string[])"/>
+    public string[]? VaryByHeaders { get; set; }
+}
+
+/// <summary>
+/// Collections of extensions to configure OutputCache
+/// </summary>
+public static class OutputCacheConfigExtensions
```
MihaZupan, line 65: Nit: I think we've mostly stuck with one type per file so far

lqlive:
Hope you don't mind me asking - looking at this PR, it seems like dynamic operations on `OutputCacheConfig` (such as adding, modifying, or deleting configurations) aren't currently supported. Is this something that's been thought about?

src/ReverseProxy/Configuration/Middlewares/OutputCacheConfig.cs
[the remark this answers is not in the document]
MihaZupan: I suppose it's a question of whether we see this being valuable to YARP users outside of Aspire.
I don't have a problem with it shipping from YARP initially, we could always take the breaking change eventually if ASP.NET starts to include it.

Good point re: package update, if we're treating this as more experimental, then I think it'd be more flexible for us if we copy-paste it in Aspire instead, and keep it just in the container here for now.

src/ReverseProxy/Configuration/Middlewares/OutputCacheConfig.cs
MihaZupan: This type specifically should likely be in `namespace Microsoft.Extensions.DependencyInjection` (doesn't matter if it's not public)

test/ReverseProxy.Tests/Configuration/OutputCacheConfigTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        var app = builder.Build();
+
+        var policies = app.Services.GetRequiredService<IYarpOutputCachePolicyProvider>();
+        var test1 = await policies.GetPolicyAsync("test1");
+        var test2 = await policies.GetPolicyAsync("test2");
+        var test3 = await policies.GetPolicyAsync("test3");
+
+        Assert.NotNull(test1);
+        Assert.NotNull(test2);
+        Assert.NotNull(test3);
```
MihaZupan, line 91: I take it there's no good way to assert that the relevant properties were applied correctly?

PR #2920, opened 2025-08-08
Add non-prod variant of azure-pipelines.yml

MihaZupan:
The diff between the two is just official => unofficial: https://www.diffchecker.com/7doC3Q3q/


karelz:
Thanks for the diff, it was useful! Pity it is not including source links / built-in GH comparison feature.

PR #2933, opened 2025-08-26
Set initial capacity for a couple of HashSets in ProxyConfigManager

j3parker:
Noticed a few easy cases here while browsing the code.

### DIFF
--- a/src/ReverseProxy/Management/ProxyConfigManager.cs
+++ b/src/ReverseProxy/Management/ProxyConfigManager.cs
@@ -772,7 +772,8 @@ internal sealed class ProxyConfigManager : EndpointDataSource, IProxyStateLookup
 
         foreach (var incomingRoute in incomingRoutes)
         {
-            desiredRoutes.Add(incomingRoute.RouteId);
+            var added = desiredRoutes.Add(incomingRoute.RouteId);
+            Debug.Assert(added);
 
             // Note that this can be null, and that is fine. The resulting route may match
             // but would then fail to route, which is exactly what we were instructed to do in this case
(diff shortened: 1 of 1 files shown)


src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -609,11 +609,11 @@ private async Task<bool> ApplyConfigAsync(IReadOnlyList<RouteConfig> routes, IRe
         return (configuredClusters, errors);
     }
 
-    private void UpdateRuntimeClusters(IEnumerable<ClusterConfig> incomingClusters)
+    private void UpdateRuntimeClusters(IReadOnlyDictionary<string, ClusterConfig> incomingClusters)
     {
-        var desiredClusters = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        var desiredClusters = new HashSet<string>(incomingClusters.Count, StringComparer.OrdinalIgnoreCase);
```
j3parker, line 614: The `Debug.Assert(added)` below shows that we will always add at least this many

src/ReverseProxy/Management/ProxyConfigManager.cs
```diff
@@ -711,7 +711,7 @@ private void UpdateRuntimeClusters(IEnumerable<ClusterConfig> incomingClusters)
 
     private bool UpdateRuntimeDestinations(IReadOnlyDictionary<string, DestinationConfig>? incomingDestinations, ConcurrentDictionary<string, DestinationState> currentDestinations)
     {
-        var desiredDestinations = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        var desiredDestinations = new HashSet<string>(incomingDestinations?.Count ?? 0, StringComparer.OrdinalIgnoreCase);
```
j3parker, line 714: Other than the `?`/`??` aspect there is again a `Debug.Assert(added)` below

PR #2978, opened 2025-11-22
Add switch for ServeUnknownFileTypes in container

aromaa:


MihaZupan:
Thank you for looking to contribute here, but I don't think it's a good idea to start exposing everything through environment variables. Even #2885 noted that the support is "experimental".

I'd recommend suggesting and discussing such additions in https://github.com/dotnet/aspire/issues/10221 first.

PR #2992, opened 2026-01-29
Add a notice about ingress annotations that affect cluster level options

specialforest:
Update docs to include a notice on ingress annotations that affect cluster level options and can conflict between multiple ingresses.

### DIFF
--- a/samples/KubernetesIngress.Sample/README.md
+++ b/samples/KubernetesIngress.Sample/README.md
@@ -93,17 +93,19 @@ The table below lists the available annotations.
 |yarp.ingress.kubernetes.io/output-cache-policy|string|
 |yarp.ingress.kubernetes.io/backend-protocol|string|
 |yarp.ingress.kubernetes.io/cors-policy|string|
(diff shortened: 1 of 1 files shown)


specialforest:
The initial prompt:
```
Let's add a notice for annotations that correspond the following YarpIngressOptions

    ```
    cluster.LoadBalancingPolicy = ingressContext.Options.LoadBalancingPolicy;
    cluster.SessionAffinity = ingressContext.Options.SessionAffinity;
    cluster.HealthCheck = ingressContext.Options.HealthCheck;
    cluster.HttpClientConfig = ingressContext.Options.HttpClientConfig;
    cluster.HttpRequest = ingressContext.Options.HttpRequest;
    ```

The notice is about a fact that these options may be shared by multiple ingress rules, because they translate to properties of a cluster. Multiple ingress rules can share the same cluster (cluster key is defined by UpstreamName function). All ingress rules that end up sharing the same cluster need to specify the same values for these options to be deterministic, otherwise values from one ingress rules will overwrite values from another. Formulate wording for the notice yourself to be clear and concise.
```

MihaZupan (approved):
Thank you. I think it looks fine within the same table, but no hard preference there.

PR #3024, opened 2026-04-28
Use `WatchListXXXForAllNamespacesAsync` to list Kubernetes resources and refactor `WatchAsync`

Xulei-NL:
This PR
1. Replace `WatchListXXXForAllNamespaces` with `WatchListXXXForAllNamespacesAsync`. For instace `WatchListEndpointsForAllNamespaces` with `WatchListEndpointsForAllNamespacesAsync`. If so, we could get watch results asynchrounously.
2. Refactor `WatchAsync` to handle returned `IAsyncEnumerable(<WatchEventType, TResource>)` in step 1. In this way we could remove `TaskCompletionSource` which simulates a Task and rely on `await foreach`.

Appreciate every feedback!

Closes #3008

### DIFF
--- a/src/Kubernetes.Controller/Client/ResourceInformer.cs
+++ b/src/Kubernetes.Controller/Client/ResourceInformer.cs
@@ -283,7 +283,7 @@ public abstract class ResourceInformer<TResource, TListResource> : BackgroundHos
     }
 
     // Encapsulate LastEventUtcTicks into a class to eliminate IDE warning `Captured variable is modified in the outer scope`.
-    private record WatchState
+    private class WatchState
     {
         public long LastEventUtcTicks =  DateTime.UtcNow.Ticks;
     }
@@ -296,7 +296,7 @@ public abstract class ResourceInformer<TResource, TListResource> : BackgroundHos
             typeof(TResource).Name,
             _lastResourceVersion);
 
-        var linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        using var linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
 
         var watchState =  new WatchState();
         // reconnect if no events have arrived after a certain time
@@ -312,7 +312,14 @@ public abstract class ResourceInformer<TResource, TListResource> : BackgroundHos
                         EventId(EventType.DisposingToReconnect),
                         "Disposing watcher for {ResourceType} to cause reconnect.",
                         typeof(TResource).Name);
-                    linkedCancellationTokenSource.Cancel();
+                    try
+                    {
+                        linkedCancellationTokenSource.Cancel();
+                    }
+                    catch (ObjectDisposedException)
+                    {
+                        // The outer scope has already disposed of the cancellation token source.
+                    }
                 }
             },
             state: null,
--- a/src/Kubernetes.Controller/Client/ResourceInformer.cs
+++ b/src/Kubernetes.Controller/Client/ResourceInformer.cs
@@ -197,7 +197,7 @@ public abstract class ResourceInformer<TResource, TListResource> : BackgroundHos
     }
 
     protected abstract Task<HttpOperationResponse<TListResource>> RetrieveResourceListAsync(string resourceVersion = null, ResourceSelector<TResource> resourceSelector = null, CancellationToken cancellationToken = default);
-    protected abstract Watcher<TResource> WatchResourceListAsync(string resourceVersion = null, ResourceSelector<TResource> resourceSelector = null, Action<WatchEventType, TResource> onEvent = null, Action<Exception> onError = null, Action onClosed = null);
+    protected abstract IAsyncEnumerable<(WatchEventType, TResource)> WatchResourceListAsync(string resourceVersion = null, ResourceSelector<TResource> resourceSelector = null, Action<Exception> onError = null);
 
     private static EventId EventId(EventType eventType) => new EventId((int)eventType, eventType.ToString());
 
@@ -282,6 +282,12 @@ public abstract class ResourceInformer<TResource, TListResource> : BackgroundHos
             typeof(TResource).Name);
     }
 
+    // Encapsulate LastEventUtcTicks into a class to eliminate IDE warning `Captured variable is modified in the outer scope`.
+    private record WatchState
+    {
+        public long LastEventUtcTicks =  DateTime.UtcNow.Ticks;
+    }
+
     private async Task WatchAsync(CancellationToken cancellationToken)
     {
         Logger.LogInformation(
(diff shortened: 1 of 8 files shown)


Xulei-NL:
The PR for the issue #3008.

Xulei-NL:
@dotnet-policy-service agree

src/Kubernetes.Controller/Client/V1IngressResourceInformer.cs
```diff
@@ -28,9 +29,9 @@ protected override Task<HttpOperationResponse<V1IngressList>> RetrieveResourceLi
         return Client.NetworkingV1.ListIngressForAllNamespacesWithHttpMessagesAsync(resourceVersion: resourceVersion, fieldSelector: resourceSelector?.FieldSelector, cancellationToken: cancellationToken);
     }
 
-    protected override Watcher<V1Ingress> WatchResourceListAsync(string resourceVersion = null, ResourceSelector<V1Ingress> resourceSelector = null, Action<WatchEventType, V1Ingress> onEvent = null, Action<Exception> onError = null, Action onClosed = null)
+    protected override IAsyncEnumerable<(WatchEventType, V1Ingress)> WatchResourceListAsync(string resourceVersion = null, ResourceSelector<V1Ingress> resourceSelector = null, Action<Exception> onError = null)
     {
-        return Client.NetworkingV1.WatchListIngressForAllNamespaces(resourceVersion: resourceVersion,
-            fieldSelector: resourceSelector?.FieldSelector, onEvent: onEvent, onError: onError, onClosed: onClosed);
+        return Client.NetworkingV1.WatchListIngressForAllNamespacesAsync();
```
MihaZupan: I take it we should still be passing the parameters through?

src/Kubernetes.Controller/Client/ResourceInformer.cs
```diff
@@ -290,77 +296,71 @@ private async Task WatchAsync(CancellationToken cancellationToken)
...
+            // The loop was canceled safely, either by the timer or the original cancellationToken
         }
     }
 
+    private void OnError(Exception error)
+    {
+        if (error is KubernetesException kubernetesError)
+        {
+            // deal with this non-recoverable condition "too old resource version"
+            if (string.Equals(kubernetesError.Status.Reason, "Expired", StringComparison.Ordinal))
```
MihaZupan:
```suggestion
            if (kubernetesError.Status.Reason == "Expired")
```

src/Kubernetes.Controller/Client/ResourceInformer.cs
```diff
@@ -290,77 +296,71 @@ private async Task WatchAsync(CancellationToken cancellationToken)
...
             dueTime: TimeSpan.FromSeconds(45),
             period: TimeSpan.FromSeconds(45));
 
-        using var registration = cancellationToken.Register(watcher.Dispose);
         try
         {
-            await watcherCompletionSource.Task;
+            await foreach (var (watchEventType, item) in WatchResourceListAsync(resourceVersion: _lastResourceVersion,
+                                   resourceSelector: _selector, onError: OnError)
+                               .WithCancellation(linkedCancellationTokenSource.Token))
```
MihaZupan:
```suggestion
            await foreach (var (watchEventType, item) in WatchResourceListAsync(_lastResourceVersion, _selector, OnError)
                .WithCancellation(linkedCancellationTokenSource.Token))
```

src/Kubernetes.Controller/Client/ResourceInformer.cs
```diff
@@ -290,77 +296,71 @@ private async Task WatchAsync(CancellationToken cancellationToken)
...
         // reconnect if no events have arrived after a certain time
-        using var checkLastEventUtcTimer = new Timer(
+        await using var checkLastEventUtcTimer = new Timer(
             _ =>
             {
-                var lastEvent = DateTime.UtcNow - lastEventUtc;
-                if (lastEvent > TimeSpan.FromMinutes(9.5))
+                var currentTicks = Interlocked.Read(ref watchState.LastEventUtcTicks);
+                var timeSinceLastEvent = TimeSpan.FromTicks(DateTime.UtcNow.Ticks - currentTicks);
+                if (timeSinceLastEvent > TimeSpan.FromMinutes(9.5))
```
MihaZupan: I think this can be simplified a bit if you use `Stopwatch.GetTimeStamp`, `Stopwatch.GetElapsedTime` instead.

src/Kubernetes.Controller/Client/ResourceInformer.cs
```diff
@@ -282,6 +282,12 @@ private async Task ListAsync(CancellationToken cancellationToken)
             typeof(TResource).Name);
     }
 
+    // Encapsulate LastEventUtcTicks into a class to eliminate IDE warning `Captured variable is modified in the outer scope`.
+    private class WatchState
```
MihaZupan:
```suggestion
    private sealed class WatchState
```

MihaZupan (commented):
Thank you

src/Kubernetes.Controller/Client/V1IngressResourceInformer.cs
Xulei-NL: Sorry this is an oversight. Will fix.

src/Kubernetes.Controller/Client/ResourceInformer.cs
Xulei-NL: Thanks today I read the api doc about `Stopwatch`. I have switched to `Stopwatch` and simplified the code a bit.

src/Kubernetes.Controller/Client/V1IngressResourceInformer.cs
Xulei-NL: have fixed it and updated this PR.

src/Kubernetes.Controller/Client/ResourceInformer.cs
```diff
@@ -282,6 +283,18 @@ private async Task ListAsync(CancellationToken cancellationToken)
...
+    private sealed class WatchState
+    {
+        // Stopwatch timestamp of the most recent watch event.
+        // Access only through Interlocked operations.
+        public long LastEventStopwatchTimestamp = Stopwatch.GetTimestamp();
+
+        // Used as an atomic boolean to ensure the re-connect cancellation path is triggered only once.
+        // 0 means cancellation has not been requested; 1 means cancellation has already been requested.
+        // Access only through Interlocked operations.
+        public int CancellationRequested;
```
MihaZupan: Why is this needed? You're only calling `.Cancel()` on the CTS, there's no harm in doing so multiple times

Xulei-NL: Dear @MihaZupan ,

indeed calling `.Cancel()` many times has no harm in the timer. The code currently on the master branch uses `lastEventUtc = DateTime.MaxValue;` to make sure the cancellation happened once. See https://github.com/dotnet/yarp/pull/3024/changes/BASE..1462c1a1f4822994206edbb8084a65fe8bb93be0#diff-d890e91050758eee4329344d227734ac5b1d3e99f2e27a9dead7c564f206c8acL339.

I regarded this PR as a refactoring PR so I wanted to make sure the log below was only printed once:

```csharp
                    Logger.LogDebug(
                        EventId(EventType.DisposingToReconnect),
                        "Disposing watcher for {ResourceType} to cause reconnect.",
                        typeof(TResource).Name);
```

I'm also totally fine if we remove the `public int CancellationRequested`. Do you think we need to remove it?

MihaZupan: Yeah I'd remove it. If your cancellation took 45 seconds to kick in your process is very dead anyway

Xulei-NL: Clear. I have removed it and meanwhile refactored the comments. Will request you a new review now.

MihaZupan (approved):

PR #3031, opened 2026-05-22
Set proxy.forwarder activity DisplayName to request path

### LINKED ISSUE #2667: Telemetry - use the request path for name of trace

When using Yarp and OpenTelemetry with the Aspire dashboard, all requests show in the dashboard with the route match path as the display name.   This means that you have to drill into each one to find the request you are looking for.   It would be better if they showed the original request path.

Example

Image

A change will enable a more useful experience when viewing telemetry.

LindqvistMartin:
Fixes #2667.

The proxy.forwarder activity is created with a constant name, so every YARP request shows up under the same name in Aspire / OTel dashboards. This sets DisplayName to `Request.Path` so requests are distinguishable. `OperationName` stays `"proxy.forwarder"` so anything filtering by activity name isn't affected. Empty paths keep the default.

Added `Invoke_SetsActivityDisplayName_FromRequestPath` in `ProxyPipelineInitializerMiddlewareTests`. Verified it fails on main and passes with the change.

The issue is tagged Type: Documentation — let me know if you'd rather close this in favor of a docs-only fix instead.

### DIFF
--- a/src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
+++ b/src/ReverseProxy/Model/ProxyPipelineInitializerMiddleware.cs
@@ -63,6 +63,11 @@ internal sealed class ProxyPipelineInitializerMiddleware
 
         var activity = Observability.YarpActivitySource.CreateActivity("proxy.forwarder", ActivityKind.Server);
 
+        if (activity is not null && context.Request.Path.Value is { Length: > 0 } path)
+        {
+            activity.DisplayName = path;
+        }
+
         return activity is null
             ? _next(context)
             : AwaitWithActivity(context, activity);
--- a/test/ReverseProxy.Tests/Model/ProxyPipelineInitializerMiddlewareTests.cs
+++ b/test/ReverseProxy.Tests/Model/ProxyPipelineInitializerMiddlewareTests.cs
@@ -160,6 +160,43 @@ public class ProxyPipelineInitializerMiddlewareTests : TestAutoMockBase
         }
     }
 
+    [Fact]
+    public async Task Invoke_SetsActivityDisplayName_FromRequestPath()
+    {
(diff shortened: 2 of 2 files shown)


LindqvistMartin:
@dotnet-policy-service agree

LindqvistMartin:
Friendly ping. The red CI is unrelated flaky functional tests (HTTP/2 reset / socket abort), not from this change. Can rebase to re-trigger if useful.

MihaZupan:
> The issue is tagged Type: Documentation — let me know if you'd rather close this in favor of a docs-only fix instead.

Yeah, my concern from the issue hasn't changed - https://github.com/dotnet/yarp/issues/2667#issuecomment-2543122603. If setting the name to the full path was a good idea to do by default, then why aren't we doing it in ASP.NET Core instead.

This PR is also changing the activity that YARP emits and not the wrapping one ASP.NET creates, so presumably the dashboard would still show you lots of the same name as top-level traces, but with more informative child traces?

LindqvistMartin:
Yeah, agreed on both.

The rename only touches the `proxy.forwarder` activity, which is a child of the ASP.NET Core hosting activity. The dashboard's trace list is keyed off the hosting activity, so it doesn't change. Only the child span ends up with the path, so this doesn't really fix what the issue was about.

And full path as the default name is the wrong call cardinality-wise, same reason ASP.NET Core stays away from it. Worse here since YARP routes are usually catch-all. Path's already on the activity as a tag anyway, so the data's there, it's just the default name.

Since it's labelled Documentation, probably cleaner to add a short doc on setting DisplayName yourself from a custom middleware or an ActivityListener, for anyone who wants per-path names and is ok with the cardinality cost. I can write that up, or just close this, whatever you prefer.

MihaZupan:
Adding that documentation is valuable, but it'd go in the https://github.com/dotnet/AspNetCore.Docs repo, so I'll close this PR for now

